PR #7, opened 2020-11-05
implement binary_to_shaped_buffer

seanmor5:


seanmor5:
Shows basic functionality of going from Elixir binary to XLA Shaped Buffer on an XLA Device...Also doing some dependency clean up

examples/binaries_as_tensors.exs
```diff
@@ -0,0 +1,25 @@
+Exla.get_or_create_local_client(:host)
+
+# Under the hood, tensors are just blocks of memory with special indexing schemes
+# Ideally, we can represent tensors as binaries with a shape and dtype
+bin =
+  <<0::float, 1::float, 2::float, 3::float, 4::float, 5::float, 6::float, 7::float, 8::float>>
+  |> :binary.decode_unsigned()
+  |> :binary.encode_unsigned(:little)
```
josevalim, line 8:
```suggestion
  <<0::float-little, 1::float-little, 2::float-little, 3::float-little, 4::float-little, 5::float-little, 6::float-little, 7::float, 8::float-little>>
```

josevalim (approved):

examples/binaries_as_tensors.exs
josevalim: It seems that the GPUs are little endian so this is another concern we will need to have in mind when sending data in and out. A couple additional questions we should look into:

1. Do you need to do the same if you run it on the CPU?
2. Do we need to worry about the endianess of integers if we replace this by `::64`?
3. Do we need to worry about the endianess of scalars? We may be fine for scalars because converting an elixir int/float to C's probably handles it for us

PR #8, opened 2020-11-06
Library refactoring

seanmor5:
- [x] Separate utilities/helpers for resource creation, argument retrieval, etc. from `exla.cc` into external `exla_nif_util.cc`
- [x] Ensure proper handling of arguments (both counts and types) in each NIF
- [x] Check for NULL client in each NIF where client is used
- [x] Appropriately handle `xla::StatusOr` type
- [ ] Properly handle all types in shape/literal/constant/buffer creation


src/exla/exla.cc
```diff
@@ -79,103 +76,75 @@ static int load(ErlNifEnv* env, void** priv, ERL_NIF_TERM load_info){
...
-    return xla::PrimitiveType::C64;
-  } else if(atom_str.compare("complex128") == 0){
-    return xla::PrimitiveType::C128;
-  } else if(atom_str.compare("tuple") == 0){
-    return xla::PrimitiveType::TUPLE;
-  } else if(atom_str.compare("opaque") == 0){
-    return xla::PrimitiveType::OPAQUE_TYPE;
-  } else if(atom_str.compare("token") == 0){
-    return xla::PrimitiveType::TOKEN;
+  if(platform_str.compare("host") == 0){
```
josevalim, line 87: I wonder if we should consider other approaches here when it comes to our atom handling. Some options are:

1. Convert atoms into strings in Elixir itself to reduce the code here
2. Convert atoms into integers in Elixir itself and match on integers here

In both cases, I would like the validation of the data to belong to Elixir as much as possible, so we get both good error messages and reduce the amount of C++ code. This will be handy when we do the changes we discussed today of having a EXLA.NIF module. What do you think?

For now, you can either:

1. Leave this as is but add a TODO

2. Rename this function to get_int_platform and introduce a Exla.get_platform in Elixir that does the conversion and call Exla.get_int_plataform (until we introduce the Exla.NIF module)

src/exla/exla_nif_util.cc
```diff
@@ -0,0 +1,42 @@
+#include "exla/exla_nif_util.h"
+
+int ExlaNifUtil::get(ErlNifEnv* env, ERL_NIF_TERM term, int &var){ return enif_get_int(env, term, &var); }
+int ExlaNifUtil::get(ErlNifEnv* env, ERL_NIF_TERM term, long int &var){ return enif_get_int64(env, term, &var); }
+int ExlaNifUtil::get(ErlNifEnv* env, ERL_NIF_TERM term, long long int &var){ return enif_get_int64(env, term, (long int*) &var); }
+
+int ExlaNifUtil::get(ErlNifEnv* env, ERL_NIF_TERM term, std::string &var){
```
josevalim, line 7: I am a bit torn about this one. Are we trying to convert either a charlist or a binary into a C++ string? What is the use case for passing any of them?

Similarly to the atoms stuff, I would prefer to ensure only a certain type of them is given to C++ string and any conversion or normalize happens in the Elixir entrypoint. :)

josevalim (commented):
The bazel stuff looks fantastic, the C++ stuff too. I have added some comments, they are mostly about removing the data handling/normalization/conversion from the C++ layer as much as possible!

README.md
```diff
@@ -4,31 +4,32 @@ Elixir XLA Client for compiling and running Elixir code on CPU/GPU/TPU.
...
 Then to run:
 
-```
-$ docker run -it \
-    -v $PWD:$PWD \
-    -w $PWD \
-    --rm exla:cuda10.1 bash
+```shell
+docker run -it \
+  -v $PWD:$PWD \
```
jackalcooper, line 19:
```suggestion
  -v $PWD:$PWD \
  -e TEST_TMPDIR=$PWD/bazel_cache \
```
add this to save bazel cache on the host

Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN wget https://repo.hex.pm/installs/1.10.0/hex-0.20.6.ez && \
     mix archive.install ./hex-0.20.6.ez --force && \
     rm ./hex-0.20.6.ez
 
-RUN python3 -m pip install ${pip_args} numpy
+# RUN python3 -m pip install ${pip_args} numpy
```
jackalcooper, line 31:
```suggestion

```

PR #9, opened 2020-11-09
check in macros and cleanup TODOs

seanmor5:


src/exla/exla_nif_util.h
```diff
@@ -17,6 +17,20 @@ namespace exla {
...
+  ERL_NIF_TERM ok(ErlNifEnv* env, ERL_NIF_TERM term);
+
+  /*
+   * Helper for returning `:ok` from NIF.
+   */
+  ERL_NIF_TERM ok(ErlNifEnv* env);
+
+  /*
+   * Helper for returning `{:ok, term}` from NIF.
+   */
```
josevalim, line 32: Left-over comment? :)

src/exla/exla_nif_util.h
```diff
@@ -44,6 +58,65 @@ namespace exla {
...
+  template <typename T>
+  ERL_NIF_TERM get(ErlNifEnv* env, ERL_NIF_TERM term, T* &var){
+    return enif_get_resource(env, term, resource_object<T>::type, (void **) &var);
+  }
+
+  template <typename T>
+  ERL_NIF_TERM make(ErlNifEnv* env, T &var){
+    void* ptr = enif_alloc_resource(resource_object<T>::type, sizeof(T));
+    new(ptr) T(std::move(var));
+    ERL_NIF_TERM ret = enif_make_resource(env, ptr);
```
josevalim, line 93: Remember we need to call `enif_release_resource`, so we should probably add a TODO to this function and everywhere else we use enif_make_resource.

src/exla/BUILD
```diff
@@ -19,7 +19,15 @@ cc_library(
     "@org_tensorflow//tensorflow/stream_executor:stream_executor_headers",
     "@org_tensorflow//tensorflow/compiler/xla:statusor",
   ],
-  linkstatic = 1,
+)
+
+# I don't like the NIF dependency
```
josevalim, line 24: Is this a TODO? :) Can we do something about it?

src/exla/exla.cc
```diff
@@ -439,14 +367,15 @@ ERL_NIF_TERM compile(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[]){
...
-  std::vector<std::unique_ptr<xla::LocalExecutable>> executables = exec_status.ConsumeValueOrDie();
+  EXLA_ASSIGN_OR_RETURN(std::vector<std::unique_ptr<xla::LocalExecutable>> executables,
+                        xla_objects->client->Compile(*computation, argument_layouts, options), env);
+
   ERL_NIF_TERM exec_refs[executables.size()];
   int i = 0;
   for(auto it=std::begin(executables);it!=std::end(executables);++it){
     exec_refs[i++] = exla::make<xla::LocalExecutable>(env, executables.at(i));
   }
+  // TODO: This should return the vector. There is an executable for every partition, usually 1.
```
josevalim, line 378: What is a partition in this case?

src/exla/exla_nif_util.h
seanmor5: I'll add it...I added those in and started testing with it earlier and ran into a strange problem. I don't know the underlying implementation of inspecting References, but I assume it shows the memory address of the resource?

Basically, when I added in `enif_release_resource` after `enif_make_resource` I would get valid references returned back, but they weren't valid addresses...sometimes the reference would be: `#Reference<0.0.49482.0>` or something similar that I knew was garbage....

src/exla/exla.cc
seanmor5: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/client/executable_build_options.h#L76

It's possible to partition operations between two devices...if you had a very large model and needed to split it, I think you could specify the number of partitions and then `Compile` takes care of partitioning the computation for you, and then you can run the executables you get back on different devices

src/exla/BUILD
seanmor5: Let me mark it as a TODO...

src/exla/exla.cc
josevalim: I see, thanks for explaining! So we would need to explicitly request it to split it, right? Maybe for now it will be good enough to just raise an error if it returns more than one entry? Something like:

```
if(execs.length > 1) {
  enif_raise_exception(env, enif_make_atom(env, "too_many_partitions"))
}
```

And then whenever, if ever, we need partitioning, we see how to tackle it. :)

PR #10, opened 2020-11-10
Scaffold a basic example with a start of High-Level API

seanmor5:


lib/exla/client.ex
```diff
@@ -0,0 +1,37 @@
...
+      {:error, _msg} -> {:error, :normal}
+    end
+  end
+
+  @impl true
+  def handle_call(:get_device_count, _from, client = %Client{}) do
+    device_count = Exla.NIF.get_device_count(client.ref)
+    {:reply, device_count, client}
+  end
+end
```
josevalim, line 37: Sorry as I was not clear before. I don't think EXLA itself should be the one providing the GenServer, but rather its users. For example, if we take the INK project and the JIT mode, we will have this:

```elixir
defn mult_and_add(a, b, c) do
  a * b + abs(c)
end
```

being compiled to this:

```
def mult_and_add(a, b, c) do
  INK.CachedComputation.run({__MODULE__, :mult_and_add, 3}, ...defnast..., [a, b, c])
end
```

The `INK.CachedComputation` will be its own gen server, as we don't want to compile this multiple times as we have multiple invocations. Therefore, in order to avoid a GenServer that calls another GenServer, I think it would be best if EXLA did not prescribe any client architecture at all. :)

In other words, the code for EXLA will be plain boring:

```elixir
client = create_client(...)
EXLA.add(client, ..., ...)
```

But I think that's fine, because people are not supposed to be invoked EXLA by hand anyway. :) WDYT?

src/exla/exla.cc
```diff
@@ -15,9 +15,10 @@
 
 // TODO: Synchronize access.
 // TODO: Separate client from global object.
+// TODO: It might be more informative on the Elixir side to replace `enif_make_badarg` with something like `{:error, reason}`. Thoughts?
```
josevalim, line 18: Yes, correct. let's always return `{:error, reason}`.

src/exla/exla_nif_util.cc
```diff
@@ -52,7 +52,31 @@ namespace exla {
...
+    if(!enif_get_map_value(env, term, exla::make(env, "number_of_replicas"), &number_of_replicas_arg)) return 0;
+    if(!enif_get_map_value(env, term, exla::make(env, "intra_op_parallelism_threads"), &intra_op_parallelism_threads_arg)) return 0;
+
+    stream_executor::Platform* platform;
+    int number_of_replicas, intra_op_parallelism_threads;
+
+    // TODO: Adjust this to accept `int` which is passed from Elixir
+    if(!exla::get_platform(env, platform_arg, platform)) return 0;
+    if(!exla::get(env, number_of_replicas_arg, number_of_replicas)) return 0;
+    if(!exla::get(env, intra_op_parallelism_threads_arg, intra_op_parallelism_threads)) return 0;
```
josevalim, line 72: Similar to the other discussions, should we simply call the NIF as: `EXLA.NIF.get_client(number_of_replicas, intra_op_parallelism_threads, ...)` instead of passing a map? This would allow us to normalize and validate the data within Elixir instead.

lib/exla/client.ex
seanmor5: Okay I see, I think that makes sense. Because this is low level, we want to leave it up to the user and leave everything as flexible/simple as possible. That will be an easy adjustment to make

src/exla/exla_nif_util.cc
seanmor5: So rather in `Exla.Client.create_client` do:

```elixir
def create_client(options = %LocalClientOptions{}) do
   Exla.NIF.get_or_create_local_client(options.device, options.replicas, ....)
```

I see what you mean, this will simplify the C++ implementations as well

src/exla/exla_nif_util.cc
```diff
@@ -52,16 +52,7 @@ namespace exla {
...
-    size_t map_size;
-
-    if(!enif_get_map_size(env, term, &map_size) || map_size != 4) return 0;
-
-    ERL_NIF_TERM platform_arg, number_of_replicas_arg, intra_op_parallelism_threads_arg
-
-    if(!enif_get_map_value(env, term, exla::make(env, "platform"), &platform_arg)) return 0;
-    if(!enif_get_map_value(env, term, exla::make(env, "number_of_replicas"), &number_of_replicas_arg)) return 0;
-    if(!enif_get_map_value(env, term, exla::make(env, "intra_op_parallelism_threads"), &intra_op_parallelism_threads_arg)) return 0;
+  int get_options(ErlNifEnv* env, ERL_NIF_TERM platform_arg, ERL_NIF_TERM number_of_replicas_arg, ERL_NIF_TERM intra_op_parallelism_threads_arg, xla::LocalClientOptions &options){
```
josevalim, line 55: Should this function even be here now? It is no longer polymorphic on the different options and, as far as I know, it is now very specific to the client configuration which is only used in one place? :)

josevalim: Maybe if you want to keep this function, we can pass the whole `argv` and then let this function take it out! This way at least all get_options functions have the same signature!

seanmor5: I think I'm going to keep it as it separates the work of constructing the options struct from the actual work the NIF needs to do. I'll have it match on `argv` rather than on individual arguments

josevalim (approved):

PR #12, opened 2020-11-10
Scaffold

seanmor5:


seanmor5:
This is pretty much the process I was going to repeat for each resource. Wrap the reference in a struct, store fields that might be useful later, and then wrap the NIFs for each XLA resource in the module.

lib/exla/client.ex
```diff
@@ -5,7 +5,7 @@ defmodule Exla.Client do
   @enforce_keys [:ref]
   defstruct [:ref]
 
-  def create_client(options = %LocalClientOptions{}) do
+  def create_client(options \\ %LocalClientOptions{}) do
     case Exla.NIF.get_or_create_local_client(
```
josevalim, line 9: Btw, do we have a `create_local_client` API that doesn't return an existing one? Also, if we have different client configurations, is XLA guaranteeing each of them return a different client instance?

lib/exla/op.ex
```diff
@@ -0,0 +1,24 @@
...
+
+  def parameter(i, shape = %Shape{}, name) when is_integer(i) and i >= 0 and is_binary(name) do
+    case Exla.NIF.parameter(i, shape.ref, name) do
+      {:ok, ref} -> {:ok, %Op{ref: ref}}
+      {:error, msg} -> {:error, msg}
+    end
+  end
+
+  def parameter(_i, _shape, _name),
+    do: {:error, "Invalid arguments passed to `#{__MODULE__}.parameter/3.`"}
```
josevalim, line 16: For invalid arguments we typically don't return error tuples, we just let it blow up. So you can remove this clause. 👍

lib/exla/shape.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Exla.Shape do
+  alias __MODULE__, as: Shape
+
+  @enforce_keys [:dims, :dtype, :ref]
+  defstruct [:dims, :dtype, :ref]
+
+  # TODO: convert dtype to integer representation to use in NIF
+  def make_shape(dtype, dims) when is_atom(dtype) and is_tuple(dims) do
```
josevalim, line 8: Should dims be a list or a tuple? Since it can have arbitrary length... maybe a list is better? What would be the biggest number of dimensions we would work with? 10 or 1000?

josevalim (commented):
Hi @seanmor5, I like this a lot! My only question at the moment is if `parameter` is the best abstraction, given they need to have unique integers. Instead we may want to have a parameters function that builds all of them, but this is something we can decide later. :)

lib/exla/shape.ex
seanmor5: For shapes I think a Tuple is better. In deep learning the highest number of dimensions I've seen is 5 for working with video data (batch size, frame, height, width, color channels).

It's not very common to go above that. I guess it's possible but most of the time you're working with 2, 3, or 4 dimensions.

I thought a list wouldn't necessarily be needed because you have to individually define the size of a dimension so I can't imagine anybody defining something with 10, 100, or 1000 dimensions

lib/exla/client.ex
seanmor5: We do not. The current one wraps the GetOrCreate XLA function. I was looking at this earlier and realized that if we continue calling this method it will return new references to the same object on the XLA side. So unless we explicitly create a new one it just keeps returning different references to the same thing

seanmor5:
I thought about that too but I figured we could create something like a parameter server on top of these methods that handles parameters with a higher-level API like within INK.

lib/exla/client.ex
josevalim: >  So unless we explicitly create a new one it just keeps returning different references to the same thing

Does XLA allow us to do that? Also what happens if I call this function with different options, are they different clients on the C++ side?

In theory it is ok if we are using the same... as long as it is thread safe?

seanmor5: We can explicitly create a new one if we specify a different platform. For example if I have a CUDA client and a CPU client, I can create two clients for interacting with each locally. But, if I tried to create 2 clients on my host device with a different `num_replicas` option, it would ignore the options and return the already instantiated client on the host device.

There's also different types of clients that need to be handled at some points like: gRPC clients and compile-only clients. It seems, though, that local clients are the most common so I'm going to only focus on those until we have more built out.

It's not explicitly marked as being thread-safe and going through the code it doesn't seem to be. The local client is supposed to be a singleton object wrapping a platform/service, and trying to use two on one platform can create some issues especially when dealing with some of the mutable attributes you can access through the client.

josevalim: I see, so after-all we may want to model this using Elixir's Registry + GenServer, but I think we can wait with moving ahead with this. I think for now I would just leave a TODO:

```suggestion
    # TODO: Rename this function to get_local_client. It is a singleton,
    # non-thread-safe resource in XLA so we need to mimic the same
    # in Elixir. We should also have distinct steps for configuring and for
    # getting it.
    case Exla.NIF.get_or_create_local_client(
```

josevalim:
> I thought about that too but I figured we could create something like a parameter server on top of these methods that handles parameters with a higher-level API like within INK.

Sounds good. I think the overall goal is to keep the API low-level but not low-level enough that we can segfault stuff, but I think it is better to focus on the low level wrapping for now and we can do the high level one alongside INK. :)

josevalim (approved):

src/exla/exla_nif_util.h
```diff
@@ -89,7 +89,7 @@ namespace exla {
     void* ptr = enif_alloc_resource(resource_object<T>::type, sizeof(T));
     new(ptr) T(std::move(var));
     ERL_NIF_TERM ret = enif_make_resource(env, ptr);
-    enif_release_resource(ptr);
+    // enif_release_resource(ptr);
```
seanmor5, line 92: Unfortunately, I was getting random segmentation faults, and commenting this out fixed them. The weird thing is, everything works fine if I do it in IEX, but switching to running the tests causes segmentation faults in unpredictable ways.

josevalim: That's because `iex` most likely holds those references in its history.

I think I actually know why the segfault happens! Imagine this code:

```
a = EXLA.parameter(0)
b = EXLA.parameter(1)
res = EXLA.add(a, b)
```

After we call `add`, we no longer reference `a` and `b`, so the GC may garbage collect it... although `res` still points to them on the C side of things.

Maybe the solution here is, as we call `EXLA.add`, something like this:

```elixir
def add(%Op{refs: a}, %Op{refs: b}) do
  case EXLA.NIF.add(a, b) do
    {:ok, ref} -> %Op{refs: [ref] ++ a ++ b}
    {:error, ...} ->
  end
end
```

Other options are:

1. build the tree on Elixir and we call EXLA only once at the end once the tree is built
2. track the references in C++ but I think that will be trickier

seanmor5:
Passes all tests up to running an executable. Doesn't handle arguments and inputs yet. Also doesn't handle custom build/run options. I plan to handle arguments and inputs in my next PR and then options after that. This is basically the scaffold that the rest of the project will follow

lib/exla/op.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  def parameter(%Builder{ref: builder}, i, %Shape{ref: shape}, name)
+      when is_integer(i) and i >= 0 and is_binary(name) do
+    case Exla.NIF.parameter(builder, i, shape, name) do
+      {:ok, ref} -> {:ok, %Op{builder: builder, ref: ref}}
+      {:error, msg} -> {:error, msg}
+    end
+  end
+
+  # TODO: builder is redundant here because it's contained within each op
```
josevalim, line 27: You are right. We can probably skip the builder as long as the builder in both are equal!

lib/exla/shape.ex
```diff
@@ -0,0 +1,14 @@
...
+    case Exla.NIF.make_shape(dtype, dims) do
+      {:ok, ref} -> {:ok, %Shape{ref: ref, dims: dims, dtype: dtype}}
+      {:error, msg} -> {:error, msg}
```
josevalim, line 11: Btw, do you have any ideas which kind of errors we may have here?

josevalim (approved):

lib/exla/shape.ex
seanmor5: As I've been debugging, I've been changing some of the argument checks within the NIF to output which argument specifically it doesn't like. It's much more informative then just throwing a badarg on that end. Once everything is cleaned up, it should be basically impossible to get an error from the NIF because everything should be thrown from the Elixir side. For development though it makes it nice to know where something is going wrong

josevalim: @seanmor5 if that's the case, we can probably rewrite it to this:

```elixir
      {:ok, ref} = Exla.NIF.make_shape(dtype, dims) do
      %Shape{ref: ref, dims: dims, dtype: dtype}
```

We still get an error but without pushing bureaucracy up the stack. WDYT?

seanmor5: Okay I see, so it throws an error on the bad match and we still get to see what the problem is clearly but there's no need to keep passing that error along. This is the case for pretty much every NIF so I'll go back and edit them to match this

josevalim (approved):

PR #13, opened 2020-11-11
Add parameter support in compilation, running

seanmor5:


lib/exla/client.ex
```diff
@@ -31,8 +43,19 @@ defmodule Exla.Client do
...
+    # the refs of each Shape. To simplify the handling of `absl::Span` on the NIF side
+    # I only read spans in as Tuples. This is important because things like the dimensions
+    # of a shape, broadcast dimensions, etc. naturally fit well with Tuples, but other things
+    # that use spans such as this argument here work better with lists. I suppose I could create
+    # two distinct methods for handling lists and tuples.
+    shape_refs =
+      argument_shapes
+      |> Tuple.to_list()
+      |> Enum.map(&(&1.ref))
+      |> List.to_tuple()
```
josevalim, line 57: I think this is a strong indicator we should be using lists after all? :)

josevalim (approved):

seanmor5:
@josevalim Just checked in the new Tensor module, let me know what you think of following that kind of design

josevalim (approved):

src/exla/exla_nif_util.cc
```diff
@@ -74,6 +74,39 @@ namespace exla {
...
+    xla::ShapedBuffer* data[num_elems];
+    for(int i=0;i<num_elems;i++){
+      xla::ShapedBuffer* elem;
+      if(!get<xla::ShapedBuffer>(env, elems[i], elem)) return 0;
+      data[i] = elem;
+    }
+    span = absl::Span<xla::ShapedBuffer*>(data, num_elems);
+    // TODO: For some reason, this prevents `run` from SegFaulting...
+    for(auto i=span.begin();i!=span.end();++i){
+      LOG(ERROR) << (**i);
```
seanmor5, line 105: This is another one of those SegFaults that happens under random conditions. We don't ever SegFault with 0 arguments. Without this for loop, we SegFault with 1 argument. With this for loop, we don't SegFault with 1 argument, but we SegFault with 2 arguments.

josevalim: Since we suspect a lot of the segfaults are related to the GC, you can use `:erlang.garbage_collect(self())` before and after some of these operations to see if it makes it more consistent.

seanmor5: As it turns out, it was due to how I was using the `absl::Span` to handle argument inputs. I switched everything up to use an `std::vector` instead and it worked.

seanmor5:
This handles fully handles parameters and output from running XLA Executables without any issues. Eventually we'll want to separate `Tensor` from the rest of the Exla things, but it's nice as a prototype right now. The next step I'm going to do is work on handling executable build and run options similar to how OneFlow handles them: https://github.com/Oneflow-Inc/oneflow/blob/master/oneflow/xrt/xla/xla_executable_context.cpp

josevalim (approved):

PR #14, opened 2020-11-12
Executable context

seanmor5:
Builds on https://github.com/seanmor5/exla/pull/13 by adding something similar to https://github.com/Oneflow-Inc/oneflow/blob/master/oneflow/xrt/xla/xla_executable_context.cpp within `LocalExecutable.run/3`. The idea is that we can work mostly in Elixir to ensure the execution context is set up before the executable is actually run. The context takes care of executable run options, populates input buffers (transfers tensors onto the device), ensures compatibility between client and specified device, etc. The idea is that it can be extended to take care of any context related task within Elixir.

Also adds `client` field to LocalExecutable to ensure the client isn't prematurely garbage collected and to facilitate client-related tasks within the executable. Adds `platform` field to client for easy pattern matching to ensure compatibility between client and requested actions.


seanmor5:
Ahhh I messed this PR up by including all of the stuff from https://www.github.com/seanmor5/exla/pull/13

examples/basic_addition.exs
```diff
@@ -2,41 +2,30 @@
...
-z = Exla.NIF.add(Exla.NIF.add(x, y), c)
+z = Exla.Op.add(Exla.Op.add(x, y), c)
 
 # Build computation from root
-comp = Exla.NIF.build(z)
+comp = Exla.Builder.build(z)
 
 # Compile to a local executable, we must include argument shapes so it can be built in to the executable
-exec = Exla.NIF.compile(comp, {arg1_shape, arg2_shape}, %{})
+exec = Exla.Client.compile(client, comp, {arg1_shape, arg2_shape})
```
josevalim, line 23: This is interesting... why do we need to pass the arg1_shape and arg2_shape both when defining the computations parameters and when compiling? Shouldn't XLA be inferring based on the parameters that those are our shapes? 

Btw, parameters list in Elixir is typically done as a list (such as `apply(Kernel :+ [1, 2, 3])`), so should they be a list here? 🤔

examples/basic_addition.exs
```diff
@@ -2,41 +2,30 @@
...
+t1 = Exla.Tensor.scalar(1, :int32)
+t2 = Exla.Tensor.scalar(1, :int32)
 
-# Place both of these on the XLA device
-# You need to specify a device ordinal, which is basically the id of the device
-# You also can specify a memory allocator to use, for now we leave it nil which
-# uses the default allocator
-a_buffer = Exla.NIF.literal_to_shaped_buffer(a, 0, nil)
-b_buffer = Exla.NIF.literal_to_shaped_buffer(b, 0, nil)
+result = Exla.LocalExecutable.run(client, exec, {t1, t2})
```
josevalim, line 29: As mentioned, this should probably be a list too:

```suggestion
result = Exla.LocalExecutable.run(client, exec, [t1, t2])
```

examples/basic_addition.exs
josevalim: Another question, when we compile, does XLA tell us what is going to be outgoing shape? Or just when we execute it?

lib/exla/tensor.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Exla.Tensor do
+  alias __MODULE__, as: Tensor
+  alias Exla.Client
+  alias Exla.Shape
+  # TODO: Different namespace?
```
josevalim, line 5: Do you mean if the struct fields should be in another namespace? No, I think this is the correct place. :)

seanmor5: I meant the Tensor module as a whole. Just because everything else wraps something specific in XLA...more of something to think about later on separating XLA specifics from the higher level stuff

examples/basic_addition.exs
seanmor5: It doesn't tell us the shape, but OneFlow does some things to save memory by reusing buffers with the output, so I was planning on doing most of the same.

lib/exla/tensor.ex
josevalim: Ah, sure!

PR #17, opened 2020-11-13
Use tensorflow as a submodule

josevalim:
This PR uses a different approach to compile exla. However, it breaks two things:

1. The compile database should be assembled differently
2. The Docker file has to be updated

I don't have enough context on any of those, so we would need to do it after merging this (if desired at all). Also note one should no longer invoke the Makefile directly, and use `mix compile` instead.


josevalim:
Dockerfile has been updated.

PR #18, opened 2020-11-14
Use a global tensorflow checkout

josevalim:
This allows EXLA to share most of the build steps between versions when installed as a package.


josevalim:
The downside of this PR is that, whenever we bump the tensorflow version, we will have to recompile everything... but I think that's fine?

seanmor5:
Realistically, how often will we need to bump the TensorFlow version? I'd say bumping the TF dependency will occur far less often than us bumping the EXLA version

PR #19, opened 2020-11-14
Refactor lib/exla/nif.ex

wojtekmach:


seanmor5 (approved):
Much cleaner!

lib/exla/nif.ex
```diff
@@ -8,178 +8,188 @@ defmodule Exla.NIF do
     :erlang.load_nif(path, 0)
   end
 
+  defmacrop nif_error() do
+    quote do
```
josevalim, line 12: What about having it call nif_error(__ENV__.function) which then builds the error message? That’s so we don’t build the same string over and over again.

PR #32, opened 2020-11-21
Zero-Copy Transfers

seanmor5:
Only optimized on CPU right now, benchmark showing virtually no difference between "pre-loaded" and not pre-loaded:

```
Name                             ips        average  deviation         median         99th %
xla cpu dot pre-loaded        2.58 K      387.98 μs    ±11.64%      380.02 μs      618.88 μs
xla cpu dot                   2.51 K      398.85 μs    ±12.73%      388.67 μs      633.52 μs
elixir dot                  0.0235 K    42494.06 μs    ±20.26%    41933.57 μs    61825.09 μs

Comparison: 
xla cpu dot pre-loaded        2.58 K
xla cpu dot                   2.51 K - 1.03x slower +10.87 μs
elixir dot                  0.0235 K - 109.53x slower +42106.08 μs
```

### DIFF
--- a/c_src/exla/BUILD
+++ b/c_src/exla/BUILD
@@ -5,9 +5,10 @@ package(default_visibility=["//visibility:private"])
 cc_library(
   name = "exla_allocator",
   srcs = ["exla_allocator.cc"],
-  hdrs = ["exla_allocator.h"],
+  hdrs = ["exla_allocator.h"] + glob(["erts/**/*.h"]),
   deps = [
-    "@org_tensorflow//tensorflow/stream_executor:stream_executor_headers",
+    "@org_tensorflow//tensorflow/core/framework:allocator",
+    "@org_tensorflow//tensorflow/core:lib",
     "@org_tensorflow//tensorflow/compiler/xla:statusor",
   ],
 )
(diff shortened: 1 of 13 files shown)


josevalim (approved):
Beautiful! Given the 16-byte alignment, if we create a vector of size 3 and type int8, does it mean we will have to copy?

seanmor5:
I think so but I still have to test. I think it might be helpful to make sure it all still works when using binaries that aren't well-aligned

seanmor5:
This is a pretty rough implementation but this PR achieves:

1) Zero-copy transfers when using a CPU backend. In both specifying input and retrieving the output of a computation, it'll always just use an Erlang binary as the underlying data representation.

2) Ability to write result of CUDA execution to binary or retain the result on the device.  The transfer works by using the underlying transfer manager to transfer the data from the GPU into an `xla::Literal`. We then create a binary that points to the underlying `xla::Literal` data so the data is only moved once from GPU to the host platform.

I think we can remove all of the little endianess stuff now because once we read the result back it's in a form the binary expects it to be in.

I want to wrap all of these implementations into an `ExlaBuffer` class which owns both an `ErlNifBinary` and an `xla::ScopedShapedBuffer` and contains all of the move semantics between the VM and the accelerators, but I'll save that for another PR.

josevalim (approved):
Amazing!

PR #103, opened 2020-12-17
AOT Compilation

seanmor5:
Here is how I would imagine the basics of AOT Compilation:

1) User runs a mix task specifying a module/file plus some flags
2) We traverse the module and compile all defn
3) We generate a NIF .so which wraps every compiled function and they can load it in their code

At a lower level this is how it works:

1) For each function, we need to build a computation, matching the constraints expected by the aot/codegen (one example is that the codegen expects the computation result to be a Tuple, something to do with sharding I believe)

2) While building the computation for a function, we need to track inputs and outputs in order to generate a Subgraph Configuration. This just specifies the names and shapes of each input and output: https://www.tensorflow.org/xla/tfcompile#step_1_configure_the_subgraph_to_compile. It's a protobuf and I'm pretty sure there are a few protobuf libs in Elixir already, or we can just build it ourselves.

3) We pass both the Computation and the config protobuf to a NIF. The NIF generates: an object file, a header file, and a metadata file for a single function. So we'd end up with multiple object/header/metadata files for multiple functions.

4) For each compiled function, we need to create a NIF interface that: includes all of the compiled function headers, creates NIF functions for each compiled function. We know the shape and how many arguments are expected, so we can read the data in as binaries/lists and do whatever normalization we need to from within the generated NIF. The NIF needs to target a module, so this is probably a user specified config.

5) At the end we have: headers/objects for each compiled function and a NIF source file that provides an Elixir interface to these functions. The final step is to use Bazel to generate a single shared object from each of the objects/headers and the source. The headers/objects depend on TensorFlow source, so this is probably going to be one of the biggest pain points.

I've included an example of using a pre-built computation and hardcoded subgraph configuration to generate a compiled function (header/object files).


josevalim:
I am continuing this work on a jv-aot branch. :)

PR #220, opened 2021-02-17
Simplify -0.0 workaround

wojtekmach:


josevalim:
String.to_integer is inlined by the compiler which is then further inlined directly into the integer by the erlang compiler. So we are being very indirect on purpose. :)

wojtekmach:
yikes, I was pretty sure I tested it on otp 23 and it worked correctly but I probably just checked from IEx. Sorry!

PR #225, opened 2021-02-17
improve Nx documentation

AlexJuca:
This PR adds several improvements and corrects a few spelling mistakes.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -71,7 +71,7 @@ defmodule Nx do
         [2, 4, 6]
       >
 
-  In practice, broadcasting is not restricted only to scalars. It
+  In practice, broadcasting is not restricted only to scalars; it
   is a general algorithm that applies to all dimensions of a tensor.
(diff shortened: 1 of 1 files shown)


nx/lib/nx.ex
```diff
@@ -71,7 +71,7 @@ defmodule Nx do
         [2, 4, 6]
       >
 
-  In practice, brodcasting is not restricted only to scalars, it
+  In practice, broadcasting is not restricted only to scalars. It
```
seanmor5, line 74:
```suggestion
  In practice, broadcasting is not restricted only to scalars; it
```

I think this might read better

seanmor5 (changes requested):

nx/lib/nx.ex
AlexJuca: I agree.

AlexJuca (commented):
This looks ok.

seanmor5 (approved):
Thanks!

Issue #157, opened 2021-01-20
Support common LinAlg primitives


seanmor5:
For now, at least:

- [x] QR decomposition
  - [x] (EXLA implementation missing)
- [x] LU decomposition
- [x] Cholesky decomposition
- [x] SVD
- [x] Eigen decomposition (hermitian)
- [ ] Eigen decomposition (general)
- [x] Triangular Solve
- [x] Norms (L0, L1, L2)
  - [x] ord: -2 for 2D missing
  - [x] nuclear missing

maxastyler:
I'm wondering if it's best to start name-spacing off these new features instead of putting them in the main Nx module?
Numpy's got their numpy.linalg namespace for all these functions.

billylanchantin:
I agree with @maxastyler. Jax also seems to have that structure:

https://github.com/google/jax/blob/master/jax/_src/lax/linalg.py

Also, is this potentially blocked by #118? For instance, the Cholesky decomposition is supposed to be `LL*` where `L*` is the conjugate transpose.

Or, do we want to just deal with the real-valued cases first and come back to the general case as part of #118?

josevalim:
We can start only real-valued ones. We will have to revisit a bunch of stuff once we add complex numbers.

also, we can always tackle namespacing later too!

polvalente:
For implementation of things like LU and QR decomposition, should contributors try to write it by hand? I ask this because I looked into the Numpy source code and it basically defers most things to the LAPACK interface.

polvalente:
For implementation of things like LU and QR decomposition, should contributors try to write it by hand? I ask this because I looked into the Numpy source code and it basically defers most things to the LAPACK interface.

seanmor5:
@polvalente Yes, you'll have to write the implementation by hand. The implementation of Cholesky decomposition is probably something you'll want to look at. Check out: https://github.com/elixir-nx/nx/blob/main/nx/lib/nx/binary_backend.ex to see how we do the work on binaries.

polvalente:
Ok! Any tips on how to tackle iterating and assigning values to a given column of a matrix?

Yesterday I played around a bit but ended up needing to use a combination of `comcatenate`, `transpose` and `reshape` to convert the list of Q columns and R columns in a Gram-Schmidt decomposition which ended up looking ugly haha

seanmor5:
[...]

PR #226, opened 2021-02-17
Implement p-norm definitions

polvalente:
This PR aims to implement the 0-norm, 1-norm and 2-norm cited in #157 via the `p_norm/2` function

### DIFF
--- a/exla/test/exla/defn_expr_test.exs
+++ b/exla/test/exla/defn_expr_test.exs
@@ -2392,8 +2392,7 @@ defmodule EXLA.DefnExprTest do
     test "works with norm calls" do
       t = Nx.tensor([3.0, 4.0, 0.0])
       assert compare_tensors!(normalize(t), Nx.tensor([0.6, 0.8, 0.0]))
-
-      assert compare_tensors!(Nx.dot(t, t), Nx.power(Nx.norm(t), 2))
+      assert compare_tensors!(Nx.norm(normalize(t)), Nx.tensor(1.0))
     end
   end
 
--- a/exla/test/exla/defn_expr_test.exs
+++ b/exla/test/exla/defn_expr_test.exs
@@ -2388,11 +2388,25 @@ defmodule EXLA.DefnExprTest do
 
   describe "norm" do
     defn normalize(t), do: t / Nx.norm(t)
+    defn l2_cols(t), do: Nx.norm(t, ord: 2, axes: [0])
+    defn l2_rows(t), do: Nx.norm(t, ord: 2, axes: [1])
+    defn l2_cols_keep_axes(t), do: Nx.norm(t, ord: 2, axes: [0], keep_axes: true)
+    defn l2_rows_keep_axes(t), do: Nx.norm(t, ord: 2, axes: [1], keep_axes: true)
 
     test "works with norm calls" do
       t = Nx.tensor([3.0, 4.0, 0.0])
-      assert compare_tensors!(normalize(t), Nx.tensor([0.6, 0.8, 0.0]))
-      assert compare_tensors!(Nx.norm(normalize(t)), Nx.tensor(1.0))
+      compare_tensors!(normalize(t), Nx.tensor([0.6, 0.8, 0.0]))
+      compare_tensors!(Nx.norm(normalize(t)), Nx.tensor(1.0))
+    end
+
+    test "accepts axes and keep_axes arguments" do
+      t = Nx.tensor([[5.0, 12.0, 0.0], [12.0, 5.0, 0.0]])
+
+      compare_tensors!(l2_cols(t),  Nx.tensor([13.0, 13.0, 0.0]))
+      compare_tensors!(l2_rows(t),  Nx.tensor([13.0, 13.0]))
+
+      compare_tensors!(l2_cols_keep_axes(t),  [13.0, 13.0, 0.0] |> Nx.tensor() |> Nx.reshape({1, 3}))
+      compare_tensors!(l2_rows_keep_axes(t),  [13.0, 13.0] |> Nx.tensor() |> Nx.reshape({2, 1}))
     end
   end
 
--- a/exla/test/exla/defn_expr_test.exs
+++ b/exla/test/exla/defn_expr_test.exs
@@ -2386,6 +2386,17 @@ defmodule EXLA.DefnExprTest do
     end
   end
 
+  describe "norm" do
+    defn normalize(t), do: t / Nx.norm(t)
+
+    test "works with norm calls" do
+      t = Nx.tensor([3.0, 4.0, 0.0])
+      assert compare_tensors!(normalize(t), Nx.tensor([0.6, 0.8, 0.0]))
+
+      assert compare_tensors!(Nx.dot(t, t), Nx.power(Nx.norm(t), 2))
+    end
+  end
+
   # We need to round the floats because of imprecision between platforms
   defp compare_tensors!(
          %{type: {:f, size}, data: %{state: left_data} = lhs} = left,
--- a/exla/test/exla/defn_expr_test.exs
+++ b/exla/test/exla/defn_expr_test.exs
@@ -2393,6 +2393,8 @@ defmodule EXLA.DefnExprTest do
     defn l2_cols_keep_axes(t), do: Nx.norm(t, ord: 2, axes: [0], keep_axes: true)
     defn l2_rows_keep_axes(t), do: Nx.norm(t, ord: 2, axes: [1], keep_axes: true)
 
+    defn non_zero_element_count(t), do: Nx.norm(t, ord: 0)
+
     test "works with norm calls" do
       t = Nx.tensor([3.0, 4.0, 0.0])
       compare_tensors!(normalize(t), Nx.tensor([0.6, 0.8, 0.0]))
@@ -2407,6 +2409,16 @@ defmodule EXLA.DefnExprTest do
 
       compare_tensors!(l2_cols_keep_axes(t),  [13.0, 13.0, 0.0] |> Nx.tensor() |> Nx.reshape({1, 3}))
       compare_tensors!(l2_rows_keep_axes(t),  [13.0, 13.0] |> Nx.tensor() |> Nx.reshape({2, 1}))
+    end
+
+    test "works with ord: 0" do
+      assert_raise RuntimeError, "ord 0 not implemented for 2-D tensor.", fn ->
+        non_zero_element_count(Nx.tensor([[0, 1, -1, 2]]))
+      end
+
+      %{type: type} = result = non_zero_element_count(Nx.tensor([0, 1, -1, 2]))
+
+      compare_tensors!(result, Nx.tensor(3, type: type))
     end
   end
 
--- a/exla/test/exla/defn_expr_test.exs
+++ b/exla/test/exla/defn_expr_test.exs
@@ -2412,7 +2412,7 @@ defmodule EXLA.DefnExprTest do
     end
 
     test "works with ord: 0" do
-      assert_raise RuntimeError, "ord 0 not implemented for 2-D tensor.", fn ->
+      assert_raise ArgumentError, "expected 1-D tensor for ord: 0, got a 2-D tensor", fn ->
         non_zero_element_count(Nx.tensor([[0, 1, -1, 2]]))
       end
 
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -6737,6 +6737,15 @@ defmodule Nx do
       5.0
     >
 
+    iex> Nx.norm(Nx.tensor([[3, 4], [0, -4]]), axes: [1], keep_axes: true)
+    #Nx.Tensor<
+      f64[2][1]
+      [
+        [5.0],
+        [4.0]
+      ]
+    >
+
   ### Caveats
 
   For big values of p, f64 rounding errors come into play
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -6855,8 +6855,9 @@ defmodule Nx do
     raise ArgumentError, "invalid :ord for 2-D tensor, got: #{inspect(ord)}"
   end
 
-  defp norm_integer(t, ord, opts) when is_integer(ord) do
-    inv_ord = tensor(1 / ord)
+  defp norm_integer(%{type: type} = t, ord, opts) when is_integer(ord) do
+    output_type = Nx.Type.to_floating(type)
+    inv_ord = tensor(1 / ord, type: output_type)
 
     # We extract this result to a variable because it's used both for
     # getting the normalization coefficient and for the main pipe chain
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -6663,6 +6663,26 @@ defmodule Nx do
 
   For the 0-norm, the norm is the number of non-zero elements in the tensor.
 
+  ### Options
+
+  - `:axes` - defines the axes upon which the norm will be calculated. Default: `nil`.
+  - `:keep_axes` - defines if the resulting tensor should have the same dimensions as the input. Default: `false`.
+  - `:ord` - defines which norm will be calculated according to the table below. Default: `2`
+
+  | ord          | 2-D                                       | 1-D                               |
+  | ------------ | ----------------------------------------- | --------------------------------- |
+  | `nil`        | Frobenius norm                            | 2-norm                            |
+  | `:frobenius` | Frobenius norm                            | -                                 |
+  | `:nuclear`   | Nuclear norm (not implemented)            | -                                 |
+  | `:inf`       | `max(sum(abs(x), axes: 1))`               | `max(abs(x))`                     |
+  | `:neg_inf`   | `min(sum(abs(x), axes: 1))`               | `min(abs(x))`                     |
+  | 0            | -                                         | Number of non-zero elements       |
+  | 1            | `max(sum(abs(x), axes: 0))`               | as below                          |
+  | -1           | `min(sum(abs(x), axes: 0))`               | as below                          |
+  | 2            | 2-norm                                    | as below                          |
+  | -2           | smallest singular value (not implemented) | as below                          |
+  | other        | -                                         | power(sum(power(abs(x), p)), 1/p) |
+
   ### Examples
 
     iex> Nx.norm(Nx.tensor([3, 4]))
@@ -6695,11 +6715,11 @@ defmodule Nx do
       2
     >
 
-    iex> Nx.norm(Nx.tensor([3, 4]), ord: :fro)
+    iex> Nx.norm(Nx.tensor([3, 4]), ord: :frobenius)
     ** (RuntimeError) expected a 2-D tensor. Got a 1-D tensor.
 
-    iex> Nx.norm(Nx.tensor([3, 4]), ord: :nuc)
-    ** (RuntimeError) expected a 2-D tensor. Got a 1-D tensor.
+    iex> Nx.norm(Nx.tensor([3, 4]), ord: :nuclear)
+    ** (RuntimeError) nuclear norm not implemented yet.
 
     iex> Nx.norm(Nx.tensor([[3, -1], [2, -4]]), ord: -1)
     #Nx.Tensor<
@@ -6725,7 +6745,7 @@ defmodule Nx do
       6
     >
 
-    iex> Nx.norm(Nx.tensor([[3, 0], [0, -4]]), ord: :fro)
+    iex> Nx.norm(Nx.tensor([[3, 0], [0, -4]]), ord: :frobenius)
     #Nx.Tensor<
       f64
       5.0
@@ -6782,17 +6802,14 @@ defmodule Nx do
       {nil, _} ->
         2
 
-      {:fro, 2} ->
+      {:frobenius, 2} ->
         2
 
-      {:fro, _} ->
+      {:frobenius, _} ->
         raise "expected a 2-D tensor. Got a #{dims}-D tensor."
 
-      {:nuc, 2} ->
-        :nuc
-
-      {:nuc, _} ->
-        raise "expected a 2-D tensor. Got a #{dims}-D tensor."
+      {:nuclear, _} ->
+        raise "nuclear norm not implemented yet."
 
       {ord, 2} when ord in [:inf, :neg_inf] ->
         ord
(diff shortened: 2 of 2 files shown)


seanmor5:
Hey thanks for the PR!

A few notes before we can merge:

- We generally try to adhere to NumPy semantics where it makes sense, I would take a look at [np.linalg.norm](https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html). I think in this case it's okay to adhere to their semantics. I would use `:ord`, `:keep_axes`, and `:axes` as optional keywords in `opts`. Sum supports both `keep_axes` and `axes` as well so you should be able to just plug those in there. You'll also want to take a look at the usage of `assert_keys`.

- You should assert upfront that the input `t` is a tensor, it will make the call stack clearer if somebody passes invalid input to this function, rather than it getting caught down the line:

```elixir
def p_norm(tensor, opts \\ []) do
  %T{type: type} = tensor = tensor!(tensor)
   ...
```
- To go along with the discussion of numpy semantics, you'll want to do some shape checks on the inputs. I believe numpy only supports 1-D/2-D inputs depending on `:ord`, so you'll need to do that validation, unless you think there's a reason not to.

- Be careful with automatic promotion to `f64`, it's not supported on all accelerators (TPU), and it's generally really slow on both CPU/GPU. Instead I would do something like:

```
def p_norm(tensor, opts \\ []) do
  %T{type: type} = tensor = tensor!(tensor)
  output_type = Nx.Type.to_floating(type)
  ...
  |> Nx.as_type(output_type)
end
```

That will ensure you're not aggressively upcasting to `f64`.

- This is a good implementation, and will make p_norm available automatically for EXLA and other compilers. If you're up for it, you can also write some tests for the EXLA compiler in `exla/test/exla/defn_api_test.exs`. There are lots of examples there. The tests should pass without any additions to EXLA.

polvalente:
@seanmor5 sure, I'll try to adapt to the NumPy interface :)

At first, I don't see a reason to restrict dimensions, since my implementation basically returns a scalar metric. However, upon defining `axes` and such this might change. I'll keep this in mind!

Thanks for the tip on `f64`. My idea was to make all norms return the same datatype, but returning the same type as the input for `0-norm` makes sense!

AlexJuca:
@polvalente I see you have a PR for this already. Looks like I will leave it to you to handle. It's a great start.

polvalente:
@seanmor5 I think I've done everything you asked! I had some trouble supporting the `axes` and `keep_axes` options, but I think I got it. The only problem that I couldn't solve is the implementation for `norm(t, -2)` when `t` is a 2-D tensor, because the numpy docs say it returns the minimum singular value for the matrix.

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,217 @@ defmodule Nx do
...
+      s ->
+        raise "expected 1-D or 2-D tensor. Received a tensor with shape #{inspect(s)} instead."
+    end
+
+    rank = rank(t)
+    ord = get_norm_ord!(input_opts, rank)
+
+    default_axes_opts = [axes: nil, keep_axes: false]
+
+    axes_opts = Keyword.merge(default_axes_opts, Keyword.take(input_opts, [:axes, :keep_axes]))
```
josevalim, line 6773: There is a function called `assert_keys!` that you should use to validate the inputs.

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,217 @@ defmodule Nx do
...
+      f64[2][1]
+      [
+        [5.0],
+        [4.0]
+      ]
+    >
+
+  ### Caveats
+
+  For big values of p, f64 rounding errors come into play
```
josevalim, line 6751: Could we use a numerically stable implementation? https://timvieira.github.io/blog/post/2014/11/10/numerically-stable-p-norms/

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,217 @@ defmodule Nx do
     impl!(tensor).cholesky(out, tensor)
   end
 
+  @doc """
+  Calculates the p-norm of a tensor.
+
+  For the 0-norm, the norm is the number of non-zero elements in the tensor.
+
```
josevalim, line 6665: Please add a description of the supported options. Note that if `:fro` and `:nuc` are abbreviations, prefer to use full words instead. :)

nx/lib/nx.ex
josevalim, line 6773: Also, those are default values, so this should be fine:

```suggestion
    axes_opts = Keyword.take(input_opts, [:axes, :keep_axes])
```

josevalim:
Looks great, thank you! :heart: I have added some comments too! Btw, should we call this norm or p_norm?

AlexJuca:
@josevalim @polvalente  I would suggest following the numpy semantics and have it be the more generic `norm()`.

nx/lib/nx.ex
polvalente: I'll take a further look into it

nx/lib/nx.ex
polvalente: That's a great suggestion! I'll try make the proper adjustments for this.
Do we have a function that yields the maximum value for each datatype?
If not, maybe we should leave this as a technical debt to avoid hardcoding any values. I'm thinking of something like `Nx.max_value(1.0)` yielding `1.0e300`

nx/lib/nx.ex
polvalente: Sure! I forgot to update the docs after adding said options. I'll improve naming as well.

polvalente:
> @josevalim @polvalente I would suggest following the numpy semantics and have it be the more generic norm.

I'm with @AlexJuca. `p_norm` would imply only the base case of `power(sum(power(abs(x), p)), 1/p)`, whereas `norm` includes the special cases documented here: https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html

By the way, @josevalim is there a good way to draw tables in docs? Maybe I should also port the norm definitions table itself to our doc.

nx/lib/nx.ex
josevalim: Yes, there is `Nx.Type.min_value_binary` which you can put into a tensor by calling `Nx.from_binary`.

polvalente:
@seanmor5 @josevalim @AlexJuca I've made the latest changes with better docs. I only have some questions on how to deal with `ord: -2` for 2-D Tensors and with `ord: :nuc`, both of which depend on SVD to work.

AlexJuca:
@polvalente both cases can always be deferred for later. You can raise saying that the operation is not supported yet. What you think?

polvalente:
> @polvalente both cases can always be deferred for later. Seeing that they are dependent on SVD you can raise saying that the operation is not supported yet. What you think?

Sounds like a good plan!

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,251 @@ defmodule Nx do
...
+      {ord, _} ->
+        ord
+    end
+  end
+
+  defp do_p_norm(t, 0, 1, opts) do
+    t
+    |> equal(0)
+    |> logical_not()
+    |> sum(opts)
```
seanmor5, line 6849: Do all of NumPy's cases return a float? If not, ignore this :)

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,251 @@ defmodule Nx do
...
+    |> function.()
+  end
+
+  defp do_p_norm(t, ord, 1, axes_opts) when ord in [:inf, :neg_inf] do
+    function = if ord == :inf, do: &reduce_max/2, else: &reduce_min/2
+
+    t
+    |> Nx.abs()
+    |> function.(axes_opts)
+  end
```
seanmor5, line 6873: I think you can combine these two:

```elixir
defp do_p_norm(t, ord, rank, opts) when ord in [:inf, :neg_inf] and rank == 1 or rank == 2 do
  function = if ord == :inf, do: &reduce_max/1, else: &reduce_min/1
  reduction = if rank == 1, do: & &1, else: &sum(&1, [keep_axes: opts[:keep_axes], axes: [1]])

  t
  |> abs()
  |> reduction.()
  |> function.()
end
```

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,251 @@ defmodule Nx do
...
+        &reduce_max/2
+      else
+        &reduce_min/2
+      end
+
+    t
+    |> Nx.abs()
+    |> sum(axes: [0])
+    |> function.(Keyword.take(opts, [:keep_axes]))
+  end
```
seanmor5, line 6887: I think you might be able to combine this with the case above as well :)

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,251 @@ defmodule Nx do
...
+
+  defp do_p_norm(_t, ord, 2, _opts) when ord not in -1..2 do
+    raise "invalid ord for 2-D tensor. Got: #{inspect(ord)}"
+  end
+
+  defp do_p_norm(%{type: type} = t, ord, _, opts) do
+    inv_ord = divide(1, ord)
```
seanmor5, line 6894: I would be careful using `divide` here because it might cause unnecessary type promotion later on. Instead I would do:

```elixir
inv_ord = tensor(1 / ord, type: type)
```

Also, is it possible `ord` is 0? Maybe it was caught earlier, but if not, be careful of division by 0 :)

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,251 @@ defmodule Nx do
...
+
+    # This coefficient is introduced for better numerical stability
+    # The idea is that by dividing the tensor by it, large values of
+    # tensor entries and large values of p are reduced, which in turn
+    # avoids numerical overflow.
+    {:ok, numerical_stability_coefficient} =
+      type
+      |> Nx.Type.max_value_binary()
+      |> Nx.from_binary(type)
+      |> Nx.Tensor.fetch(0)
```
seanmor5, line 6904: You don't need the fetch here, the max value scalar will work fine in subsequent operations!

seanmor5 (commented):
Hey @polvalente great work! I understand you need SVD to get this finished. For now, if you're up for it, you can try to implement a somewhat naive version of SVD and I can finish the rest for you on the EXLA side. If not, raising is fine for now and I will just add a note to the issue tracker.

polvalente:
> Hey @polvalente great work! I understand you need SVD to get this finished. For now, if you're up for it, you can try to implement a somewhat naive version of SVD and I can finish the rest for you on the EXLA side. If not, raising is fine for now and I will just add a note to the issue tracker.

Sure! Do you want me to add the naive SVD in this PR, or should we merge this first and I add it in a subsequent one? I lean towards this second option because L2-norm could help with QR decomposition and other things like that.

nx/lib/nx.ex
polvalente: The intention here was to avoid having a [1]-shaped Tensor  instead of a pure scalar (altough they are almost the same) in some of the test results.

nx/lib/nx.ex
polvalente: There are special function clauses for `ord: 0`. The numpy docs specified it in a similar manner to what I had in the original draft: returns the count of non-zero elements in the tensor

nx/lib/nx.ex
polvalente: ooh this seems nice :)
I'll look into this!

seanmor5:
@polvalente You can go ahead and get this finished up and worry about SVD later! I would raise on the unsupported cases for now then, and once this is merged I'll add a note to the issues tracker

exla/test/exla/defn_expr_test.exs
```diff
@@ -2386,6 +2386,16 @@ defmodule EXLA.DefnExprTest do
     end
   end
 
+  describe "norm" do
+    defn normalize(t), do: t / Nx.norm(t)
+
+    test "works with norm calls" do
+      t = Nx.tensor([3.0, 4.0, 0.0])
+      assert compare_tensors!(normalize(t), Nx.tensor([0.6, 0.8, 0.0]))
+      assert compare_tensors!(Nx.norm(normalize(t)), Nx.tensor(1.0))
```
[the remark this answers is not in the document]
polvalente: Sure!

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,251 @@ defmodule Nx do
...
+    #Nx.Tensor<
+      f64[2][1]
+      [
+        [5.0],
+        [4.0]
+      ]
+    >
+
+    iex> Nx.norm(Nx.tensor([0, 1]), my_opt: 1)
+    ** (RuntimeError) unknown key :my_opt in [my_opt: 1], expected one of [:ord, :axes, :keep_axes]
```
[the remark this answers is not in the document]
polvalente: Sounds good. Should I really keep the error cases as doctests/examples? Or maybe add them to nx_test.exs?

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,251 @@ defmodule Nx do
...
+
+  ### Caveats
+
+  For big values of p, f64 rounding errors come into play
+  """
+
+  def norm(t_raw, input_opts \\ []) when is_list(input_opts) do
```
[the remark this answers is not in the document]
polvalente: no problems!

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,251 @@ defmodule Nx do
...
+    t = tensor!(t_raw)
+    assert_keys!(input_opts, [:ord, :axes, :keep_axes])
+
+    default_axes_opts = [axes: nil, keep_axes: false, ord: nil]
+    opts = Keyword.merge(default_axes_opts, input_opts)
```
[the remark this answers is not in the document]
polvalente: I tend to use this approach so all defaults are specified in a single place. Do you feel the need for this change?

nx/lib/nx.ex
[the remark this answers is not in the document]
seanmor5: I lean towards keeping at least 1 error case in the doctests, just for documentation purposes!

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,251 @@ defmodule Nx do
...
+      s ->
+        raise "expected 1-D or 2-D tensor. Received a tensor with shape #{inspect(s)} instead."
+    end
+
+    rank = rank(t)
+    ord = get_norm_ord!(opts, rank)
+
+    axes_opts = Keyword.take(opts, [:axes, :keep_axes])
+
+    do_p_norm(t, ord, rank, axes_opts)
```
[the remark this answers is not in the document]
polvalente: I feared this would be an anti-pattern for Nx, since the Tensor is `inspect`ed as an opaque struct. Since this is ok, I'll refactor this as well!

nx/lib/nx.ex
polvalente: Some of them return 1-D tensors for 2-D inputs

nx/lib/nx.ex
polvalente: I ran into some problems using `type: type` because `type` can represent an integer type. I've refactored this line to `tensor(1/p)` instead

nx/lib/nx.ex
polvalente: This one is possible, but I think it removes some readability because it's only defined for matrices

nx/lib/nx.ex
polvalente: I've simplified this __a lot__, so we can disregard my comment :)

nx/lib/nx.ex
```diff
@@ -6769,6 +6769,12 @@ defmodule Nx do
     iex> Nx.norm(Nx.tensor([0, 1]), my_opt: 1)
     ** (RuntimeError) unknown key :my_opt in [my_opt: 1], expected one of [:ord, :axes, :keep_axes]
 
+    iex> Nx.norm(Nx.tensor([[0], [1]]), ord: :nuclear)
+    ** (RuntimeError) nuclear norm not implemented yet.
+
+    iex> Nx.norm(Nx.tensor([[0], [1]]), ord: -2)
+    ** (RuntimeError) ord: -2 for 2-D tensor not implemented yet.
```
josevalim, line 6779: Nitpick: don't end error messages with dot. :)

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,218 @@ defmodule Nx do
...
+  | 0            | -                                         | Number of non-zero elements       |
+  | 1            | `max(sum(abs(x), axes: 0))`               | as below                          |
+  | -1           | `min(sum(abs(x), axes: 0))`               | as below                          |
+  | 2            | 2-norm                                    | as below                          |
+  | -2           | smallest singular value (not implemented) | as below                          |
+  | other        | -                                         | power(sum(power(abs(x), p)), 1/p) |
+
+  ## Examples
+
+    ### Vector norms
```
josevalim, line 6688:
```suggestion
  ### Vector norms
```

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,218 @@ defmodule Nx do
...
+
+    iex> Nx.norm(Nx.tensor([3, 4]))
```
josevalim, line 6690: Make sure the examples are indented with 4 spaces from the heredoc beginning indentation, as that's required by Markdown. Otherwise they won't be highlighted as code snippets. TL;DR - indent all examples by two spaces. :)

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,218 @@ defmodule Nx do
...
+    %{shape: s} = t = tensor!(tensor)
+    assert_keys!(opts, [:ord, :axes, :keep_axes])
+
+    default_axes_opts = [axes: nil, keep_axes: false, ord: nil]
+    opts = Keyword.merge(default_axes_opts, opts)
+
+    rank = rank(t)
+
+    unless rank == 1 or rank == 2,
+      do: raise("expected 1-D or 2-D tensor, got tensor with shape #{inspect(s)}")
```
josevalim, line 6792:
```suggestion
      do: raise(ArgumentError, "expected 1-D or 2-D tensor, got tensor with shape #{inspect(s)}")
```

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,218 @@ defmodule Nx do
...
+
+    do_p_norm(t, ord, axes_opts)
+  end
+
+  defp do_p_norm(t, nil, opts), do: do_p_norm(t, 2, opts)
+
+  defp do_p_norm(_t, :nuclear, _opts), do: raise("nuclear norm not implemented yet.")
+
+  defp do_p_norm(%{shape: {_}}, :frobenius, _opts),
+    do: raise("expected a 2-D tensor. Got a 1-D tensor.")
```
josevalim, line 6804:
```suggestion
    do: raise(ArgumentError, ":frobenius expects a 2-D tensor, got a 1-D tensor")
```

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,218 @@ defmodule Nx do
...
+  defp do_p_norm(%{shape: {_, _}} = t, :frobenius, opts), do: do_p_norm(t, 2, opts)
+
+  defp do_p_norm(%{shape: {_}} = t, 0, opts) do
+    t
+    |> not_equal(0)
+    |> sum(opts)
+  end
+
+  defp do_p_norm(%{shape: {_, _}}, 0, _opts) do
+    raise "expected 1-D tensor for ord: 0. Got: 2-D tensor."
```
josevalim, line 6815:
```suggestion
    raise ArgumentError, "expected 1-D tensor for ord: 0, got a 2-D tensor"
```

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,218 @@ defmodule Nx do
...
+      do: raise("expected 1-D or 2-D tensor, got tensor with shape #{inspect(s)}")
+
+    {ord, axes_opts} = Keyword.pop(opts, :ord)
+
+    do_p_norm(t, ord, axes_opts)
+  end
+
+  defp do_p_norm(t, nil, opts), do: do_p_norm(t, 2, opts)
+
+  defp do_p_norm(_t, :nuclear, _opts), do: raise("nuclear norm not implemented yet.")
```
josevalim, line 6801: If we don't implement it yet, maybe we should simply not support this option. :)

josevalim, line 6801:
```suggestion
  defp do_p_norm(_t, :nuclear, _opts), do: raise("nuclear norm not implemented yet")
```

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,218 @@ defmodule Nx do
...
+    |> sum(axes: [0])
+    |> function.(Keyword.take(opts, [:keep_axes]))
+  end
+
+  defp do_p_norm(%{shape: {_, _}}, -2, _opts) do
+    raise "ord: -2 for 2-D tensor not implemented yet."
+  end
+
+  defp do_p_norm(%{shape: {_, _}}, ord, _opts) when ord not in -1..2 or ord == 0 do
+    raise "invalid ord for 2-D tensor. Got: #{inspect(ord)}"
```
josevalim, line 6849:
```suggestion
    raise ArgumentError, "invalid :ord for 2-D tensor, got: #{inspect(ord)}"
```

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,218 @@ defmodule Nx do
...
+      end
+
+    t
+    |> Nx.abs()
+    |> sum(axes: [0])
+    |> function.(Keyword.take(opts, [:keep_axes]))
+  end
+
+  defp do_p_norm(%{shape: {_, _}}, -2, _opts) do
+    raise "ord: -2 for 2-D tensor not implemented yet."
```
josevalim, line 6845:
```suggestion
    raise "ord: -2 for 2-D tensor not implemented yet"
```

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,218 @@ defmodule Nx do
...
+    opts = Keyword.merge(default_axes_opts, opts)
+
+    rank = rank(t)
+
+    unless rank == 1 or rank == 2,
+      do: raise("expected 1-D or 2-D tensor, got tensor with shape #{inspect(s)}")
+
+    {ord, axes_opts} = Keyword.pop(opts, :ord)
+
+    do_p_norm(t, ord, axes_opts)
```
josevalim, line 6796: I think we can slightly cleanup the code by handling the major ord branches here:

```
case ord do
  :frobinius -> norm_frobinius(t, axes_opts)
  :nuclear -> norm_nuclear(t, axes_opts)
  _ when ord in [:inf, :neg_inf] -> norm_inf(t, ord, axes_opts)
  _ when is_integer(ord) -> norm_ord(t, ord, axes_opts)
  _ -> raise ArgumentError, "unknown ord #{inspect(ord)}"
end
```

josevalim (commented):
I have added some minor nitpicks to the error messages to make them consistent. I have also made a proposal to reduce the branching inside do_p_norm. We are almost there I think, thank you!

nx/lib/nx.ex
```diff
@@ -6658,6 +6658,224 @@ defmodule Nx do
...
+  defp norm_integer(%{shape: {_, _}}, -2, _opts) do
+    raise "ord: -2 for 2-D tensor not implemented yet"
+  end
+
+  defp norm_integer(%{shape: {_, _}}, ord, _opts) when ord not in -1..2 or ord == 0 do
+    raise ArgumentError, "invalid :ord for 2-D tensor, got: #{inspect(ord)}"
+  end
+
+  defp norm_integer(t, ord, opts) when is_integer(ord) do
+    inv_ord = tensor(1 / ord)
```
seanmor5, line 6859: Sorry I lead you astray last time, this will promote to `f64` rather quickly which you want to avoid. It can be fixed on the user end, but in general it's better to make sure we're not doing it unnecessarily if possible. This is what you want instead: 

```suggestion
  defp norm_integer(%{type: type} = t, ord, opts) when is_integer(ord) do
    output_type = Nx.Type.to_floating(type)
    inv_ord = tensor(1 / ord, type: output_type)
```

seanmor5 (approved):
Just the 1 comment on type promotion, once that's fixed this looks good to me :)

nx/lib/nx.ex
polvalente: No worries!

PR #228, opened 2021-02-18
Add math ops

seanmor5:
Just missing `erf_inv` implementation in Elixir


josevalim (commented):
LGTM! Don't forget to add the new functions to the list of pending grads!

PR #233, opened 2021-02-18
Improve defn.ex docs

AlexJuca:
no function `Nx.multipy` exists, doc should updated to use `Nx.multiply` instead


josevalim:

PR #235, opened 2021-02-19
no type NX.Tensor.axes() exists

qrede:
no type NX.Tensor.axes() exits, should be Nx.Tensor.axes() instead


josevalim:

Issue #161, opened 2021-01-22
Add missing gradients


seanmor5:
Currently missing gradients for these functions:

- [x] Product
- [ ] Scatter Window Max
- [ ] Scatter Window Min
- [x] Scatter Add
- [x] LU Decomposition
- [x] Triangular Solve
- [x] SVD

polvalente:
https://github.com/tensorflow/tensorflow/pull/45475
https://github.com/tensorflow/tensorflow/issues/6504

This issue and Pull Request are related to QR grad. Might be relevant for who tackles this later.
In the PR comments there's a paper related to backpropagation of QR and LQ factorizations.

Benjamin-Philip:
I believe that gradients for some of these functions have been implemented. Could you update the list @seanmor5 ?

josevalim:
@Benjamin-Philip i recently updated the list and it should be good. If anything is missing, please let me know :)

Benjamin-Philip:
@josevalim, you're right! I remember seeing some completed gradients a few days ago. I did not realise that it was updated. My bad 🙂.

polvalente:
I'm starting work on triangular solve gradient

PR #237, opened 2021-02-19
gradient for tan(x)

t-rutten:
These changes add the gradient for `Nx.tan/1` per #161.

### DIFF
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -416,6 +416,11 @@ defmodule Nx.Defn.Grad do
     to_grad(x, g, cache)
   end
 
+  defp grad(:tan, [x], _ans, g, cache) do
+    g = Nx.divide(g, Nx.multiply(Nx.cos(x), Nx.cos(x)))
+    to_grad(x, g, cache)
+  end
+
   defp grad(:reduce, _, _, _, _) do
     raise ArgumentError, """
     cannot compute gradient for Nx.reduce/4.
--- a/nx/test/nx/defn/grad_test.exs
+++ b/nx/test/nx/defn/grad_test.exs
@@ -468,6 +468,8 @@ defmodule Nx.Defn.GradTest do
 
     test "computes gradient" do
       for _ <- 1..100 do
+          # check_grads!/4 fails for values close to the asymptotes
+          # of tan's gradient, so we select t to avoid them.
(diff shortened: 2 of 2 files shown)


nx/test/nx/defn/grad_test.exs
```diff
@@ -463,6 +463,19 @@ defmodule Nx.Defn.GradTest do
...
 
+  describe "tan" do
+    defn grad_tan(t), do: grad(t, Nx.tan(t))
+
+    test "computes gradient" do
+      for _ <- 1..100 do
+          multiplier = Nx.random_uniform({}, 0, 10, type: {:u, 32})
+          offset = Nx.random_uniform({}, -1.5, 1.5, type: {:f, 64})
+          t = 3.14159 |> Nx.multiply(multiplier) |> Nx.add(offset)
+          check_grads!(&Nx.tan/1, &grad_tan/1, t)
```
t-rutten, line 474: The gradient of tan(t) looks like this:
tan
`check_grads!/4` fails for values close to the asymptotes, so we select `t` to avoid them.

This is why we test tan's gradient separately from the `tuples` above, where `t` is selected by
`t = Nx.random_uniform({}, 0.1, 10.0, type: {:f, 64})`

nx/lib/nx/defn/grad.ex
```diff
@@ -416,6 +416,11 @@ defmodule Nx.Defn.Grad do
     to_grad(x, g, cache)
   end
 
+  defp grad(:tan, [x], _ans, g, cache) do
+    g = Nx.divide(g, Nx.multiply(Nx.cos(x), Nx.cos(x)))
```
josevalim, line 420:
```suggestion
    g = Nx.divide(g, Nx.power(Nx.cos(x), 2))
```

?

nx/test/nx/defn/grad_test.exs
josevalim: Looks great! Can you please add these notes as a comment to the test above? Thank you!

nx/lib/nx/defn/grad.ex
t-rutten: That works!

nx/test/nx/defn/grad_test.exs
t-rutten: Added!

josevalim:

josevalim:
Thank you, contribution around the gradients are very appreciated. :)

PR #238, opened 2021-02-19
Add dot with batching - dot/6

### LINKED ISSUE #182: Make Nx.dot/4 aware of batch dimensions
This is related to #174 so perhaps it's more appropriate to move this into the discussion in that issue.
 
Right now, the dot product is only aware of contracting dimensions. One example of something we can't support right now is a bilinear transformation like in: https://pytorch.org/docs/stable/nn.functional.html#bilinear

Given input shapes: `{a, b}, {z, b, c}, {a, c}`, for a bilinear transformation the output shape should be: `{a, z}` given by:

```elixir
input1
|> Nx.dot(weight)
|> Nx.dot(input2)
```

However, even if we configure the contracting dimensions correctly, we'll end up with shape `{a, z, a}`. Instead what we need is to treat the first dimension as a batch dimension so the final dot product between shapes `{a, z, c}` and `{a, c}` is really a batched dot product between shapes `{z, c}` and `{c}` resulting in an output of `{a, z}`

elbow-jason:
This PR resolves #182 by adding `Nx.dot/6` which is a dot operation that supports batching. This PR also changes the `Nx.Backend`'s  `dot/5` callback to `dot/7` in order to allow batching axis info to be passed to a Backend.

This is a WIP at the moment, but I wanted some feedback before it got too much further along.

`dot/7` is implemented for `Nx.BinaryBackend`. Tests are passing, and I have verified the outputs of batching the right tensor, batching the left tensor, and batching both tensors against Jax's `lax.dot_general` functions.

After some discussion with @seanmor5 we decided to only support axis 0 for dot batching for the time being.

### TODOs

  - [ ] add documentation with examples to `Nx.dot/6`
  - [ ] add checks for scalar tensors to `Nx.dot/6` and tests
  - [ ] implement and test `dot/7` callback for EXLA
  - [ ] make sure `grad` works with batched tensors and add tests
  
### Questions

  + I added lots of `defp` helper functions in the Nx module which, other than the code from this PR, is mostly public functions and very few helpers. I noticed a `## Helpers` section at the bottom. Should I move the helpers down there? Or to a different module perhaps?
  
  + in `Nx.BinaryBackend.dot/7` I make a call to `Nx.to_batched_list` which felt dirty to me. Should I change this to call BinaryBackend's to_batched_list instead?

  + Currently, the batching axis into for the `dot/7` callback is type `0 | nil`. Using a single integer works for now, but is not very future proof. I think it might be better to use a list where an empty list would mean "no batch" instead of `nil`. Additionally, the Jax implementation `lax.dot_general` uses a tuple of tuples for their batching info which is also closer to a batch-axes-list implementation. Thoughts?


nx/lib/nx.ex
```diff
@@ -5637,7 +5637,7 @@ defmodule Nx do
       {_, 0} -> multiply(t1, t2)
       {n, 1} -> dot(t1, [n - 1], t2, [0])
       {1, m} -> dot(t1, [0], t2, [m - 2])
-      {n, m} when n >= 2 and m >= 2 -> dot(t1, [n - 1], t2, [m - 2])
+      {n, m} when n >= 2 and m >= 2 -> dot(t1, [n - 1], nil, t2, [m - 2], nil)
```
josevalim, line 5640: Please use an empty list instead of nil as that will give us more space for maneuver in the future.

nx/lib/nx/binary_backend.ex
```diff
@@ -427,7 +427,54 @@ defmodule Nx.BinaryBackend do
...
+
+    batch_out = %{out | shape: batch_shape, names: batch_names}
+
+    dotted_batches =
+      left_tensors
+      |> Stream.zip(right_tensors)
+      |> Enum.map(fn {left_batch, right_batch} ->
+        dot4(batch_out, left_batch, axes1, right_batch, axes2)
+      end)
+    concatenate(out, dotted_batches, 0)
```
josevalim, line 447: We should try to do operations on binaries as much as possible. The idea here is that the batches on both tensors will have the same dimension. So you can do something like this:

```
%{type: {_, left_size}} = left_tensor
%{type: {_, right_size}} = right_tensor

left_batch = div(Nx.size(left_tensor), batch) * left_size
right_batch = div(Nx.size(right_tensor), batch) * right_size

for i <- 0..batch-1 do
  consumed = i * left_batch
  <<_::size(consumed)-bitstring, left_inner::size(left_batch)-bitstring, _::bitstring>> = left_data
  consumed = i * right_batch
  <<_::size(consumed)-bitstring, right_inner::size(right_batch)-bitstring, _::bitstring>> = right_data
  dot(left_inner, left_axis, right_inner, right_axis)
end
```

Where the dot at the end is the dot operation that we have today. In other words, we are traversing the batch dimensions together (as if we were zipping them) and dotting them as you go. It is similar to what you are doing, but it should be more efficient because we are traversing the binary only once and not creating streams and so on.

Once you get that working, supporting multiple dimensions will be a better of changing the for comprehension to traverse on batch1*batch2 elements at the same time.  But we can cross that bridge once we get there!

elbow-jason:
Closing for now. :-)

PR #246, opened 2021-02-19
Fix Windows compilation failure due to lack of some :math functions

### LINKED ISSUE #239: Windows BEAM missing some math functions in Nx.shared
Windows BEAM does not have the following OTP math module functions: asinh, acosh, atanh, erf, and erfc.
This lack of functions causes installation failure of Nx on Windows BEAM.
Corresponding source of Nx:
https://github.com/elixir-nx/nx/blob/560ec29ce38a583fedacfe21c0c74aa7bfa4d7dc/nx/lib/nx/shared.ex#L137-L166

See maint-23 Branch of Erlang/OTP:

* https://github.com/erlang/otp/blob/maint-23/erts/emulator/beam/erl_math.c#L146-L153
* https://github.com/erlang/otp/blob/maint-23/erts/emulator/sys/win32/erl_win_sys.h#L74-L82

There might be many ways to work around with this, but I guess it's up to the Nx devs.

BTW kudos to the Nx and Exla development! 👍

elbow-jason:
This PR adds a check for potentially unsupported/non-existent `:math` functions required by Nx unary operations at compile-time that are keeping Nx from compiling on some (all?) Windows environments.

Hopefully this PR fixes #239, but I have not tested it on Windows.

I did, however, change the `math_arity` of one of the unary ops and was still able to compile and run test and a lot of them raised the new error at runtime.

@jj1bdx Could you see if this branch fixes your issue and respond in this PR?

cc @seanmor5


elbow-jason:
```elixir
iex(1)> function_exported?(:math, :acosh, 1)
true
iex(2)> :math.acosh(0)
** (UndefinedFunctionError) function :math.acosh/1 is undefined or private. Did you mean one of:
      * acosh/1
    (stdlib 3.14) :math.acosh(0)
```

What is going on with the `:math` module?! I feel like I'm taking crazy pills.

seanmor5:
Tested on Windows and fixes all compilation issues. Some EXLA tests will fail because of the lack of an implementation for some of these methods.

jj1bdx:
@seanmor5 I don't have the Windows testing environment right now, but the patch looks good to me! 👍

maccam912:
I had the acosh error that made me start following the related issue, and can confirm that this PR does indeed fix the problem for me.

josevalim:
Thanks for the PR @elbow-jason!

I would like to propose a different approach, which is to emit different formulas if those functions are not available. Especially because it is likely that those operations are available on other backends (XLA and  future ones), even on Windows, but they cannot be implemented efficiently in the Elixir ones.

So what do you think about something like this. Instead of:

    arctanh: {"inverse hyperbolic tangent", quote(do: :math.atanh(var!(x)))},

We could do:

    arctanh: {"inverse hyperbolic tangent", arctanh_formula()},

where:

    defp arctanh_formula() do
      if Code.ensure_loaded?(:math) and function_exported?(:math, :atanh, 1) do
        quote(do: :math.atanh(var!(x)))
      else
        quote(do: :math.log((1 + var!(x)) / (1 - var!(x))) / 2)
      end
    end

This way it will still work on Windows, although a more inefficient implementation.

elbow-jason:
That's a great idea. The only issue is that if the `:math` function is not supported, it is still defined in the `:math` module. When called it raises `UndefinedFunctionError`. Super confusing.

But if we substitute `math_func_exported?/2` for `function_exported?/3` it should work wonderfully.

elbow-jason:
Ah ha! Looking some more at #239 I believe this line is the source of the `UndefinedFunctionError`.

https://github.com/erlang/otp/blob/93caf7975a293bfd842f2c77ccab3bb8a4660363/erts/emulator/beam/erl_math.c#L149

elbow-jason:
Add fallback formulae:

  - [ ] asinh
  - [ ] acosh
  - [ ] atanh
  - [ ] erf
  - [ ] erfc

josevalim:
@elbow-jason oh, I see. So call the function with `0` and if it raises go with the custom implementation. :)

elbow-jason:
Phew! That erf_fallback was a killer.

nx/lib/nx/shared.ex
```diff
@@ -155,16 +155,110 @@ defmodule Nx.Shared do
...
+      0.17087277
+      |> muladd(t, -0.82215223)
+      |> muladd(t, 1.48851587)
+      |> muladd(t, -1.13520398)
+      |> muladd(t, 0.27886807)
+      |> muladd(t, -0.18628806)
+      |> muladd(t, 0.09678418)
+      |> muladd(t, 0.37409196)
+      |> muladd(t, 1.00002368)
+      |> muladd(t, 1.26551223)
```
seanmor5, line 239: You can go with this or not, it's just less code:

```elixir
coefficients = [...]
a = eval_recurrence(coefficients, t)

defp eval_recurrence(coefficients, t)
  Enum.reduce(coefficients, {t, t}, fn x, {acc, t} -> acc * t + x end)
end
```

elbow-jason: I thought of just reducing, but went with inline instead for performance.

josevalim:

PR #247, opened 2021-02-19
Remove NumPy dependency

seanmor5:
This is about as much of a hack as you can get, but basically we just comment out 4 parts of the TensorFlow build workflow so it doesn't check for the NumPy dependency up front. LLVM requires Python later on so we can't get rid of that, but we no longer need to depend on NumPy just to build.

I think this `sed` will just keep prepending comments to those lines, though.


josevalim:
This is great. Perhaps we need to be more assertive on those seds though, just in case we end-up commenting out something we should not. I don't know the exact syntax but this can give an idea:

    /python_configure(name = "local_config_python")/ s/^/#/

Perhaps it would be even better to comment out based on the patterns instead of lines. WDYT?

seanmor5:
@josevalim I think you're right. I think there's also a way to avoid adding an additional comment if one exists, I'll look into it

seanmor5:
@josevalim This is g2g

josevalim (approved):

xire28:
Really nice, one small question:
Shouldn't Makefile.win be updated as well ?

PR #249, opened 2021-02-20
AOT compilation

josevalim:
This is still WIP and we are not computing the proper results yet but we can emit the proper .so files with cross compilation support. :)


josevalim:

PR #251, opened 2021-02-20
Fix names bug on reshape with same shape

### LINKED ISSUE #250: Nx.reshape with the existing shape won't update the names.
```
>
iex(17)> t = Nx.tensor([1, 2, 3, 4])
#Nx.Tensor<
  s64[4]
  [1, 2, 3, 4]
>
iex(19)> t |> Nx.reshape({1, 4}) |> Nx.reshape({2, 2}, names: [:x, :y])
#Nx.Tensor<
  s64[x: 2][y: 2]
  [
    [1, 2],
    [3, 4]
  ]
>
iex(20)> t |> Nx.reshape({2, 2}) |> Nx.reshape({2, 2}, names: [:x, :y])
#Nx.Tensor<
  s64[2][2]
  [
    [1, 2],
    [3, 4]
  ]
>
```
The expected iex(20) result should be with names `:x` and `:y`

seanmor5:
Resolves #250

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -1262,16 +1262,6 @@ defmodule Nx do
         ]
       >
 
-      iex> t = Nx.tensor([[1, 2], [3, 4]])
-      iex> Nx.reshape(t, {2, 2}, names: [:x, :y])
-      #Nx.Tensor<
-        s64[x: 2][y: 2]
-        [
-          [1, 2],
-          [3, 4]
-        ]
-      >
-
   The shape can also be an existing tensor:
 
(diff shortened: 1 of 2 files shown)


nx/lib/nx.ex
```diff
@@ -1262,6 +1262,16 @@ defmodule Nx do
         ]
       >
 
+      iex> t = Nx.tensor([[1, 2], [3, 4]])
```
josevalim, line 1265: Since this doesn’t have much purpose as an example, perhaps an actual test?

josevalim (approved):
Looks good, just one comment!

also please double check if we don’t have the same bug on broadcast

PR #254, opened 2021-02-20
Improve docs

AlexJuca:
This PR adds a couple of improvements to Nx's documentation. 

Also, I am not sure where the correct place would be to add this comment, but I noticed that the function `Nx.shape/1` and its variants do not have any associated tests. 

Was that intentional?

If not, can the tests for this function be added?


josevalim:
The initial functions were driven by doc tests, so those are the tests. :) but we are planning to add more specific tests to _test.exs when necessary.

josevalim:

PR #259, opened 2021-02-21
Add erf_inv/1 to Nx and enable in EXLA

elbow-jason:
This PR adds `erf_inv/1` operation to `Nx`, `Nx.Backend`, and `EXLA`.

 #### SciPy:
 
 ```python
 In [2]: from scipy import special

In [3]: special.erfinv(0.999)
Out[3]: 2.3267537655135464

In [4]: special.erfinv(0.01)
Out[4]: 0.008862501280950608
 ```
 
 #### Nx:
 ```elixir
 iex(1)> Nx.erf_inv(0.999)
#Nx.Tensor<
  f64
  2.326753756865462
>
iex(2)> Nx.erf_inv(0.01) 
#Nx.Tensor<
  f64
  0.008862500728738846
>
 ```
 
 #### Resouces:

  + [erf_inv in XLA](https://github.com/tensorflow/tensorflow/blob/0e6b3a307f6363a9ee9dc9e03a8a0350c7668799/tensorflow/compiler/xla/client/lib/math.cc#L340-L390)
  
  + [erf_inv's grad in Jax](https://github.com/google/jax/blob/7e040c0a89b7b1f5c3607623d67c811d977470f4/jax/_src/lax/lax.py#L2434-L2436)

### DIFF
--- a/nx/lib/nx/shared.ex
+++ b/nx/lib/nx/shared.ex
@@ -253,7 +253,19 @@ defmodule Nx.Shared do
   @doc """
   Approximation for the inverse error function.
 
-  from Giles, M., "Approximating the erfinv function"
+  ## Examples
+
+    iex> Nx.erf_inv(0.999)
+    #Nx.Tensor<
+      f64
+      2.326753756865462
+    >
+
+    iex> Nx.erf_inv(0.01)
+    #Nx.Tensor<
+      f64
+      0.008862500728738846
+    >
   """
   def erf_inv(x) do
     w = -:math.log((1 - x) * (1 + x))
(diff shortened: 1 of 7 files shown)


nx/lib/nx/shared.ex
```diff
@@ -249,6 +250,56 @@ defmodule Nx.Shared do
...
+    #Nx.Tensor<
+      f64
+      2.326753756865462
+    >
+
+    iex> Nx.erf_inv(0.01)
+    #Nx.Tensor<
+      f64
+      0.008862500728738846
+    >
```
josevalim, line 268: Please indent with four spaces!

josevalim: Also, we shouldn't be testing the `Nx` API via `Nx.Shared` doctests. We should either test the Nx.Shared API directly or add the test to `nx/shared_test`. :)

josevalim:
Looks great, just one minor comment. :)

nx/lib/nx/shared.ex
elbow-jason: Good point. I missed that.

josevalim:

PR #264, opened 2021-02-21
Add groups to convolutions

### LINKED ISSUE #181: Support groups in convolution
Grouped Convolutions are common enough that we should support them. Here's some discussion and papers on grouped convolutions: https://paperswithcode.com/method/grouped-convolution

XLA supports grouped convolutions with it's `feature_group_count` option (right now we always set it to 1). We will just need to add support for groups in the binary implementation of `Nx.conv`. This involves an additional iteration in the binary implementation as well as some additional shape checks.

XLA also supports `batch_group_count` for grouping parts of the batch. I see less value in adding this option as I haven't been able to find any applications that are all that common.

seanmor5:
Resolves #181 

Hopefully I never have to write a convolution from scratch in my life again

### DIFF
--- a/nx/lib/nx/binary_backend.ex
+++ b/nx/lib/nx/binary_backend.ex
@@ -973,16 +973,12 @@ defmodule Nx.BinaryBackend do
   defp split_filters(kernel_data, kernel_shape, {_, kernel_size}, groups) do
     filter_group_size = div(Nx.size(kernel_shape) * kernel_size, groups)
     filter_size = div(Nx.size(kernel_shape) * kernel_size, elem(kernel_shape, 0))
-    filter_groups_with_index =
-      for i <- 0..groups - 1 do
-        offset = filter_group_size * i
-        <<_::size(offset)-bitstring, filter_group::size(filter_group_size)-bitstring, _::bitstring>> = kernel_data
-        for <<filter::size(filter_size)-bitstring <- filter_group>> do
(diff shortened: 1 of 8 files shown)


nx/lib/nx/binary_backend.ex
```diff
@@ -967,6 +970,21 @@ defmodule Nx.BinaryBackend do
...
+    filter_groups_with_index =
+      for i <- 0..groups - 1 do
+        offset = filter_group_size * i
+        <<_::size(offset)-bitstring, filter_group::size(filter_group_size)-bitstring, _::bitstring>> = kernel_data
+        for <<filter::size(filter_size)-bitstring <- filter_group>> do
+          {filter, i}
+        end
+      end
+
+    List.flatten(filter_groups_with_index)
```
josevalim, line 985:
```suggestion

    for i <- 0..groups - 1,
        offset = filter_group_size * i,
        <<_::size(offset)-bitstring, filter_group::size(filter_group_size)-bitstring, _::bitstring>> = kernel_data,
        <<filter::size(filter_size)-bitstring <- filter_group>>,
        do: {filter, i}
```

I haven't tried it, please double check!

josevalim (approved):

nx/lib/nx/binary_backend.ex
seanmor5: Wizadry

PR #265, opened 2021-02-21
Add QR decomposition

polvalente:
This PR adds an implementation for the Householder QR decomposition method. The function signature was designed for Numpy compatibility.

related to #157

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7056,107 +7056,81 @@ defmodule Nx do
 
     * `:reduced` - returns `q` and `r` with shapes `{M, K}` and `{K, N}`
     * `:complete` - returns `q` and `r` with shapes `{M, M}` and `{M, N}`
-    * `:r` - returns `r` with shape `{K, N}`
-    * `:raw` - returns `{h, tau}`, with `h` being the array of Householder reflectors and
-               `tau` being the array with the respective scaling factors.
 
   ## Examples
 
   Note: The following examples round elements of the output tensors
   to minimize rounding errors on different host machines.
 
-    iex(1)> [q, r] = Nx.qr(Nx.tensor([[12, -51, 4], [6, 167, -68], [-4, 24, -41]])) |> Tuple.to_list() |> Enum.map(fn t -> Nx.map(t, &Float.round(&1, 4)) end)
-    [
+      iex(1)> [q, r] = Nx.qr(Nx.tensor([[12, -51, 4], [6, 167, -68], [-4, 24, -41]])) |> Tuple.to_list() |> Enum.map(fn t -> Nx.map(t, &Float.round(&1, 4)) end)
+      [
+        Nx.tensor([
+          [-0.8571, 0.3943, -0.3314],
+          [-0.4286, -0.9029, 0.0343],
+          [0.2857, -0.1714, -0.9429]
+        ]),
+        Nx.tensor([
+          [-14.0, -21.0, 14.0],
+          [-0.0, -175.0, 70.0],
+          [-0.0, -0.0, 35.0]
+        ])
+      ]
+      iex(2)> Nx.dot(q, r) |> Nx.map(&Float.round(&1, 0))
       Nx.tensor([
-        [-0.8571, 0.3943, -0.3314],
-        [-0.4286, -0.9029, 0.0343],
-        [0.2857, -0.1714, -0.9429]
-      ]),
-      Nx.tensor([
-        [-14.0, -21.0, 14.0],
-        [-0.0, -175.0, 70.0],
-        [-0.0, -0.0, 35.0]
+        [12.0, -51.0, 4.0],
+        [6.0, 167.0, -68.0],
+        [-4.0, 24.0, -41.0]
       ])
-    ]
-    iex(2)> Nx.dot(q, r) |> Nx.map(&Float.round(&1, 0))
-    Nx.tensor([
-      [12.0, -51.0, 4.0],
-      [6.0, 167.0, -68.0],
-      [-4.0, 24.0, -41.0]
-    ])
 
-    iex(1)> [q, r] = Nx.qr(Nx.tensor([[1, -1, 4], [1, 4, -2], [1, 4, 2], [1, -1, 0]]), mode: :reduced) |> Tuple.to_list() |> Enum.map(fn t -> Nx.map(t, &Float.round(&1, 4)) end)
-    [
+      iex(1)> [q, r] = Nx.qr(Nx.tensor([[1, -1, 4], [1, 4, -2], [1, 4, 2], [1, -1, 0]]), mode: :reduced) |> Tuple.to_list() |> Enum.map(fn t -> Nx.map(t, &Float.round(&1, 4)) end)
+      [
+        Nx.tensor([
+          [-0.5774, 0.8165, 0.0],
+          [-0.5774, -0.4082, 0.7071],
+          [-0.5774, -0.4082, -0.7071],
+          [0.0, 0.0, 0.0]
+        ]),
+        Nx.tensor([
+          [-1.7321, -4.0415, -2.3094],
+          [0.0, -4.0825, 3.266],
+          [0.0, 0.0, -2.8284]
+        ])
+      ]
+      iex(2)> Nx.dot(q, r) |> Nx.map(&Float.round(&1, 0))
       Nx.tensor([
-        [-0.5774, 0.8165, 0.0],
-        [-0.5774, -0.4082, 0.7071],
-        [-0.5774, -0.4082, -0.7071],
-        [0.0, 0.0, 0.0]
-      ]),
-      Nx.tensor([
-        [-1.7321, -4.0415, -2.3094],
-        [0.0, -4.0825, 3.266],
-        [0.0, 0.0, -2.8284]
+        [1.0, -1.0, 4.0],
+        [1.0, 4.0, -2.0],
+        [1.0, 4.0, 2.0],
+        [0.0, -0.0, 0.0]
       ])
-    ]
-    iex(2)> Nx.dot(q, r) |> Nx.map(&Float.round(&1, 0))
-    Nx.tensor([
-      [1.0, -1.0, 4.0],
-      [1.0, 4.0, -2.0],
-      [1.0, 4.0, 2.0],
-      [0.0, -0.0, 0.0]
-    ])
 
-    iex(1)> [q, r] = Nx.qr(Nx.tensor([[1, -1, 4], [1, 4, -2], [1, 4, 2], [1, -1, 0]]), mode: :complete) |> Tuple.to_list() |> Enum.map(fn t -> Nx.map(t, &Float.round(&1, 4)) end)
-    [
+      iex(1)> [q, r] = Nx.qr(Nx.tensor([[1, -1, 4], [1, 4, -2], [1, 4, 2], [1, -1, 0]]), mode: :complete) |> Tuple.to_list() |> Enum.map(fn t -> Nx.map(t, &Float.round(&1, 4)) end)
+      [
+        Nx.tensor([
+          [-0.5, 0.5, -0.5, -0.5],
+          [-0.5, -0.5, 0.5, -0.5],
+          [-0.5, -0.5, -0.5, 0.5],
+          [-0.5, 0.5, 0.5, 0.5]
+        ]),
+        Nx.tensor([
+          [-2.0, -3.0, -2.0],
+          [0.0, -5.0, 2.0],
+          [-0.0, -0.0, -4.0],
+          [0.0, 0.0, 0.0]
+        ])
+      ]
+      iex(2)> Nx.dot(q, r) |> Nx.map(&Float.round(&1, 0))
       Nx.tensor([
-        [-0.5, 0.5, -0.5, -0.5],
-        [-0.5, -0.5, 0.5, -0.5],
-        [-0.5, -0.5, -0.5, 0.5],
-        [-0.5, 0.5, 0.5, 0.5]
-      ]),
-      Nx.tensor([
-        [-2.0, -3.0, -2.0],
-        [0.0, -5.0, 2.0],
-        [-0.0, -0.0, -4.0],
-        [0.0, 0.0, 0.0]
+        [1.0, -1.0, 4.0],
+        [1.0, 4.0, -2.0],
+        [1.0, 4.0, 2.0],
+        [1.0, -1.0, 0.0]
       ])
-    ]
-    iex(2)> Nx.dot(q, r) |> Nx.map(&Float.round(&1, 0))
-    Nx.tensor([
-      [1.0, -1.0, 4.0],
-      [1.0, 4.0, -2.0],
-      [1.0, 4.0, 2.0],
-      [1.0, -1.0, 0.0]
-    ])
-
-    iex(1)> {_q, _r, reflectors} = Nx.qr(Nx.tensor([[1, -1, 4], [1, 4, -2], [1, 4, 2], [1, -1, 0]]), mode: :raw)
-    iex(2)> Enum.map(reflectors, fn t -> Nx.map(t, &Float.round(&1, 4)) end)
-    [
-      Nx.tensor([
-        [1.0, 0.0, 0.0, 0.0],
-        [0.0, 1.0, 0.0, 0.0],
-        [0.0, 0.0, -0.6, 0.8],
-        [0.0, 0.0, 0.8, 0.6]
-      ]),
-      Nx.tensor([
-        [1.0, 0.0, 0.0, 0.0],
-        [0.0, -0.6667, -0.6667, 0.3333],
-        [0.0, -0.6667, 0.7333, 0.1333],
-        [0.0, 0.3333, 0.1333, 0.9333]
-      ]),
-      Nx.tensor([
-        [-0.5, -0.5, -0.5, -0.5],
-        [-0.5, 0.8333, -0.1667, -0.1667],
-        [-0.5, -0.1667, 0.8333, -0.1667],
-        [-0.5, -0.1667, -0.1667, 0.8333]
-      ])
-    ]
 
   ## Error cases
 
-    iex> Nx.qr(Nx.tensor([[1, 1, 1, 1], [-1, 4, 4, -1], [4, -2, 2, 0]])) |> Tuple.to_list() |> Enum.map(fn t -> Nx.map(t, &Float.round(&1, 4)) end)
-    ** (ArgumentError) tensor must have at least as many rows than columns, got shape: {3, 4}
+      iex> Nx.qr(Nx.tensor([[1, 1, 1, 1], [-1, 4, 4, -1], [4, -2, 2, 0]])) |> Tuple.to_list() |> Enum.map(fn t -> Nx.map(t, &Float.round(&1, 4)) end)
+      ** (ArgumentError) tensor must have at least as many rows than columns, got shape: {3, 4}
   """
   @doc type: :linalg
   def qr(tensor, opts \\ []) do
@@ -7176,20 +7150,14 @@ defmodule Nx do
 
     output_type = Nx.Type.to_floating(type)
 
-    {q_shape, {k, n}} = Nx.Shape.qr(shape, opts)
+    {q_shape, r_shape} = Nx.Shape.qr(shape, opts)
 
-    q = iota(q_shape, type: output_type)
-
-    mode = opts[:mode]
-
-    r =
-      if mode == :reduced do
-        tensor[[0..(k - 1), 0..(n - 1)]]
-      else
-        tensor
-      end
-
-    impl!(tensor).qr(q, r, opts)
+    impl!(tensor).qr(
+      {%{tensor | type: output_type, shape: q_shape},
+       %{tensor | type: output_type, shape: r_shape}},
+      tensor,
+      opts
+    )
   end
 
   ## Helpers
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7046,57 +7046,151 @@ defmodule Nx do
     impl!(tensor).sort(tensor, tensor, axis: axis, comparator: comparator)
   end
 
-  def qr(tensor) do
+  @doc """
+  Calculates the QR decomposition of an 2-D tensor with shape `{M, N}` via the Householder method.
+
+  ## Options
+
+  `:mode` - Can be one of `:reduced`, `:complete`, `:r` or `:raw`. Defaults to `:reduced`
+    For the following, `K = min(M, N)`
+
+    * `:reduced` - returns `q` and `r` with shapes `{M, K}` and `{K, N}`
+    * `:complete` - returns `q` and `r` with shapes `{M, M}` and `{M, N}`
+    * `:r` - returns `r` with shape `{K, N}`
+    * `:raw` - returns `{h, tau}`, with `h` being the array of Householder reflectors and
+               `tau` being the array with the respective scaling factors.
+
+  ## Examples
+
+  Note: The following examples round elements of the output tensors
+  to minimize rounding errors on different host machines.
+
+    iex(1)> [q, r] = Nx.qr(Nx.tensor([[12, -51, 4], [6, 167, -68], [-4, 24, -41]])) |> Tuple.to_list() |> Enum.map(fn t -> Nx.map(t, &Float.round(&1, 4)) end)
+    [
+      Nx.tensor([
+        [-0.8571, 0.3943, -0.3314],
+        [-0.4286, -0.9029, 0.0343],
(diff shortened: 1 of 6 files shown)


nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,139 @@ defmodule Nx.BinaryBackend do
...
+      |> Nx.multiply(Nx.outer(v, v))
+
+    subtract(
+      projection,
+      eye,
+      projection
+    )
+  end
+
+  defp identity(%{type: {_, num_bits} = type} = tensor, shape: {m, n} = shape) do
```
polvalente, line 1139: I wasn't sure on how to proceed with this identity matrix function. I needed one for the implementation of the householder method, but there wasn't any available. Should I also add the public interface for compatibility with the `eye` function from Numpy?

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,139 @@ defmodule Nx.BinaryBackend do
...
+      for col <- 0..(n - 1), row <- 0..(m - 1), reduce: data do
+        data ->
+          value = if col == row, do: 1, else: 0
+          set_value_at_index(data, type, shape, [row, col], value)
+      end
+
+    from_binary(%{tensor | shape: {m, n}}, identity_binary)
+  end
+
+  defp set_value_at_index(tensor_binary, {_, num_bits} = type, shape, index, value)
```
polvalente, line 1154: This can be an useful utility for other implementations on the binary backend. Should I change this into a `def` with `@doc false` for thorough testing?

nx/lib/nx/defn/expr.ex
```diff
@@ -675,6 +675,12 @@ defmodule Nx.Defn.Expr do
     expr(out, tensor.data.context, :cholesky, [tensor])
   end
 
+  @impl true
```
polvalente, line 678: What should I have written here? How can I test this piece of code?

nx/lib/nx.ex
```diff
@@ -7046,6 +7046,152 @@ defmodule Nx do
...
+
+  ## Options
+
+  `:mode` - Can be one of `:reduced`, `:complete`, `:r` or `:raw`. Defaults to `:reduced`
+    For the following, `K = min(M, N)`
+
+    * `:reduced` - returns `q` and `r` with shapes `{M, K}` and `{K, N}`
+    * `:complete` - returns `q` and `r` with shapes `{M, M}` and `{M, N}`
+    * `:r` - returns `r` with shape `{K, N}`
+    * `:raw` - returns `{h, tau}`, with `h` being the array of Householder reflectors and
```
polvalente, line 7060: Almost forgot to ask: does anyone know what the `tau` parameter means?

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,139 @@ defmodule Nx.BinaryBackend do
...
+      projection,
+      eye,
+      projection
+    )
+  end
+
+  defp identity(%{type: {_, num_bits} = type} = tensor, shape: {m, n} = shape) do
+    data_size = m * num_bits * n
+    # allocate empty data
+    data = <<0::size(data_size)>>
```
elbow-jason, line 1142: Instead of building a full-sized binary of 0s and then changing the indices that are 1s it is much more preferred to create a binary by accumulating.

Building binaries by appending to them is fast as it only requires reallocation when the underlying binary object is at full capacity.

Changing the contents of an exist binary results a totally new binary each time so it's very very slow.

Here are some erlang docs about [constructing binaries](https://erlang.org/doc/efficiency_guide/binaryhandling.html#constructing-binaries)

polvalente: Is there a preference over appending or prepending regarding preformance? I know that for lists prepending is faster.

elbow-jason: For binaries, appending to an accumulator that has been appended to previously will be faster than just about any other kind of change that can happen to a binary - it results in fewer allocations than other changes.

nx/lib/nx.ex
```diff
@@ -7046,6 +7046,152 @@ defmodule Nx do
...
+    valid_modes = [:reduced, :complete, :r, :raw]
+
+    unless mode in valid_modes do
+      raise ArgumentError,
+            "invalid :mode received. Expected one of #{valid_modes}, received: #{mode}"
+    end
+
+    output_type = Nx.Type.to_floating(type)
+
+    {q_shape, {k, n}} = Nx.Shape.qr(shape, opts)
```
seanmor5, line 7179: Nice, note you should also include a name rule here. Because this is a factorization of `A` into the product `QR`, the names should inherit from `A` such that the names of `Q . R` are equal to the names in `A`. I think it would be something like:

```
a_names = [n1, n2]
q_names = [n1, nil]
r_names = [nil, n2]
```

And perhaps we can provide an option for naming those unnamed dimensions? The other option is to just set names of the new tensor to nil

nx/lib/nx/binary_backend.ex
polvalente: Thanks! That's nice to know

nx/lib/nx.ex
```diff
@@ -7046,6 +7046,152 @@ defmodule Nx do
...
+    mode = opts[:mode]
+
+    r =
+      if mode == :reduced do
+        tensor[[0..(k - 1), 0..(n - 1)]]
+      else
+        tensor
+      end
+
+    impl!(tensor).qr(q, r, opts)
```
seanmor5, line 7192: So really when dispatching to the backend implementation you want to give it information about the output, the actual input, and then the options. So I wouldn't call out to other functions to build `q` and `r`. You should just provide the type/shape/name information, the input tensor, and the options and let the backend do the work. You can accomplish the `iota` and `slicing` from within `binary_backend`.

nx/lib/nx.ex
seanmor5: For simplicity, we can just omit `:r` and `:raw` modes and only return `:reduced` and `:complete`, wdyt? Note I'm referring to what the [jax.lax.linalg.qr](https://github.com/google/jax/blob/master/jax/_src/lax/linalg.py#L147) primitive does for this one!

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,138 @@ defmodule Nx.BinaryBackend do
...
+  def qr(%{shape: {_m, k} = shape_q} = q, %{shape: {k, n}} = r, opts) do
+    mode = opts[:mode]
+
+    keep_reflectors = mode == :raw
+    calculate_q = mode in [:reduced, :complete, :raw]
+
+    q =
+      if calculate_q do
+        identity(q, shape: shape_q)
+      end
```
seanmor5, line 1051: As @elbow-jason mentioned it's expensive to build this and reduce on the pre-built result. I'm not familiar with the algorithm, but if possible you should try and iteratively build the binary by appending to it.

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,138 @@ defmodule Nx.BinaryBackend do
...
+                set_value_at_index(
+                  acc,
+                  h.type,
+                  shape_h,
+                  [row + i, col + i],
+                  reflector[[row, col]]
+                )
+            end
+
+          h = from_binary(h, h_bin)
```
seanmor5, line 1074: Keep this in the raw binary form, that way when you go to do the dot product, you can use `bin_zip_reduce` rather than `Nx.dot`.

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,138 @@ defmodule Nx.BinaryBackend do
...
+            for col <- 0..(reflector_m - 1), row <- 0..(reflector_n - 1), reduce: to_binary(h) do
+              acc ->
+                set_value_at_index(
+                  acc,
+                  h.type,
+                  shape_h,
+                  [row + i, col + i],
+                  reflector[[row, col]]
+                )
+            end
```
seanmor5, line 1072: I think this is all going to end up being really expensive. Is there a way you can turn the algorithm into an aggregation/reduction rather than building the binary AOT?

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,138 @@ defmodule Nx.BinaryBackend do
...
+            else
+              reflectors
+            end
+
+          q =
+            if calculate_q do
+              Nx.dot(q, h)
+            end
+
+          r = Nx.dot(h, r)
```
seanmor5, line 1088: From above as well, we try to avoid calls to the `Nx` namespace in the backend if at all possible. You can use `bin_zip_reduce` for this if it's just matrix multiplication. If you keep `q`, `r` and `h` as raw binaries then you can copy the matrix multiplication case from the `dot` implementation in this module!

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,138 @@ defmodule Nx.BinaryBackend do
...
+
+  defp householder_reflector(%{shape: {n}} = a) do
+    a_0 = a[0]
+
+    v =
+      %{type: {_, num_bits} = type_v} =
+      if Nx.sign(a_0) == -1 do
+        Nx.divide(a, Nx.subtract(a_0, Nx.norm(a)))
+      else
+        Nx.divide(a, Nx.add(a_0, Nx.norm(a)))
```
seanmor5, line 1116: Just a note for this stuff, we try to avoid calls to the `Nx` namespace, so you should either try to work on the raw binaries OR you can use the implementations in this module. It will just save you the overhead of dispatching here.

nx/lib/nx/defn/expr.ex
elbow-jason: I'm not sure testing this would be useful. It's up to a backend to handle (or not) an expression correctly. Effectively, what expr/4 is doing is adding the `:qr` step and the given tensors and options to the expression graph.

seanmor5: This is fine! You did it right :)

seanmor5 (commented):
Hey @polvalente nice work! There are a few things we should try and change. I've left comments but the biggest thing is you should try and write the algorithm such that you're iteratively building the binary rather than all at once! Also, we can limit the modes to just `:reduced` and `:complete` for simplicity. Finally, we will need a name rule for this implementation! I dropped a comment in my review, but I'm in favor of something like:

```elixir
a_names = [n1, n2]
q_names = [n1, nil]
r_names = [nil, n2]
```

with the option to name the unnamed dimensions. Thanks for your hard work, this is a tough one :)

nx/lib/nx.ex
polvalente: Sounds like a good idea! I think that keeping the names when they are given is good. Should we require names for the unnamed dimensions, or maybe provide default ones when input names are given?

nx/lib/nx.ex
polvalente: I see! How would I go about defining `q` and `r` then? Should I build the `Tensor` struct by hand with empty data?

seanmor5: You can do something like:

```elixir
q = %{tensor | shape: q_shape, names: q_names, type: output_type}
r = %{tensor | shape: r_shape, names: r_names, type: output_type}
```

It might help to think of the public facing API as a contract for shapes, types, and names. So all you should worry about doing is enforcing the shape, type, and name rules, and then let the backend worry about implementation details!

nx/lib/nx/binary_backend.ex
polvalente: This algorithm relies heavily on pre and post matrix multiplication to yield Q and R. I'll try to think of a more iterative way so we can build the binary dinamically.

nx/lib/nx/binary_backend.ex
polvalente: This may be possible, although I'll have to sit on this a little bit longer since I'm not _that_ familiar with the algorithm. I'll also keep in mind approaching QR factorization from the Gram-Schmidt and Given Rotations algorithms as well, if Householder ends up leading nowhere.

seanmor5: If Household relies on a lot of in place updates, then it's probably not going to be the most efficient at least for the Binary backend.

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,157 @@ defmodule Nx.BinaryBackend do
     from_binary(out, output_data)
   end
 
+  @impl true
+  def qr(
```
polvalente, line 1042: @seanmor5 I've managed to refactor this function for operating directly on binaries and to avoid memory pre-allocation.
All tests continue to pass, so that's something haha

There may be one last trick we can apply on the construction of Q due to the nature of the reflector matrices, but I don't think I can optimize this implementation further.

The main idea behind the latest refactor (aside from operating upon binaries) is that the reflector can be constructed in a single K*K loop, where we can both calculate the reflection and the "identity matrix wrapper" easily

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,157 @@ defmodule Nx.BinaryBackend do
...
+        a_0 + :math.sqrt(norm_a_squared)
+      else
+        a_0 - :math.sqrt(norm_a_squared)
+      end
+
+    prefix_threshold = target_k - num_rows_a
+    prefix_size = prefix_threshold * num_bits
+    v_bin = <<0::size(prefix_size)>> <> number_to_binary(v_0, type) <> tail
+
+    # dot(v, v) = norm_v_squared, which can be calculated from norm_a as: norm_v_squared = norm_a_squared - a_0^2 + v_0^2
```
polvalente, line 1158: I don't know if I should document this better or it's legible enough. I'm open to suggestions.

The main trick here is that we have already needed to calculate `<a, a> = norm_a^2`. Since `v` is basically `a` with one element changed, `norm_v^2` can be calculated by subtracting the square of the original element and adding the square of the new one, as described in the comment above.

This avoids having to iterate over `v` a second time.

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,157 @@ defmodule Nx.BinaryBackend do
...
+
+    norm_v_squared = norm_a_squared - a_0 * a_0 + v_0 * v_0
+
+    scale = 2 / norm_v_squared
```
polvalente, line 1162: It's worth noting that this line makes use of the identity `v_t.v = ||v||^2`

nx/lib/nx/backend.ex
```diff
@@ -75,7 +75,7 @@ defmodule Nx.Backend do
   @callback sort(out :: tensor, tensor, keyword) :: tensor
 
   @callback cholesky(out :: tensor, tensor) :: tensor
-
+  @callback qr(q :: tensor, r :: tensor, keyword) :: tensor
```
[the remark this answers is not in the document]
josevalim: Yes, I like the tuple approach!

nx/lib/nx.ex
```diff
@@ -7046,6 +7046,120 @@ defmodule Nx do
...
+  """
+  @doc type: :linalg
+  def qr(tensor, opts \\ []) do
+    %T{type: type, shape: shape} = tensor = tensor!(tensor)
+
+    assert_keys!(opts, [:mode])
+
+    opts = Keyword.merge([mode: :reduced], opts)
+
+    mode = opts[:mode]
```
josevalim, line 7143:
```suggestion
    mode = opts[:mode] || :reduced
```

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,159 @@ defmodule Nx.BinaryBackend do
...
+      ) do
+    mode = opts[:mode]
+
+    r =
+      if mode == :reduced do
+        tensor[[0..(k - 1), 0..(n - 1)]]
+      else
+        tensor
+      end
+      |> Nx.as_type(type)
```
josevalim, line 1056: Don't do the type conversion upfront if possible. Instead keep the type as is and it should transparently become the proper type as you do number_to_binary at the end.

nx/lib/nx.ex
```diff
@@ -7046,6 +7046,120 @@ defmodule Nx do
...
+    valid_modes = [:reduced, :complete, :r, :raw]
```
josevalim, line 7144:
```suggestion
    valid_modes = [:reduced, :complete]
```

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,159 @@ defmodule Nx.BinaryBackend do
...
+    # We receive the target shape so the binary is already pre-generated with 1s in the
+    # main diagonal for rows < k - n
+    <<head::bitstring-size(num_bits), tail::bitstring>> = a_bin
+
+    norm_a_squared =
+      for <<x_bin::bitstring-size(num_bits) <- a_bin>>, reduce: 0 do
+        acc ->
+          x = binary_to_number(x_bin, type)
+          acc + x * x
+      end
```
josevalim, line 1143: Generally speaking, we want to discourage using `binary_to_number`. It is more efficient to use the match_types syntax:

```suggestion
      match_types [input_type] do
        for <<match!(x, 0) <- a_bin>>, reduce: 0 do
          acc ->
            x = read!(x, 0)
            acc + x * x
        end
      end
```

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,159 @@ defmodule Nx.BinaryBackend do
...
+
+          result =
+            if row >= prefix_threshold and col >= prefix_threshold do
+              identity_element -
+                scale * binary_to_number(col_factor, type) * binary_to_number(row_factor, type)
+            else
+              identity_element
+            end
+
+          acc = acc <> scalar_to_binary(result, type)
```
josevalim, line 1182: `<>` copies both sides, so if it is often batter to build cons lists:

```suggestion
          acc = [acc | scalar_to_binary(result, type)]
```

then at the end: `IO.iodata_to_binary(reflector_binary)`.

nx/lib/nx/binary_backend.ex
josevalim: Or better yet, convert `a_bin` to `a_list` at the beginning of the function with this:

    for <<match!(x, 0) <- a_bin>>, do: read!(x, 0)

And then keep a_bin and v_bin as lists. I think it will simplify and speed up the rest of the code considerably.

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,159 @@ defmodule Nx.BinaryBackend do
...
+                  if col + 1 == n do
+                    {row + 1, 0}
+                  else
+                    {row, col + 1}
+                  end
+
+                {data <> item, num_rows_a, row, col}
+            end
+
+          h_bin = householder_reflector(a, num_rows_a, type, k)
```
josevalim, line 1085: I think it will be simpler if you build a list of values here and then work with lists in householder_reflector. Something like this:

```suggestion
          {a, num_rows_a, _, _} =
            match_types [input_type] do
              for <<match!(x, 0) <- r>>, reduce: {[], 0, 0, 0} do
                {acc, num_rows_a_prev, row, col} ->
                  {acc, num_rows_a} =
                    if row >= i and col == i do
                      {[read!(x, 0) | acc], num_rows_a_prev + 1}
                    else
                      {acc, num_rows_a_prev}
                    end

                  {row, col} =
                    if col + 1 == n do
                      {row + 1, 0}
                    else
                      {row, col + 1}
                    end

                  {acc, num_rows_a, row, col}
              end
            end

          h_bin = householder_reflector(Enum.reverse(a), num_rows_a, type, k)
```

It will also be more performant because you will avoid converting to number multiple times in the household function.

nx/lib/nx.ex
```diff
@@ -7046,6 +7046,123 @@ defmodule Nx do
     impl!(tensor).sort(tensor, tensor, axis: axis, comparator: comparator)
   end
 
+  @doc """
+  Calculates the QR decomposition of an 2-D tensor with shape `{M, N}` via the Householder method.
```
seanmor5, line 7050: Just a nit, some backends may not use the Householder method, so you don't need to mention it!Also "an" should be "a" :)

nx/lib/nx.ex
```diff
@@ -7046,6 +7046,123 @@ defmodule Nx do
...
+
+  ## Options
+
+  `:mode` - Can be one of `:reduced`, `:complete`, `:r` or `:raw`. Defaults to `:reduced`
```
seanmor5, line 7054: Make sure to update these docs to only include the 2 modes!

nx/lib/nx.ex
```diff
@@ -7046,6 +7046,123 @@ defmodule Nx do
...
+    For the following, `K = min(M, N)`
+
+    * `:reduced` - returns `q` and `r` with shapes `{M, K}` and `{K, N}`
+    * `:complete` - returns `q` and `r` with shapes `{M, M}` and `{M, N}`
+
+  `:q_axes` - Defines the axes for the `q` tensor
```
seanmor5, line 7060: We use `names` elsewhere so I would just call these `q_names` and `r_names` respectively

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,152 @@ defmodule Nx.BinaryBackend do
...
+          h_bin = householder_reflector(a_reverse, num_rows_a, k, output_type)
+
+          # If we haven't allocated Q yet, let Q = H1
+          q =
+            if is_nil(q) do
+              padding_length = (m - k) * k
+
+              zero = number_to_binary(0, output_type)
+
+              h_bin <> String.duplicate(zero, padding_length)
```
seanmor5, line 1095: I think it'll be more efficient to do a `List.duplicate` and cons `h_bin` and then call `IO.iodata_to_binary`

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,152 @@ defmodule Nx.BinaryBackend do
...
+
+    q = from_binary(q_holder, q_bin)
+    r = from_binary(r_holder, r_bin)
+
+    case mode do
+      :complete ->
+        {q, r}
+
+      :reduced ->
+        {q, r[[0..(k - 1), 0..(n - 1)]]}
```
seanmor5, line 1113: Is it possible to do the slicing on the binary before building the output tensor? If you look at `Nx.slice` it's just a shape change then a weighted traverse.

nx/lib/nx/binary_backend.ex
```diff
@@ -1038,6 +1038,152 @@ defmodule Nx.BinaryBackend do
...
+        {%{shape: {m, k} = shape_q, type: output_type} = q_holder,
+         %{shape: {k, n} = shape_r, type: output_type} = r_holder},
+        %{type: input_type} = tensor,
+        opts
+      ) do
+    mode = opts[:mode]
+
+    r =
+      if mode == :reduced do
+        tensor[[0..(k - 1), 0..(n - 1)]]
```
seanmor5, line 1052: You can slice the binary in the same way as `Nx.slice` instead of having to dispatch outside here, but it's not that much overhead

seanmor5 (commented):
@polvalente This looks really, really good! I left a few comments, but they are relatively minor, awesome job!

polvalente:
> @polvalente This looks really, really good! I left a few comments, but they are relatively minor, awesome job!

Thank you so much! I'll make the requested changes and I'll request a review afterwards

nx/lib/nx/binary_backend.ex
polvalente: It turns out that:
a) This particular slicing was much more efficient on the binary because it happens to just be a "take N bits from the beggining of the binary" algorithm
b) Only this first slicing is needed. The other one at the end of the function was redundant in the end.

nx/lib/nx.ex
```diff
@@ -7046,6 +7046,123 @@ defmodule Nx do
...
+    * `:reduced` - returns `q` and `r` with shapes `{M, K}` and `{K, N}`
+    * `:complete` - returns `q` and `r` with shapes `{M, M}` and `{M, N}`
+
+  `:q_names` - Defines the axes for the `q` tensor
+
+  `:r_names` - Defines the axes for the `r` tensor
+
+  ## Examples
+
+  Note: The following examples round elements of the output tensors
```
seanmor5, line 7066: Sorry I missed this in the first, one I saw you moved the test so you can remove this!

seanmor5 (approved):
Great job! I have no other issues, sorry I missed one thing but it's just a documentation thing, otherwise I have no complaints :)

Thanks so much for contributing these!

nx/lib/nx.ex
polvalente: Done!

elbow-jason (approved):
Looks good to me. :-)

josevalim:

PR #269, opened 2021-02-22
fix warnings when `mix docs` invoked

kentaro:
I fixed the warnings that `mix docs` generates as below:

```
$ mix docs
Generating docs...
warning: documentation references "Nx.Defn.Kernel.max_unsigned_type/2" but it is undefined or private
  lib/nx/defn.ex:2: Nx.Defn

warning: documentation references "Nx.Defn.Kernel.max_signed_type/2" but it is undefined or private
  lib/nx/defn.ex:2: Nx.Defn

warning: documentation references "Nx.Defn.Kernel.max_float_type/2" but it is undefined or private
  lib/nx/defn.ex:2: Nx.Defn

View "html" docs at "doc/index.html"
```

### DIFF
--- a/nx/lib/nx/defn.ex
+++ b/nx/lib/nx/defn.ex
@@ -95,11 +95,12 @@ defmodule Nx.Defn do
 
   ### Options
 
-  The `Nx.Defn` compiler supports the following options:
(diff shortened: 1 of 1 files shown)


josevalim:

Issue #268, opened 2021-02-22
When a tensor is square (NxN), what can I do to invert it without writing my own implementation?


ArthurHypersphere:

elbow-jason:
The word "invert" here is ambiguous. Can you give an example of what your square tensor looks like before and what your square tensor looks like after it is inverted?

ArthurHypersphere:
I mean it in the sense of an invertible matrix, i.e.

AB == BA == (identity matrix of the same size)

seanmor5:
I don't think it's possible with our current API. In NumPy you would do something like:

```python
a = np.array([[2, 1, 2], [3, 2, 3], [4, 5, 6]], dtype=np.float32)
np.linalg.inv(a)
```

Given the expected API after we resolve #157,  under the hood, I think it'd be something like:

```elixir
defn invert_matrix(a) do
  solve(a, Nx.eye(a)) # solve `a` for the identity matrix
end

defn solve(a) do
  lu = Nx.lu(a)
  Nx.triangular_solve(lower)
end
```

I don't know how long it's going to be until we are able to getting around to tackling those implementations. @polvalente has been pretty active in handling those. He might be able to help more

polvalente:
@ArthurHypersphere Depending on your use-case, QR-factorization could remove the need to invert a matrix. 

If you're trying to find `x` for `Ax = b`, if you have `A = QR`, where Q is  orthonormal and R is upper triangular, you can turn the problem into `Rx = transpose(Q)b`, which is easily solvable through back substitution. QR decomposition is being implemented in #265 and I think there'll be a pull request for `triangular_solve` coming soon. If not, an implementation would be something akin to:

```elixir

for row <- n-1..0, col <- n-1..row do
```

polvalente:
@ArthurHypersphere Depending on your use-case, QR-factorization could remove the need to invert a matrix. 

If you're trying to find `x` for `Ax = b`, if you have `A = QR`, where Q is  orthonormal and R is upper triangular, you can turn the problem into `Rx = transpose(Q)b`, which is easily solvable through back substitution. QR decomposition is being implemented in #265 and I think there'll be a pull request for `triangular_solve` coming soon.

The back substitution algorithm itself can be implemented as a placeholder while `triangular_solve` isn't available. If this can solve your problem, I might be able to help with it!

ArthurHypersphere:
[...]

PR #272, opened 2021-02-22
implement triangular solve

AlexJuca:
This PR aims to add triangular solve to #157


nx/lib/nx.ex
```diff
@@ -7142,6 +7142,42 @@ defmodule Nx do
     impl!(tensor).sort(tensor, tensor, axis: axis, comparator: comparator)
   end
 
+  @doc """
+    Solve the equation a x = b for x, assuming a is a triangular matrix.
```
josevalim, line 7146:
```suggestion
  Solve the equation a x = b for x, assuming a is a triangular matrix.
```

nx/lib/nx.ex
```diff
@@ -7142,6 +7142,42 @@ defmodule Nx do
...
+
+  # Examples
```
josevalim, line 7148:
```suggestion
  ## Examples
```

nx/lib/nx.ex
```diff
@@ -7142,6 +7142,42 @@ defmodule Nx do
...
+
+  iex> Nx.triangular_solve([[3, 0, 0, 0], [2, 1, 0, 0], [1, 0, 1, 0], [1, 1, 1, 1]], [4, 2, 4, 2])
```
josevalim, line 7150: Please indent all examples with four spaces.

nx/lib/nx.ex
```diff
@@ -7142,6 +7142,42 @@ defmodule Nx do
...
+  ### Error cases
+
+  iex> Nx.triangular_solve([[3, 0, 0, 0], [2, 1, 0, 0], [1, 0, 1, 0], [1, 1, 1, 1]], [4, 2, 4, 2], trans: 9)
+  ** (ArgumentError) unknown trans 9, expected 0 or 'N' for a x = b
+
+  """
+  @doc type: :linalg
+  def triangular_solve(a, b, opts \\ []) do
+    a = tensor(a)
+    b = tensor(b)
```
josevalim, line 7165:
```suggestion
    a = tensor!(a)
    b = tensor!(b)
```

`tensor!` is a bit more strict (it doesn't accept lists) and more optimized.

AlexJuca: @josevalim Thanks, also is there a way I can check if a tensor is square? 

I know I can use Nx.shape/1 but I am not sure how to actually validate any square matrix.

josevalim: See the eye/2 function, it will have an example.

nx/lib/nx.ex
AlexJuca: @josevalim will do, I believe `mix format` will do this for me correct?

josevalim: No, it doesn’t change doctests at the moment.

nx/lib/nx.ex
```diff
@@ -7142,6 +7142,61 @@ defmodule Nx do
...
+      nil ->
+        triangular_solve_n(a, b)
+
+      trans ->
+        raise ArgumentError, "unknown trans #{inspect(trans)}, expected 0 or 'N' for a x = b"
+    end
+  end
+
+  defp triangular_solve_n(a, b) do
+    # TODO: write implementation
```
polvalente, line 7197: @AlexJuca I wrote a draft [here](https://github.com/elixir-nx/nx/issues/268#issuecomment-783352908) that might be of interest to you. My draft had an upper triangular matrix in mind, but with some minor changes I think the idea remains the same for lower triangular ones

AlexJuca: @polvalente Hey thanks, I will have a look at it and by end of the day I will have an implementation.

AlexJuca: @josevalim Does this particular implementation need to be done via the binary backend?

josevalim: Yes! The QR decomposition is a good example.

nx/lib/nx.ex
AlexJuca: @josevalim when using `%T{shape: s1 = {m, n}} = a = tensor!(a)`, I get the following error: expected exception ArgumentError but got FunctionClauseError with message "no function clause matching in Nx.tensor!/1"

I am not sure what the issue here is.

nx/lib/nx.ex
AlexJuca: > Yes! The QR decomposition is a good example.

Thanks

nx/lib/nx.ex
seanmor5: @AlexJuca your test cases are lists, they need to be Tensors that's why you're getting the error

AlexJuca: > @AlexJuca your test cases are lists, they need to be Tensors that's why you're getting the error

I totally missed that. Thanks for the heads up.

nx/lib/nx.ex
```diff
@@ -7164,21 +7164,32 @@ defmodule Nx do
...
+  defp triangular_solve_n(a, b, opts) do
+    %T{shape: {m, n}} = tensor!(a)
+    %T{type: type, shape: shape} = tensor!(b)
+
+    output_type = Nx.Type.to_floating(type)
+
+    impl!(b).triangular_solve(
+      {
+        %{b | type: output_type, shape: shape}
+      },
```
polvalente, line 7189:
```suggestion
      %{b | type: output_type, shape: shape},
```

Since you only have a single output tensor, no tuples are needed

AlexJuca: @polvalente Thanks for the heads up.

nx/lib/nx/binary_backend.ex
```diff
@@ -2033,6 +2051,36 @@ defmodule Nx.BinaryBackend do
...
+        b
+      ) do
+    a_matrix = a |> to_binary() |> binary_to_matrix(input_type, {rows, rows})
+    b_vector = b |> to_binary() |> binary_to_matrix(input_type, {0, 1}) |> List.flatten()
+
+    solution = List.duplicate(0, rows)
+
+    Enum.with_index(a_matrix)
+    |> Enum.zip(b_vector)
+    |> Enum.reduce([], fn {current_row_and_idx, current_b}, previous_y ->
```
polvalente, line 2067:
```suggestion
    |> Enum.reduce(solution, fn {current_row_and_idx, current_b}, previous_y ->
```

nx/lib/nx/binary_backend.ex
```diff
@@ -2033,6 +2051,36 @@ defmodule Nx.BinaryBackend do
...
+      row = elem(current_row_and_idx, 0)
+
+      y =
+        cond  do
+          idx == 0 -> current_b / Enum.at(row, idx)
+          idx in 1..rows - 1 -> current_b - Enum.at(row, idx) * Enum.at(previous_y, 0)
+          idx == rows - 1 -> current_b - Enum.at(row, idx) * Enum.sum(previous_y)
+        end
+
+      solution = replace_vector_element(solution, idx, y)
```
polvalente, line 2078:
```suggestion
      y = (current_b - dot_matrix(row, previous_y)) / Enum.at(row, idx)

      replace_vector_element(previous_y, idx, y)
```

nx/lib/nx/binary_backend.ex
```diff
@@ -2033,6 +2051,36 @@ defmodule Nx.BinaryBackend do
...
+    end)
+
+    solution
```
polvalente, line 2081:
```suggestion
```

nx/lib/nx/binary_backend.ex
AlexJuca: @polvalente Thanks, this helped!

nx/lib/nx/binary_backend.ex
AlexJuca: @polvalente This will surely help deal with all the cases.

nx/lib/nx.ex
```diff
@@ -7842,6 +7842,70 @@ defmodule Nx do
     impl!(tensor).sort(tensor, tensor, axis: axis, comparator: comparator)
   end
 
+  @doc """
+  Solve the equation a x = b for x, assuming a is a triangular matrix.
+
```
polvalente, line 7847: What do you think about adding a section on the valid options?

nx/lib/nx.ex
```diff
@@ -7842,6 +7842,70 @@ defmodule Nx do
...
+    %T{shape: {q}} = b = tensor!(b)
+
+    case shape(s1) do
+      {n, n} -> {n, n}
+      other -> raise ArgumentError, "expected a square matrix, got: #{inspect(other)}"
+    end
+
+    if m != q, do: raise(ArgumentError, "incompatible dimensions")
+
+    assert_keys!(opts, [:trans])
```
polvalente, line 7879: I think Nx's pattern is to have more extensive names instead of abbreviations for the options

nx/lib/nx.ex
AlexJuca: Sounds good.

nx/lib/nx/binary_backend.ex
```diff
@@ -2033,6 +2051,29 @@ defmodule Nx.BinaryBackend do
...
+
+    solution = List.duplicate(0, rows)
+
+    x =
+      Enum.with_index(a_matrix)
+      |> Enum.zip(b_vector)
+      |> Enum.reduce(solution, fn {{row, idx}, current_b}, previous_y ->
+        y = (current_b - dot_matrix(row, previous_y)) / Enum.at(row, idx)
+        replace_vector_element(previous_y, idx, y)
+      end)
```
polvalente, line 2071:
```suggestion
    x =
      Enum.with_index(a_matrix)
      |> Enum.zip(b_vector)
      |> Enum.reduce([], fn {{row, idx}, current_b}, previous_y ->
        y = (current_b - dot_matrix(row, previous_y)) / Enum.at(row, idx)
        previous_y ++ [y]
      end)
```

This works and avoids preallocation `solution` and replacing items in the list (which ends up being costly due to lists in elixir/erlang being linked-lists)

This only works because of the way `dot_matrix` is defined, dropping the remaining elements of the longer list (which for dot products is the same as dotting with a same-dimension matrix where the missing entries are 0)

nx/lib/nx.ex
AlexJuca: @polvalente I was going for a something like that available in scipy. 

https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve_triangular.html

What do you think of following that route?

polvalente: Sounds good. I'd rename this to something more descriptive like `system_type`, and only accepting either atoms or numbers as the values. @josevalim might be able to guide us towards the best solution here.

seanmor5: I would maybe hold off on adding this until we add support for complex numbers. It's supposed to represent `transpose` with the options of no transpose, transpose, and conjugate transpose, correct? In that case, is it the same to just do: `triangular_solve(transpose(a))`? If the results are the same, then I would hold off on adding it for the time being and then we can revisit with complex numbers.

nx/lib/nx/binary_backend.ex
AlexJuca: @polvalente This is great. Much more efficient. Thanks much!

nx/lib/nx.ex
polvalente: Seems reasonable. Until we have complex numbers there's really not much of a difference (even with respect to implementation, where some algorithms change a bit)

AlexJuca: @seanmor5 I believe that setting trans to 0 or 'N' is the default. So I am only dealing with the default case for when the system a x = b

No other cases are supported.

AlexJuca:
@josevalim I would love your feedback. 😃

josevalim:

nx/lib/nx.ex
```diff
@@ -7842,6 +7842,95 @@ defmodule Nx do
...
+
+  ## Options
+
+    * `:trans` - Type of system to solve, if none is provided defaults to: a x = b - Optional
+
+  | trans                    | system                        |
+  | ------------------------ | ------------------------------|
+  | `nil` or 0 or 'N'        | a x = b                       |
+
+    * `:lower` - Use only data contained in the lower triangle of a. Default is to use lower triangle, Optional
```
josevalim, line 7856: I have removed the options for now since they always have no effect. We can add them back once implemented!

nx/lib/nx/binary_backend.ex
```diff
@@ -2033,6 +2051,29 @@ defmodule Nx.BinaryBackend do
...
+    b_vector = b |> to_binary() |> binary_to_matrix(input_type, {0, 1}) |> List.flatten()
+
+    x =
+      Enum.with_index(a_matrix)
+      |> Enum.zip(b_vector)
+      |> Enum.reduce([], fn {{row, idx}, current_b}, previous_y ->
+        if Enum.at(row, idx) == 0, do: raise(ArgumentError, "can't solve for singular matrix")
+
+        y = (current_b - dot_matrix(row, previous_y)) / Enum.at(row, idx)
+        previous_y ++ [y]
```
josevalim, line 2070: I wonder if there is a way to implement this without the `++ [y]`, which can be expensive... 🤔

polvalente: I think I have an idea, which we can try out later. Pseudo code follows:

```elixir
a |> Enum.zip(b) |> Enum.reduce(%{row_idx: 0}, fn {row, current_b}, %{row_idx: row_idx} = y ->
  y = current_b - Enum.reduce_while(row, {0, 0}, fn 
    {x, col}, acc when col < row_idx -> {:cont, x * y[col] + acc}
    _, acc -> {:halt, acc}
end)
y |> Map.put(:row_idx, row_idx + 1) |> Map.put(row_idx, y)
end)
```

polvalente: The idea is to use maps for a constant-time `data[index]` access and update, so we can avoid needing to `Enum.reverse` either `previous_y` or `row`

nx/lib/nx.ex
AlexJuca: @josevalim Understood!

nx/lib/nx/binary_backend.ex
AlexJuca: @polvalente Seems interesting. Is there a way to run memory benchmarks in Elixir or see how the allocations are going on behind the scenes?

polvalente: You could try Benchee or something similar

PR #273, opened 2021-02-23
Add defndelegate for delegating defn functions

elbow-jason:
Earlier today in a conversation with Sean I suggested he use `defdelegate` to delegate his numerical functions, and he responded saying that one could not use `defdelegate` with numerical functions. So I made `defndelegate` so Sean can delegate his numerical functions.

cc @seanmor5 

@josevalim I didn't want to end up with raw ast, but despite my efforts I could not get a normal `quote do` block to behave.

Suggestions welcome!

### DIFF
--- a/nx/lib/nx/defn.ex
+++ b/nx/lib/nx/defn.ex
@@ -270,7 +270,7 @@ defmodule Nx.Defn do
   @doc """
   Defines a public numerical function that delegates to another module.
 
-  Function signatures must be valid defn function signatures - see defn.
+  Function signature must target defn function signature.
 
   ## Examples
 
@@ -299,11 +299,15 @@ defmodule Nx.Defn do
       {:__aliases__, _, _} ->
         Macro.expand(module, %{__CALLER__ | function: {:__info__, 1}})
       _ ->
-        raise ArgumentError, "expected :to to be a module, got #{Macro.to_string(module)}"
+        raise ArgumentError, "defndelegate expected :to to be a module, got #{Macro.to_string(module)}"
     end
-    func = Keyword.get(opts, :as) || name
-    arity = length(args)
 
+    func = Keyword.get(opts, :as, name)
+    if not is_atom(func) do
+      raise ArgumentError, "defndelegate expected :as to be an atom, got #{Macro.to_string(func)}"
+    end
+    
+    arity = length(args)
     defn_call = {:., [], [{:__aliases__, [alias: false], [:Nx, :Defn]}, :defn]}
 
     {:__block__, [], [
--- a/nx/lib/nx/defn.ex
+++ b/nx/lib/nx/defn.ex
@@ -267,6 +267,52 @@ defmodule Nx.Defn do
     define(:defp, call, block, __CALLER__)
   end
 
+  @doc """
+  Defines a public numerical function that delegates to another module.
+
+  Function signatures must be valid defn function signatures - see defn.
+
+  ## Examples
+
+      defmodule MyImpl do
+        import Nx.Defn
+
+        defn calc(a, b), do: a * b + 1
+        defn calc_pair({a, b}), do: calc(a, b)
+      end
+
+      defmodule MyMath do
+        import Nx.Defn
+
+        defndelegate calc(a, b), to: MyImpl
+
+        defndelegate calc_pair({a, b}), to: MyImpl
+
+        defndelegate calc_w_other_name(a, b), to: MyImpl, as: :calc
+
+        defndelegate calc_pair_tuple({a, b}), to: MyImpl, as: :calc_pair
+      end
+  """
+  defmacro defndelegate({name, meta, args}, opts) do
+    module = Keyword.fetch!(opts, :to)
+    module_name = case module do
+      {:__aliases__, _, _} ->
+        Macro.expand(module, %{__CALLER__ | function: {:__info__, 1}})
+      _ ->
+        raise ArgumentError, "expected :to to be a module, got #{Macro.to_string(module)}"
+    end
+    func = Keyword.get(opts, :as) || name
+    arity = length(args)
+
+    defn_call = {:., [], [{:__aliases__, [alias: false], [:Nx, :Defn]}, :defn]}
+
+    {:__block__, [], [
+      {:@, [], [{:doc, [],[[delegate_to: {:{}, [], [module_name, func, arity]}]]}]},
+      {:require, [context: Elixir], [{:__aliases__, [alias: false], [:Nx, :Defn]}]},
+      {defn_call, meta, [{name, meta, args}, [do: {{:., [], [module, func]}, meta, args}]]}
+    ]}
+  end
+
   ## Callbacks
 
   defp define(kind, call, block, env) do
--- a/nx/lib/nx/defn.ex
+++ b/nx/lib/nx/defn.ex
@@ -268,9 +268,13 @@ defmodule Nx.Defn do
   end
 
   @doc """
-  Defines a public numerical function that delegates to another module.
+  Defines a public numerical function that delegates to another
+  module.
 
-  Function signature must target defn function signature.
+  Delegated function signature patterns must match target defn
+  function signature patterns. For example, if the target function
+  has a 2-tuple as its first arg then the delegated function
+  signature must also have a 2-tuple as its first arg.
 
   ## Examples
 
--- a/nx/lib/nx/defn.ex
+++ b/nx/lib/nx/defn.ex
@@ -326,20 +326,16 @@ defmodule Nx.Defn do
 
     defaults = for {{:\\, _, [_, _]}, i} <- Enum.with_index(args), do: i
 
-    body =
-      quote do
-        transform(unquote(parameters_tuple), fn unquote(parameters_tuple) ->
-          apply(@__delegate_to, @__delegate_as, unquote(parameters_list))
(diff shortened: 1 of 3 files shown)


nx/lib/nx/defn.ex
```diff
@@ -267,6 +267,60 @@ defmodule Nx.Defn do
...
+        defndelegate calc(a, b), to: MyImpl
+
+        defndelegate calc_pair({a, b}), to: MyImpl
+
+        defndelegate calc_w_other_name(a, b), to: MyImpl, as: :calc
+
+        defndelegate calc_pair_tuple({a, b}), to: MyImpl, as: :calc_pair
+      end
+  """
+  defmacro defndelegate({name, meta, args}, opts) do
```
josevalim, line 300: Unfortunately, if we want to support all cases found in `Kernel`, which is what users would likely expect, this is slightly more complex. Perhaps it is best to copy and paste the implementation from Kernel and then replace `def` here by `defn`:

https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/kernel.ex#L5282

Just to give one example of the differences, we can do this in Elixir:

    foo = MyMod
    defdelegate bar(arg1, arg2), to: foo

We can also delegate with default arguments and so on. The good news is that Elixir's implementation has some code that has been deprecated, so we can remove those deprecated APIs for `defn`.

elbow-jason: For some reason it didn't occur to me to try the `defdelegate` implementation. I'll swap implementations and try it out.

nx/test/nx/defn_test.exs
```diff
@@ -1079,4 +1079,162 @@ defmodule Nx.DefnTest do
...
+      defmodule VarModuleDefndelegate do
+        import Nx.Defn
+
+        impl = VarModuleImpl
+        ex_fun = :elixir_call
+        nx_fun = :max_sum
+
+        defdelegate delegated_elixir_call(a, b), to: impl, as: ex_fun
+
+        defndelegate maximum_sum(a, b), to: impl, as: nx_fun
```
elbow-jason, line 1223: @josevalim the test for variable options :-)

nx/lib/nx/defn.ex
```diff
@@ -267,6 +267,136 @@ defmodule Nx.Defn do
...
+      else
+        {:{}, [], parameters_list}
+      end
+
+    defaults = for {{:\\, _, [_, _]}, i} <- Enum.with_index(args), do: i
+
+    body =
+      quote do
+        transform(unquote(parameters_tuple), fn unquote(parameters_tuple) ->
+          apply(@__delegate_to, @__delegate_as, unquote(parameters_list))
```
elbow-jason, line 332: @josevalim When the `to` and/or the `as` are variables (an not literals) I can't seem to get the variables into the scope of the newly defined function. My solution was to use module attributes.

 It feels hacky to me. Got a better idea?

nx/test/nx/defn_test.exs
```diff
@@ -1079,4 +1079,189 @@ defmodule Nx.DefnTest do
...
+      a = Nx.tensor([1, 2, 1])
+      b = Nx.tensor([2, 1, 1])
+      # apply to avoid compiler warnings
+      out = apply(VarModuleDefndelegate, :maximum_sum, [a, b])
+      assert Nx.to_scalar(out) == 5
+    end
+
+    import Nx.Defn
+
+    test "works with default arguments" do
```
elbow-jason, line 1242: The test for default args.

nx/lib/nx/defn.ex
josevalim: You should be able to generate the function call directly:

```elixir
iex(1)> mod = Nx 
Nx
iex(2)> fun = :add
:add
iex(3)> defmodule Foo do
...(3)> import Nx.Defn
...(3)> defn hello(arg) do
...(3)> unquote(mod).unquote(fun)(arg)
...(3)> end
...(3)> end
{:module, Foo,
 <<70, 79, 82, 49, 0, 0, 9, 208, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 1, 114,
   0, 0, 0, 33, 10, 69, 108, 105, 120, 105, 114, 46, 70, 111, 111, 8, 95, 95,
   105, 110, 102, 111, 95, 95, 10, 97, 116, ...>>, {:hello, 1}}
```

In other words, skip the apply and stuff. Instead, generate the `defn` function call as you would do by hand.

---

I just realized we have one big issue here: Nx requires your tuples to be matched upfront. So for example, you can't delegate to a function like `foo({a, b, c}, d)` by doing:

    defn calls_foo(tuple, d), do: foo(tuple, d)

You need to match on the tuple upfront too:

    defn calls_foo({a, b, c}, d), do: foo({a, b, c}, d)

It may be that we will need a different defdelegate function or at least check the types - which adds another dimension to the problem. In any case, I would go ahead with this patch, doing the simplification above, and have it around until we can solve this problem.

elbow-jason: The tuple-patterns-must-match issue is in the documentation.

elbow-jason: I tried many permutations trying to get this to work. 

I could only get one case to work.

The pertinent code that is being used:

```elixir
  to_ast = opts[:to]
  as_ast = opts[:as] || name

  quote do
    to = unquote(to_ast)
    as = unquote(as_ast)

    unquote(__MODULE__).__define__(
      __MODULE__,
      :defn,
      name,
      arity,
      defaults
    )

    @doc delegate_to: {to, as, arity}

    @__delegate_to to
    @__delegate_as as

    def unquote(name)(unquote_splicing(args)) do
      use Nx.Defn.Kernel
      # <case code goes here>
    end
  end
```

### Working cases

#### Use of transform with apply and module attributes

case:

```elixir
transform(unquote(parameters_tuple), fn unquote(parameters_tuple) ->
  apply(@__delegate_to, @__delegate_as, unquote(parameters_list))
end)
```

It works!

### Non-working cases

#### Direct use of module-level vars

case:

```elixir
to.as(unquote_splicing(parameters_list))
```

error:

```
** (CompileError) test/nx/defn_test.exs:1096: invalid numerical expression:

to().as({a, b})
```

#### Use of apply with module-level vars

case:

```elixir
apply(to, as, unquote(parameters_list))
```

error:

```
== Compilation error in file test/nx/defn_test.exs ==
** (CompileError) test/nx/defn_test.exs:1096: invalid numerical expression:

    to()
```

#### Use direct use of unquoting ast

case:

```elixir
unquote(to_ast).unquote(as_ast)(unquote_splicing(parameters_list))
```

error:

```
1) test defndelegate works with :to and :as as variables at call site (Nx.DefnTest)
test/nx/defn_test.exs:1202
** (CompileError) nofile:20: invalid numerical expression:

    impl().nx_fun(a, b)
```


#### Use transform with direct use of module-scoped variables

case:

```elixir
transform(unquote(parameters_tuple), fn unquote(parameters_tuple) ->
  to.as(unquote_splicing(parameters_list))
end)
```

error:

```
** (CompileError) test/nx/defn_test.exs:1096: undefined function to/0
```

#### Use transform with direct use of unquoting ast

case:

```elixir
transform(unquote(parameters_tuple), fn unquote(parameters_tuple) ->
  unquote(to_ast).unquote(as_ast)(unquote_splicing(parameters_list))
end)
```

error:

```
test/nx/defn_test.exs:1202
** (CompileError) nofile:20: undefined function impl/0
```

#### Use of transform with apply and module-scoped variables

case:

```elixir
transform(unquote(parameters_tuple), fn unquote(parameters_tuple) ->
  apply(to, as, unquote(parameters_list))
end)
````

error:

```
** (CompileError) test/nx/defn_test.exs:1096: undefined function as/0
```

elbow-jason: It seems like the module-level variables are not considered numerical and are also not in scope when used with `transform`.

josevalim: Thanks, I will have to play with it in order to understand exactly why the simplest approach does not work.

The other thing is that I am not sure asking people to repeat the tuple pattern is a good idea. It creates a coupling where you now have to change the delegator whenever the implementation changes, which goes against the purpose of defdelegate? I am thinking we can maybe have a mechanism to store the argument shape for `defn` functions and then automatically recover it for `defndelegate`.

This whole thing may also become more complex once we add supports for maps and datasets too. :/

elbow-jason: I think storing the signatures of `defn` functions as AST would solve the matching-signature issues. We could "zip" the target `defn` signature AST with the `defndelegate` signature AST replacing names of named parameters of the `defndelegate` with the named parameters/patterns of the `defn` and keep the defaults of the `defndelegate`.

WRT the complexity of other arg types/patterns, I think the signature-AST-zipping approach above would work for this as well.

If you think that we could use the approach described above it might be a good idea to close this PR for now and try again once the signature info is available. Specifically, this: `defdelegate/2 only accepts function parameters`. We want parity to how `defdelegate` is used and we can't do it right now without the signature info.

Thoughts?

josevalim: Also note we can solve this by removing the requirement tuples have to be matched upfront. So there are some uncertainties. Better to wait and decide later. :)

elbow-jason:
Closing for now.

Revisit when function signature restrictions for `defn` change or more info is stored/available for `defn` functions.

PR #275, opened 2021-02-23
Sv pytorch

josevalim:


nx/Makefile
```diff
@@ -0,0 +1,25 @@
+default_target: all
```
josevalim, line 1: Instead of changing the `nx` project, let's create a separate project at the root, a sibling to `nx` and `exla`. We just need to settle on the name.

nx/Makefile
```diff
@@ -0,0 +1,25 @@
...
+
+ERL_INCLUDE_PATH = $(shell erl -eval 'io:format("~s", [lists:concat([code:root_dir(), "/erts-", erlang:system_info(version), "/include"])])' -s init stop -noshell)
```
josevalim, line 3: elixir_make should set this environment variable for you.

nx/Makefile
```diff
@@ -0,0 +1,25 @@
...
+LIBTORCH_DIR = ../../libtorch
```
josevalim, line 4: Let's expect this environemnt to be set. If it is not, we can raise.

nx/Makefile
```diff
@@ -0,0 +1,25 @@
...
+	LDFLAGS += -flat_namespace -undefined suppress
+else  # Linux
+	CFLAGS += -shared
+	LDFLAGS += -lm
+endif
+
+
+all: c_src/pytorch.cpp
+	mkdir -p priv
+	@$(CXX) $(CFLAGS) c_src/pytorch.cpp -o priv/pytorch.so $(LDFLAGS)
```
josevalim, line 21: It is recommended to write to the app path priv directory instead. here are the recommendations from elixir make:

https://hexdocs.pm/elixir_make/Mix.Tasks.Compile.ElixirMake.html#module-compilation-artifacts-and-working-with-priv-directories

nx/Makefile
```diff
@@ -0,0 +1,25 @@
...
+
+
+clean:
+	rm priv/*.so
```
josevalim, line 25: If you store it in the mix path above, you don't need clean, because elixir will clean automatically for you.

nx/lib/nx/pytorch_backend.ex
```diff
@@ -0,0 +1,177 @@
...
+        inspect: 2
+      ]
+
+  for {fun, arity} <- funs do
+    args = Macro.generate_arguments(arity, __MODULE__)
+
+    def unquote(fun)(unquote_splicing(args)) do
+      raise "#{unquote(fun)}() is not supported"
+    end
+  end
```
josevalim, line 35: There is a cool trick you can do which is to move this at the end and skip the functions that were defined: https://github.com/elixir-nx/nx/blob/main/exla/lib/exla/device_backend.ex#L54-L60

nx/lib/nx/pytorch_backend.ex
```diff
@@ -0,0 +1,177 @@
...
+
+  ## Creation
+
+  @impl true
+  def tensor(tensor) do
```
josevalim, line 40: Please rebase because this callback no longer exists. :)

nx/lib/nx/pytorch_backend.ex
```diff
@@ -0,0 +1,177 @@
+defmodule Nx.PytorchBackend do
+  @behaviour Nx.Backend
+
+  defstruct [:ref]
+
+  alias Nx.Pytorch.NIF
+
+  import Nx.Shared
```
josevalim, line 8: You are not supposed to use this module once you move this to a separate, because it is private for Nx. I see you have using it for iota though, so I will have a suggestion on how to refactor it.

nx/lib/nx/pytorch_backend.ex
```diff
@@ -0,0 +1,177 @@
...
+
+    data =
+      for _ <- 1..cycles,
+          i <- 0..(dim - 1),
+          _ <- 1..repeat_blocks,
+          into: "",
+          do: number_to_binary(i, type)
+
+    t = NIF.from_blob(data, shape, torch_type(type))
+    from_ref(out, t)
```
josevalim, line 92: So there is a faster implementation on top of aten which is by doing this. Imagine you have a tensor `t`, with `shape`, and you want iota on `axis`. You can do this:

```elixir
# gets the size of iota
dim = elem(shape, axis)

# build the iota in one dimension
aten = NIF.arange(0, dim)

# reshape the tensor above to be have shape where everything is 1, except for dim
reshape = Tuple.duplicate(1, Nx.rank(shape)) |> put_elem(axis, dim)
aten = NIF.reshape(aten, reshape)

# Now broadcast the tensor using the original shape
aten = NIF.broadcast(aten, shape)
```

You can see this working with Nx too:

```elixir
iex> Nx.iota({3}) |> Nx.reshape({1, 1, 3}) |> Nx.broadcast({3, 3, 3}) == Nx.iota({3, 3, 3}, axis: 2)
true

iex> Nx.iota({3}) |> Nx.reshape({3, 1, 1}) |> Nx.broadcast({3, 3, 3}) == Nx.iota({3, 3, 3}, axis: 0)
true
```

This will most likely be more efficient on aten side, especially if it lazily expands broadcasts. :)

nx/lib/nx/pytorch_backend.ex
```diff
@@ -0,0 +1,177 @@
...
+  end
+
+  ## Creation
+
+  @impl true
+  def tensor(tensor) do
+    tensor
+  end
+
+  def torch_type({:u, 8}), do: :byte
```
josevalim, line 44: Don't they support higher unsigned types? 🤔

nx/lib/nx/pytorch_backend.ex
```diff
@@ -0,0 +1,177 @@
...
+  def torch_type({:s, 32}), do: :int
+  def torch_type({:s, 64}), do: :long
+  def torch_type({:f, 16}), do: :half
+  def torch_type({:f, 32}), do: :float
+  def torch_type({:f, 64}), do: :double
+
+  @impl true
+  def iota(out, axis \\ nil)
+
+  def iota(%{shape: {}, type: type} = out, nil), do: iota(%{out | shape: {1}}, nil)
```
josevalim, line 56: This will unfortunately return a tensor with the wrong shape (the shape will be {1}, it should be {}).

nx/c_src/pytorch.cpp
```diff
@@ -0,0 +1,327 @@
...
+  at::Tensor *tensorPtr;
+
+  tensorPtr = (at::Tensor *)enif_alloc_resource(TENSOR_TYPE, sizeof(at::Tensor));
+  if (tensorPtr == NULL)
+    return enif_make_badarg(env);
+
+  new (tensorPtr) at::Tensor(tensor.variable_data());
+
+  ret = enif_make_resource(env, tensorPtr);
+  enif_keep_resource(tensorPtr);
```
josevalim, line 23: You don't need this keep reosurce call. All `enif_make_resource` are by default automatically kept, so if you have this additional `enif_keep_resource` call, it means it won't ever be garbage collected.

nx/lib/nx/pytorch_backend.ex
```diff
@@ -0,0 +1,177 @@
...
+
+  @impl true
+  def cholesky(%{type: output_type, shape: {rows, cols}} = out, tensor) do
+    t = NIF.cholesky(tensor.data.ref)
+    from_ref(out, t)
+  end
+
+  @impl true
+  def backend_deallocate(%Nx.Tensor{data: _data}) do
+    :ok
```
josevalim, line 167: In this one you are supposed to deallocate the tensor from memory in a way that any future operation with that tensor raises. :)

nx/lib/nx/pytorch_backend.ex
```diff
@@ -0,0 +1,177 @@
...
+  end
+
+  @impl true
+  def from_binary(%{type: type, shape: shape} = out, binary, _opts) do
+    t = NIF.from_blob(binary, shape, torch_type(type))
+    from_ref(out, t)
+  end
+
+  @impl true
+  def to_binary(%{data: %{ref: ref}}, _opts), do: NIF.to_blob(ref)
```
josevalim, line 125: So I have a question. Should we automatically read from memory on to_binary? Or should we force the user to do an explicit conversion? EXLA requires the user to do an explicit conversion by calling `backend_transfer`. You can see `EXLA.DeviceBackend` for an example.

If we require an explicit conversion, then inspect won't work... but i can help with that.

nx/c_src/pytorch.cpp
```diff
@@ -0,0 +1,327 @@
...
+}
+
+ERL_NIF_TERM from_blob(ErlNifEnv *env, int argc, const ERL_NIF_TERM argv[])
+{
+  ErlNifBinary blob;
+  std::vector<int64_t> shape;
+  std::string type;
+
+  if (!nx::nif::get_tuple(env, argv[1], shape))
+    return enif_make_badarg(env);
```
josevalim, line 54: If this fails is likely because the tensor has been deallocated. So try to raise something like `{:error, :deallocated}`. What we do in EXLA is that al NIF operations return `{:ok, value} | {:error, :deallocated}` and then we call a function called unwrap!() after every NIF call.

Something like:

```elixir
defp unwrap!({:ok, tensor}), do: tensor
defp unwrap!({:error, :deallocated}), do: raise(ArgumentError, "the tensor has been deallocated")
```

nx/test/nx/pytorch_test.exs
```diff
@@ -0,0 +1,25 @@
...
+      assert Nx.to_flat_list(c) == [6.0, 8.0, 10.0, 12.0]
+    end
+
+    test "dot" do
+      a = Nx.tensor([[1, 2], [3, 4]], backend: PytorchBackend)
+      b = Nx.tensor([[5, 6], [7, 8]], backend: PytorchBackend)
+
+      c = Nx.dot(a, b)
+
+      assert Nx.to_flat_list(c) == [19, 22, 43, 50]
```
josevalim, line 22: Ideally do this in the tests:

    assert Nx.dot(a, b) |> Nx.backend_transfer() == Nx.tensor([19, 22, 43, 50])

It is best to compare tensors with tensors because we will also check that the shape, names, and types are correct.

nx/lib/nx/pytorch_backend.ex
```diff
@@ -0,0 +1,177 @@
...
+  def squeeze(out, tensor, _axes) do
+    t = NIF.squeeze(tensor.data.ref)
+    from_ref(out, t)
+  end
+
+  ## Ops
+
+  @impl true
+  def add(out, left, right) do
+    t = NIF.add(left.data.ref, right.data.ref)
```
josevalim, line 143: Please add tests for mixed types. For example, what happens if we add {:u, 8} with {:s, 32}. Does aten automatically cast? Or do we need to do it?

In EXLA we need to explicitly convert the types before calling aten. So for example, if you are adding {:u, 8} and {:f, 32}, the output type in `out` is `{:f, 32}`, so you will need to convert {:u, 8} to {:f, 32} and then perform the addition.

nx/Makefile
versilov: Done, `torchx` created.

nx/Makefile
versilov: Done.

nx/lib/nx/pytorch_backend.ex
versilov: It's strange, but there are no definitions of higher unsigned types anywhere in the include files of the lib.
So, i.e. `kUInt32` is not available.

Here's what we've got in `types.h:

```
/// Fixed width dtypes.
constexpr auto kUInt8 = at::kByte;
constexpr auto kInt8 = at::kChar;
constexpr auto kInt16 = at::kShort;
constexpr auto kInt32 = at::kInt;
constexpr auto kInt64 = at::kLong;
constexpr auto kFloat16 = at::kHalf;
constexpr auto kFloat32 = at::kFloat;
constexpr auto kFloat64 = at::kDouble;
```

versilov: Correct, PyTorch has 12 data types, and only 8bit int has unsigned version: https://pytorch.org/docs/stable/tensor_attributes.html

So, I'll map Nx unsigned ints to PyTorch's signed types for now. Should we issue any warning when unsigned PyTorch tensor is requested?

josevalim: I would raise for now, if a non-existing type is requested, then we can revisit it later. :)

versilov: OK, changed to rising.

nx/c_src/pytorch.cpp
versilov: Here it fails only if not tuple was sent.
I've changed all operations to return tuples, which go into `unwrap/1`.

nx/lib/nx/pytorch_backend.ex
versilov: But I see it here: 
https://github.com/elixir-nx/nx/blob/e5d51f4ad04e7cd6612289093bc0dfbca99e9f84/nx/lib/nx/backend.ex#L46

And still don't get what's it's purpose:)

josevalim: That was a left-over. I have removed it!

nx/lib/nx/pytorch_backend.ex
versilov: libtorch does type conversion automatically, as far as I saw. Will write tests for these cases, of course.

josevalim:
Closing this in favor of the new PR!

PR #276, opened 2021-02-23
Add gradient for inverse hyperbolic functions

0urobor0s:
This commit adds gradient for the following inverse hyperbolic functions:

- Arcsinh
- Arccosh
- Arctanh

PS: Shouldn't this functions drop the c? Wikipedia has some information on this. [Notation](https://en.wikipedia.org/wiki/Inverse_hyperbolic_functions#Notation)

### DIFF
--- a/nx/test/nx/defn/grad_test.exs
+++ b/nx/test/nx/defn/grad_test.exs
@@ -509,6 +509,29 @@ defmodule Nx.Defn.GradTest do
     end
   end
 
+  describe "inverse hyperbolic functions" do
+    defn grad_arcsinh(t), do: grad(t, Nx.arcsinh(t))
+    defn grad_arccosh(t), do: grad(t, Nx.arccosh(t))
+    defn grad_arctanh(t), do: grad(t, Nx.arctanh(t))
+
+    test "computes gradient of inverse hyperbolic functions" do
+      for _ <- @iters do
+        t = Nx.random_uniform({}, -100.0, 100.0, type: {:f, 64})
+        check_grads!(&Nx.arcsinh/1, &grad_arcsinh/1, t, eps: 0.1)
+      end
+
+      for _ <- @iters do
+        t = Nx.random_uniform({}, 1.01, 100.0, type: {:f, 64})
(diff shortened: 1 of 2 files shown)


nx/test/nx/defn/grad_test.exs
```diff
@@ -509,6 +509,29 @@ defmodule Nx.Defn.GradTest do
...
 
+  describe "inverse hyperbolic functions" do
+    defn grad_arcsinh(t), do: grad(t, Nx.arcsinh(t))
+    defn grad_arccosh(t), do: grad(t, Nx.arccosh(t))
+    defn grad_arctanh(t), do: grad(t, Nx.arctanh(t))
+
+    test "computes gradient of inverse hyperbolic functions" do
+      for _ <- @iters do
+        t = Nx.random_uniform({}, -100.0, 100.0, type: {:f, 64})
+        check_grads!(&Nx.arcsinh/1, &grad_arcsinh/1, t, eps: 0.1)
```
josevalim, line 520: I assume a lower eps fails frequently?

josevalim:
> PS: Shouldn't this functions drop the c? Wikipedia has some information on this. Notation

Good call. I will rename them to asin, acos, etc after I merge this. This is the notation used by Julia, Erlang, and C's math.

josevalim:
Merging so I can do the rename, we can revisit the EPS later accordingly.

josevalim:

nx/test/nx/defn/grad_test.exs
0urobor0s: I think it's fine to lower it or remove it to use the default value.

0urobor0s:
> > PS: Shouldn't this functions drop the c? Wikipedia has some information on this. Notation
> 
> Good call. I will rename them to asin, acos, etc after I merge this. This is the notation used by Julia, Erlang, and C's math.

Apparently in Wikipedia, following the ISO 80000-2 it should be, ar<func>h, but following Erlang or C might be easier to for developers as it reduces the things to remember.

PR #281, opened 2021-02-25
Torchx

versilov:


nx/.gitignore
```diff
@@ -18,3 +18,5 @@ erl_crash.dump
 
 # Ignore package tarball (built via "mix hex.build").
 nx-*.tar
+
+/priv/*.so
```
josevalim, line 22: Please revert this change now. :D

nx/mix.exs
```diff
@@ -20,9 +20,7 @@ defmodule Nx.MixProject do
   end
 
   defp deps do
-    [
-      {:ex_doc, "~> 0.23", only: :dev}
-    ]
+    [{:elixir_make, "~> 0.6"}, {:ex_doc, "~> 0.23", only: :dev}]
```
josevalim, line 23: And revert this change! :D

nx/mix.lock
```diff
@@ -1,5 +1,6 @@
 %{
   "earmark_parser": {:hex, :earmark_parser, "1.4.12", "b245e875ec0a311a342320da0551da407d9d2b65d98f7a9597ae078615af3449", [:mix], [], "hexpm", "711e2cc4d64abb7d566d43f54b78f7dc129308a63bc103fbd88550d2174b3160"},
+  "elixir_make": {:hex, :elixir_make, "0.6.2", "7dffacd77dec4c37b39af867cedaabb0b59f6a871f89722c25b28fcd4bd70530", [:mix], [], "hexpm", "03e49eadda22526a7e5279d53321d1cced6552f344ba4e03e619063de75348d9"},
```
josevalim, line 3: And revert this change too!

torchx/.formatter.exs
```diff
@@ -0,0 +1,12 @@
...
+locals_without_parens = [
+  defn: 2,
+  defnp: 2
+]
+
+[
+  locals_without_parens: locals_without_parens,
+  export: [locals_without_parens: locals_without_parens],
+  inputs: ["{mix,.formatter}.exs", "{lib,test}/**/*.{ex,exs}"]
+]
```
josevalim, line 12: We should import the definitions from nx instead!

```suggestion
# Used by "mix format"

[
  import_deps: [:nx],
  inputs: ["{mix,.formatter}.exs", "{bench,examples,config,lib,test}/**/*.{ex,exs}"]
]
```

torchx/.gitignore
```diff
@@ -0,0 +1,22 @@
...
+# Ignore .fetch files in case you like to edit your project deps locally.
+/.fetch
+
+# If the VM crashes, it generates a dump, let's ignore it too.
+erl_crash.dump
+
+# Ignore package tarball (built via "mix hex.build").
+nx-*.tar
+
+/priv/*.so
```
josevalim, line 22: We probably don't need this because we are writing to MIX_APP_PATH, right?

torchx/.gitignore
```diff
@@ -0,0 +1,22 @@
...
+/doc/
+
+# Ignore .fetch files in case you like to edit your project deps locally.
+/.fetch
+
+# If the VM crashes, it generates a dump, let's ignore it too.
+erl_crash.dump
+
+# Ignore package tarball (built via "mix hex.build").
+nx-*.tar
```
josevalim, line 20:
```suggestion
torchx-*.tar
```

torchx/Makefile
```diff
@@ -0,0 +1,26 @@
+
+ifndef LIBTORCH_DIR
+	$(error LIBTORCH_DIR env var is not set. It should point to the directory, where libtorch is installed.)
```
josevalim, line 3: Also include the URL to download libtorch too. Something like: "You can download libtorch here: https://pytorch.org/get-started/locally/"

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,219 @@
...
+  @behaviour Nx.Backend
+
+  defstruct [:ref]
+
+  alias Torchx.NIF
+
+  def torch_type({:u, 8}), do: :byte
+
+  def torch_type({:u, 16}),
+    do: raise(ArgumentError, "PyTorch does not support unsigned 16 bit integer.")
```
josevalim, line 11: Nitpicks:

1. Use Torchx instead of PyTorch
2. Error messages in Elixir do not finish with "."

Suggestion:

```suggestion
    do: raise(ArgumentError, "Torchx does not support unsigned 16 bit integer")
```

Please change the other ones accordingly! :heart:

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,219 @@
...
+  def torch_type({:s, 64}), do: :long
+  def torch_type({:bf, 16}), do: :brain
+  def torch_type({:f, 16}), do: :half
+  def torch_type({:f, 32}), do: :float
+  def torch_type({:f, 64}), do: :double
+
+  ## Creation
+
+  @impl true
+  def eye(%{shape: {n, n}, type: type} = out) do
```
josevalim, line 31: Let's pattern match on the struct in all of the function heads below. It will help us find bugs (like mistyped field names):

```suggestion
  alias Nx.Tensor, as: T

  def eye(%T{shape: {n, n}, type: type} = out) do
```

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,219 @@
...
+    raise "Operation to_binary is not supported on Torchx.Backend. " <>
+            "You must first transfer the tensor to Elixir by calling Nx.backend_transfer/1"
+  end
+
+  defp to_blob(%{type: {_, elem_size}, data: %{ref: ref}} = tensor, limit \\ nil) do
+    if(limit,
+      do: NIF.to_blob(ref, limit),
+      else: NIF.to_blob(ref)
+    )
+  end
```
josevalim, line 97:
```suggestion
  defp to_blob(tensor, limit \\ nil)
  defp to_blob(%T{data: %{ref: ref}} = tensor, nil), do: NIF.to_blob(ref)
  defp to_blob(%T{data: %{ref: ref}} = tensor, limit), do: NIF.to_blob(ref, limit)
```

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,219 @@
...
+    else
+      raise "Tensor is located on #{device(tensor)} device, so direct to_binary/1 operation is not supported. " <>
+              "You must first transfer the tensor to Elixir by calling Nx.backend_transfer/1"
+    end
+  end
+
+  defp unwrap!({:ok, result}), do: result
+
+  defp unwrap!({:error, error}),
+    do: raise(RuntimeError, "PyTorch: " <> to_string(error))
```
josevalim, line 194:
```suggestion
    do: raise(RuntimeError, "Torchx: " <> to_string(error))
```

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,219 @@
...
+
+  defp from_ref({:ok, ref}, t), do: from_ref(ref, t)
+  defp from_ref({:error, _error} = input, _t), do: unwrap!(input)
+  defp from_ref({ref1, ref2}, {t1, t2}), do: {from_ref(ref1, t1), from_ref(ref2, t2)}
+  defp from_ref([ref | list], t), do: [from_ref(ref, t) | from_ref(list, t)]
+  defp from_ref([], _t), do: []
```
josevalim, line 200: I would have those as distinct functions, so the code is clearer: `from_pair_ref` when working with tuples, `from_list_ref` when working with lists, and so on. This would allow us to assert what is the value we are expecting from the NIF.

Examples:

    defp from_ref(maybe_ref, t), do: maybe_ref |> unwrap!() |> to_tensor(t)

    defp from_pair_ref(maybe_ref, t) do
      {left, right} = unwrap!(maybe_ref)
      {to_tensor(left, t), to_tensor(right, t)}
    end

    defp to_tensor(ref, t), do: %{t | data: %__MODULE__{ref: ref}}

torchx/mix.exs
```diff
@@ -0,0 +1,46 @@
...
+      name: "Torchx",
+      version: @version,
+      elixir: "~> 1.11",
+      deps: deps(),
+      docs: docs(),
+      compilers: [:elixir_make] ++ Mix.compilers(),
+      make_env: %{
+        "MIX_CURRENT_PATH" => File.cwd!(),
+        "ERTS_VERSION" => List.to_string(:erlang.system_info(:version))
+      }
```
josevalim, line 19: Those do not seem to be used:

```suggestion
```

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,219 @@
...
+  defp from_ref(ref, t) when is_reference(ref), do: %{t | data: %__MODULE__{ref: ref}}
+
+  defp device(%{data: %{ref: ref}}), do: NIF.device(ref) |> unwrap!() |> to_string()
```
josevalim, line 203: What is the device in this case? Is it an atom? Or an Erlang string? If the latter, use `List.to_string()` here for clarity!

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,219 @@
...
+  def inspect(%{type: {_, elem_size}} = tensor, inspect_opts) do
+    limit = if(inspect_opts.limit == :infinity, do: nil, else: inspect_opts.limit + 1)
+
+    if on_cpu?(tensor) do
+      byte_size = nbytes(tensor)
+      byte_limit = limit && limit * div(elem_size, 8)
+
+      if(min(byte_limit, byte_size) > @big_tensor_threshold) do
+        raise "Tensor is too big (#{byte_size} bytes) for to_binary/1 operation." <>
+                "You must first transfer the tensor to Elixir by calling Nx.backend_transfer/1"
```
josevalim, line 180: Instead of raising, we should probably return a message:

    "Torchx tensor is too large to inspect. Explicitly transfer the tensor by calling Nx.backend_transfer/1"

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,219 @@
...
+
+  @impl true
+  def inspect(%{type: {_, elem_size}} = tensor, inspect_opts) do
+    limit = if(inspect_opts.limit == :infinity, do: nil, else: inspect_opts.limit + 1)
+
+    if on_cpu?(tensor) do
+      byte_size = nbytes(tensor)
+      byte_limit = limit && limit * div(elem_size, 8)
+
+      if(min(byte_limit, byte_size) > @big_tensor_threshold) do
```
josevalim, line 178: Nitpick: no parens around if:

```suggestion
      if min(byte_limit, byte_size) > @big_tensor_threshold do
```

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,219 @@
...
+        raise "Tensor is too big (#{byte_size} bytes) for to_binary/1 operation." <>
+                "You must first transfer the tensor to Elixir by calling Nx.backend_transfer/1"
+      else
+        binary = to_blob(tensor, limit)
+        Nx.Backend.inspect(tensor, binary, inspect_opts)
+      end
+    else
+      raise "Tensor is located on #{device(tensor)} device, so direct to_binary/1 operation is not supported. " <>
+              "You must first transfer the tensor to Elixir by calling Nx.backend_transfer/1"
+    end
```
josevalim, line 188: I also would like to add a header to Torchx.Backend, so it is printed like this:

```
#Nx.Tensor<
  s64[2][2]
  Torchx.Backend
  [
    ...,
    ...,
  ]
>
```

You can do this by getting the result of this function right now and wrapping it:

    concat(["Torchx.Backend", line(), result])

torchx/c_src/torchx.cpp
```diff
@@ -0,0 +1,420 @@
...
+    F(squeeze, 2),
+    F(squeeze, 1),
+    F(broadcast_to, 2),
+    F(transpose, 3),
+    F(add, 2),
+    F(dot, 2),
+    F(cholesky, 1),
+    F(cholesky, 2),
+    F(qr, 1),
+    F(qr, 2)};
```
josevalim, line 418: We should split those functions. Most of them should be tagged as CPU bound Dirty NIFs, except for a handful (`type`, `device`, `nbytes`).

In the future, we want to have both CPU and IO dirty NIFs (the IO dirty ones for GPU stuff) - and we will need a way to control which one we want to call.

torchx/lib/torchx/backend.ex
versilov: The idea here was to mark libtorch native exceptions, which I catch in NIF, wrap in tuple and send out.
But your change is correct — not all error messages come from libtorch exceptions.
Should we mark them somehow, to distinguish between our own error messages and those coming from external world?

josevalim: I don't think it is relevant to users to know if it is from NIF or Elixir, so I think prefixing with Torchx is fine. :)

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,218 @@
...
+
+  ## Transfer
+
+  @impl true
+  def to_batched_list(%T{shape: shape} = out, %T{data: %TB{ref: ref}}),
+    do: NIF.split(ref, elem(shape, 0)) |> from_list_ref(out)
+
+  @impl true
+  def to_binary(_tensor, _limit \\ nil) do
+    raise "Operation to_binary is not supported on Torchx.Backend. " <>
```
josevalim, line 78:
```suggestion
    raise "operation to_binary is not supported on Torchx.Backend. " <>
```

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,218 @@
...
+      else
+        alias Inspect.Algebra, as: IA
+
+        binary = to_blob(tensor, limit)
+        result = Nx.Backend.inspect(tensor, binary, inspect_opts)
+        IA.concat(["Torchx.Backend", IA.line(), result])
+      end
+    else
+      raise "Tensor is located on #{device(tensor)} device, so direct to_binary/1 operation is not supported. " <>
+              "You must first transfer the tensor to Elixir by calling Nx.backend_transfer/1"
```
josevalim, line 178: Same as above, make sure this returns a string instead of raising:

```suggestion
              "Tensors on the GPU cannot be inspected. Explicitly transfer the tensor by calling Nx.backend_transfer/1"
```

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,218 @@
...
+      byte_limit = limit && limit * div(elem_size, 8)
+
+      if min(byte_limit, byte_size) > @big_tensor_threshold do
+        "Torchx tensor is too large to inspect. Explicitly transfer the tensor by calling Nx.backend_transfer/1"
+      else
+        alias Inspect.Algebra, as: IA
+
+        binary = to_blob(tensor, limit)
+        result = Nx.Backend.inspect(tensor, binary, inspect_opts)
+        IA.concat(["Torchx.Backend", IA.line(), result])
```
josevalim, line 174: Move this outside of the `if` block, because we always want to add this prefix. :) Maybe we even want to do this:

    IA.concat(["Torchx.Backend (#{device})", IA.line(), result])

? :)

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,218 @@
...
+      end
+    else
+      raise "Tensor is located on #{device(tensor)} device, so direct to_binary/1 operation is not supported. " <>
+              "You must first transfer the tensor to Elixir by calling Nx.backend_transfer/1"
+    end
+  end
+
```
josevalim, line 181: Nitpick:

```suggestion

  ## Helpers

```

torchx/c_src/torchx.cpp
versilov: I added two definitions for each fun — the default is CPU dirty and _io version, which is IO dirty.

torchx/lib/torchx/nif.ex
```diff
@@ -0,0 +1,70 @@
...
+  dnif(qr(tensor))
+  dnif(qr(tensor, reduced))
+
+  def type(_tensor), do: nif_error(__ENV__.function)
+  def device(_tensor), do: nif_error(__ENV__.function)
+  def nbytes(_tensor), do: nif_error(__ENV__.function)
+
+  defp nif_error({name, arity}) do
+    raise "failed to load implementation of #{inspect(__MODULE__)}.#{name}/#{arity}"
+  end
```
josevalim, line 69: I believe you got this from EXLA because EXLA was wrong. We really need to use :erlang.nif_error otherwise Dialyzer will complain. Please check EXLA master for a reference. :)

torchx/lib/torchx/nif.ex
```diff
@@ -0,0 +1,70 @@
+defmodule Torchx.NIF.Macro do
+  defmacro dnif(call) do
+    {name, args} = Macro.decompose_call(call)
+    name_io = name |> Atom.to_string() |> Kernel.<>("_io") |> String.to_atom()
```
josevalim, line 4:
```suggestion
    name_io = :"#{name}_io"
```

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,261 @@
...
+  @impl true
+  def transpose(out, tensor, axes), do: IO.inspect(axes)
+  # do: NIF.transpose(tensor, dim0, dim1) |> from_ref(out)
+
+  ## Ops
+
+  @impl true
+  def add(out, left, right) do
+    NIF.add(left.data.ref, right.data.ref) |> from_ref(out)
```
josevalim, line 140: You also need to handle the case where this happens:

    Nx.add(tensor_from_torchx, Nx.tensor([1, 2, 3]))

So my suggestion is that instead of `left.data.ref`, you should do `to_ref`. In there you will check if the backend is a Torchx.Backend, so you get the ref. If it is a Nx.BinaryBackend, you convert it to Torchx and return its ref. :)

torchx/c_src/torchx.cpp
```diff
@@ -0,0 +1,583 @@
...
+
+    at::Tensor result = at::broadcast_to(aten, shape).clone();
+
+    return nx::nif::ok(env, create_tensor_resource(env, result));
+  }
+  catch (c10::Error error)
+  {
+    return nx::nif::error(env, error.msg().c_str());
+  }
+}
```
josevalim, line 263: Personally, I would prefer for this to be implemented in Elixir and keep the NIF to be simple wrappers around libtorch. After all, it is easier to maintain Elixir code than C++ one, but there is no problem if you would prefer to keep this function as is.

torchx/test/torchx_test.exs
```diff
@@ -0,0 +1,144 @@
...
+
+    assert(Nx.backend_transfer(r) == binary_r)
+  end
+
+  describe "binary tensor-tensor ops" do
+    for op <- @ops ++ @logical_ops,
+        type_a <- @types,
+        type_b <- @types,
+        not ((op in [:divide, :remainder, :atan2, :power] and (type_a == {:bf, 16} and type_b in @bf16_and_ints)) or
+               (type_b == {:bf, 16} and type_a in @bf16_and_ints)) do
```
josevalim, line 48: Instead of having the `not`, break them in subgroups: has a module attribute with only `[:divide, :remainder, :atan2, :power]` so you can check these rules.

torchx/test/torchx_test.exs
```diff
@@ -0,0 +1,144 @@
...
+
+        assert Nx.backend_transfer(c) ==
+                 Nx.tensor([[1.0, 64.0], [2192.0, 65536.0]],
+                   type: {:bf, 16}
+                 )
+      end
+    end
+  end
+
+  describe "vectors" do
```
josevalim, line 133: What do you mean by vectors in this case? Outer only works with vectors for Pytorch?

josevalim (commented):
I have done another pass and dropped some minor comments.

I think the only change that we need to do to merge this is to support `Nx.add(torchx_backend, binary_backend)`. To support it, you should stop doing `tensor.data.ref` (and stop pattern matching on it) and instead implement a `to_ref` private function that converts the tensor to libtorch if in a BinaryBackend, otherwise it is guaranteed it is a Torchx tensor.

Then I think we can merge this PR even with an incomplete test suite. Then the next PR is to implement the tests for the other functionality that you have. :)

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,268 @@
...
+    [:add, :subtract, :multiply, :power, :remainder, :divide, :atan2, :min, :max, :quotient] ++
+      [:bitwise_and, :bitwise_or, :bitwise_xor, :left_shift, :right_shift] ++
+      [:equal, :not_equal, :greater, :less, :greater_equal, :less_equal] ++
+      [:logical_and, :logical_or, :logical_xor] ++
+      [:outer]
+
+
+  for op <- binary_ops do
+    @impl true
+    def unquote(op)(out, left, right) when is_number(right) do
```
josevalim, line 149: I don't think it will ever be a number. So we can likely remove this clause.

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,268 @@
...
+      maybe_ref
+      |> unwrap!()
+      |> Enum.map(&to_tensor(&1, t))
+
+  defp to_tensor(ref, %{type: out_type} = t),
+    do: %T{t | data: %__MODULE__{ref: maybe_cast_type(ref, out_type)}}
+
+  defp maybe_cast_type(ref, type) do
+    if from_torch_type(unwrap!(NIF.type(ref))) != type do
+      NIF.to_type(ref, torch_type(type)) |> unwrap!()
```
josevalim, line 246: When is this type casting supposed to happen? I think it would be better to not do this automatically... and only call maybe_cast_type on the few cases (for example, directly inside `def add` or wherever else expect them  we identify them to be different). Thoughts?

torchx/lib/torchx/backend.ex
versilov: So, all scalars in ops will be wrapped in tensors, like this?
```
iex(1)> Nx.tensor(5.0)
#Nx.Tensor<
  f32
  5.0
>
```

torchx/lib/torchx/backend.ex
versilov: This was added to always match with the `out` tensor expected type.
After changing type merging policies we still have some mismatches with PyTorch type merging. There are two cases for now:
1. Bitwise ops. PyTorch outputs the smallest type (e.g. `xor(s16, s64) == s16)`, and we promote to `s64`.
2. Power. PyTorch always promotes to `f64`, while we stay within general rule (e.g. `power(s16, s32) == s32`).
In both cases PyTorch policy makes sense, because we should throw away excess bits, if operands sizes do not match in bitwise. And in power we can have neg values in right operand, thus getting fractions in the result.
I agree with you, that we should add explicit type cast in these two cases.

versilov: But wait, we should not do type casts in these cases, but change the `out` tensor expected type instead!:)

torchx/lib/torchx/backend.ex
versilov: Yes, I see I can:)
Me better do not write comments while still sleeping:)

torchx/test/torchx_test.exs
versilov: Yes, only with 1-D tensors. Should we reshape input tensors into vectors for it?

versilov: Yes, only with 1-D tensors. Should we reshape input tensors into vectors for it?

torchx/c_src/torchx.cpp
versilov: Done!

nx/lib/nx/shared.ex
```diff
@@ -329,6 +329,13 @@ defmodule Nx.Shared do
   def binary_type(a, b) when is_number(b), do: Nx.Type.merge_scalar(type(a), b)
   def binary_type(a, b), do: Nx.Type.merge(type(a), type(b))
 
+  def binary_type(_a, _b, :power), do: {:f, 64}
+
+  def binary_type(a, _b, op) when op in [:bitwise_and, :bitwise_or, :bitwise_xor],
+    do: type(a)
```
josevalim, line 335: Instead of changing it here, change the code in `Nx` so bitwise_and, bitwise_or, and bitwise_xor call a new function called `element_wise_bit_op` that has the new type logic. We will also need a new `Nx.Type.merge_bitwise/2` function?

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,274 @@
...
+  def as_type(%T{type: type} = out, %T{} = t),
+    do: NIF.to_type(to_ref(t), torch_type(type)) |> from_ref(out)
+
+  @impl true
+  def squeeze(out, tensor, _axes) do
+    NIF.squeeze(to_ref(tensor)) |> from_ref(out)
+  end
+
+  @impl true
+  def transpose(_out, _tensor, axes), do: IO.inspect(axes)
```
josevalim, line 145: Should we remove this one for now since the implementation is incomplete?

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,274 @@
...
+      |> Enum.map(&to_tensor(&1, t))
+
+  defp to_ref(%T{data: %TB{ref: ref}}), do: ref
+
+  defp to_ref(%T{} = tensor),
+    do: Nx.backend_transfer(tensor, TB) |> to_ref()
+
+  # Update out tensor type here to mark the cases where our type policy mismatches with the libtorch's one.
+  defp to_tensor(ref, %T{} = t),
+    do: %T{t | data: %__MODULE__{ref: ref}, type: from_torch_type(unwrap!(NIF.type(ref)))}
```
josevalim, line 256: I think instead we want to do something like this:

```suggestion
  defp to_tensor(ref, %T{type: type} = t) do
    current_type = ref |> NIF.type() |> unwrap!() |> from_torch_type()

    if current != type do
      raise "type mismatch in Torchx: expected #{inspect(type)}, got: #{inspect(current_type)}. " <>
              "Please report this bug"
    end
    
    %{t | data: %__MODULE__{ref: ref}}
  end
```

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,274 @@
...
+  ## All remaining callbacks
+
+  funs = Nx.Backend.behaviour_info(:callbacks) -- Module.definitions_in(__MODULE__, :def)
+
+  for {fun, arity} <- funs do
+    args = Macro.generate_arguments(arity, __MODULE__)
+
+    @impl true
+    def unquote(fun)(unquote_splicing(args)) do
+      raise "operation #{unquote(fun)} is not supported on Torchx.Backend."
```
josevalim, line 271:
```suggestion
      raise "operation #{unquote(fun)} is not supported on Torchx.Backend"
```

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,274 @@
...
+
+    IA.concat(["Torchx.Backend(#{device(tensor)})", IA.line(), result])
+  end
+
+  ## Helpers
+
+  defp unwrap!({:ok, result}), do: result
+
+  defp unwrap!({:error, error}),
+    do: raise(RuntimeError, "Torchx: " <> List.to_string(error))
```
josevalim, line 234:
```suggestion
  defp unwrap!({:ok, result}), do: result
  defp unwrap!({:error, error}), do: raise("Torchx: " <> List.to_string(error))
```

torchx/test/torchx_test.exs
```diff
@@ -0,0 +1,157 @@
...
+    a = tt(data_a, type_a)
+    b = tt(data_b, type_b)
+
+    c = Kernel.apply(Nx, op, [a, b])
+
+    binary_a = Nx.backend_transfer(a, Nx.BinaryBackend)
+    binary_b = Nx.backend_transfer(b, Nx.BinaryBackend)
+    binary_c = Kernel.apply(Nx, op, [binary_a, binary_b])
+
+    assert(Nx.backend_transfer(c) == binary_c)
```
josevalim, line 41: Let's also add a test for mixed:

```suggestion
    c = Kernel.apply(Nx, op, [a, b])
    mixed_c = Kernel.apply(Nx, op, [a, binary_b])

    binary_a = Nx.backend_transfer(a, Nx.BinaryBackend)
    binary_b = Nx.backend_transfer(b, Nx.BinaryBackend)
    binary_c = Kernel.apply(Nx, op, [binary_a, binary_b])

    assert Nx.backend_transfer(c) == binary_c
    assert Nx.backend_transfer(mixed_c) == binary_c
```

torchx/c_src/torchx.cpp
```diff
@@ -0,0 +1,531 @@
...
+BINARY_OP(logical_and)
+BINARY_OP(logical_or)
+BINARY_OP(logical_xor)
+
+BINARY_OP(add)
+BINARY_OP(subtract)
+BINARY_OP(divide)
+BINARY_OP(remainder)
+BINARY_OP(multiply)
+BINARY_OP2(power, float_power)
```
josevalim, line 351: Perhaps, to avoid future confusion, at the NIF level we should name the functions after libtorch API, and not after the Nx one?

versilov: On the one hand — yes. And it would simplify C++ code a bit.
On the other — we would not be able to do binary and unary operations code generation in cycles in backend, if names differ. But we can replace function atoms with tuples in operations lists. I'll have a look at that.

For other functions (like `arange`, `randint` etc.) I use original pytorch names.

josevalim: >  On the other — we would not be able to do binary and unary operations code generation in cycles in backend, if names differ. But we can replace function atoms with tuples in operations lists. I'll have a look at that.

Ah, I see. Ok, let's go with the consistent Nx naming then!

torchx/lib/torchx/backend.ex
```diff
@@ -0,0 +1,323 @@
...
+  Do not add signature during tests to match Nx doctests.
+
+  """
+  defp maybe_add_signature(result, tensor) do
+    if Mix.env() == :test do
+      result
+    else
+      Inspect.Algebra.concat(["Torchx.Backend(#{device(tensor)})", Inspect.Algebra.line(), result])
+    end
+  end
```
josevalim, line 268: Instead do this:

```suggestion
  if Application.get_env(:torchx, :add_backend_on_inspect, true) do
    defp maybe_add_signature(result, tensor) do
      Inspect.Algebra.concat(["Torchx.Backend(#{device(tensor)})", Inspect.Algebra.line(), result])
    end
  else
    defp maybe_add_signature(result, _tensor) do
      result
    end
  end
```

And then create a config/runtime.exs with this:

```elixir
import Config
config :torchx, :add_backend_on_inspect, config_env() != :test
```

versilov: `runtime.exs` did not work, `config.exs` did.

josevalim (approved):

josevalim:

PR #283, opened 2021-02-26
Add SVD BinaryBackend implementation

polvalente:
This PR adds SVD decomposition for Nx's BinaryBackend


polvalente:
I've opened this draft for progress visibility. Feel free to comment, but the implementation is not yet complete

nx/lib/nx/binary_backend.ex
```diff
@@ -1082,6 +1083,186 @@ defmodule Nx.BinaryBackend do
...
+                  v_cols_pq,
+                  {n, 2},
+                  output_type,
+                  rot_r,
+                  shape_r,
+                  output_type,
+                  output_type
+                )
+
+              # assign d_cols_pq to d_bin
```
polvalente, line 1250: @josevalim @seanmor5 I got a little confused in this last part of the algorithm. Basically, I need a way to translate:
`d[:, [p, q]] = np.matmul(d[:, [p, q]], rot_t)`

If these were rows, I could use the same trick as before where I can slice the bitstrings out and replace them, but I'm not entirely sure if there's an easier way of doing this assignment other than something like:

```elixir
for <<match!(x, 0) <- d_bin>>, reduce: {"", 0, 0, d_bin_cols_pq} do
{result, row, col, <<match!(h, 0), tail :: bitstring>>} when col in [p, q] -> 
  # inc row and col accordingly
  {result <> h, row, col, tail}
{result, row, col, data} ->
 # inc row and col accordingly
 {result <> x, row, col, data}
end
```

Copying this left and right seems a little iffy, but maybe this can become an useful util for this module?

nx/lib/nx/binary_backend.ex
```diff
@@ -1082,6 +1083,186 @@ defmodule Nx.BinaryBackend do
...
+    max_iter = 100
+
+    {fro_norm, _diag_norm, off_diag_norm} = get_matrix_norms(d_bin, s_shape, output_type)
+
+    Enum.reduce_while(1..max_iter, {u_bin, d_bin, v_bin, off_diag_norm, fro_norm}, fn _,
+                                                                                      {u_bin,
+                                                                                       d_bin,
+                                                                                       v_bin,
+                                                                                       off_diag_norm,
+                                                                                       fro_norm} ->
```
josevalim, line 1113: You have two options here so the formatter doesn't give you a hard time.

1. Move the accumulator to its own line:

```suggestion
    acc = {u_bin, d_bin, v_bin, off_diag_norm, fro_norm}
    Enum.reduce_while(1..max_iter, acc, fn _, {u_bin, d_bin, v_bin, off_diag_norm, fro_norm} ->
```

2. Break the function head into its own clause:

```suggestion
    Enum.reduce_while(1..max_iter, {u_bin, d_bin, v_bin, off_diag_norm, fro_norm}, fn
      _, {u_bin, d_bin, v_bin, off_diag_norm, fro_norm} ->
```

WDYT?

nx/lib/nx/binary_backend.ex
```diff
@@ -1082,6 +1083,186 @@ defmodule Nx.BinaryBackend do
...
+                      {row, col} =
+                        if col + 1 == n do
+                          {row + 1, 0}
+                        else
+                          {row, col + 1}
+                        end
+
+                      acc |> Map.put(:row, row) |> Map.put(:col, col)
+                  end
+                end
```
josevalim, line 1187: The copying of binary here in a loop is going to be expensive (i.e. `acc.before_row_p <> x`). My suggestion is to move this to a list (or a list of lists) and provide a version of `dot_binary` that works on lists. Since it is always MxN, perhaps a hardcoded version will suffice. :) What do you think?

nx/lib/nx/binary_backend.ex
josevalim: I have dropped a comment which may be related to this. Since we are talking only about matrices, it may be more convenient for us to redesign some of those operations to work with lists (or list of lists or list of tuples), it will likely be simpler than working with binaries on every step along the way.

If you find it more comfortable, you can even start by modifying QR first - there is a dot_binary in there which likely can work on lists, and then bring this up.

josevalim:
Amazing work @polvalente! I am excited about this. I have dropped some comments on potential ways to move forward. :)

nx/lib/nx/binary_backend.ex
polvalente: Looks a lot better!

nx/lib/nx/binary_backend.ex
polvalente: Right, I'll think about this. One of the advantages is being able to keep everything in the numerical-type domain, serializing to binary only at the end. This would remove the need for repeated `match_types` + `for` constructs.

One concern I have is about needing to access columns and being able to replace them as well. One idea I was thinking about for random access but probably performs really poorly is having a `%{linear_index => element}` map, which could be indexed through `linear_index = row * num_cols + col`. Another problem with this approach is that I don't think that `Map.values` would be realiable for retrieving the matrix as a whole due to the fact that ordering is not guaranteed.

A list of lists works fine for reading and replacing rows at O(num_rows), assuming a list of rows, but ends up being O(num_rows * num_cols) for accessing columns.

If we go with the "list of lists" approach, I'm thinking of accessing elements using constructs like:

```elixir
#matrix[[row, 0..num_cols-1]]
row = Enum.at(matrix, row)

#matrix[[row, col]]
element = matrix |> Enum.at(row) |> Enum.at(col)

# access matrix[[0..num_rows-1, col]]
column = matrix |> List.flatten() |> Enum.drop(col) |> Enum.take_every(num_cols)

# For lists of columns maybe a `reduce` over `List.flatten(matrix)` is probably better than the following:
# matrix[0..num_rows-1, cols]

flat_matrix = List.flatten(matrix)
columns = Enum.map(cols, & &1 |> Enum.drop() |> Enum.take_every(num_cols))
```

For replacing rows, it's pretty straight forward, but for replacing columns and elements, it does sound like a nested loop of some sort will be needed.

Any thoughts on this?

josevalim: For lists of lists, could you transpose the dimensions whenever you need to access it by columns? then the dot should be a zip_with operation?

polvalente: If this is acceptable, I'll run with it. Afterwards we can optimize the code further!

polvalente:
@josevalim SVD is almost working now. There's still a bug I didn't find before converting the draft haha

I've ended up also making a few changes to QR decomposition so the Householder method works on lists of lists instead of binaries, and the householder reflector ended up being changed for an equivalent implementation which provokes fewer sign changes (and I think it's a little bit more numerically stable as well).

This one is really tough, but we're almost there! :sweat_smile: 

I've ended up leaving some of the options from [numpy.linalg.svd](https://numpy.org/doc/stable/reference/generated/numpy.linalg.svd.html) out. Let me know if any of them is needed back!

nx/lib/nx.ex
```diff
@@ -7229,6 +7231,72 @@ defmodule Nx do
...
+          [0.0, 1.0, 0.0],
+          [0.0, 0, 1.0]
+        ]),
+        Nx.tensor([1.0, 1.0, 1.0]),
+        Nx.tensor([
+          [1.0, 0.0, 0.0],
+          [0.0, 1.0, 0.0],
+          [0.0, 0, -1.0]
+        ])
+      }
```
josevalim, line 7262: Instead of checking against the tuple, you can match on it and compare each element individually:

```
      iex> {u, s, vt} = Nx.svd(Nx.tensor([[1, 0, 0], [0, 1, 0], [0, 0, -1]]), compute_uv: true)
      iex> u
      #Nx.Tensor<...>
      iex> s
      #Nx.Tensor<...>
     iex> vt
      #Nx.Tensor<...>
```

Note you can't have empty lines between them for ExUnit to consider them the same doctest. :)

nx/lib/nx.ex
```diff
@@ -7229,6 +7231,72 @@ defmodule Nx do
...
+        ]),
+        Nx.tensor([3.0, 2.0, 1.0]),
+        Nx.tensor([
+          [0.0, 1.0, 0.0],
+          [1.0, 0.0, 0.0],
+          [0.0, 0, -1.0]
+        ])
+      }
+
+      iex> Nx.svd(Nx.tensor([[2, 0, 0], [0, -3, 0], [0, 0, -1], [0, 0, 0]]))
```
josevalim, line 7280: Let's add names to the tensor so we can check the return result preserve names.

nx/lib/nx/binary_backend.ex
```diff
@@ -1012,119 +1015,323 @@ defmodule Nx.BinaryBackend do
...
+    # It's important to note that since s left-multiplies v in
+    # the SVD result, and represents a diagonal matrix,
+    # changing a sign in s implies a sign change in the
+    # corresponding row of v.
+
+    # This function also sorts singular values from highest to lowest,
+    # as this can be convenient.
+
+    if compute_uv do
+      s
```
josevalim, line 1232:
```suggestion
      #TODO: Use Enum.zip_with on Elixir v1.12
      s
```

josevalim:
@polvalente this looks great, i have dropped some final nitpicks!

Btw, I love the transpose_matrix and binary_to_matrix functions. This means if we want to try another implementation (say a list of tuples or even tuples of tuples), it will be relatively straight-forward to do so. 

> I've ended up leaving some of the options from numpy.linalg.svd out. Let me know if any of them is needed back!

Definitely, let's go with baby steps.

josevalim:
> Btw, I love the transpose_matrix and binary_to_matrix functions. This means if we want to try another implementation (say a list of tuples or even tuples of tuples), it will be relatively straight-forward to do so.

If you want to push this idea forward, you should also create helpers such as `row_matrix(m, pos)` and `elem_matrix(m, row_pos, col_pos)` and `update_matrix(m, row_pos, col_pos, value)` so all of the matrix functions are encapsulated. :)

polvalente:
> > Btw, I love the transpose_matrix and binary_to_matrix functions. This means if we want to try another implementation (say a list of tuples or even tuples of tuples), it will be relatively straight-forward to do so.
> 
> If you want to push this idea forward, you should also create helpers such as `row_matrix(m, pos)` and `elem_matrix(m, row_pos, col_pos)` and `update_matrix(m, row_pos, col_pos, value)` so all of the matrix functions are encapsulated. :)

Will do! This might help in dealing with the last bug in the code as well. For some reason, the `U` matrix isn't being calculated properly.

nx/lib/nx.ex
polvalente: Oh, I thought you had to number them like `iex(1)`, `iex(2)`, etc. Nice to know :)

polvalente:
@josevalim SVD is now almost ready! There probably are still some interface or code issues, but the SVD itself now works as expected :D

Turns out I lost some hours due to swapping rows of a matrix instead of its transpose :sweat_smile: :laughing:

josevalim:
Beautiful, I will merge it so we make it up to date to master and I can apply some nits, but it looks great! If there are any improvements you want to make, feel free to send another PR later on!

josevalim:

josevalim:
I just pushed my nits. Note I have removed compute_uv for now (it is now always true) because having a function that returns different types (a tuple or not) based on the option is actually quite complicated for us to implement on later passes. We can add the behaviour later if desired (most likely by having a separate function). :)

PR #284, opened 2021-02-27
Nx.random_uniform/2 raises in BinaryBackend for non-float types

### LINKED ISSUE #282: Nx.random_uniform/2 raises in BinaryBackend for non-float types
When the code path reaches [these lines inside Nx.BinaryBackend.random_uniform/3](https://github.com/elixir-nx/nx/blob/4acedf922b490f29972158f98f68bc8577b2d281/nx/lib/nx/binary_backend.ex#L28-L29
) with the default range of `0.0, 1.0` from `Nx.random_uniform/2`, the calls `:rand.uniform(max - min)` are passed a float. The [docs for `:rand.uniform/1`]() say that it only accepts positive integers.

There are a few ways to fix this:

  + raise when int types are passed to `Nx.random_uniform/2`
  
  + use different defaults for int types in `Nx.random_uniform/2`
  
  + cast the `min - max` in `Nx.BinaryBackend.random_uniform/3` to int e.g. `(min - max) |> Float.round |> trunc`

This would be a good place to start for a new contributor.

The error:

```elixir
iex(14)> Nx.random_uniform({2, 4, 2}, type: {:s, 32}) 
** (FunctionClauseError) no function clause matching in :rand.uniform_s/2    
    
    The following arguments were given to :rand.uniform_s/2:
    
        # 1
        1.0
    
        # 2
        {%{
           bits: 58,
           jump: #Function<3.47293030/1 in :rand."-fun.exsplus_jump/1-">,
           next: #Function<0.47293030/1 in :rand."-fun.exsss_next/1-">,
           type: :exsss,
           uniform: #Function<1.47293030/1 in :rand."-fun.exsss_uniform/1-">,
           uniform_n: #Function<2.47293030/2 in :rand."-fun.exsss_uniform/2-">
         }, [197936017285128206 | 88863623788616986]}
    
    (stdlib 3.13.2) rand.erl:326: :rand.uniform_s/2
    (stdlib 3.13.2) rand.erl:299: :rand.uniform/1
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:28: anonymous fn/2 in Nx.BinaryBackend.random_uniform/3
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:33: anonymous fn/4 in Nx.BinaryBackend.random_uniform/3
    (elixir 1.11.2) lib/enum.ex:3449: Enum.reduce_range_inc/4
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:33: Nx.BinaryBackend.random_uniform/3
```

xire28:
Closes #282

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -685,10 +685,10 @@ defmodule Nx do
     assert_keys!(opts, [:type, :names, :backend])
     shape = Nx.shape(tensor_or_shape)
     names = Nx.Shape.named_axes!(opts[:names] || names!(tensor_or_shape), shape)
-    type = Nx.Type.normalize!(opts[:type] || Nx.Type.infer(max - min))
+    range_type = Nx.Type.infer(max - min)
+    type = Nx.Type.normalize!(opts[:type] || range_type)
 
-    unless Nx.Type.float?(type) or
-           (Nx.Type.integer?(type) and is_integer(min) and is_integer(max)) do
+    unless Nx.Type.float?(type) or (Nx.Type.integer?(type) and Nx.Type.integer?(range_type)) do
       raise ArgumentError,
             "random_uniform/3 expects compatible types, got: #{inspect(type)}" <>
             " in range #{inspect(min)} to #{inspect(max)}"
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -691,7 +691,7 @@ defmodule Nx do
     unless Nx.Type.float?(type) or (Nx.Type.integer?(type) and Nx.Type.integer?(range_type)) do
       raise ArgumentError,
             "random_uniform/3 expects compatible types, got: #{inspect(type)}" <>
-            " in range #{inspect(min)} to #{inspect(max)}"
+            " with range #{inspect(range_type)}"
     end
 
     backend = opts[:backend] || Nx.BinaryBackend
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -687,9 +687,11 @@ defmodule Nx do
     names = Nx.Shape.named_axes!(opts[:names] || names!(tensor_or_shape), shape)
     type = Nx.Type.normalize!(opts[:type] || Nx.Type.infer(max - min))
 
-    unless Nx.Type.float?(type) or (Nx.Type.integer?(type) and is_integer(min) and is_integer(max)) do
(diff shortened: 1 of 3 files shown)


nx/lib/nx.ex
```diff
@@ -687,9 +687,11 @@ defmodule Nx do
     names = Nx.Shape.named_axes!(opts[:names] || names!(tensor_or_shape), shape)
     type = Nx.Type.normalize!(opts[:type] || Nx.Type.infer(max - min))
 
-    unless Nx.Type.float?(type) or (Nx.Type.integer?(type) and is_integer(min) and is_integer(max)) do
-      raise ArgumentError, "random_uniform/3 expects compatible types, got: #{inspect(type)}" <>
-                           " in range #{inspect(min)} to #{inspect(max)}"
+    unless Nx.Type.float?(type) or
+           (Nx.Type.integer?(type) and is_integer(min) and is_integer(max)) do
```
seanmor5, line 691: I believe this will also fail when `min` or `max` are outside the acceptable range for the integer type. What I would do is something like:

```elixir
min_max_type =
  type
  |> Nx.Type.merge_scalar(min)
  |> Nx.Type.merge_scalar(max)

unless Nx.Type.float?(type) or (Nx.Type.integer?(type) and type == min_max_type) do
...
end
```

xire28: Didn't think of that, should be fine now

nx/lib/nx.ex
```diff
@@ -685,7 +685,15 @@ defmodule Nx do
     assert_keys!(opts, [:type, :names, :backend])
     shape = Nx.shape(tensor_or_shape)
     names = Nx.Shape.named_axes!(opts[:names] || names!(tensor_or_shape), shape)
-    type = Nx.Type.normalize!(opts[:type] || Nx.Type.infer(max - min))
+    range_type = Nx.Type.infer(max - min)
```
seanmor5, line 688: `infer` will always infer the 64-bit version of whatever the output type of `min - max` is. If I do

```elixir
Nx.random_uniform({5, 5, 5}, -10000, 10000, type: {:s, 8})
```
It will work, but I will end up with truncated values. I'm not really sure if letting this behavior in is ideal or something we should prevent altogether. I guess since we allow the same sort of truncation in `Nx.as_type` it might not matter that much.

josevalim: Yes, we generally document that given high values will overflow/underflow. We can add more checks in the future though, it doesn't hurt.

josevalim:

PR #286, opened 2021-02-27
Support convolution configurations

### LINKED ISSUE #262: Support configuration of convolution dimension numbers
Right now, our convolutions require input and kernel to have shapes:

- `input_shape - (batch_size, input_channels, spatial_dim0, ..., spatial_dimN)`
- `kernel_shape - (output_channels, input_channels, kernel_spatial_dim0, ..., kernel_spatial_dimN)`

With a resulting output shape of:

- `output_shape - (batch_size, output_channels, output_spatial_dim0, ..., output_spatial_dimN)`

While this is probably a pretty good general configuration, some backends or compilers may enforce other shape constraints. It would be best to support a configuration option that allows users to choose how to configure their convolutions. As an example, here's a JAX convolution dimension number configuration that matches the semantics described above:

```python
dimnos = ('NCHW', 'OIHW', 'NCHW')
```

There are two options for supporting this configuration. The first is to add an option to `Nx.conv` which interprets the configuration to compute shape constraints, padding configuration, output shape, etc. and then passes the same configuration to the backend or compiler to handle. In the case of an unsupported configuration, the backend/compiler will simply need to raise.

The other option is to have backends/compilers configure a default convolution dimension configuration and always dispatch to that implementation. The advantage here, I think, is portability of numerical definitions. Rather than needing to change inputs to `Nx.conv` based on a compiler/backend, you'd only need to set your backends default convolution configuration.

Implementation note for both EXLA and Nx: EXLA supports general configurations should implementation should be straightforward. For the Binary backend, I'm in favor of raising on shapes that don't match our current configuration, mainly to save the overhead of having to refactor and rewrite a convolution from scratch.

seanmor5:
Resolves #262 

I think this is the most flexible way to support different kinds of convolutions. I tried to make the explanation as easy to understand as possible. Please let me know if it's confusing

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -273,7 +273,7 @@ defmodule EXLA.Defn do
     kernel_dilation = opts[:kernel_dilation]
     groups = opts[:groups]
 
-    %{type: output_type, shape: shape} = ans
+    %{type: output_type} = ans
 
     # Build general conv dims
     input_permutation = List.to_tuple(opts[:input_permutation])
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -274,15 +274,20 @@ defmodule EXLA.Defn do
     groups = opts[:groups]
 
     %{type: output_type, shape: shape} = ans
-    rank = tuple_size(shape)
 
     # Build general conv dims
-    input_dims = List.to_tuple(for i <- 0..(rank - 1), do: i)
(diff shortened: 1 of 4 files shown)


josevalim (approved):
Except for the parens in `defn`, everything else looks great to me!

elbow-jason (commented):
Looks good. I much prefer the word "permutation" over the other names we've encountered.

elbow-jason (approved):

seanmor5:
The behavior was not quite what I had described, but I fixed that and added some more checks.

PR #287, opened 2021-02-28
Set AllocMode to avoid copying input data on AOT

josevalim:
@seanmor5 I have ran some tests locally and I didn't  see any segfaults. Thoughts?


seanmor5:
I can pull and see what happens, I tried this approach as well earlier and ran in to issues

josevalim:
FWIW, I ran this:

    IO.puts Enum.map(1..1000, fn _ -> AOTSoftmax.softmax_32(t32) end) |> Enum.uniq() |> length()

And all results where the same. :)

PR #290, opened 2021-02-28
Do not increase type size on int with float merging. Make float32 default float type

versilov:


nx/lib/nx/shared.ex
```diff
@@ -134,6 +134,11 @@ defmodule Nx.Shared do
   defp shared_bin_modifier(var, :f, size),
     do: quote(do: unquote(var) :: float - native - size(unquote(size)))
 
+  def to_float32(f64) when is_float(f64) do
```
josevalim, line 137:
```suggestion
  @doc """
  Converts an Erlang float (float64) to float32 precision.
  """
  def to_float32(f64) when is_float(f64) do
```

josevalim:

PR #292, opened 2021-02-28
Add note about ElixirLS per request from Sean. Add note about gitref vs local path.

TheFirstAvenger:
Per the below slack request from @seanmor5 I have added a section on compiling in ElixirLS. I also added a note about gitref vs local path, as this was a mistake I made early on that cost me many hours in unnecessary recompilation.

Screen Shot 2021-02-28 at 5 55 19 PM

### DIFF
--- a/exla/README.md
+++ b/exla/README.md
@@ -46,6 +46,11 @@ The first compilation will take a long time, as it needs to compile parts of Ten
       * asdf global bazel 3.1.0
   * ElixirLS on VSCode
     * Make sure that your Python installation is available globally, as ElixirLS won't know how to activate Python
+  * When importing `:exla` into a project, make sure you use a git ref, and *not* `path: "../exla"`. Using a local reference will trigger a full build of `:exla` every time you compile your project.
+
+#### Compiling in ElixirLS
+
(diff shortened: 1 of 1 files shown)


exla/README.md
```diff
@@ -46,6 +46,11 @@ The first compilation will take a long time, as it needs to compile parts of Ten
       * asdf global bazel 3.1.0
   * ElixirLS on VSCode
     * Make sure that your Python installation is available globally, as ElixirLS won't know how to activate Python
+  * When importing `:exla` into a project, make sure you use a git ref, and *not* `path: "../exla"`. Using a local reference will trigger a full build of `:exla` every time you compile your project.
```
josevalim, line 49: So the reason I think you were seeing full rebuilds is because ElixirLS and the regular code were racing each other (i.e. they were trying to compile to the same output dir). Were you using the same Erlang/OTP versions for both?

TheFirstAvenger: @josevalim Multiple processes running the make file simultaneously, which from what I can tell deletes part of the cache as its first step, was definitely complicating the issue, yes. The singleton process feature, which recognizes multiple simultaneous builds and serializes them, happens too late in the process. This means that multiple builds of the exact same version both have to go through the full compile as long as all are started before the first one ends. Erlang/OTP were the same version. The workaround of using a git ref as you suggested severely reduced the number of times :exla was compiled, which meant that once a full compile was completed cleanly, even with ElixirLS rapid-firing subsequent compiles, they were fast because they didn't need to run the make file for :exla.

Issue #294, opened 2021-03-01
What is ERTS_SYM_DIR?


jordanhubbard:
Not suggesting the following diff, but it's the only thing that moves the compile past the fact that the erts subdirectory is not checked or called out as an explicit dependency in README.md (not sure why erlang's erts directory is being referenced at all).

```mix run``` works, on the other hand.
 
```
@@ -37,8 +37,8 @@ all: symlinks
 symlinks: $(TENSORFLOW_DIR)
        @ rm -f $(TENSORFLOW_EXLA_DIR)
        @ ln -s "$(MIX_CURRENT_PATH)/$(EXLA_DIR)" $(TENSORFLOW_EXLA_DIR)
-       @ rm -f $(ERTS_SYM_DIR)
-       @ ln -s "$(ERTS_INCLUDE_DIR)" $(ERTS_SYM_DIR)
+#      @ rm -f $(ERTS_SYM_DIR)
+#      @ ln -s "$(ERTS_INCLUDE_DIR)" $(ERTS_SYM_DIR)

```

seanmor5:
To compile the NIF we need to link ERTS includes into the TF directory because Bazel doesn't work very well with global dependencies. 

The directory is included (usually) in standard Erlang installations. Sometimes you'll need to install `erlang-dev` as a separate dependency. Try installing `erlang-dev` and everything should work out!

jordanhubbard:
Sorry, I think I somehow failed to make the more important fundamental points I was trying to make:

1.  I'm compiling all of this from source, including OTP, so there is no package dependency chain here.  The standard erlang install target, up to and including some of the more battery-included options like "install-docs", does not put an erts directory anywhere in /usr/local.
2. Even if it (or the erlang-dev package) did, this is not called out as a dependency for exla anywhere in the documentation so it's a "failed dependency" from that standpoint, at least.
3. Even with all of the above aside, the ```symlinks``` target in the Makefile never has the opportunity to succeed because ERTS_INCLUDE_DIR is not set or documented as needing to be set.   Ironically, relying on ERL_TOP, which *is* documented, would have a much greater chance of working, though this is now dragging in erlang to be even more explicit.

TL;DR:  If the erlang runtime is needed explicitly, why not just check it out in the same fashion as tensorflow (hardly a tiny dep) is being checked out?

seanmor5:
[...]

PR #295, opened 2021-03-01
Make various make targets symmetrical and usable from the top level.

jordanhubbard:
This essentially describes (and fixes) the problem we're discussing in https://github.com/elixir-nx/nx/issues/294

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,33 @@
+SUBDIRS=	nx exla
+
 all:	setup
-	cd exla && mix compile
+	@for _dir in ${SUBDIRS}; do \
+		(cd $${_dir} && mix compile); \
+	done
 
 setup:
-	cd nx && mix deps.get
-	cd exla && mix deps.get
+	@for _dir in ${SUBDIRS}; do \
+		(cd $${_dir} && mix deps.get); \
+	done
 
 test:
-	 cd nx && mix test
-	 cd exla && mix test
+	@for _dir in ${SUBDIRS}; do \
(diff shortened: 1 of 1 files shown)


jordanhubbard:
Also, QQ:  What audience *is* the all target in the exla directory for?  Internal developers who have some type of environment (variable) setup in advance or is it actually just an alias for `mix compile` with another, internal, target taking over from "all"?  I ask only because targets like clean, all, and install are generally considered "public API" and they should therefore just work with reasonable defaults.

seanmor5:
Ah yes this makes sense to me now!

So both `all` and `clean` are only ever intended to be invoked by `elixir_make` on calls to `mix compile` and `mix clean` respectively. They're the default targets `elixir_make` looks for in these cases, although I'm not sure if that can be changed. I do see where the confusion is coming in with respect to the environment variables coming out of nowhere. An easy fix is to set them from within the Makefile rather than on waiting for Elixir to do it, although I can't think of a reason to compile without mix, unless you had something in mind?

As for the PR, for completeness it might make sense to add:

`cd nx && mix compile` as well. But we have plans to split the repository up, so any adjustment to the build process in this case should probably be made directly to EXLA rather than at this level. I will defer to @josevalim on this one :)

jordanhubbard:
Please see updated commit - this covers the concerns you raised and rounds out the makefile's default target coverage a bit more.

josevalim:
This was meant to be used only internally but adding some conveniences doesn’t hurt. Thank you!

josevalim:

PR #297, opened 2021-03-01
Update README for asdf installed pythons

chrismcg:
I couldn't get exla to build with asdf installed python but found a workaround that uses the homebrew installed version of Python.

This adds a note to the README for others in the same position.


josevalim:

PR #302, opened 2021-03-02
feat: add missing norms

polvalente:
Adds the missing norms for #157, which needed SVD to be ready


seanmor5 (approved):
@polvalente Amazing :D

PR #305, opened 2021-03-03
Add and use allclose? when testing

seanmor5:
This is what JAX uses to test against NumPy, a bit more precise than float rounding. Also updates `check_grads!` to support n-D output so we can use finite differences for all of the gradient tests rather than just comparing against JAX output

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -4137,6 +4137,40 @@ defmodule Nx do
   end
 
   @doc """
+  Returns a scalar tensor of value 1 if element-wise values
+  of a are within tolerance of b, value 0 otherwise.
+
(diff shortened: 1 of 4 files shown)


nx/lib/nx.ex
```diff
@@ -4136,6 +4136,40 @@ defmodule Nx do
...
+
+      iex> Nx.allclose?(Nx.tensor([1.0e-8, 1.0e-8]), Nx.tensor([1.0e-8, 1.0e-9]))
+      #Nx.Tensor<
+        u8
+        1
+      >
+
+  """
+  @doc type: :aggregation
+  def allclose?(a, b, opts \\ []) do
```
josevalim, line 4166:
```suggestion
  def all_close?(a, b, opts \\ []) do
```

?

PR #306, opened 2021-03-04
feat: add lu decomposition

polvalente:
This PR implements LU decomposition for #157, as per the [scipy.linalg.lu](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.lu.html).

No options are supported besides our own `:eps` and `*_names` because:

- overwrite_a: this doesn't make sense for Elixir
- permute_l: makes the function's return type polymorphic so it wasn't implemented
- check_finite: we don't support infinities yet

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7656,6 +7656,125 @@ defmodule Nx do
     )
   end
 
+  @doc """
+  Calculates the (P)LU decomposition of a 2-D tensor with shape `{N, N}`.
+
+  ## Options
+    * `:p_names` - Defines the names for the `p` tensor
+
+    * `:l_names` - Defines the names for the `l` tensor
+
+    * `:u_names` - Defines the names for the `u` tensor
+
+    * `:eps` - Rounding error threshold that can be applied during the factorization
+
+  ## Examples
+
+      iex> {p, l, u} = Nx.lu(Nx.tensor([[1, 2, 3], [4, 5, 6], [7, 8, 9]]))
+      iex> p
+      #Nx.Tensor<
+        s64[3][3]
+        [
+          [0, 0, 1],
+          [0, 1, 0],
+          [1, 0, 0]
+        ]
+      >
+      iex> l
+      #Nx.Tensor<
+        f32[3][3]
+        [
+          [1.0, 0.0, 0.0],
+          [0.5714285969734192, 1.0, 0.0],
+          [0.1428571492433548, 2.0, 1.0]
+        ]
+      >
+      iex> u
+      #Nx.Tensor<
+        f32[3][3]
+        [
+          [7.0, 8.0, 9.0],
+          [0.0, 0.4285714328289032, 0.8571428656578064],
+          [0.0, 0.0, 0.0]
+        ]
+      >
+      iex> p |> Nx.dot(l) |> Nx.dot(u)
+      #Nx.Tensor<
+        f32[3][3]
+        [
+          [1.0, 2.0, 3.0],
+          [4.0, 5.0, 6.0],
+          [7.0, 8.0, 9.0]
+        ]
+      >
+
+      iex> {p, l, u} = Nx.lu(Nx.tensor([[1, 0, 1], [-1, 0, -1], [1, 1, 1]]))
+      iex> p
+      #Nx.Tensor<
+        s64[3][3]
+        [
+          [1, 0, 0],
+          [0, 0, 1],
+          [0, 1, 0]
+        ]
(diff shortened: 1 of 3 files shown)


nx/lib/nx.ex
```diff
@@ -7656,6 +7656,125 @@ defmodule Nx do
...
 
+  @doc """
+  Calculates the A = PLU decomposition of a 2-D tensor A with shape `{N, N}`.
+
+  ## Options
+    * `:p_names` - Defines the names for the `p` tensor
+
+    * `:l_names` - Defines the names for the `l` tensor
+
+    * `:u_names` - Defines the names for the `u` tensor
```
josevalim, line 7667: So I have a question:

1. For svd, we are using names of the initial tensor

2. For qr and lu, we have `_names` option

Is this intentional? Or should we try to unify them? Or maybe should we always use the names from the input tensor?

/cc @polvalente @seanmor5

josevalim:
Just one comment, everything else looks perfect. thank you @polvalente!

nx/lib/nx.ex
seanmor5: I think we should try to use the names of the input tensor whenever possible, but in some cases it might not make sense. I believe this is of the form:

`A = LU` so can we preserve inner names like a dot product and then accept options to fill in the `nil`?

josevalim: It is PLU, unfortunately. Which means we could use the names for QR but it doesn't help here. Especially we would have to choose if we want to name PL, PU, or LU. The other question is, do we want to pass names as arguments or do we want to call `Nx.rename` (or a similar function) immediately after?

seanmor5: I see. My thought process is always to retain the names where it will allow us to recover the original names given whatever property holds. So if it is `A = P.L.U` as you said you can choose? Maybe then that's the configuration option w.r.t names: which tensors retain which names. And then backtracking on the other implementations, maybe we should force the renaming to be explicit after a call to `lu` and other functions like this.

josevalim: So our options are:

1. Infer as much as possible (and potentially make the inference configurable)
2. Fallback to the `_names` option
3. Ask users to explicitly rename

I am thinking we should be fine with 1 and 3. So I vote to merge this and then to a short clean up afterwards. WDYT?

josevalim:

nx/lib/nx.ex
polvalente: I think 1 and 3 could be combined. What do you think?

For instance, in A = P L U, we could have:

[a_rows, a_cols] = A.names

P_names = [a_rows, nil] (because P is just a matrix that swaps the rows of L.U to restore A)
L_names = [a_rows, nil] (because we want L.U to have the same names as A, since L.U is just A with rows swapped)
U_names = [nil, a_cols]

And for SVD we could have:

A = U.S.Vt

U_names = [a_rows, nil]
S_names = [nil, nil] (I can see [a_rows, a_cols] making sense here as well, but I'd rather keep nils because S scales both Vt and U to recompose A)
Vt_names = [nil, a_cols]

In both cases the recomposed matrix will also have the original names recomposed, as I've tested here:
```elixir
iex(1)> a = Nx.tensor([[0,1],[1,0]], names: [:rows, nil])
#Nx.Tensor<
  s64[rows: 2][2]
  [
    [0, 1],
    [1, 0]
  ]
>
iex(2)> b = Nx.tensor([[1,2],[3,4]], names: [nil, :cols])
#Nx.Tensor<
  s64[2][cols: 2]
  [
    [1, 2],
    [3, 4]
  ]
>
iex(3)> Nx.dot(a, b)
#Nx.Tensor<
  s64[rows: 2][cols: 2]
  [
    [3, 4],
    [1, 2]
  ]
>
iex(4)> Nx.dot(a, Nx.tensor([[1, 0], [0, 1]], names: [nil, nil])) |> Nx.dot(b)
#Nx.Tensor<
  s64[rows: 2][cols: 2]
  [
    [3, 4],
    [1, 2]
  ]
>
```

josevalim: My concern is that those choices seem to be arbitrary. PLU could also emit `[nil, nil], [nil, nil], [col, row]`. And at the moment I am not sure which configuration to pick up from all possible ones. :(

polvalente: I see! I thought that it only kept the dimensions according to the outermost axis, but it does make sense that it just overrides nils from the last result.

With that in mind, I'd rather leave no defaults and optional options for the result names

PR #313, opened 2021-03-05
Slice & sum

versilov:


nx/lib/nx.ex
```diff
@@ -4208,8 +4208,11 @@ defmodule Nx do
   precision to make sure the sum does not overflow:
 
       iex> Nx.sum(Nx.tensor([[101, 102], [103, 104]], type: {:u, 8}, names: [:x, :y]))
+      ** (RuntimeError) Torchx does not support unsigned 64 bit integer (explicitly cast the input tensor to a signed integer before taking sum)
```
josevalim, line 4211: If we change this like this, the `Nx` suite will fail. My suggestion: remove this doctest and add it to test/nx_test.exs instead.

torchx/.iex.exs
```diff
@@ -0,0 +1,4 @@
+
```
josevalim, line 1:
```suggestion
```

torchx/.iex.exs
```diff
@@ -0,0 +1,4 @@
...
+if :code.module_status(Nx) == :loaded do
```
josevalim, line 2:
```suggestion
if Code.ensure_loaded?(Nx) do
```

torchx/lib/torchx/backend.ex
```diff
@@ -150,6 +150,44 @@ defmodule Torchx.Backend do
     NIF.transpose(to_ref(t), 0, 1) |> from_ref(out)
   end
 
+  @impl true
+  def slice(%T{shape: shape} = out, %T{} = t, start_indices, lengths, strides) do
+    ref = to_ref(t)
+
+    for dim <- 0..(length(start_indices) - 1), reduce: ref do
+      ref -> NIF.narrow(ref, dim, Enum.at(start_indices, dim), Enum.at(lengths, dim)) |> unwrap!()
+    end
```
josevalim, line 159: Don't use `0..(length(start_indices) - 1)` because it will be wrong for scalars (tensors with no dimensions) as the range will have one element but there is no actual dimension. Prefer to use `Nx.axes(shape)` instead.

This code can also be more efficient if you avoid narrowing when you know there is nothing to narrow. Here is how I would implement it:

    t
    |> to_ref()
    |> narrow(start_indices, lengths, 0, shape)

And then:

    defp narrow(ref, [start | starts], [length | lengths], axis, shape) do
      dim = elem(shape, axis)

      # Nothing to narrow
      if start == 0 and length == dim do
        narrow(ref, starts, lengths, axis + 1, shape)
      else
        ref
        |> NIF.narrow(ref, axis, start, length)
        |> narrow(starts, lengths, axis + 1, shape)
      end
    end

    defp narrow(ref, [], [], _axis, _shape), do: ref

I haven't tested it. :)

torchx/lib/torchx/backend.ex
```diff
@@ -150,6 +150,44 @@ defmodule Torchx.Backend do
...
+    |> NIF.as_strided(shape, steps_to_strides(lengths, strides), 0)
```
josevalim, line 160: Similarly, there is nothing to stride if `Enum.all?(strides, & &1 == 1)`, so it may be worth adding such check.

torchx/lib/torchx/backend.ex
```diff
@@ -150,6 +150,44 @@ defmodule Torchx.Backend do
...
+
+    axes = opts[:axes] || []
+    keep_axes = opts[:keep_axes] || false
+
+    NIF.sum(to_ref(t), axes, keep_axes) |> from_ref(out)
+  end
+
+  defp check_out_type!(type, hint) do
+    torch_type(type)
+  rescue
```
josevalim, line 186: Avoid rescuing! Instead pass the hint to torch_type and raise the correct thing in the first place. :)

torchx/lib/torchx/backend.ex
```diff
@@ -291,14 +328,21 @@ defmodule Torchx.Backend do
     do: Nx.backend_transfer(tensor, TB) |> to_ref()
 
   # Update out tensor type here to mark the cases where our type policy mismatches with the libtorch's one.
-  defp to_tensor(ref, %T{type: type} = t) do
+  defp to_tensor(ref, %T{type: type, shape: shape} = t) do
```
josevalim, line 331: What do you think about adding another flag such as `Application.get_env(:torchx, :check_backend_shape_and_type, false)` that we set to true only for tests? So we only run those checks during development? :)

nx/lib/nx.ex
```diff
@@ -4208,11 +4208,8 @@ defmodule Nx do
   precision to make sure the sum does not overflow:
 
       iex> Nx.sum(Nx.tensor([[101, 102], [103, 104]], type: {:u, 8}, names: [:x, :y]))
-      ** (RuntimeError) Torchx does not support unsigned 64 bit integer (explicitly cast the input tensor to a signed integer before taking sum)
-
-      iex> Nx.sum(Nx.tensor([[101, 102], [103, 104]], type: {:s, 8}, names: [:x, :y]))
       #Nx.Tensor<
-        s64
+        u64
```
josevalim: You want to keep the other doctest. :)

torchx/test/torchx/nx_doctest_test.exs
```diff
@@ -38,7 +38,8 @@ defmodule Torchx.NxDoctestTest do
         norm_integer: 3,
         all_close?: 3,
         atan2: 2,
-        lu: 2
+        lu: 2,
+        sum: 2
```
josevalim, line 42: Sorry, to be clear, we don’t want to exclude sum. Change the doctest to keep only s8-s64 and we should be good. :)

versilov: That's nice! I was ready to rewrite all `sum/2` doctests in regular form for Torchx:)

josevalim (approved):

PR #316, opened 2021-03-06
Do not rebuild libexla if already built

hauleth:
This should reduce build time when running the application in different environments. On my machine the XLA was rebuilt in test environment, even if it was already built in `dev` environment. This caused unnecessary work when using ElixirLS which is building the whole project in `test` environment. This change will simply assume that project is built if there already is `libexla.so`.

### DIFF
--- a/exla/Makefile
+++ b/exla/Makefile
@@ -28,11 +28,15 @@ TENSORFLOW_DIR = $(EXLA_CACHE)/$(TENSORFLOW_NS)/erts-$(ERTS_VERSION)
 TENSORFLOW_EXLA_NS = tensorflow/compiler/xla/exla
 TENSORFLOW_EXLA_DIR = $(TENSORFLOW_DIR)/$(TENSORFLOW_EXLA_NS)
 
-all: symlinks
+XLA_SO = $(TENSORFLOW_DIR)/bazel-bin/$(TENSORFLOW_EXLA_NS)/libexla.so
+
+all: symlinks $(XLA_SO)
+	mkdir -p priv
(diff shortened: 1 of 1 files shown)


exla/Makefile
```diff
@@ -28,11 +28,15 @@ TENSORFLOW_DIR = $(EXLA_CACHE)/$(TENSORFLOW_NS)/erts-$(ERTS_VERSION)
...
 TENSORFLOW_EXLA_DIR = $(TENSORFLOW_DIR)/$(TENSORFLOW_EXLA_NS)
 
-all: symlinks
+XLA_SO = $(TENSORFLOW_DIR)/bazel-bin/$(TENSORFLOW_EXLA_NS)/libexla.so
+
+all: symlinks $(XLA_SO)
+	mkdir -p priv
+	cp -f $(XLA_SO) $(EXLA_SO)
+
+$(XLA_SO): $(TENSORFLOW_DIR)
```
josevalim, line 37: Thanks @hauleth!

I believe this is a welcome change but unfortunately it will break our development experience and code updates because if we change exla.cc or any other file, it won't be rebuilt. And unfortunately the only way to rebuild it is by asking bazel.

josevalim: Also, I believe the reason ElixirLS is building again is because Bazel is detecting it is being built in a different environment, and builds it from scratch. We should try to find a way to address this without breaking the development flow. Does ELIXIR_LS set an environment variable? Maybe we can execute the code you added here if said env var is set.

josevalim:
Closing this for now. If someone has an idea on how to solve the problems above, we will be glad to reopen!

PR #317, opened 2021-03-06
fix: change col and row for heatmatp inspect

polvalente:
`@hauleth` reported in the `#nx` that the Heatmap inspect had an unexpected behavior for asymetric tensors (i.e. 4x2). This PR aims to fix this problem.


josevalim:

PR #318, opened 2021-03-08
32- and 64-bit float values NaN checks

versilov:
Am I moving in the right direction?
Would add brain float checks as well.

### DIFF
--- a/nx/lib/nx/backend.ex
+++ b/nx/lib/nx/backend.ex
@@ -204,12 +204,13 @@ defmodule Nx.Backend do
     end
   end
 
+  import Nx.Shared, only: [is_nan32: 1, is_nan64: 1]
+
   defp inspect_float(data, 32) do
     case data do
       <<0xFF800000::32-native>> -> "-Inf"
       <<0x7F800000::32-native>> -> "Inf"
-      <<0xFF800001::32-native>> -> "NaN"
-      <<0xFFC00001::32-native>> -> "NaN"
+      <<nan::32-native>> when is_nan32(nan) -> "NaN"
       <<x::float-32-native>> -> Float.to_string(x)
     end
   end
@@ -218,8 +219,7 @@ defmodule Nx.Backend do
     case data do
       <<0xFFF0000000000000::64-native>> -> "-Inf"
       <<0x7FF0000000000000::64-native>> -> "Inf"
-      <<0x7FF0000000000001::64-native>> -> "NaN"
-      <<0x7FF8000000000001::64-native>> -> "NaN"
+      <<nan::64-native>> when is_nan64(nan) -> "NaN"
       <<x::float-64-native>> -> Float.to_string(x)
     end
   end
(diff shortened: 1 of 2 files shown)


nx/lib/nx/backend.ex
```diff
@@ -218,8 +219,7 @@ defmodule Nx.Backend do
     case data do
       <<0xFFF0000000000000::64-native>> -> "-Inf"
       <<0x7FF0000000000000::64-native>> -> "Inf"
-      <<0x7FF0000000000001::64-native>> -> "NaN"
-      <<0x7FF8000000000001::64-native>> -> "NaN"
+      <<nan::64-native>> when is_nan64(nan) -> "NaN"
```
josevalim, line 222: I think a more efficient version would be something like:

    <<0x7FF::12-native, _>> -> "NaN"
    <<0xFFF::12-native, _>> -> "NaN"

No guards. :)

josevalim: Wait, wait. That won't work. We will need to check if it is little or big endian.

josevalim: You will need something like this:

```elixir
if System.endianness() == :little do
  defp inspect_float(data, 64) do
    case data do
      <<0xFFF0000000000000::64-native>> -> "-Inf"
      <<0x7FF0000000000000::64-native>> -> "Inf"
      <<_::52, 0x7FF::12-native>> -> "NaN"
      <<_::52, 0xFFF::12-native>> -> "NaN"
      <<x::float-64-native>> -> Float.to_string(x)
    end
  end
else
  defp inspect_float(data, 64) do
    case data do
      <<0xFFF0000000000000::64-native>> -> "-Inf"
      <<0x7FF0000000000000::64-native>> -> "Inf"
      <<0x7FF::12-native, _::52>> -> "NaN"
      <<0xFFF::12-native, _::52>> -> "NaN"
      <<x::float-64-native>> -> Float.to_string(x)
    end
  end
end
```

nx/lib/nx/backend.ex
```diff
@@ -204,23 +204,39 @@ defmodule Nx.Backend do
...
-      <<0xFF800000::32-native>> -> "-Inf"
-      <<0x7F800000::32-native>> -> "Inf"
-      <<0xFF800001::32-native>> -> "NaN"
-      <<0xFFC00001::32-native>> -> "NaN"
-      <<x::float-32-native>> -> Float.to_string(x)
+  defp inspect_float(<<0xFF800000::32-native>>, 32), do: "-Inf"
+  defp inspect_float(<<0x7F800000::32-native>>, 32), do: "Inf"
+
+  defp inspect_float(<<0x7FF0000000000000::64-native>>, 64), do: "Inf"
+  defp inspect_float(<<0xFFF0000000000000::64-native>>, 64), do: "-Inf"
```
josevalim, line 211: So the reason why the code was written with `case` previously is because the pattern matching is more efficient. First we matched on the size and then on the binary. With the current approach, the code may end-up trying the clauses linearly and check the different sizes. Can you please merge them as it was before into a single case? I  know there will be a bit of repetition but I think it is fine. :)

josevalim (commented):
One last comment and we can ship it!

josevalim (approved):

PR #320, opened 2021-03-09
Examples

versilov:
Adapted from EXLA.


nx/lib/nx/shared.ex
```diff
@@ -81,6 +81,45 @@ defmodule Nx.Shared do
...
+      iex> (fn f when is_nan64(f) -> true end).(nan)
+      true
+
+  """
+  defguard is_nan64(val)
+           when is_integer(val) and
+                  (val in 0x7FF0000000000001..0x7FF7FFFFFFFFFFFF or
+                     val in 0xFFF0000000000001..0xFFF7FFFFFFFFFFFF or
+                     val in 0x7FF8000000000000..0x7FFFFFFFFFFFFFFF or
+                     val in 0xFFF8000000000000..0xFFFFFFFFFFFFFFFF)
```
josevalim, line 121: I think this is old work and it should be removed, right?

torchx/lib/torchx.ex
```diff
@@ -1,2 +1,35 @@
 defmodule Torchx do
+  @moduledoc """
+
+  Torchx behaviour that is different from BinaryBackend:
```
josevalim, line 4: Maybe we should put this in the Torchx.Backend module instead?

torchx/lib/torchx/backend.ex
```diff
@@ -51,7 +51,7 @@ defmodule Torchx.Backend do
   def iota(out, axis \\ nil)
 
   def iota(%T{shape: {}, type: type} = out, nil) do
-    NIF.scalar_tensor(0, torch_type(type)) |> from_ref(out)
+    NIF.scalar_tensor(0.0, torch_type(type)) |> from_ref(out)
```
josevalim, line 54: Does this work for both `iota(s64)` and `iota(f64)`?

torchx/lib/torchx/backend.ex
```diff
@@ -202,6 +202,22 @@ defmodule Torchx.Backend do
     NIF.sum(to_ref(t), axes, keep_axes) |> from_ref(out)
   end
 
+  @impl true
+  def argmax(%T{type: out_type} = out, %T{} = t, opts) do
+    axe = opts[:axis] || 0
+    keep_axes = opts[:keep_axes] || false
+
```
josevalim, line 209: Let's raise if the tiebreak is not the supported one!

torchx/lib/torchx/backend.ex
```diff
@@ -202,6 +202,22 @@ defmodule Torchx.Backend do
     NIF.sum(to_ref(t), axes, keep_axes) |> from_ref(out)
   end
 
+  @impl true
+  def argmax(%T{type: out_type} = out, %T{} = t, opts) do
+    axe = opts[:axis] || 0
```
josevalim, line 207: To be clear, not having an axis is not the same as setting the default axis to zero! Here is an example:

```
iex(4)> Nx.iota({2, 3}, axis: 1) |> Nx.argmax()
#Nx.Tensor<
  s64
  2
>
iex(5)> Nx.iota({2, 3}, axis: 0) |> Nx.argmax()
#Nx.Tensor<
  s64
  3
>
iex(6)> Nx.iota({2, 3}) |> Nx.argmax()         
#Nx.Tensor<
  s64
  5
>
```

PS: singular of axes is axis. :)

torchx/test/torchx/nx_doctest_test.exs
```diff
@@ -13,33 +13,31 @@ defmodule Torchx.NxDoctestTest do
       |> Kernel.++(
         to_binary: 2,
         default_backend: 2,
-        add: 2,
         round: 1,
```
josevalim, line 16: Let's add documentation to why we are not including some doctests. For example

  # Those are not callbacks
  default_backend: 2,

  # We require explicit transfer for those
  to_binary: 2,
  to_scalar: 1,

  # We round differently than Nx for now
  round: 1,

```

and so on!

File: torchx/lib/torchx/backend.ex (hunk 3 above)
Comment (versilov) [reply torchx/lib/torchx/backend.ex:54]: Yes.

File: torchx/examples/mnist.exs [hunk 7]
```diff
@@ -0,0 +1,187 @@
...
+    batch
+    |> Nx.dot(w1)
+    |> Nx.add(b1)
+    |> Nx.logistic()
+    |> Nx.dot(w2)
+    |> Nx.add(b2)
+    |> softmax()
+  end
+
+  defp update(
```
josevalim, line 29: Btw, this example can still use `defn`! The `defn` evaluator should also default to using the default backend. :) So my suggestion is to keep this as close as possible to the EXLA one!

torchx/lib/torchx/backend.ex
```diff
@@ -1,6 +1,38 @@
 defmodule Torchx.Backend do
   @moduledoc """
+
```
josevalim, line 3: No newlines at the beginning of heredoc. :)

```suggestion
```

torchx/lib/torchx/backend.ex
```diff
@@ -202,6 +234,37 @@ defmodule Torchx.Backend do
...
+    axis = opts[:axis] || -1
+    keep_axes = opts[:keep_axes] || false
+
+    NIF.argmin(to_ref(t), axis, keep_axes) |> from_ref(out)
+  end
+
+  defp unsupported_option!(opts, key, acceptable_default \\ nil),
+    do:
+      if(opts[key] != acceptable_default, do: raise("#{inspect(key)} option is not supported in #{caller()}"))
```
josevalim, line 259:
```suggestion
  defp unsupported_option!(opts, key, acceptable_default) do
    if(opts[key] != acceptable_default, do: raise("#{inspect(key)} option is not supported in #{caller()}"))
  end
```

torchx/test/torchx/nx_doctest_test.exs
```diff
@@ -1,44 +1,100 @@
...
+  ## Result mismatch in less significant digits (precision issue)
+
+  * logistic/1
+
+  ## Callback or test requires explicity backend transfer to BinaryBackend
+
+  * to_binary/2
+  * to_flat_list/2
+  * random_uniform/4
+  * random_normal/4
```
josevalim, line 23: Those should work now with the default backend!

torchx/test/torchx/nx_doctest_test.exs
```diff
@@ -1,44 +1,100 @@
...
+  * atan2/2
```
josevalim, line 24: This should work... no?

torchx/test/torchx/nx_doctest_test.exs
```diff
@@ -1,44 +1,100 @@
...
+
+  ## Not all options are supported
+
+  * argmax/2
+  * argmin/2
+
+  ## Unsigned ints not supported by Torch
+
+  * quotient/2
+  * mean/2
```
josevalim, line 34: For those we can move the tests that require unsigned integers to `test/nx_test_exs` so we can use them.

torchx/test/torchx/nx_doctest_test.exs
```diff
@@ -1,44 +1,100 @@
...
+
+  ## Output mismatch
+
+  * qr/2
+  * dot/4
+  * cholesky/1
```
josevalim, line 40: We need to add tests for all operations that are missing doctests. So this means tests for: qr, dot, cholesky, quotient, mean, argmin, argmax, logistic, round, atan2, random_uniform and random_normal. :)

torchx/test/torchx/nx_doctest_test.exs
versilov: They do, but tests fail, because they use `Nx.to_binary/1` to check the result.

torchx/test/torchx/nx_doctest_test.exs
versilov: Same story: tests use `Nx.to_binary/1`, which is banned.

```elixir
     iex> t = Nx.atan2(pos_and_neg_zero_x, pos_and_neg_zero_y)
      iex> Nx.to_binary(t)
      <<-3.141592653589793::float-32-native, (-1.0*(Integer.parse("0")|>elem(0)))::float-32-native,
        3.141592653589793::float-32-native, 0.0::float-32-native>>
```

torchx/examples/mnist.exs
```diff
@@ -0,0 +1,189 @@
...
+
+    step = Nx.tensor(@step)
+
+    {{
+       Nx.subtract(w1, Nx.multiply(grad_w1, step)),
+       Nx.subtract(b1, Nx.multiply(grad_b1, step)),
+       Nx.subtract(w2, Nx.multiply(grad_w2, step)),
+       Nx.subtract(b2, Nx.multiply(grad_b2, step))
+     }, avg_loss, avg_accuracy}
+  end
```
josevalim, line 82: You implemented the grad by hand but now you can rollback and use `grad` as usual! My suggestion is to copy the example from EXLA again and mostly use it as is, simply removing the `@defn_compiler` directives. :)

josevalim: We need to update the regression example too!

torchx/test/torchx/nx_doctest_test.exs
josevalim: I see, thanks!

torchx/examples/mnist.exs
```diff
@@ -0,0 +1,163 @@
+defmodule MNIST do
+  import Nx.Defn
+
+
```
josevalim, line 4:
```suggestion
```

torchx/lib/torchx/backend.ex
```diff
@@ -202,6 +245,39 @@ defmodule Torchx.Backend do
...
+    axis = opts[:axis] || -1
+    keep_axes = opts[:keep_axes] || false
+
+    NIF.argmin(to_ref(t), axis, keep_axes) |> from_ref(out)
+  end
+
+  defp unsupported_option!(opts, key, acceptable_default) do
+    if(opts[key] != acceptable_default,
+      do: raise("#{inspect(key)} option is not supported in #{caller()}")
+    )
```
josevalim, line 271:
```suggestion
    if opts[key] != acceptable_default do
      raise "#{inspect(key)} option is not supported in #{caller()}"
    end
```

torchx/lib/torchx/backend.ex
```diff
@@ -143,6 +172,20 @@ defmodule Torchx.Backend do
     NIF.squeeze(to_ref(t)) |> from_ref(out)
   end
 
+  def expand(%T{} = t, shape) do
```
josevalim, line 175: Do we use `expand` anywhere?

torchx/examples/mnist.exs
```diff
@@ -0,0 +1,163 @@
...
+    |> Nx.logistic()
+    |> Nx.dot(w2)
+    |> Nx.add(b2)
+    |> softmax()
+  end
+
+  defn accuracy({w1, b1, w2, b2}, batch_images, batch_labels) do
+    Nx.mean(
+      Nx.equal(
+        Nx.argmax(Nx.as_type(batch_labels, {:s, 8}), axis: :output),
```
josevalim, line 30: We don't need this conversion here as argmax should work without casting.

torchx/examples/mnist.exs
```diff
@@ -0,0 +1,163 @@
...
+params = MNIST.init_random_params()
+
+IO.puts("Training MNIST for 10 epochs...\n\n")
+final_params = MNIST.train(train_images, train_labels, params, epochs: 10)
+
+IO.puts("The result of the first batch")
+IO.inspect(MNIST.predict(final_params, hd(train_images)) |> Nx.argmax(axis: :output))
+
+IO.puts("Labels for the first batch")
+IO.inspect(hd(train_labels) |> Nx.as_type({:s, 8}) |> Nx.argmax(axis: :output))
```
josevalim, line 163: We shouldn't need this conversion here either. :)

torchx/examples/mnist_native.exs
```diff
@@ -0,0 +1,188 @@
...
+    z2 = Nx.dot(a1, w2) |> Nx.add(b2)
+    preds = softmax(z2)
+
+    batch_loss =
+      Nx.sum(Nx.mean(Nx.multiply(Nx.log(preds), batch_labels), axes: [:output])) |> Nx.negate()
+
+    batch_accuracy =
+      Nx.mean(
+        Nx.equal(
+          Nx.argmax(Nx.as_type(batch_labels, {:s, 8}), axis: :output),
```
josevalim, line 49: We shouldn't need this conversion!

torchx/examples/mnist_native.exs
```diff
@@ -0,0 +1,188 @@
...
+IO.puts("Initializing parameters...\n")
+params = MNIST.init_random_params()
+
+final_params = MNIST.train(train_images, train_labels, params)
+
+IO.puts("The result of the first batch")
+IO.inspect(MNIST.predict(final_params, hd(train_images)) |> Nx.argmax(axis: :output))
+
+IO.puts("Labels for the first batch")
+IO.inspect(hd(train_labels) |> Nx.as_type({:s, 8}) |> Nx.argmax(axis: :output))
```
josevalim, line 188: We shouldn't need this conversion!

torchx/examples/mnist.exs
versilov: PyTorch does not implement argmax for unsigned byte.
Compilation fails without this type cast:

```
== Compilation error in file examples/mnist.exs ==
** (RuntimeError) Torchx: "argmax_cpu" not implemented for 'Bool' in NIF.argmax/3
```

torchx/examples/mnist.exs
versilov: Same reason as above.
Should we add type check in Torchx argmax and issue a warning or raise more imperative error message?

torchx/lib/torchx/backend.ex
versilov: I introduced it while trying to implement broadcasting along axis, which does not have direct analogue in PyTorch.

torchx/examples/mnist.exs
josevalim: I see, thank you! My suggestion is for then to convert them to s8 before we call the neural network! I think this is for labels, which could be s8 from the beginning.

torchx/lib/torchx/backend.ex
```diff
@@ -143,6 +172,20 @@ defmodule Torchx.Backend do
...
+    NIF.expand(to_ref(t), shape) |> from_bare_ref()
+  end
```
josevalim, line 177: Let's remove expand for now since we aren't going to use it and then we can merge it. :)

josevalim (approved):

PR #325, opened 2021-03-11
feat: add Nx.tile/2 function

polvalente:
Implements the `Nx.tile/2` function as per the [Numpy docs](https://numpy.org/doc/stable/reference/generated/numpy.tile.html)

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -1455,7 +1455,7 @@ defmodule Nx do
   ## Examples
 
       iex> a = Nx.tensor([0, 1, 2])
-      iex> Nx.tile(a, 2)
+      iex> Nx.tile(a, [2])
       #Nx.Tensor<
         s64[6]
         [0, 1, 2, 0, 1, 2]
@@ -1497,7 +1497,7 @@ defmodule Nx do
       >
 
       iex> b = Nx.tensor([[1,2],[3,4]])
-      iex> Nx.tile(b, 2)
+      iex> Nx.tile(b, [2])
       #Nx.Tensor<
         s64[2][4]
         [
@@ -1539,38 +1539,22 @@ defmodule Nx do
   ### Error cases
 
       iex> Nx.tile(Nx.tensor([1,2]), 1.0)
-      ** (ArgumentError) repetitions must be either an integer or a list of integers. Got: 1.0
+      ** (ArgumentError) repetitions must be a list of integers. Got: 1.0
 
       iex> Nx.tile(Nx.tensor([1,2]), [1, 1.0])
-       ** (ArgumentError) repetitions must be either an integer or a list of integers. Got: [1, 1.0]
+      ** (ArgumentError) repetitions must be a list of integers. Got: [1, 1.0]
 
       iex> Nx.tile(Nx.tensor([1,2]), nil)
-      ** (ArgumentError) repetitions must be either an integer or a list of integers. Got: nil
+      ** (ArgumentError) repetitions must be a list of integers. Got: nil
   """
   @doc type: :shape
   def tile(tensor, repetitions) do
     %T{shape: old_shape} = tensor = tensor!(tensor)
 
-    invalid_repetitions_message =
-      "repetitions must be either an integer or a list of integers. Got: #{inspect(repetitions)}"
-
-    repetitions =
-      case repetitions do
-        reps when is_list(reps) ->
-          if Enum.any?(reps, &(not is_integer(&1))) do
-            raise ArgumentError, invalid_repetitions_message
-          end
-
-          reps
-
-        reps when is_integer(reps) and reps > 0 ->
-          reps
-
-        _ ->
-          raise ArgumentError, invalid_repetitions_message
-      end
-
-    repetitions = List.wrap(repetitions)
+    unless tile_valid_repetitions?(repetitions) do
+      raise ArgumentError,
+            "repetitions must be a list of integers. Got: #{inspect(repetitions)}"
+    end
 
     num_dims = tuple_size(old_shape)
     length_reps = length(repetitions)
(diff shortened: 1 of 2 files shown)


nx/lib/nx.ex
```diff
@@ -1440,6 +1440,176 @@ defmodule Nx do
     end
   end
 
+  @doc """
+  Creates a new tensor by repeating the input tensor
+  along the given axes.
+
+  If the `tensor` has less dimensions than the repetitions given,
```
josevalim, line 1447:
```suggestion
  `repetitions` is either a number, which means all dimensions
  will be repeated by the given number, or a list of dimensions
  to repeat.

  If the `tensor` has less dimensions than the repetitions given,
```

nx/lib/nx.ex
```diff
@@ -1440,6 +1440,176 @@ defmodule Nx do
...
+      >
+
+      iex> b = Nx.tensor([[1,2],[3,4]])
+      #Nx.Tensor<
+        s64[2][2]
+        [
+          [1, 2],
+          [3, 4]
+        ]
+      >
```
josevalim, line 1506:
```suggestion
```

nx/lib/nx.ex
```diff
@@ -1440,6 +1440,176 @@ defmodule Nx do
...
+          [1, 2, 1, 2],
+          [3, 4, 3, 4]
+        ]
+      >
+
+      iex> c = Nx.tensor([1,2,3,4])
+      #Nx.Tensor<
+        s64[4]
+        [1, 2, 3, 4]
+      >
```
josevalim, line 1538:
```suggestion
```

nx/lib/nx.ex
```diff
@@ -1440,6 +1440,176 @@ defmodule Nx do
...
+    ### Error cases
+
+    iex> Nx.tile(Nx.tensor([1,2]), 1.0)
+    ** (ArgumentError) repetitions must be either an integer or a list of integers. Got: 1.0
+
+    iex> Nx.tile(Nx.tensor([1,2]), [1, 1.0])
+    ** (ArgumentError) repetitions must be either an integer or a list of integers. Got: [1, 1.0]
+
+    iex> Nx.tile(Nx.tensor([1,2]), nil)
+    ** (ArgumentError) repetitions must be either an integer or a list of integers. Got: nil
```
josevalim, line 1559:
```suggestion
  ### Error cases

      iex> Nx.tile(Nx.tensor([1,2]), 1.0)
      ** (ArgumentError) repetitions must be either an integer or a list of integers. Got: 1.0

      iex> Nx.tile(Nx.tensor([1,2]), [1, 1.0])
       ** (ArgumentError) repetitions must be either an integer or a list of integers. Got: [1, 1.0]

      iex> Nx.tile(Nx.tensor([1,2]), nil)
      ** (ArgumentError) repetitions must be either an integer or a list of integers. Got: nil
```

nx/lib/nx.ex
```diff
@@ -1440,6 +1440,176 @@ defmodule Nx do
...
+  """
+  @doc type: :shape
+  def tile(tensor, repetitions) do
+    %T{shape: old_shape} = tensor = tensor!(tensor)
+
+    invalid_repetitions_message =
+      "repetitions must be either an integer or a list of integers. Got: #{inspect(repetitions)}"
```
josevalim, line 1566: This means you are always building the error message, even if you never raise.

Perhaps it is better to write it like this:

```elixir
unless valid_repetition?(repetitions) do
  raise ...
end
```

nx/lib/nx.ex
```diff
@@ -1440,6 +1440,176 @@ defmodule Nx do
...
+          reps
+
+        reps when is_integer(reps) and reps > 0 ->
+          reps
+
+        _ ->
+          raise ArgumentError, invalid_repetitions_message
+      end
+
+    repetitions = List.wrap(repetitions)
```
josevalim, line 1584: If `repetitions` as an integer is the same as `[repetition]`, then I propose to not accept integer inputs. Alternatively, we can support integer inputs with a special meaning (it means repeating all dimensions with that given value).

nx/lib/nx.ex
```diff
@@ -1440,6 +1440,176 @@ defmodule Nx do
...
+
+    # grow the dimensionality of the tensor by new_dims_count
+    shape_grow_count = Kernel.max(length_reps - num_dims, 0)
+    resized_shape_list = List.duplicate(1, shape_grow_count) ++ Tuple.to_list(old_shape)
+
+    repetitions_grow_count = Kernel.max(num_dims - length_reps, 0)
+    resized_repetitions = List.duplicate(1, repetitions_grow_count) ++ repetitions
+
+    zipped_shapes = Enum.zip(resized_repetitions, resized_shape_list)
+    broadcast_shape = zipped_shapes |> Enum.flat_map(&Tuple.to_list/1) |> List.to_tuple()
```
josevalim, line 1597: This is probably better done with a private function?

```elixir
defp alternate([h1|t1], [h2|t2]), do: [h1, h2 | alternate(t1, t2)]
defp alternate([], []), do: []
```

nx/lib/nx.ex
```diff
@@ -1440,6 +1440,176 @@ defmodule Nx do
...
+  def tile(tensor, repetitions) do
+    %T{shape: old_shape} = tensor = tensor!(tensor)
+
+    invalid_repetitions_message =
+      "repetitions must be either an integer or a list of integers. Got: #{inspect(repetitions)}"
+
+    repetitions =
+      case repetitions do
+        reps when is_list(reps) ->
+          if Enum.any?(reps, &(not is_integer(&1))) do
```
josevalim, line 1571: We should validate the integers are all `>= 1`.

nx/lib/nx.ex
```diff
@@ -1440,6 +1440,176 @@ defmodule Nx do
...
+    resized_repetitions = List.duplicate(1, repetitions_grow_count) ++ repetitions
+
+    zipped_shapes = Enum.zip(resized_repetitions, resized_shape_list)
+    broadcast_shape = zipped_shapes |> Enum.flat_map(&Tuple.to_list/1) |> List.to_tuple()
+
+    tensor_reshape =
+      [nil | resized_shape_list]
+      |> Enum.intersperse(1)
+      |> tl()
+      |> List.to_tuple()
```
josevalim, line 1603:
```suggestion
      [1 | Enum.intersperse(resized_shape_list, 1)]
      |> List.to_tuple()
```

?

josevalim (commented):
Looks great! I have added some comments.

Btw, to keep this module clean, we tend to move a lot of the validation and shapes computation to the `Nx.Shape` module. So you could do something like this:

```elixir
{tensor_reshape, broadcast_shape, result_shape} = Nx.Shape.tile(shape, repetitions)

tensor
|> reshape(tensor_reshape)
|> broadcast(broadcast_shape)
|> reshape(result_shape)
```

nx/lib/nx.ex
polvalente: Yes, it's the same. I'm ok with not accepting integers. Since most of the API is akin to Numpy, having a completely different behavior for a valid input set in each one might confuse newcomers. Any thoughts?

nx/lib/nx.ex
polvalente: :sweat_smile: a lot better

nx/lib/nx.ex
polvalente: @josevalim aside from this discussion, the PR is ready for review :)
For now, it just rejects anything that isn't a list of integers

josevalim: >  For now, it just rejects anything that isn't a list of integers

Agreed, let's start with a restricted API. We can always add conveniences later.

nx/lib/nx.ex
```diff
@@ -1440,6 +1440,136 @@ defmodule Nx do
...
+      iex> Nx.tile(Nx.tensor([1,2]), nil)
+      ** (ArgumentError) repetitions must be a list of integers. Got: nil
+  """
+  @doc type: :shape
+  def tile(tensor, repetitions) do
+    tensor = tensor!(tensor)
+
+    unless tile_valid_repetitions?(repetitions) do
+      raise ArgumentError,
+            "repetitions must be a list of integers. Got: #{inspect(repetitions)}"
```
josevalim, line 1556: For consistency:

```suggestion
      iex> Nx.tile(Nx.tensor([1,2]), 1.0)
      ** (ArgumentError) repetitions must be a list of integers, got: 1.0

      iex> Nx.tile(Nx.tensor([1,2]), [1, 1.0])
      ** (ArgumentError) repetitions must be a list of integers, got: [1, 1.0]

      iex> Nx.tile(Nx.tensor([1,2]), nil)
      ** (ArgumentError) repetitions must be a list of integers, got: nil
  """
  @doc type: :shape
  def tile(tensor, repetitions) do
    tensor = tensor!(tensor)

    unless tile_valid_repetitions?(repetitions) do
      raise ArgumentError,
            "repetitions must be a list of integers, got: #{inspect(repetitions)}"
```

josevalim:

PR #327, opened 2021-03-12
Torchx device specification

versilov:


torchx/.formatter.exs
```diff
@@ -1,6 +1,5 @@
 # Used by "mix format"
 [
-  locals_without_parens: [dnif: 1],
-  import_deps: [:nx],
+  locals_without_parens: [dnif: 1, defn: 1],
```
josevalim, line 3:
```suggestion
  locals_without_parens: [dnif: 1, defn: 2],
```

torchx/lib/torchx/backend.ex
```diff
@@ -477,9 +507,23 @@ defmodule Torchx.Backend do
...
+  defp parse_torch_device_str(str) when is_binary(str) do
+    str
+    |> String.split(":")
+    |> (fn
+          [type, index] ->
+            {String.to_existing_atom(type), String.to_integer(index)}
+
+          [type] ->
+            String.to_existing_atom(type)
+        end).()
```
josevalim, line 522:
```suggestion
    |> case do
          [type, index] ->
            {String.to_existing_atom(type), String.to_integer(index)}

          [type] ->
            String.to_existing_atom(type)
        end
```

torchx/lib/torchx/backend.ex
```diff
@@ -69,6 +69,28 @@ defmodule Torchx.Backend do
...
+    opencl: 4,
+    ideep: 5,
+    hip: 6,
+    fpga: 7,
+    msnpu: 8,
+    xla: 9,
+    vulkan: 10,
+    metal: 11,
+    xpu: 12
+  ]
```
josevalim, line 88:
```suggestion
  @devices {
    cpu: 0,
    cuda: 1,
    mkldnn: 2,
    opengl: 3,
    opencl: 4,
    ideep: 5,
    hip: 6,
    fpga: 7,
    msnpu: 8,
    xla: 9,
    vulkan: 10,
    metal: 11,
    xpu: 12
  }
```

torchx/test/torchx/device_test.exs
```diff
@@ -0,0 +1,21 @@
...
+
+  setup do
+    Nx.default_backend(Torchx.Backend)
+    :ok
+  end
+
+  describe "creation" do
+    test "from_binary" do
+      t = Nx.tensor([1, 2, 3], backend_options: [device: :cpu])
+      Nx.backend_transfer(t, TB, backend_options: [device: :cpu])
```
josevalim, line 14: For backend transfer, the options are already the backend options:

```suggestion
      Nx.backend_transfer(t, TB, [device: :cpu])
```

Btw, is there a way for us to ask Torch which devices are enabled?

torchx/lib/torchx/backend.ex
versilov: This is not going to compile

torchx/test/torchx/device_test.exs
versilov: Yes, can check, but for CUDA only.

torchx/lib/torchx/backend.ex
versilov: Or did you mean map?

torchx/lib/torchx/backend.ex
```diff
@@ -69,6 +69,29 @@ defmodule Torchx.Backend do
   defp from_torch_type(:float), do: {:f, 32}
   defp from_torch_type(:double), do: {:f, 64}
 
+  def device_option(opts), do: (opts[:backend_options] || [])[:device] || opts[:device] || :cpu
```
josevalim, line 72: This should be:

```suggestion
  def device_option(opts), do: opts[:device] || :cpu
```

And then you should do `opts[:backend_options]` on all places except the backend_transfer function. :)

torchx/lib/torchx/backend.ex
```diff
@@ -477,9 +508,23 @@ defmodule Torchx.Backend do
     defp check_shape_and_type!(ref, _, _), do: ref
   end
 
-  defp device(%T{data: %TB{ref: ref}}), do: NIF.device(ref) |> unwrap!() |> List.to_string()
+  def device(%T{data: %TB{ref: ref}}),
```
josevalim, line 511:
```suggestion
  @doc false
  def device(%T{data: %TB{ref: ref}}),
```

torchx/lib/torchx.ex
```diff
@@ -1,2 +1,21 @@
 defmodule Torchx do
+  alias Torchx.NIF
+
+  @doc """
+  Check if device of the given type is available for Torchx.
+  Device atom can be any of
+  [:cpu, :cuda, :mkldnn, :opengl, :opencl, :ideep, :hip, :fpga, :msnpu, :xla, :vulkan, :metal, :xpu].
+
+  But only :cuda is supported for now.
```
josevalim, line 9: Only cuda is supported or only cuda can be checked? :)

torchx/lib/torchx.ex
```diff
@@ -1,2 +1,21 @@
...
+  """
+
```
josevalim, line 11:
```suggestion
```

josevalim (approved):

torchx/lib/torchx.ex
versilov: That's a very good question! I think it's the latter.

PR #328, opened 2021-03-13
feat: concatenate grad

polvalente:
Adds the grad implementation for `concatenate/2` for #161


polvalente:
@josevalim I've pushed this initial draft because I got stumped by an infinite loop, so maybe I accidentally recursed into `grad(:concatenate, ...)` somewhere.

I plan on continuing this soon, but some more pointers are welcome!

The good news is that the test at least runs without raising a "no function clause matching" or "argument error" somewhere :sweat_smile:

nx/lib/nx/defn/grad.ex
```diff
@@ -424,6 +424,82 @@ defmodule Nx.Defn.Grad do
...
+    starts
+    |> Enum.zip(limits)
+    |> Enum.map(fn {start, limit} ->
+      {start, len} =
+        start
+        |> Enum.zip(limit)
+        |> Enum.map(fn {s, lim} -> {s, lim - s + 1} end)
+        |> Enum.unzip()
+
+      {Nx.slice(ans, start, len), g}
```
polvalente, line 498: Should this be Expr.slice instead?

nx/test/nx/defn/grad_test.exs
```diff
@@ -60,6 +60,51 @@ defmodule Nx.Defn.GradTest do
...
+    end
+
+    defn concatenate_grad_power(t) do
+      grad(
+        t,
+        stop_grad(Nx.sum(Nx.sum(Nx.concatenate([Nx.power(t, 2), Nx.power(t, 3)]), axes: [0])))
+      )
+    end
+
+    test "computes concatenate/1 grad for various input shapes" do
```
polvalente, line 74: Ok, so after some investigation which involved inserting `print` statements in my local copy of Jax, I managed to get some results which make sense with intermediate `"result"` which is the printed `return ...` from `jax/_src/lax/lax.py -> _concatenate_transpose_rule`, but I'm not sure if these tests should actually be passing or if there's something wrong.

My main concern now is that the final output of `jax.grad` isn't the same as my `grad` calls, and if I remove `stop_grad`, everything goes into an infinite loop.

nx/test/nx/defn/grad_test.exs
```diff
@@ -1214,24 +1259,21 @@ defmodule Nx.Defn.GradTest do
...
-                         [0.0, 0.0],
-                         [0.0, 0.0]],
-                        [[0.0, 1.0],
-                         [0.0, 1.0],
-                         [0.0, 1.0]],
-                        [[0.0, 1.0],
-                         [0.0, 1.0],
-                         [0.0, 1.0]]]])
+
+      rhs =
```
polvalente, line 1263: Should I revert this formatting?

nx/test/nx/defn/grad_test.exs
josevalim: Here is a diff that fixes the loop:

```diff
diff --git a/nx/lib/nx/defn/grad.ex b/nx/lib/nx/defn/grad.ex
index 85b2360..c4fd586 100644
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -491,18 +491,17 @@ defmodule Nx.Defn.Grad do
     IO.inspect(starts, label: "starts")
     IO.inspect(limits, label: "limits")
 
-    pre_result = Nx.concatenate(tensors, axis: axis)
-
     starts
     |> Enum.zip(limits)
-    |> Enum.map(fn {start, limit} ->
+    |> Enum.zip(tensors)
+    |> Enum.map(fn {{start, limit}, t} ->
       {start, len} =
         start
         |> Enum.zip(limit)
         |> Enum.map(fn {s, lim} -> {s, lim - s} end)
         |> Enum.unzip()
 
-      {pre_result, Nx.slice(pre_result, start, len)}
+      {t, Nx.slice(ans, start, len)}
     end)
     |> IO.inspect(label: "slices")
     |> grad_pairs(g, cache)
diff --git a/nx/test/nx/defn/grad_test.exs b/nx/test/nx/defn/grad_test.exs
index ebf3229..8775341 100644
--- a/nx/test/nx/defn/grad_test.exs
+++ b/nx/test/nx/defn/grad_test.exs
@@ -61,13 +61,13 @@ defmodule Nx.Defn.GradTest do
     end
 
     defn concatenate_grad(t) do
-      grad(t, stop_grad(Nx.sum(Nx.sum(Nx.concatenate([t, t]), axes: [0]))))
+      grad(t, Nx.sum(Nx.sum(Nx.concatenate([t, t]), axes: [0])))
     end
 
     defn concatenate_grad_power(t) do
       grad(
         t,
-        stop_grad(Nx.sum(Nx.sum(Nx.concatenate([Nx.power(t, 2), Nx.power(t, 3)]), axes: [0])))
+        Nx.sum(Nx.sum(Nx.concatenate([Nx.power(t, 2), Nx.power(t, 3)]), axes: [0]))
       )
     end
 ```

File: nx/test/nx/defn/grad_test.exs [hunk 4]
```diff
@@ -60,6 +60,51 @@ defmodule Nx.Defn.GradTest do
...
     end
 
+    defn concatenate_grad(t) do
+      grad(t, stop_grad(Nx.sum(Nx.sum(Nx.concatenate([t, t]), axes: [0]))))
+    end
+
+    defn concatenate_grad_power(t) do
+      grad(
+        t,
+        stop_grad(Nx.sum(Nx.sum(Nx.concatenate([Nx.power(t, 2), Nx.power(t, 3)]), axes: [0])))
```
josevalim, line 70: No need for double sum. Also, please move concatenate tests alongside the tests for reshape, squeeze, etc. They have their own describe IIRC.

nx/test/nx/defn/grad_test.exs
josevalim: No, keep it formatted like you just did :)

nx/lib/nx/defn/grad.ex
```diff
@@ -424,6 +424,90 @@ defmodule Nx.Defn.Grad do
...
+
+    operand_shapes = Enum.map(tensors, & &1.shape)
+    IO.inspect(operand_shapes, label: "operand shapes")
+
+    limit_points =
+      Enum.reduce(operand_shapes, [], fn
+        x, [] -> [elem(x, axis)]
+        x, [h | _] = acc -> [elem(x, axis) + h | acc]
+      end)
+      |> Enum.reverse()
```
josevalim, line 467:
```suggestion
    limit_points = Enum.scan(tensors, 0, fn t, acc -> elem(t.shape, axis) + acc end)
```

Or if you want to be super concise:

```suggestion
    limit_points = Enum.scan(tensors, 0, &elem(&1.shape, axis) + &2)
```

nx/lib/nx/defn/grad.ex
```diff
@@ -424,6 +424,90 @@ defmodule Nx.Defn.Grad do
...
+
+    n_tensors = length(tensors)
+    ndim = tuple_size(ans_shape)
+
+    starts =
+      Enum.map(0..(n_tensors - 1), fn row ->
+        Enum.map(0..(ndim - 1), fn col ->
```
josevalim, line 474: `0..(ndim - 1)` will be wrong if `ndim` is zero but I assume we can't concat scalar tensor anyways. In any case, the preferred approach in those cases is to use Nx.axes:


```suggestion
    n_tensors = length(tensors)
    axes = Nx.axes(ans_shape)

    starts =
      Enum.map(0..(n_tensors - 1), fn row ->
        Enum.map(axes, fn col ->
```

nx/lib/nx/defn/grad.ex
```diff
@@ -424,6 +424,90 @@ defmodule Nx.Defn.Grad do
...
+          if col == axis and row > 0 do
+            Enum.at(limit_points, row - 1)
```
josevalim, line 476:
```suggestion
            Enum.fetch!(limit_points, row - 1)
```

nx/lib/nx/defn/grad.ex
```diff
@@ -424,6 +424,90 @@ defmodule Nx.Defn.Grad do
...
+          else
+            0
+          end
+        end)
+      end)
+
+    limits =
```
josevalim, line 483:
```suggestion
    # TODO: Use Enum.zip_with/3 on Elixir v1.12
    limits =
```

nx/test/nx/defn/grad_test.exs
polvalente: Thank you!

nx/test/nx/defn/grad_test.exs
```diff
@@ -1580,6 +1574,57 @@ defmodule Nx.Defn.GradTest do
...
+    defn concatenate_grad(t) do
+      grad(t, Nx.sum(Nx.concatenate([t, t], axis: 0)))
+    end
+
+    defn concatenate_grad_power(t) do
+      grad(
+        t,
+        Nx.sum(Nx.concatenate([Nx.power(t, 2), Nx.power(t, 3)], axis: 0))
+      )
+    end
```
polvalente, line 1587: If I remove all sums I get a "** (ArgumentError) can only compute gradients of expressions that return scalars, got shape: {2}" message. I assume the issue you commented about was that the outer sum was a no-op?

Other than this, there seems to be something wrong with the grad when applied to the `Nx.power` calls.
The expected result is still documented in the relevant test assertion, but the actual result is `Nx.tensor([5.0, 112.0], [783.0, 3.2e3]])`. I'll keep investigating

josevalim: Sum should sum across all dimensions by default and return a scalar, no?

polvalente: Yes, that's what I meant by "no-op": sum(sum(....sum(x))) === sum(x) when using the default options. At first I thought that the sum would happen across a single axis, and thus the need for nesting sums.

nx/lib/nx/defn/grad.ex
```diff
@@ -424,6 +424,50 @@ defmodule Nx.Defn.Grad do
...
+
+    # TODO: Use Enum.zip_with/3 on Elixir v1.12
+    limits =
+      ans_shape
+      |> Tuple.to_list()
+      |> List.duplicate(n_tensors)
+      |> Enum.zip(limit_points)
+      |> Enum.map(fn {row, p} -> List.replace_at(row, axis, p) end)
+
+    ones = Nx.map(ans, fn _ -> 1 end)
```
polvalente, line 454: @josevalim Upon inspecting the following code from Jax, I found that there was a `1`s array with shape `ans.shape` arising somewhere.

```python
jax.make_jaxpr(jax.grad(lambda x: jnp.sum((jnp.concatenate((jnp.power(x, 2), jnp.power(x, 3)), 0)))))(np.array([[1.0, 2.0], [3.0, 4.0]]))

{{ lambda  ; a.
  let b = integer_pow[ y=2 ] a
      c = integer_pow[ y=1 ] a
      d = mul 2.0 c
      e = integer_pow[ y=3 ] a
      f = integer_pow[ y=2 ] a
      g = mul 3.0 f
      h = concatenate[ dimension=0 ] b e
      _ = reduce_sum[ axes=(0, 1) ] h
      i = broadcast_in_dim[ broadcast_dimensions=(  )
                            shape=(4, 2) ] 1.0
      j = slice[ limit_indices=(2, 2)
                 start_indices=(0, 0)
                 strides=None ] i
      k = slice[ limit_indices=(4, 2)
                 start_indices=(2, 0)
                 strides=None ] i
      l = mul k g
      m = mul j d
      n = add_any l m
  in (n,) }
```

After defining this array as the one we apply the slices, everything is working as expected. I'm not really sure if this is the correct fix, but if it isn't, this might give some insight on what the fix is. Any thoughts?

After we settle on this aspect, I plan on refactoring some of the code because I think that we can build `starts` and `limits` iteratively in a `Enum.reduce(tensors, ...)` call

polvalente: Update: this implementation makes the tests pass, which makes me believe I have gotten something *really* wrong somewhere:

```elixir
defp grad(:concatenate, [tensors, _axis], _ans, g, cache) do
  tensors
  |> Enum.map(fn t ->
    {t, Nx.map(t, fn _ -> 1 end)}
  end)
  |> grad_pairs(g, cache)
end
```

polvalente: After giving this some thought, I think this makes sense...

If we define a 2 tensor concatenation as
```
concat(A, B) = [[eye(shape(A)) zeros(shape(B))], [zeros]] dot concat(A, 0) + [[zeros], [zeros(shape(A)) ones(shape(B))]] dot concat(0, B)
                     = C1 dot A' + C2 dot B'
```

Then we could calculate grad(concat(A,B)) as follows

```
grad(concat(A, B)) = grad(C1) dot A' + C1 dot grad(A') + grad(C2) dot B' + C2 dot grad(B')
                              # assuming grad(C1) = grad(C2) = 0
                              = C1 dot grad(A') + C2 dot grad(B')
                              = concat(grad(A), grad(B))
```

This would mean that `grad(sum(concat([a, b]))` would result in `grad(a) + grad(b)`, which is indeed the actual result we get in the test cases. For example:

```elixir
defn my_grad_power(t, n \\ 2) do
  grad(t, Nx.sum(Nx.power(t, n)))
end

defn concatenate_grad_power(t) do
  grad(
    t,
    Nx.sum(Nx.concatenate([Nx.power(t, 2), Nx.power(t, 3)], axis: 0))
  )
end

test "computes grad for powers of a {2, 2}-tensor" do
  IO.inspect(my_grad_power(Nx.tensor([[1.0, 2.0], [3.0, 4.0]])))
  IO.inspect(my_grad_power(Nx.tensor([[1.0, 2.0], [3.0, 4.0]]), 3))

   assert concatenate_grad_power(Nx.tensor([[1.0, 2.0], [3.0, 4.0]])) ==
      Nx.tensor([[5.0, 16.0], [33.0, 56.0]])
end
```

Passes and prints:
```elixir
#Nx.Tensor<
  f32[2][2]
  [
    [2.0, 4.0],
    [6.0, 8.0]
  ]
>
#Nx.Tensor<
  f32[2][2]
  [
    [3.0, 12.0],
    [27.0, 48.0]
  ]
>
```

I'm not sure if I followed the code correctly, but `grad_pairs` seems to end up either ignoring this value (`x`) calling `Nx.multiply(t, x)`, which would yield `t` if `x == ones(shape(t))`

Any thoughts on this?

nx/lib/nx/defn/grad.ex
```diff
@@ -424,6 +424,48 @@ defmodule Nx.Defn.Grad do
...
+    |> Enum.zip(limits)
+    |> Enum.zip(tensors)
+    |> Enum.map(fn {{start, limit}, t} ->
+      {start, len} =
+        start
+        |> Enum.zip(limit)
+        |> Enum.map(fn {s, lim} -> {s, lim - s} end)
+        |> Enum.unzip()
+
+      {t, Nx.slice(g, start, len)}
```
polvalente, line 464: Does this make sense?

nx/lib/nx/defn/grad.ex
```diff
@@ -424,6 +424,24 @@ defmodule Nx.Defn.Grad do
...
+      t_len = elem(t.shape, axis)
+      current_limit = t_len + limit
+      start = List.replace_at(zero_axes, axis, limit)
+      len = List.replace_at(ans_shape_list, axis, t_len)
+
+      result = [{t, Nx.slice(g, start, len)} | result]
+      {current_limit, result}
+    end)
+    |> elem(1)
+    |> grad_pairs(g, cache)
```
polvalente, line 442:
```suggestion
    |> elem(1)
    |> Enum.reverse()
    |> grad_pairs(g, cache)
```

This should be applied if `grad_pairs` depends on the ordering of the tensors. The tests kept passing even without the reverse, so I assumed it wasn't needed.

nx/lib/nx/defn/grad.ex
```diff
@@ -424,6 +424,24 @@ defmodule Nx.Defn.Grad do
...
+    ans_shape_list = Tuple.to_list(ans_shape)
+
+    tensors
+    |> Enum.reduce({0, []}, fn t, {limit, result} ->
+      t_len = elem(t.shape, axis)
+      current_limit = t_len + limit
+      start = List.replace_at(zero_axes, axis, limit)
+      len = List.replace_at(ans_shape_list, axis, t_len)
+
+      result = [{t, Nx.slice(g, start, len)} | result]
```
polvalente, line 438: This basically means we travel along the concatenation axis, retrieving the gradient corresponding to the current tensor and associating it to the tensor itself

nx/lib/nx/defn/grad.ex
josevalim: The ordering doesn't matter. grad_pairs will basically call Nx.add/2 after grading each pair.

josevalim, line 442: My take:

```suggestion
    {pairs, _} = 
      Enum.map_reduce(tensors, limit, fn t, limit ->
        t_len = elem(t.shape, axis)
        current_limit = t_len + limit
        start = List.replace_at(zero_axes, axis, limit)
        len = List.replace_at(ans_shape_list, axis, t_len)
        {{t, Nx.slice(g, start, len)}, current_limit}
      end)

    grad_pairs(pairs, g, cache)
```

I am not a big fan to piping to elem, I tend to prefer pattern matching, but piping here would be fine too.

nx/lib/nx/defn/grad.ex
josevalim: Yup, sounds correct to me!

josevalim (commented):
Looks great, I have added just one minor nitpick to use map_reduce and it looks good to go!

nx/lib/nx/defn/grad.ex
polvalente: Is there an specific reason? I chose to use it here because the output always has the same format, but I can see this being an issue when the reduce can return tagged tuples `{:ok, _}` | `{:error, _}`

josevalim:

PR #331, opened 2021-03-13
feat: accept prng seed for random_* functions

polvalente:
As per the discussion in Slack, this is a proposal for accepting a random seed option in `Nx.random_*` functions.

I'm not sure where to modify to implement this for Torchx and ExLA, so pointers are welcome!

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -705,7 +705,6 @@ defmodule Nx do
     * `:backend` - the backend to allocate the tensor on. It is either
       an atom or a tuple in the shape `{backend, options}`. This option
       is ignored inside `defn`
-
   """
   @doc type: :random
   def random_uniform(tensor_or_shape, min, max, opts \\ [])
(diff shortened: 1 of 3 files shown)


seanmor5:
We definitely need the ability to seed RNGs, and I'm confident this approach will work for Torchx and other backends, but I'm not 100% sure how it will extend to EXLA and other Tensor compilers that can't depend on a stateful RNG. As an example, we can set the RNG seed as an executable run option in EXLA, but that would apply to the entire execution and not individual calls to `random_uniform` within `defn`. It would also need to be passed as a compile option rather than directly to calls to `random_x`.

I think it probably is best to move forward instead with Jax-style stateless PRNGs because we can implement them with our current API and have a solution that extends to every compiler and backend. The JAX PRNGs also tout themselves as perfect for distribution/parallel computing, and I think that aligns with some of our future goals. In order to rework this from the EXLA perspective, it would probably involve using `RngBitGenerator` and reimplementing our current random functions in terms of that and other primitives.

polvalente:
@seanmor5 thanks for the feedback! Should we close this and create an issue instead?
It seems that a more in-depth discussion is needed for the behavior one stateless PRNG would need to expose

seanmor5:
@polvalente I think this is definitely worth opening an issue on first and then we can revisit the PR later. It may end up that we go with this approach and then you already have the solution, but it's also possible we go a different direction. Either way this is still a valuable contribution :)

Issue #223, opened 2021-02-17
Support dynamic indexing on slices and dynamic updates


josevalim:

billylanchantin:
@josevalim I've been thinking about this issue since it came out of #221 (though I've been preoccupied by a contribution to ElixirLS), and I've done a little research into the various options. It seems that we likely want to support some version of the [gather/scatter paradigm](https://en.wikipedia.org/wiki/Gather-scatter_%28vector_addressing%29). But how those operations generalize to tensors is under defined.

I found a decent breakdown of various approaches that different deep learning frameworks have taken to the generalization [here](https://www.pathpartnertech.com/gather-scatter-operation-in-deep-learning-framework/) (warning: a _lot_ of index-chasing; you might want to just skim the comparison tables at first). I'm confident that we can find an API that works for our purposes, though I think a little experimentation is inevitable.

It's my hope to start working on this in my free time, but I wanted to gut-check the idea before I got too far.

Note: I also think that we can build a solution to #229 around the "scatter" portion of this.

seanmor5:
@billylanchantin From the XLA perspective, I can confirm that gather/scatter is pretty much how JAX implements indexing.

See: https://github.com/google/jax/blob/master/jax/_src/numpy/lax_numpy.py#L5389 and https://github.com/google/jax/blob/master/jax/_src/numpy/lax_numpy.py#L4352

From a brief overview it looks like they rewrite indexing and index updates as gather/scatter ops.

See also: https://github.com/google/jax/blob/84f55f87ecd4b1a69ed7eac48b144b57df9d3f11/jax/_src/ops/scatter.py

billylanchantin:
@seanmor5

> From the XLA perspective, I can confirm that gather/scatter is pretty much how JAX implements indexing.

Awesome, thanks!

> From a brief overview it looks like they rewrite indexing and index updates as gather/scatter ops.

Yeah that's what I'm seeing across the board, not just with JAX (though my investigations have also been brief). AFAICT, we really have to nail gather/scatter regardless of how we want the higher-level access/update API to work.

BTW, my plan is to copy what JAX does as much as I can. I just want to be cognizant of how gather/scatter works outside of XLA while I do so.

Issue #221, opened 2021-02-17
Support Boolean-indexing


billylanchantin:
## Background

See here for a reference on "Advance Indexing" in NumPy: https://numpy.org/doc/stable/reference/arrays.indexing.html#advanced-indexing

## Description

There are 2 types of advanced indexing in NumPy: integer-based and boolean-based. From my (cursory) reading of `Nx.slice/4`, integer-indexing is essentially covered. What I've not seen yet is boolean-indexing.

A common thing to do in NumPy is to access and/or manipulate an array based on some criteria. E.g.

```python
>>> import numpy as np
>>> a = np.arange(5)
array([0, 1, 2, 3, 4])
>>> b = a >= 3
array([False, False, False,  True,  True], dtype=bool)
>>> a[b]
array([3, 4])
>>> a[b] = 100
array([  0,   1,   2, 100, 100])
```

If you access an array `a` with a boolean array `b` (where `b`'s shape can be broadcast to `a`'s shape), NumPy knows to do things with elements of `a` whose corresponding element in `b` is `True`.

`Nx.select/3` seems to cover the result of doing `a[a >= 3]` fairly well (thanks for [pointing that out](https://elixir-lang.slack.com/archives/C01M2B96EF9/p1613583744229600), @seanmor5!). But the boolean-indexing counter part isn't there. I personally find boolean-indexing very powerful. IMO it's a type of de-functionalization for operations like select/replace, but with really powerful bonuses like being able to do boolean operations on the intermediate objects.

Is boolean-indexing something Nx could/should support? I'd be willing to try and implement it if so.

seanmor5:
I think this would be generalized under something of a dynamic slicing API. `Nx.slice/4` covers slicing pretty well so far, (integer and range support), but we'd need to extend that to accept tensors. I think we could then resolve `a >= 3` to a tensor of indices and do the slicing that way. I think the binary implementation would be pretty straightforward, but the EXLA implementation might need something like [Gather](https://www.tensorflow.org/xla/operation_semantics#gather) or perhaps it could all be implemented in terms of `Nx.select/3`?

I'm for the addition of advanced/boolean-indexing. I think this is a good issue to gauge support from others as well.

josevalim:
It is worth noting that at the moment we don't have a predicates type. I assume that would be a requirement for us to add this feature because otherwise it is impossible to know if we are passing a tensor of u8 to access certain positions or doing predicates.

[...]

PR #338, opened 2021-03-16
wip: add sort grad

polvalente:


nx/lib/nx/defn/grad.ex
```diff
@@ -440,6 +440,77 @@ defmodule Nx.Defn.Grad do
...
+    # iotas = []
+    # for dim, size in enumerate(shape):
+    #   dtype = np.int32 if size < np.iinfo(np.int32).max else np.int64
+    #   iotas.append(broadcasted_iota(dtype, shape, dim))
+    # sorted_tensor = sort_p.bind(*(tensor + (iotas[dimension],)), dimension=dimension,
+    #                       is_stable=is_stable, num_keys=num_keys)
+    # idx = tuple(sorted_tensor[-1] if i == dimension else iotas[i]
+    #             for i in range(len(shape)))
+    # g_out = tuple(t if type(t) is ad_util.Zero else t[idx] for t in g)
+    # return tuple(sorted_tensor[:-1]), g_out
```
polvalente, line 454: This is the code I tried to translate from Jax.
It seems that the basic idea is to retrieve the corresponding grads for each "line" of the sorted tensor, by logical indexing.

However, we don't have logical indexing available, so I tried another approach by concatenating a sequence of numbers to the tensor, slicing it afterwards. However, I couldn't find a way to convert the resulting {n} tensor into a list of integers, which in turn makes the construction of `starts` impossible.

Any tips on how we could approach this?

josevalim: What is logical indexing in this case? Are you selecting only some lines and ignoring the other ones? Or are you re-ordering?

polvalente: From what I gathered from the code, it basically reorders the gradient tensor according to the ordering applied to the original tensor

polvalente:
I'm closing this PR for now because it'll be a lot easier to implement this grad when #223 or #221 are available

PR #339, opened 2021-03-16
Move some responsibilities to Torchx

versilov:
I am not sure we should go on with `{device, ref}` tuples.
Getting device from ref is a cheap operation. And working with bare refs is a way more straightforward.


torchx/lib/torchx.ex
```diff
@@ -17,4 +19,71 @@ defmodule Torchx do
   """
   def device_count(:cuda), do: NIF.cuda_device_count()
   def device_count(_), do: raise("Only CUDA devices can be counted for now.")
+
+  # LibTorch API bindings
+
+  ## Creation
+
+  def arange(from, to, step \\ 1, opts \\ []) do
+    type = opts[:type] || Nx.Type.infer([from, to])
```
josevalim, line 28: I think the API in this module should work directly with torch types instead of Nx types. WDYT?

torchx/lib/torchx.ex
```diff
@@ -17,4 +19,71 @@ defmodule Torchx do
...
+    device = opts[:device] || :cpu
+
+    NIF.arange(from, to, step, torch_type(type), torch_device(device))
+    |> wrap_with_device()
```
josevalim, line 32: In my mind we would do this:

```suggestion
    device = opts[:device] || :cpu

    NIF.arange(from, to, step, torch_type(type), torch_device(device))
    |> wrap_with_device(device)
```

We would always pass the device in.

torchx/lib/torchx.ex
```diff
@@ -17,4 +19,71 @@ defmodule Torchx do
...
+
+  def tensordot(left_ref, right_ref, left_axes, right_axes) do
+    NIF.tensordot(
+      left_ref,
+      right_ref,
+      left_axes,
+      right_axes
+    )
+    |> wrap_with_device()
+  end
```
josevalim, line 45: Similarly this function would look like this:

```suggestion
  def tensordot(left_ref, right_ref, left_axes, right_axes) do
    {device, refs} = to_refs([left_ref, right_ref])
    NIF.tensordot(
      left_ref,
      right_ref,
      left_axes,
      right_axes
    ) |> wrap_device(device)
  end
  
  defp to_refs([{device, ref} | t]) do
    refs =
      Enum.map(t, fn
        {^device, ref} -> ref
        {other_device, _ref} ->
          raise "cannot perform operations on across devices: #{inspect(device)} and #{inspect(other_device)}
      end)
      
    {device, [ref | refs]}
  end
```

This way we would always check the device and always know the device upfront to wrap in. WDYT?

torchx/lib/torchx.ex
```diff
@@ -17,4 +19,69 @@ defmodule Torchx do
...
+    |> wrap_with_device(device)
+  end
+
+  ## Utils
+
+  @doc false
+  def type_of(ref), do: NIF.scalar_type(ref) |> unwrap!()
+
+  @doc false
+  def device_of(ref),
```
josevalim, line 50: Note this should be `{device, ref}` too. But we can handle this later. :)

torchx/lib/torchx.ex
```diff
@@ -1,6 +1,8 @@
 defmodule Torchx do
   alias Torchx.NIF
 
+  import Torchx.Backend, only: [torch_type: 1, torch_device: 1]
```
josevalim, line 4: We don't use torch_type here anymore, so we can remove it.

We also handle all of the device logic here, so perhaps move torch_device here too? 🤔

torchx/lib/torchx.ex
versilov: Made a shortcut for now.

josevalim (approved):

PR #345, opened 2021-03-20
feat: add matrix inversion

polvalente:
Adds matrix inversion

Indirectly related to #157 and unblocks some of the grads in #161

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -683,8 +683,8 @@ defmodule Nx do
 
     unless min_shape == {} and max_shape == {} do
       raise ArgumentError,
-             "random_uniform/3 expects min and max to be scalars, got:" <>
-             " min shape: #{inspect(min_shape)} and max shape: #{inspect(max_shape)}"
+            "random_uniform/3 expects min and max to be scalars, got:" <>
+              " min shape: #{inspect(min_shape)} and max shape: #{inspect(max_shape)}"
     end
 
     unless Nx.Type.float?(type) or (Nx.Type.integer?(type) and Nx.Type.integer?(range_type)) do
@@ -797,13 +797,15 @@ defmodule Nx do
     type = Nx.Type.normalize!(opts[:type] || {:f, 32})
 
     unless mu_shape == {} and sigma_shape == {} do
-      raise ArgumentError, "random_normal/3 expects mu and sigma to be scalars" <>
-                           " got: mu shape: #{inspect(mu_shape)} and sigma shape: #{inspect(sigma_shape)}"
+      raise ArgumentError,
+            "random_normal/3 expects mu and sigma to be scalars" <>
+              " got: mu shape: #{inspect(mu_shape)} and sigma shape: #{inspect(sigma_shape)}"
     end
 
     unless Nx.Type.float?(mu_type) and Nx.Type.float?(sigma_type) do
-      raise ArgumentError, "random_normal/3 expects mu and sigma to be float types," <>
-                           " got: mu type: #{inspect(mu_type)} and sigma type: #{inspect(sigma_type)}"
+      raise ArgumentError,
+            "random_normal/3 expects mu and sigma to be float types," <>
+              " got: mu type: #{inspect(mu_type)} and sigma type: #{inspect(sigma_type)}"
     end
 
     unless Nx.Type.float?(type) do
@@ -7887,7 +7889,7 @@ defmodule Nx do
         f64[3]
         [1.0, 1.0, -1.0]
       >
-      
+
   ### Error cases
 
       iex> Nx.triangular_solve(Nx.tensor([[3, 0, 0, 0], [2, 1, 0, 0]]), Nx.tensor([4, 2, 4, 2]), trans: 0)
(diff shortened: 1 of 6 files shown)


seanmor5:
Hi, @polvalente! This looks good, I am only hesitant to add this into the API because it looks pretty easily implementable as a routine in defn, but it is probably useful if it is going to be a common method. I don't work much with matrix inversion or any of the other linalg operators for that matter that often, so I can't speak much on how often it will be used. Are there a lot of use cases?

It might also be time for a discussion on how we are going to split out other linear algebra operators, random operators, etc. that are built on top of our existing primitives.

polvalente:
> Hi, @polvalente! This looks good, I am only hesitant to add this into the API because it looks pretty easily implementable as a routine in defn, but it is probably useful if it is going to be a common method. I don't work much with matrix inversion or any of the other linalg operators for that matter that often, so I can't speak much on how often it will be used. Are there a lot of use cases?
> 
> It might also be time for a discussion on how we are going to split out other linear algebra operators, random operators, etc. that are built on top of our existing primitives.

My main motivation for this is because matrix inversion is needed for some of the grads I explored throughout this week (namely cholesky, QR and LU). Since this was basically fully implemented using the existing API, I'm ok with having it as a private function in the Nx.Grad module.

However, other people might find this useful. For instance, we can use inversion for reversing computer graphics operations (i.e. combined rotation and transalation matrices). 

As you said, maybe we could have this as a part of a documented linalg tutorial which could be implemented by users as needed. The comments included in this also show how to use QR together with triangular_solve for solving any (sovable) linear system.

Another approach we could take with inversion specifically is to turn this into a binary backend implementation so we can delegate this to the proper implementation in other backends.

nx/lib/nx.ex
```diff
@@ -7919,6 +7921,101 @@ defmodule Nx do
...
+  @doc type: :linalg
+  def invert(tensor) do
+    %T{shape: {m, n}} = tensor = tensor!(tensor)
+
+    if m != n,
+      do:
+        raise(
+          ArgumentError,
+          "can only invert square matrices, got: {#{m}, #{n}}"
+        )
```
josevalim, line 7981: This is unnecessarily verbose in this format. Maybe it should be a regular do block?

```suggestion
    if m != n do
      raise ArgumentError,
              "can only invert square matrices, got: {#{m}, #{n}}"
    end
```

nx/lib/nx.ex
```diff
@@ -7919,6 +7921,101 @@ defmodule Nx do
...
+    |> Enum.map(fn i ->
+      b = identity |> slice([0, i], [n, 1]) |> Nx.reshape({n})
+
+      y =
+        try do
+          triangular_solve(l, b)
+        rescue
+          ArgumentError ->
+            raise ArgumentError, "can't invert a singular matrix"
+        end
```
josevalim, line 8012: This try/catch won't work when invoking inverse inside a `defn` so perhaps we should remove it? :) What is its purpose?

polvalente: Better error messages. Triangular solve will return something like "can't solve singular system", which isn't the most semantic message, but this isn't necessary for the algorithm itself

josevalim:
Hi @polvalente, in JAX they implement inverse on top of triangular_solve. This is because both numpy/jax allow triangular_solve for a and b having the same rank. Here is the code: https://github.com/google/jax/blob/f8c36d9c0a0920c5d68f578cbf3e05d3a5aaa182/jax/_src/numpy/linalg.py#L331-L336

My proposal to add inverse therefore is to solve this using `triangular_solve`. `triangular_solve` is most likely optimized by almost all backends which will make sure this is efficient.

WDYT?

polvalente:
> Hi @polvalente, in JAX they implement inverse on top of triangular_solve. This is because both numpy/jax allow triangular_solve for a and b having the same rank. Here is the code: https://github.com/google/jax/blob/f8c36d9c0a0920c5d68f578cbf3e05d3a5aaa182/jax/_src/numpy/linalg.py#L331-L336
> 
> My proposal to add inverse therefore is to solve this using `triangular_solve`. `triangular_solve` is most likely optimized by almost all backends which will make sure this is efficient.
> 
> WDYT?

Sounds better! It seems they use `solve` but we can use all these tricks I've used in inverting for solving a general `A . B = C` system in binary backend (which in turn would use `triangular_solve`), delegating `solve` to the proper implementations in other backends.

IMO we should also use this opportunity to expose `solve` as part of the API, since it's a more general case then `triangular_solve`. 

Other than that, this also improves the AST in `Expr`, because using this implementation the code actually shows the inner workings of `invert` instead of showing `invert a` as a step.

How does that sound?

nx/lib/nx.ex
josevalim: Ok. The issue is that this error message won't work inside `defn`, so perhaps we should improve the error message in the BinaryBackend if possible.

josevalim:
Ah, you are correct, `inverse` is using solve and not triangular_solve. I mixed them up. In this case, I would not expose `solve` for now and go ahead with `inverse`. Also note their `solve` stops computing the gradient [for the input here](https://github.com/google/jax/blob/f8c36d9c0a0920c5d68f578cbf3e05d3a5aaa182/jax/_src/lax/linalg.py#L264). We may need to do the same to compute the gradient of inverse but I am not sure. Perhaps we need some tests on the gradient of inverse. :)

nx/lib/nx.ex
AlexJuca: @josevalim Just for curiosity, what are current limitations of `defn` and what do we expect to have solved in terms of limitations by the time the API becomes stable?

nx/lib/nx.ex
```diff
@@ -7903,22 +7919,90 @@ defmodule Nx do
...
-      raise ArgumentError, "incompatible dimensions for a and b on tringular solve"
+    case b_shape do
+      {^m, ^m} ->
+        nil
+
+      {^m} ->
+        nil
+
+      _ ->
+        raise ArgumentError, "incompatible dimensions for a and b on tringular solve"
```
AlexJuca, line 7937: I believe this should be "triangular" and not "tringular"

AlexJuca:
```suggestion
        raise ArgumentError, "incompatible dimensions for a and b on triangular solve"
```

AlexJuca (changes requested):

nx/lib/nx.ex
polvalente: yeap

nx/lib/nx/defn/expr.ex
```diff
@@ -643,6 +643,13 @@ defmodule Nx.Defn.Expr do
     composite({u, s, vt}, context, &expr(&1, context, :svd, [{u, s, vt}, tensor, opts]))
   end
 
+  @impl true
+  def invert(output, input) do
+    input = to_expr(input)
+    context = input.data.context
+    composite(output, context, &expr(&1, context, :invert, [output, input]))
```
josevalim, line 650: No need to call composite as it never returns a tuple:

```suggestion
    expr(output, context, :invert, [output, input])
```

nx/lib/nx/binary_backend.ex
```diff
@@ -1115,15 +1117,52 @@ defmodule Nx.BinaryBackend do
...
+    y = triangular_solve(l, l, b, [])
+
+    dot(output_holder, qt, [1], y, [0])
+  end
+
+  @impl true
+  def invert(output, %T{shape: {n, n}} = tensor) do
+    identity = eye(tensor, [])
+    solve(output, tensor, identity)
+  end
```
josevalim, line 1164: Given both solve and invert work exclusively on existing operations, I think we can move this implementations to Nx. WDYT? This way we don't add any new function to the backend. :)

nx/lib/nx/binary_backend.ex
```diff
@@ -1115,15 +1117,52 @@ defmodule Nx.BinaryBackend do
...
+      when b_shape == {rows, rows} or b_shape == {rows} do
     a_data = to_binary(a)
     b_data = to_binary(b)
-    out_bin = B.Decomposition.ts(a_data, a_type, b_data, b_type, rows, output_type, opts)
+    out_bin = B.Decomposition.ts(a_data, a_type, b_data, b_type, b_shape, output_type, opts)
     from_binary(out, out_bin)
   end
 
+  @doc false
+  def solve(%T{type: output_type} = output_holder, %T{shape: {n, n}} = a, %T{shape: b_shape} = b)
```
josevalim, line 1131: Let's also expose this on `Nx` as `solve`. :) And given it works on existing tensors, there is no need for a backend function as well!

josevalim (commented):
This looks great! I have dropped some minor comments about code organization and we can ship it.

josevalim:
I tried to resist but we should probably have our own `Nx.LinAlg` namespace. :) We can focus on that after this PR is merged.

AlexJuca:
@polvalente @josevalim I would love to handle the move to the `Nx.LinAlg` namespace, after these changes are merged.

nx/lib/nx/binary_backend.ex
polvalente: For this function specifically I think we can benefit of existing implementations on XLA and other backends. Matrix inversion has lots of optimized versions (with divide and conquer algorithms and such), so I'd rather delegate it to the low-level backends whenever possible

nx/lib/nx/binary_backend.ex
polvalente: This has the same issue as invert, albeit in a milder way. There are some algorithms which can be more efficient for solving non-triangularized matrices.

If we're exposing `solve`, I'd rather expose it delegating the implementation to each backend. Are there any major downsides in doing this?

josevalim:
@polvalente i know backends can optimize those implementations but we don’t quite support it yet. So let’s move them to Nx and we will add optional callbacks to the future. The biggest upside right now of implementing them on Nx is that we get grad for free.

polvalente:
> @polvalente i know backends can optimize those implementations but we don’t quite support it yet. So let’s move them to Nx and we will add optional callbacks in the future. The biggest upside right now of implementing them on Nx is that we get grad for free.

I see! On it, then :)

josevalim:

PR #349, opened 2021-03-21
Fix select grad

seanmor5:
This resolves some issues with select grad

### DIFF
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -234,11 +234,14 @@ defmodule Nx.Defn.Grad do
     {elem(tuple, index), cache}
   end
 
-  defp grad(:select, [pred, on_true, on_false], _ans, g, cache) do
-    {d_on_true, cache} = to_grad(on_true, g, cache)
-    {d_on_false, cache} = to_grad(on_false, g, cache)
-    result = Nx.select(pred, d_on_true, d_on_false)
-    {result, cache}
+  defp grad(:select, [pred, on_true, on_false], ans, g, cache) do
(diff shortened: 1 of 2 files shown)


josevalim (approved):
Can you come up with a test that fails in the old new but not in the new one? If it is not straight-forward, ship it anyway!

polvalente:
@seanmor5 @josevalim IIRC `sum(power(select(&greater(&1, 0), power(t, 2), power(t, 3)), 2))` failed with the previous implementation

nx/lib/nx/defn/grad.ex
```diff
@@ -234,11 +234,14 @@ defmodule Nx.Defn.Grad do
...
   end
 
-  defp grad(:select, [pred, on_true, on_false], _ans, g, cache) do
-    {d_on_true, cache} = to_grad(on_true, g, cache)
-    {d_on_false, cache} = to_grad(on_false, g, cache)
-    result = Nx.select(pred, d_on_true, d_on_false)
-    {result, cache}
+  defp grad(:select, [pred, on_true, on_false], ans, g, cache) do
+    gs = Nx.broadcast(g, ans)
+    zeros = Nx.broadcast(Expr.tensor(0.0), ans)
```
polvalente, line 239: Is there a practical difference between using `ans`, `on_true` and `on_false` here? I'd guess not since they are all the same shape

seanmor5: I think since `select` broadcasts `on_true` and `on_false`, it could be different if the shapes of `on_true` and `on_false` differ, but if they we're successfully broadcasted in the result, they'd successfully broadcast here as well.

PR #351, opened 2021-03-25
Convert tensor to scalar

versilov:


torchx/lib/torchx/backend.ex
```diff
@@ -148,16 +148,22 @@ defmodule Torchx.Backend do
   @impl true
   def random_uniform(%T{type: {s, _} = type, shape: shape} = out, min, max, backend_options)
       when s in [:u, :s] do
+    min = to_scalar(min) |> trunc()
+    max = to_scalar(max) |> trunc()
```
josevalim, line 152: I think for the binary backend, min and max have to be integers if the type is u/s. Are you sure we need this `trunc()` call here? 🤔

versilov: Yes, I've got confused with C++ type casts :) Fixed.

josevalim (approved):

PR #355, opened 2021-03-26
Add missing :compiler opts to Defn.aot calls

justincjohnson:
Without this change the examples are currently broken.


josevalim:

PR #367, opened 2021-04-05
feat: add Nx.argsort/2

### LINKED ISSUE #342: Returns the indices after sorting a tensor
Hi everybody,
I am trying to implement a function that receives a tensor and returns the indices of each value after the tensor has been sorted. Something exactly like the argsort  function from Numpy (https://numpy.org/doc/stable/reference/generated/numpy.argsort.html).
I even called @polvalente to help me but we ran out of ideas.

Anyone has any idea how this can be implemented using Nx?

polvalente:
This PR aims to add the sort_indices option so we can support "argsort" functionality

closes #342


polvalente:
This isn't ready because EXLA doesn't implement the option yet.
@seanmor5 any pointers on how we can achieve this?

I read the XLA docs on sort (https://www.tensorflow.org/xla/operation_semantics#sort) and it seems that if we provide 2 arguments, they will be sorted simultaneously. However, I had some trouble on how to build the corresponding index tensor inside the NIF.

I'm leaning towards using something akin to the code below for build the second argument: 
```elixir
iex(7)> tensor = Nx.tensor([[[[4,3,2],[1,0,-1]], [[1,2,3],[4,5,2]]]])
#Nx.Tensor<
  s64[1][2][2][3]
  [
    [
      [
        [4, 3, 2],
        [1, 0, -1]
      ],
      [
        [1, 2, 3],
        [4, 5, 2]
      ]
    ]
  ]
>
iex(8)> Nx.iota(tensor, axis: Nx.axes(tensor) |> Enum.at(-1))        
#Nx.Tensor<
  s64[1][2][2][3]
  [
    [
      [
        [0, 1, 2],
        [0, 1, 2]
      ],
      [
        [0, 1, 2],
        [0, 1, 2]
      ]
    ]
  ]
>
```

However, I'm not sure how it will behave, if it'll actually need a arity-4 function for sorting (as implied by the documentation linked above).

josevalim:
Awesome job @polvalente!

Some notes:

1. Generally speaking, we prefer to expose the EXLA operations as close to "native" as possible. I.e. EXLA.Op should expose the C API instead of providing conveniences on top of it. For this purpose, it is probably better to implement `variadic_sort`, pretty much like we implemented `variadic_reduce`.

2. At the NX level, I would rather introduce an argsort function instead of having options inside `sort`. :)

seanmor5:
@polvalente I looked into implementing argsort in EXLA a little bit ago. What I would do is change the sort NIF to always accept a list of Tensors. Then you can implement argsort in EXLA.Lib and sort the input tensor side by side with an iota operation.

exla/lib/exla/defn.ex
```diff
@@ -702,6 +702,25 @@ defmodule EXLA.Defn do
...
+    # variadic_sort underneath
+    path = [Access.key!(:data), Access.key!(:args)]
+    [[arg_1, arg_2], expr, fun] = get_in(comparator, path)
+
+    comparator_4 =
+      put_in(comparator, path, [
+        [arg_1, arg_2, arg_1, arg_2],
+        expr,
+        fn x, y, _, _ -> fun.(x, y) end
+      ])
```
polvalente, line 718: I'm not sure if there's a better way to achieve this. Basically, the comparison function needed for variadic_sort always has 2*num_args arity, so I needed to extend the arity-2 comparator to an arity-4 one which only uses the first 2 args (which are the current comparison elements for the first of the args passed to sort)

nx/lib/nx.ex
```diff
@@ -7299,8 +7299,20 @@ defmodule Nx do
 
   If no axis is given, defaults to `0`.
 
+  ### Options
+
+  * `:axis`: The name or number of the corresponding axis on which the sort should be applied.
+  * `:return_indices`: If `true`, returns the sorted indices instead of the values. Defaults to `false`
```
polvalente, line 7305:
```suggestion
```

nx/lib/nx.ex
```diff
@@ -7390,13 +7430,165 @@ defmodule Nx do
...
+  """
+  @doc type: :ndim
+  def argsort(tensor, opts \\ []) do
+    opts = keyword!(opts, axis: 0, comparator: :desc)
+    comparator = opts[:comparator]
+
+    %T{shape: shape, names: names} = tensor = to_tensor(tensor)
+    axis = Nx.Shape.normalize_axis(shape, opts[:axis], names)
+
+    impl!(tensor).argsort(%{tensor | type: {:s, 64}}, tensor,
```
polvalente, line 7587: I wasn't able to make argsort work with `{:u, 64}` type on EXLA, which would be ideal, so I defaulted to `{:s, 64}`

nx/lib/nx/binary_backend.ex
```diff
@@ -1572,17 +1573,32 @@ defmodule Nx.BinaryBackend do
...
+            a <= b
+          end
 
         :asc ->
-          &>/2
+          fn a, b ->
+            a = binary_to_number(a, type)
+            b = binary_to_number(b, type)
+            a >= b
+          end
```
polvalente, line 1601: These changes are two-fold. First because the original comparators didn't work with negative numbers, and secondly I've used`<=`  and `>=` so the sort is stable

nx/lib/nx/binary_backend.ex
```diff
@@ -1634,10 +1657,20 @@ defmodule Nx.BinaryBackend do
             end
           end
 
-        IO.iodata_to_binary(Enum.sort(data, comparator))
+        sorted =
```
polvalente, line 1660: I'm not sure if the `match_types` call on line 1654 is needed. Perhaps we should remove it or refactor the `for`-loop somehow to use it correctly

exla/lib/exla/defn.ex
josevalim: I think you can wrap that function inside the existing function. Something like this:

```elixir
    # Grow the comparator to arity 4 because argsort uses
    # variadic_sort underneath
    [[arg1, arg2], _expr, fun] = comparator.data.args

    comparator_4 = Expr.fun([arg1, arg2, arg1, arg2], fn x, y, _, _ -> fun.(x, y) end)
```

nx/lib/nx.ex
```diff
@@ -7390,13 +7429,165 @@ defmodule Nx do
...
+  Sorts the tensor along the given axis with the
+  given comparator and returns the corresponding indices
+  of the original tensor in the new sorted positions.
+
+  If no axis is given, defaults to `0`.
+
+  ### Options
+
+  * `:axis`: The name or number of the corresponding axis on which the sort should be applied.
+  * `:comparator`: Can be `:asc`, `:desc` or a arity-2 function for comparison between two tensor elements. Defaults to `:asc`
```
josevalim, line 7454:
```suggestion
  * `:axis` - The name or number of the corresponding axis on which the sort
    should be applied
  * `:comparator` - Can be `:asc`, `:desc` or a arity-2 function for comparison
    between two tensor elements. Defaults to `:asc`
```

nx/lib/nx.ex
```diff
@@ -7390,13 +7429,165 @@ defmodule Nx do
...
+        s64[x: 2][y: 3]
+        [
+          [1, 0, 1],
+          [0, 1, 0]
+        ]
+      >
+  """
+  @doc type: :ndim
+  def argsort(tensor, opts \\ []) do
+    opts = keyword!(opts, axis: 0, comparator: :desc)
```
josevalim, line 7581: The default here seems to be `:desc`. We probably want to change both sort and argsort to be asc.

nx/lib/nx.ex
```diff
@@ -7390,13 +7429,165 @@ defmodule Nx do
...
+      iex> Nx.argsort(Nx.tensor([[3, 1, 7], [2, 5, 4]], names: [:x, :y]), axis: :y, comparator: :asc)
+      #Nx.Tensor<
+        s64[x: 2][y: 3]
+        [
+          [2, 0, 1],
+          [1, 2, 0]
+        ]
+      >
+
+      iex> tensor = Nx.tensor([[[4, 5], [2, 5], [5, 0]], [[1, 9], [2, 1], [2, 1]], [[0, -1], [-1, 0], [0, -1]], [[-1, 0], [0, -1], [-1, 0]]], names: [:x, :y, :z])
```
josevalim, line 7491: This example is probably too big to follow as a doctest. Let's move it to `test/nx_test.exs` instead?

nx/lib/nx.ex
josevalim: Which error do you get? Looking at the code it feels like it should work... on the other, numpy and jax both return int64 too.

josevalim: Torchx also return int64, so let's keep this as is.

nx/lib/nx.ex
polvalente: Ok!

nx/lib/nx.ex
polvalente: The error was something along the lines of "The comparator function expected 0-th argument to have shape s64[], got u64[] instead"

nx/lib/nx.ex
polvalente: It seems that the comparator definitions were swapped between desc and asc. My latest commit changed this

nx/lib/nx.ex
```diff
@@ -7390,13 +7429,139 @@ defmodule Nx do
...
+  If no axis is given, defaults to `0`.
+
+  ### Options
+
+  * `:axis` - The name or number of the corresponding axis on which the sort
+    should be applied
+  * `:comparator` - Can be `:asc`, `:desc` or a arity-2 function for comparison
+    between two tensor elements. Defaults to `:asc`
+
+  ### Examples
```
josevalim, line 7458:
```suggestion
  ## Options

    * `:axis` - The name or number of the corresponding axis on which the sort
      should be applied
    * `:comparator` - Can be `:asc`, `:desc` or a arity-2 function for comparison
      between two tensor elements. Defaults to `:asc`

  ## Examples
```

nx/lib/nx/binary_backend.ex
josevalim: It isn't, you can remove it.

josevalim (approved):
Two minor comments and ship it! 🚢

PR #370, opened 2021-04-07
feat: add qr decomposition grad

polvalente:


nx/lib/nx/defn/grad.ex
```diff
@@ -1022,4 +1035,24 @@ defmodule Nx.Defn.Grad do
   defp up_to(_, _), do: []
 
   defp argsort(list), do: list |> Enum.with_index() |> Enum.sort() |> Enum.map(&elem(&1, 1))
+
+  # copies the lower triangle over the upper triangle of a 2x2 tensor
+  defp copyltu(tensor) do
```
polvalente, line 1040: Should we expose this in `Nx`? I've seen this operation referenced in more than one place for grads, but since this one is fully Nx-native and I've never heard of it in another context, maybe we can just point to this snippet when its needed

nx/test/nx/defn/grad_test.exs
```diff
@@ -1646,6 +1646,20 @@ defmodule Nx.Defn.GradTest do
...
+      grad(t, fn tensor ->
+        {q, r} = Nx.LinAlg.qr(tensor)
+
+        Nx.sum(Nx.concatenate([q, r]))
+      end)
+    end
+
+    test "computes grad for tensor" do
+      assert qr_grad(Nx.tensor([[1.0, 2.0], [1.0, -1.0]])) == Nx.tensor([[1, 1], [-1, 1]])
+    end
```
polvalente, line 1660: This test is giving me a weird exception which I suspect has to do with `{q, r} = ...`

```elixir
  1) test qr computes grad for tensor (Nx.Defn.GradTest)
     test/nx/defn/grad_test.exs:1658
     ** (ArgumentError) argument error
     code: assert qr_grad(Nx.tensor([[1.0, 2.0], [1.0, -1.0]])) == Nx.tensor([[1, 1], [-1, 1]])
     stacktrace:
       :erlang.element(1, #Nx.Tensor<
  f32[2][2]
  
  Nx.Defn.Expr
  parameter a                                f32[2][2]
  tensor g                                   u8[2][2]
  tensor o                                   s64[2][2]
  tensor q                                   s64[2][2]
  tensor r                                   s64[2][2]
  tensor bd                                  s64[2][2]
  b = metadata [ a, [Nx.Defn.Grad] ]         f32[2][2]
  c = slice [ 1.0, [0, 0], [2, 2], [1, 1] ]  f32[2][2]
  d = multiply [ b, c ]                      f32[2][2]
  e = metadata [ d, [Nx.Defn.Grad] ]         f32[2][2]
  f = add [ e, e ]                           f32[2][2]
  h = multiply [ b, c ]                      f32[2][2]
  i = metadata [ h, [Nx.Defn.Grad] ]         f32[2][2]
  j = transpose [ i, [1, 0] ]                f32[2][2]
  k = dot [ b, [1], j, [0] ]                 f32[2][2]
  l = transpose [ e, [1, 0] ]                f32[2][2]
  m = dot [ l, [1], b, [0] ]                 f32[2][2]
  n = subtract [ k, m ]                      f32[2][2]
  p = select [ g, n, o ]                     f32[2][2]
  s = select [ q, n, r ]                     f32[2][2]
  t = add [ p, s ]                           f32[2][2]
  u = transpose [ t, [1, 0] ]                f32[2][2]
  v = add [ u, p ]                           f32[2][2]
  w = dot [ f, [1], v, [0] ]                 f32[2][2]
  x = transpose [ b, [1, 0] ]                f32[2][2]
  y = transpose [ x, [1, 0] ]                f32[2][2]
  z = qr [ {y, y}, y, mode: :reduced ]       tuple2
  ba = elem [ z, 0, 2 ]                      f32[2][2]
  bb = elem [ z, 1, 2 ]                      f32[2][2]
  bc = transpose [ bb, [1, 0] ]              f32[2][2]
  be = triangular_solve [ bc, bd, [] ]       f32[2][2]
  bf = dot [ ba, [1], be, [0] ]              f32[2][2]
  bg = transpose [ bf, [1, 0] ]              f32[2][2]
  bh = dot [ w, [1], bg, [0] ]               f32[2][2]
  bi = metadata [ bh, [Nx.Defn.Grad] ]       f32[2][2]
>)
       (nx 0.1.0-dev) lib/nx/defn/grad.ex:234: Nx.Defn.Grad.grad/5
       (nx 0.1.0-dev) lib/nx/defn/grad.ex:161: anonymous fn/3 in Nx.Defn.Grad.to_grad/3
       (elixir 1.11.2) lib/enum.ex:1521: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (nx 0.1.0-dev) lib/nx/defn/grad.ex:195: Nx.Defn.Grad.grad_pairs/3
       (nx 0.1.0-dev) lib/nx/defn/grad.ex:161: anonymous fn/3 in Nx.Defn.Grad.to_grad/3
       (nx 0.1.0-dev) lib/nx/defn/grad.ex:24: Nx.Defn.Grad.transform/2
       (nx 0.1.0-dev) lib/nx/defn/kernel.ex:232: Nx.Defn.Kernel.grad/2
       test/nx/defn/grad_test.exs:1650: anonymous fn/3 in Nx.Defn.GradTest.qr_grad/1
       (nx 0.1.0-dev) lib/nx/defn/evaluator.ex:21: Nx.Defn.Evaluator.__jit__/4
       test/nx/defn/grad_test.exs:1659: (test)
```

Keep in mind that since `Nx.LinAlg.invert` is Nx-native, its underlying expressions are all present in the expr's introspection

polvalente: @josevalim do you have any idea on what might be going on here? This is the same error I mentioned the other day

nx/lib/nx/defn/grad.ex
josevalim: Do we have an equivalent in scipy/numpy we could use as reference? Would this belong to the LinAlg module?

nx/lib/nx/defn/grad.ex
```diff
@@ -1022,4 +1045,22 @@ defmodule Nx.Defn.Grad do
...
+    # the main diagonal is 1, and the rest is 0
+    selector = Nx.greater(Nx.iota(tensor, axis: 0), Nx.iota(tensor, axis: 1))
+
+    zeros = Nx.broadcast(0, tensor)
+
+    # Fetch the lower triangle of the matrix without the diagonal
+    lower = Nx.select(selector, tensor, zeros)
+
+    # Get the diagonal
+    diag = Nx.select(identity, tensor, zeros)
```
josevalim, line 1061: `identity` here is undefined. :)

nx/test/nx/defn/grad_test.exs
josevalim: I believe this is because `qr` is meant to return a tuple. So the grad of `qr` has to return a tuple but it is currently returning a single element. For example, you could make it so it returns this:

```
{res, cache} = to_grad(input, da, cache)
{{res, res}, cache}
```

but i don't know if this is the correct result. :)

nx/lib/nx/defn/grad.ex
polvalente: ooops, it went missing after a minor refactor :sweat_smile:

nx/test/nx/defn/grad_test.exs
polvalente: Thanks! I'll look into this and see what it yields

nx/lib/nx/defn/grad.ex
polvalente: The closest functions I could find were https://numpy.org/doc/stable/reference/generated/numpy.triu.html and https://numpy.org/doc/stable/reference/generated/numpy.tril.html#numpy.tril, which could be used as part of this implementation.

Other than that, "copyltu" only yields either papers or implementations of said papers for autograds

nx/test/nx/defn/grad_test.exs
polvalente: This removed the exception, although the result isn't correct, but this might be a problem in the calculation itself

polvalente: For reference, this is Jax's implementation:
```python
def qr_jvp_rule(primals, tangents, full_matrices):
  # See j-towns.github.io/papers/qr-derivative.pdf for a terse derivation.
  x, = primals
  dx, = tangents
  q, r = qr_p.bind(x, full_matrices=False)
  *_, m, n = x.shape
  if full_matrices or m < n:
    raise NotImplementedError(
      "Unimplemented case of QR decomposition derivative")
  dx_rinv = triangular_solve(r, dx)  # Right side solve by default
  qt_dx_rinv = jnp.matmul(_H(q), dx_rinv)
  qt_dx_rinv_lower = jnp.tril(qt_dx_rinv, -1)
  do = qt_dx_rinv_lower - _H(qt_dx_rinv_lower)  # This is skew-symmetric
  # The following correction is necessary for complex inputs
  do = do + jnp.eye(n, dtype=do.dtype) * (qt_dx_rinv - jnp.real(qt_dx_rinv))
  dq = jnp.matmul(q, do - qt_dx_rinv) + dx_rinv
  dr = jnp.matmul(qt_dx_rinv - do, r)
  return (q, r), (dq, dr)
```

I'm not sure if this is mathematically equivalent to the definition I was using, but I'll aim for this implementation instead

polvalente: @josevalim I've basically ported line by line this definition, with the difference that because Jax jvps receive the grad, I had to use `grad/5` instead. However, the result isn't what is expected.

Any ideas on what where the issue could be?

nx/lib/nx/defn/grad.ex
```diff
@@ -476,6 +476,41 @@ defmodule Nx.Defn.Grad do
...
+    dr = c |> Nx.add(Nx.transpose(c)) |> Nx.divide(2) |> Nx.multiply(e) |> Nx.dot(r)
+    dq = g |> Nx.subtract(Nx.dot(q, dr)) |> Nx.dot(r_inv)
+
+    # The code below yields the following thensor for the test at  test/nx/defn/grad_test.exs:1658
+    # #Nx.Tensor<\n  f32[2][2]\n  [\n    [1.9999998807907104, -0.33333325386047363],\n    [-0.3333333134651184, -0.9999999403953552]\n  ]\n>
+    # While the code above yields
+    # #Nx.Tensor<\n  f32[2][2]\n  [\n    [2.0, -0.3333331346511841],\n    [-0.33333325386047363, -0.9999999403953552]\n  ]\n>
+
+    # The code below was translated from jax, while the code above was translated from
+    # https://arxiv.org/pdf/2009.10071.pdf, page 5
```
polvalente, line 499: @josevalim I've just found this, which may tell us that we're on the right track but I'm missing something.

Basically, two different codes which are based on two different sources are yielding equivalent tensors (but not the expected output when compared to Jax).

The only thing I could think of now is that there is some fundamental difference between `grad/5` and `jvp/3` which would make sense of this.

Any thoughts?

nx/lib/nx/defn/grad.ex
```diff
@@ -498,6 +498,44 @@ defmodule Nx.Defn.Grad do
     [{input, dl}]
   end
 
+  defp grad(:qr, [{q, r}, input, _opts], ans, g) do
+    {q, r} = Nx.Defn.Expr.tuple({q, r}, ans)
```
polvalente, line 502: @josevalim this function doesn't exist anymore. What should I use here instead?

josevalim: @polvalente please make the function public again? The API is slightly different `.tuple(ans, [q, r])` but it should still be in Expr.

nx/lib/nx/defn/expr.ex
```diff
@@ -816,7 +816,8 @@ defmodule Nx.Defn.Expr do
     end)
   end
 
-  defp tuple(%T{type: {:tuple, size}, data: %{context: context}} = expr, list)
+  @doc false
```
josevalim, line 819: Please move it to the top of this module (around `def metadata`) and add a public doc to it. I think it is a good practice to treat `grad` as a public consumer: i.e. it should not use any private API.

nx/lib/nx/defn/grad.ex
```diff
@@ -248,11 +248,17 @@ defmodule Nx.Defn.Grad do
     grads
   end
 
-  @reduced_grads [:add, :multiply, :power]
+  @reduced_grads [:add, :multiply, :power, :qr]
```
josevalim:
```suggestion
  @reduced_grads [:add, :multiply, :power]
```

nx/lib/nx/defn/grad.ex
```diff
@@ -709,6 +738,13 @@ defmodule Nx.Defn.Grad do
     [{x, Nx.divide(Nx.multiply(g, 0.5), ans)}]
   end
 
+  # g(x) = (f(x))^2 -> 2f(x) * f'(x)
+  # g(x) = (x^3 + 1)^2 -> 2(x^3 + 1) * 3x
+
+  # ans = f(x)^1/2 -> 1/2 * f'(x) / (f(x)^(1/2))
+
+  # qr(f(x)) -> qr'(f(x)) * f'(x)
```
josevalim: no longer needed :D

nx/test/support/case.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Nx.Case do
```
josevalim, line 1: We already have a Nx.GradHelpers, perhaps it is worth unifying both into `Nx.Helpers`? Also, it doesn't have to be a case template. I think a module that you import as needed is enough :)

nx/test/nx/defn/grad_test.exs
```diff
@@ -56,13 +56,13 @@ defmodule Nx.Defn.GradTest do
   end
 
   describe "tokens" do
-    defn grad_token(t), do: value_and_grad(t, fn t -> hook(Nx.power(t, 2), :grad) end)
+    defn grad_token(t), do: grad(t, fn t -> hook(Nx.power(t, 2), :grad) end)
```
josevalim, line 59: Accidental change.

nx/test/nx/defn/grad_test.exs
```diff
@@ -1366,6 +1366,47 @@ defmodule Nx.Defn.GradTest do
     end
   end
 
+  describe "outer rule" do
```
josevalim, line 1369: Brought back by accident too. :) The outer rule is no longer tested because it is no longer a primitive.

nx/lib/nx/defn/expr.ex
polvalente: Oh, ok. Any suggestions for the doc?

polvalente: I've moved this. There's also defp tuple_out. Should I leave it where it is?

josevalim: You can bring it close to tuple but also leave it as is, your call.

nx/lib/nx/defn/expr.ex
```diff
@@ -96,6 +96,26 @@ defmodule Nx.Defn.Expr do
     expr(expr, expr.data.context, :metadata, [expr, metadata])
   end
 
+  @doc """
+  Creates a tuple with elements in `list` that points to tuple
+  expression `expr`.
+
```
josevalim, line 102:
```suggestion

```

nx/lib/nx/defn/expr.ex
```diff
@@ -96,6 +96,25 @@ defmodule Nx.Defn.Expr do
     expr(expr, expr.data.context, :metadata, [expr, metadata])
   end
 
+  @doc """
+  Creates a tuple with elements in `list` that points to tuple
+  expression `expr`.
```
josevalim, line 101:
```suggestion
  expression `expr`.

```

nx/test/support/grad_helpers.ex
```diff
@@ -14,9 +14,26 @@ defmodule Nx.GradHelpers do
     approx_equal?(comp_grad, est_grad, x, atol, rtol)
   end
 
+  def approx_equal?(lhs, rhs, opts \\ []) do
+    atol = opts[:atol] || 1.0e-4
+    rtol = opts[:rtol] || 1.0e-4
+
+    unless Nx.all_close(lhs, rhs, atol: atol, rtol: rtol) == Nx.tensor(1, type: {:u, 8}) do
```
josevalim, line 21: This should delegate to the private function?

josevalim (approved):

nx/test/nx/defn/grad_test.exs
```diff
@@ -1,5 +1,5 @@
 defmodule Nx.Defn.GradTest do
-  use ExUnit.Case, async: true
+  use Nx.Case, async: true
```
josevalim, line 2:
```suggestion
  use ExUnit.Case, async: true
```

josevalim:
HUGE!

PR #371, opened 2021-04-10
Device tuple

versilov:
Still WIP, just to check if I'm moving in the right direction?


torchx/lib/torchx.ex
```diff
@@ -37,23 +114,74 @@ defmodule Torchx do
...
+  deftorch broadcast_to(tensor, shape)
+  deftorch transpose(tensor, dim0, dim1)
+  deftorch permute(tensor, dims)
+  deftorch split(tensor, split_size)
+  deftorch narrow(tensor, dim, start, length)
+  deftorch as_strided(tensor, size, strides, offset)
+
+  deftorch sum(tensor, axes, keep_axes)
+  deftorch argmax(tensor, axe, keep_axes)
+  deftorch argmin(tensor, axe, keep_axes)
```
josevalim, line 147: I getr confused by this all the time but the singular of axes is axis :)

torchx/lib/torchx.ex
```diff
@@ -1,7 +1,84 @@
...
+      Module.register_attribute(Torchx, :torch_function, accumulate: true)
+    end
+  end
+
+  defmacro deftorch(call) do
+    {name, args} = Macro.decompose_call(call)
+
+    args_names = Enum.map(args, &elem(&1, 0))
+
+    if Enum.any?(args, &match?({:device, _, nil}, &1)) do
```
josevalim, line 14: Instead of using checking for the device variable, I would say it is cleaner to have two macros: maybe `deftorch` and `defdevice` or similar?

torchx/lib/torchx.ex
```diff
@@ -1,7 +1,84 @@
...
+    end
+  end
+
+  defmacro deftorch(op, _t) do
+    quote do
+      @torch_function {var!(op), [:tensor]}
+      def unquote(op)(tensor) do
+        {device, prepared_args} = prepare_args([tensor])
+
+        Torchx.NIF.call(unquote(op), device, prepared_args)
```
josevalim, line 53: Instead of calling `Torchx.NIF.call` and have that call `apply`, I would select the proper IO/CPU operation here, which has the benefit that we can build both ops at compilation time. Something like this:

```elixir
    op_io = :"#{op}_io"

    quote do
      @torch_function {var!(op), [:tensor]}
      def unquote(op)(tensor) do
        case prepare_args([tensor]) do
          {:cpu, prepared_args} -> Torchx.NIF.unquote(op)(unquote_splicing(prepared_args))
          {:io, prepared_args} -> Torchx.NIF.unquote(op_io)(unquote_splicing(prepared_args))
        end
      end
    end
```

It will be slightly faster too. :)

torchx/lib/torchx.ex
```diff
@@ -37,23 +114,74 @@ defmodule Torchx do
...
 
   def tensordot(left, right, left_axes, right_axes) do
     {device, [left_ref, right_ref]} = to_refs([left, right])
 
     NIF.call(:tensordot, device, [left_ref, right_ref, left_axes, right_axes])
-    |> wrap_with_device(device)
+    |> wrap(device)
   end
 
+  def torch_functions(), do: @torch_function
```
josevalim, line 183: Shall we keep this with underscore so others do not call it?

```suggestion
  def __torch__, do: @torch_function
```

josevalim:
I think this is a solid direction! :100: I have added some comments!

josevalim:

PR #375, opened 2021-04-15
Add leftover options to to_batched_list

### LINKED ISSUE #374: Add wrapping options in `Nx.to_batched_list`
In real datasets, it's common that your preferred batch size is not a multiple of the size of your dataset. In these cases it's useful to either:

1) Repeat the first n values until a complete batch is made
2) Truncate the last n values

I believe we should add an option similar to [Enum.chunk_every](https://hexdocs.pm/elixir/Enum.html#chunk_every/4) which determines what to do with the leftover values.

seanmor5:
Resolves #374

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -1221,7 +1221,7 @@ defmodule Nx do
         ]
       >
 
-      If the batch size would result in uneven batches, you can repeat or discard excess data:
+  If the batch size would result in uneven batches, you can repeat or discard excess data:
 
       iex> [first, second, third] = Nx.to_batched_list(Nx.iota({5, 2}, names: [:x, :y]), 2)
       iex> first
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -1165,7 +1165,11 @@ defmodule Nx do
 
   The first dimension (axis 0) is divided by `batch_size`.
   In case the dimension cannot be evenly divided by
-  `batch_size`, it raises an `ArgumentError`.
+  `batch_size`, you may specify what to do with leftover
+  data using `:leftover`. `:leftover` must be one of `:repeat`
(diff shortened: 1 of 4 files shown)


nx/lib/nx.ex
```diff
@@ -1217,29 +1221,64 @@ defmodule Nx do
         ]
       >
 
-  In case the dimension cannot be evenly divided by `batch_size`,
-  it raises an `ArgumentError`:
+      If the batch size would result in uneven batches, you can repeat or discard excess data:
```
josevalim, line 1224:
```suggestion
  If the batch size would result in uneven batches, you can repeat or discard excess data:
```

nx/lib/nx/binary_backend.ex
```diff
@@ -156,11 +156,36 @@ defmodule Nx.BinaryBackend do
...
+          <<wrapped::bitstring-size(size_needed), _::bitstring>> = even_batches
+          last_batch = from_binary(out, IO.iodata_to_binary([leftover, wrapped]))
+          [last_batch | tensors]
+
+        :discard ->
+          tensors
+      end
+    else
+      tensors
     end
```
josevalim, line 189:
```suggestion
    case opts[:leftover] do
      _ when excess == 0 ->
        tensors

      :repeat ->
        size_needed = (elem(output_shape, 0) - excess) * bitsize
        <<wrapped::bitstring-size(size_needed), _::bitstring>> = even_batches
        last_batch = from_binary(out, IO.iodata_to_binary([leftover, wrapped]))
        [last_batch | tensors]

      :discard ->
        tensors
    end
```

josevalim:
No, ignore me, to_batched_list isn't valid inside the compiler.

josevalim (approved):
Just some minor comments and we can ship it!

PR #378, opened 2021-04-21
feat: support batching in Nx.dot through Nx.dot/6

polvalente:
This PR solves #182

Code was mostly imported from #238, with minor changes to make it pass tests without importing the binary backend traversal features. I've added @elbow-jason as a coauthor in one of the commits because of this.


nx/lib/nx.ex
```diff
@@ -6408,17 +6408,219 @@ defmodule Nx do
...
+            batch_vs_contract_conflict_message("left", batch_axes1, b1, contract_axes1, c1)
+    end
+
+    # ensure there is no conflict between right batch axis and right contract axes
+    if right_batched? and 0 in c2 do
+      raise ArgumentError,
+            batch_vs_contract_conflict_message("right", batch_axes2, b2, contract_axes2, c2)
+    end
+
+    {output_shape, output_names} = dot_batch_output_shape(s1, c1, names1, b1, s2, c2, names2, b2)
```
seanmor5, line 6536: I think most of this should probably be combined into a dot shape rule. I've found it to be helpful for Axon to have some of the Nx shape rules available, and it might be necessary here as well

nx/lib/nx.ex
```diff
@@ -6408,17 +6408,219 @@ defmodule Nx do
...
+         batch_norm,
+         contract_axes,
+         contract_norm
+       ) do
+    "dot batch axes #{render_normalized_axes(batch_axes, batch_norm)} for the #{side} tensor cannot " <>
+      "be in the contract axes #{render_normalized_axes(contract_axes, contract_norm)}"
+  end
+
+  defp render_normalized_axes(same, same), do: inspect(same)
+  defp render_normalized_axes(axes, norm), do: "#{inspect(axes)} (norm: #{inspect(norm)})"
```
seanmor5, line 6622: Same as comment above, I'd move this to Nx.Shape :)

nx/lib/nx/binary_backend.ex
```diff
@@ -476,8 +476,101 @@ defmodule Nx.BinaryBackend do
...
+    {right_batch_shape, right_batch_names} =
+      Nx.Shape.contract(right_shape, right_batch_axes, right_names, false)
+
+    left_batch_size = Nx.size(left_batch_shape)
+    right_batch_size = Nx.size(right_batch_shape)
+
+    left_batch_count = Nx.Shape.batch_count(left_shape, left_batch_axes)
+    right_batch_count = Nx.Shape.batch_count(right_shape, right_batch_axes)
+
+    {batch_shape_out, _output_names} =
```
seanmor5, line 514: I think you should have the output shape in `out`

nx/lib/nx/binary_backend.ex
```diff
@@ -476,8 +476,101 @@ defmodule Nx.BinaryBackend do
...
+
+    left_count = Nx.size(left)
+    right_count = Nx.size(right)
+
+    count = max(left_count, right_count)
+
+    max_batch_count = max(left_batch_count, right_batch_count)
+    batch_count = div(count, max_batch_count)
+
+    range = if max_batch_count == 0, do: [], else: Enum.uniq(0..(max_batch_count - 1))
```
seanmor5, line 532: What's the need for Enum.uniq here?

nx/lib/nx/binary_backend.ex
```diff
@@ -476,8 +476,101 @@ defmodule Nx.BinaryBackend do
...
+        right_batch_contract_axes,
+        right_batch_names
+      )
+
+    left_count = Nx.size(left)
+    right_count = Nx.size(right)
+
+    count = max(left_count, right_count)
+
+    max_batch_count = max(left_batch_count, right_batch_count)
```
seanmor5, line 529: Can we enforce the batch size is the same to skip this?

seanmor5 (commented):
This is beautiful @polvalente! I dropped a few minor comments, mostly about moving things to Nx.Shape

nx/lib/nx/binary_backend.ex
polvalente: For some reason I have it stuck in my head that `0..0 |> Enum.to_list() === [0,0]`. I'll change this :)

nx/lib/nx/binary_backend.ex
polvalente: Sure. This will simplify things!

nx/lib/nx/binary_backend.ex
polvalente: Perhaps the name is wrong. This is the shape for each batch item, not the shape for the batch itself

polvalente: Any suggestions for the name?

polvalente: I changed it to `batch_item_output_shape`

nx/lib/nx/binary_backend.ex
polvalente: Actually, I think that without this we won't be able to have "dot(batched_tensor, normal_tensor)" or "dot(normal_tensor, batched_tensor)"

polvalente: Yes, this is the case

nx/lib/nx.ex
```diff
@@ -6408,15 +6408,160 @@ defmodule Nx do
 
   """
   @doc type: :ndim
-  def dot(t1, axes1, t2, axes2) do
+  def dot(t1, axes1, t2, axes2), do: dot(t1, axes1, [], t2, axes2, [])
+
+  @doc """
+  Computes the dot-product of two batched tensors, throughout the batch dimensions.
```
seanmor5, line 6414: I would do something like: "Computes a general dot product with batch axes and contracting axes, where both tensors are vectorized along the given batch axes and contracted along the given contracting axes"

nx/lib/nx.ex
```diff
@@ -6408,15 +6408,160 @@ defmodule Nx do
...
+
+  ## Examples
+
+    ### Dot product between two batched tensors
```
seanmor5, line 6418:
```suggestion
  ### Dot product between two batched tensors
```

nx/lib/nx.ex
```diff
@@ -6408,15 +6408,160 @@ defmodule Nx do
...
+          [
+            [6]
+          ],
+          [
+            [16]
+          ]
+        ]
+      >
+
+    ### Dot product between a batched tensor and a non-batched tensor
```
seanmor5, line 6450:
```suggestion
  ### Dot product between a batched tensor and a non-batched tensor
```

nx/lib/nx.ex
```diff
@@ -6408,15 +6408,160 @@ defmodule Nx do
...
+            [15.0, 18.0, 21.0, 24.0]
+          ],
+          [
+            [30.0, 36.0, 42.0, 48.0],
+            [45.0, 54.0, 63.0, 72.0]
+          ]
+        ]
+      >
+
+    ### Dot product between a non-batched tensor and a batched tensor
```
seanmor5, line 6469:
```suggestion
  ### Dot product between a non-batched tensor and a batched tensor
```

nx/lib/nx.ex
```diff
@@ -6408,15 +6408,160 @@ defmodule Nx do
 
   """
   @doc type: :ndim
-  def dot(t1, axes1, t2, axes2) do
+  def dot(t1, axes1, t2, axes2), do: dot(t1, axes1, [], t2, axes2, [])
```
seanmor5, line 6411: We will probably eventually work this out, but it's okay for now so it doesn't break anything and we can handle it in a later PR. Just documenting for now :)

nx/lib/nx/binary_backend.ex
seanmor5: I believe JAX (and XLA) requires the number of batch dimensions to be the same, so both tensors will have to be batched: 

```python
a = lax.iota(np.float32, 32)
a = lax.reshape(a, (2, 4, 4))
b = lax.iota(np.float32, 16)
b = lax.reshape(b, (4, 4))
lax.dot_general(a, b, (((2,), (0,)), ((0,), ())))
```

```
TypeError: dot_general requires equal numbers of lhs_batch and rhs_batch dimensions, got lhs_batch (0,) and rhs_batch ().
```

nx/lib/nx.ex
polvalente: Some day I'll remember which is the correct indentation :laughing:

nx/lib/nx.ex
```diff
@@ -6408,15 +6408,158 @@ defmodule Nx do
...
+
+  @doc """
+  Computes the generalized dot product bewteen two tensors, given the contracting
+  and batch axes.
+
+
```
josevalim, line 6417:
```suggestion
```

nx/lib/nx.ex
```diff
@@ -6408,15 +6408,158 @@ defmodule Nx do
...
+  defp shift_left_for_batch(shape, contract_axes, names) do
+    {shift_left_shape(shape), shift_left_axes(contract_axes), tl(names)}
+  end
+
+  defp shift_left_shape(shape) do
+    shape
+    |> Tuple.to_list()
+    |> tl()
+    |> List.to_tuple()
```
josevalim, line 6554: There is a Tuple.delete_at you can use instead. :)

nx/lib/nx/defn/grad.ex
```diff
@@ -380,7 +380,7 @@ defmodule Nx.Defn.Grad do
     end)
   end
 
-  defp grad(:dot, [x, axes_x, y, axes_y], ans, g, cache) do
+  defp grad(:dot, [x, axes_x, _x_batch_axes, y, axes_y, _y_batch_axes], ans, g, cache) do
```
josevalim, line 383: As a reference: https://github.com/google/jax/blob/75c898e4f00d3811094c41bfc5e72d9f9576d0c2/jax/_src/lax/lax.py#L3199

nx/lib/nx/shape.ex
```diff
@@ -322,6 +322,32 @@ defmodule Nx.Shape do
     end
   end
 
+  @doc """
+  Calculates the number of batches of a shape for the given axes.
+  It expects the batch_axes to have been normalized.
+  ## Examples
```
josevalim, line 328:
```suggestion

  ## Examples

```

nx/lib/nx/shape.ex
```diff
@@ -322,6 +322,32 @@ defmodule Nx.Shape do
...
+    12
+  """
+  def batch_count(shape, batch_axes) do
+    Enum.reduce(batch_axes, 1, fn axis, total -> elem(shape, axis) * total end)
+  end
+
+  @doc """
+  Shifts all the axes by n.
+  Used to re-normalize axes for batching.
+  ## Examples
```
josevalim, line 343:
```suggestion

  ## Examples

```

nx/lib/nx/defn/grad.ex
polvalente: I was planning on dealing with this in a follow-up PR. Should I add a `to-do` comment or maybe create an issue?

polvalente: btw, thanks for the reference!

nx/lib/nx.ex
```diff
@@ -6408,15 +6408,158 @@ defmodule Nx do
...
+
     output_type = binary_type(t1, t2)
-    %T{shape: s1, names: names1} = t1 = to_tensor(t1)
-    %T{shape: s2, names: names2} = t2 = to_tensor(t2)
-    axes1 = Nx.Shape.normalize_axes(s1, axes1, names1)
-    axes2 = Nx.Shape.normalize_axes(s2, axes2, names2)
-    {output_shape, output_names} = Nx.Shape.zip_reduce(s1, axes1, names1, s2, axes2, names2)
+
+    {s1, names1, c1, b1, s2, names2, c2, b2} =
+      Nx.Shape.dot(t1, contract_axes1, batch_axes1, t2, contract_axes2, batch_axes2)
```
josevalim, line 6495: Btw, I would expect a function named `Nx.Shape.dot` to return the output shape of `dot`. It can also perform validation, but generally speaking it computes the shape. You can also make it compute the `{output_shape, batch_each_shape}` so you can reuse it elsewhere.

nx/lib/nx/binary_backend.ex
```diff
@@ -476,8 +476,89 @@ defmodule Nx.BinaryBackend do
...
+          _::bitstring>> = right_binary
+
+        batch_item_output_template
+        |> dot4(
+          from_binary(left_batch_item_template, left_batch_item_binary),
+          left_batch_contract_axes,
+          from_binary(right_batch_item_template, right_batch_item_binary),
+          right_batch_contract_axes
+        )
+        |> to_binary()
```
josevalim, line 554: I would change bin_zip_reduce and dot4 (maybe call it `bin_dot4`) so it returns a binary instead of a tensor. This means you will no longer need to compute the `batch_item_output_template`.

Perhaps you should also change bin_reduce to return a binary too, for consistency.

nx/lib/nx/defn/grad.ex
josevalim: Let's add a TODO, but, just to be safe, we should match on empty lists.

Also, we need to fix EXLA too, I believe it will be broken after this change. Do you have EXLA running on your machine?

nx/lib/nx.ex
polvalente: This makes sense. Perhaps some of the validation will need to be duplicated if the intermediate values aren't to be returned. I'll look into maybe changing the function signature to accept the normalized axes instead. WDYT?

josevalim: Yes, typically axes normalization is done outside of Nx.Shape.

nx/lib/nx/defn/grad.ex
polvalente: > Let's add a TODO, but, just to be safe, we should match on empty lists.

Ok, will do.

> Also, we need to fix EXLA too, I believe it will be broken after this change. Do you have EXLA running on your machine?

Yes, I do. I'll also handle the "not breaking" part of EXLA, leaving batching for a follow-up PR (I'm thinking one for grad and another for EXLA). Sounds good?

josevalim: My understand is that this PR will break EXLA though - the current tests will no longer pass, and I would prefer to not do that. So I think EXLA needs to be handled now. :S

polvalente: I meant that I'll work on EXLA as far as making non-batching dot products work normally at least. 

If you feel batching should also be handled in this PR, I'll be happy to work on that too.

josevalim: I may be missing something, but I am assuming that EXLA will fail because we are now calling dot/6 and we haven't changed it to handle dot/6 accordingly, even if to ignore the batch dimensions. But if the tests pass as is, then it is fine by me to be handled later. :)

polvalente: Some changes will be needed. I'm thinking more on the terms of hard-matching on `[]` for the batch axes for now, handling them properly later

josevalim: Sounds good to me!

exla/lib/exla/defn.ex
```diff
@@ -300,9 +300,9 @@ defmodule EXLA.Defn do
     EXLA.Op.get_tuple_element(op, index)
   end
 
-  defp to_operator(:dot, [left, axes1, right, axes2], %{type: type}, state) do
+  defp to_operator(:dot, [left, contract_axes1, _batch_axes1, right, contract_axes2, _batch_axes2], %{type: type}, state) do
     precision = state.precision
-    EXLA.Op.dot_general(to_type(left, type), to_type(right, type), {axes1, axes2}, precision)
+    EXLA.Op.dot_general(to_type(left, type), to_type(right, type), {contract_axes1, contract_axes2}, precision)
```
polvalente, line 305: It turns out that EXLA already impĺemented dot through dot_general. 

In exla/c_src/exla/exla_nif_util.cc there is the `get_dot_dimension_numbers` function in which we need to deal with receiving a tuple of tuples instead of a tuple of axes lists. I think that changing this, everything should work.

However, I don't yet know what should be changed in there for this to work

exla/lib/exla/defn.ex
```diff
@@ -300,11 +300,16 @@ defmodule EXLA.Defn do
...
   end
 
+  defp to_operator(:dot, _, _, _) do
+    raise ArgumentError, "dot operation not implemented for batched tensors"
+  end
+
+
```
josevalim, line 312: Just let it crash on function clause instead. :)

```suggestion
```

nx/lib/nx.ex
```diff
@@ -6408,15 +6408,151 @@ defmodule Nx do
...
+        batch_dim = elem(s1, 0)
+        batch_name = hd(names1)
+        {s1, c1, names1} = shift_left_for_batch(s1, c1, names1)
+        {batch_dim, batch_name, s1, c1, names1, s2, c2, names2}
+
+      {[], [0]} ->
+        batch_dim = elem(s2, 0)
+        batch_name = hd(names2)
+        {s2, c2, names2} = shift_left_for_batch(s2, c2, names2)
+        {batch_dim, batch_name, s1, c1, names1, s2, c2, names2}
```
josevalim, line 6546: Should we raise a nice error message here? It seems only the highest dimension can be batched after all?

nx/lib/nx/binary_backend.ex
```diff
@@ -476,8 +476,77 @@ defmodule Nx.BinaryBackend do
...
+          from_binary(right_batch_item_template, right_batch_item_binary),
+          right_batch_contract_axes,
+          out.type
+        )
+      end
+
+    from_binary(out, bin_result)
+  end
+
+  defp dot4(%{type: t1} = left, contract_axes1, %{type: t2} = right, contract_axes2, type) do
```
josevalim, line 548: nitpick: call it `bin_dot`, so it is consistent in that it returns a binary. :)

josevalim (approved):
Just a couple other minor comments and ship it :)

nx/lib/nx.ex
polvalente: You're right. Actually `b1` and `b2` can only be of the same length as well, so these clauses don't make sense anymore

polvalente: Nx.Shape.dot already raises in unsupported cases, so we can just not match on anything else here

PR #379, opened 2021-04-22
feat: concatenate scalar tensors

polvalente:
This was brought up in the slack channel. Basically, it would be nice to `Nx.concatenate([Nx.tensor(1), Nx.tensor(2)])` into `Nx.tensor([1, 2])`

This first attempt started by changing some Nx.Shape functions which ended up being sufficient for Nx.BinaryBackend to work. However, I ended up stumbling across the same problem in EXLA (in which scalar tensors have no axes).

The main question here is: should I treat this at the Backend level or should I just reshape scalar tensors into `{1}`-shaped tensors for concatenation in the `nx.ex` `def concatenate` itself?

### DIFF
--- a/exla/test/exla/defn_expr_test.exs
+++ b/exla/test/exla/defn_expr_test.exs
@@ -2412,6 +2412,10 @@ defmodule EXLA.DefnExprTest do
     defn concatenate2(t1, t2, t3), do: Nx.concatenate([t1, t2, t3], axis: 2)
     defn concatenate1_inp(t1), do: Nx.concatenate([t1], axis: 2)
 
+    test "works on list of scalar tensors" do
+      assert concatenate0(Nx.tensor(1), Nx.tensor(2), Nx.tensor(3)) == Nx.tensor([1, 2, 3])
+    end
+
     test "works 0th axis" do
       t1 = Nx.iota({2, 2, 2})
       t2 = Nx.iota({1, 2, 2})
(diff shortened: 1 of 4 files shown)


josevalim:
To make sure we are not reinventing the wheel here, does Numpy concatenate does the same? If no, what would be the API to achieve the same in Numpy and how is it implemented in JAX?

seanmor5:
So it looks like both JAX and NumPy do not support concatenating scalars:

```python
lax.concatenate([np.array(0), np.array(1), np.array(2)], 0)
concatenate dimension out of bounds: dimension 0 for shapes (), (), ().

jnp.concatenate([np.array(0), np.array(1), np.array(2)])
Zero-dimensional arrays cannot be concatenated.

np.concatenate(np.array(0), np.array(1), np.array(2))
iteration over a 0-d array
```

It does seem a bit counter-intuitive, but I think the idea is that concatenate shouldn't do any rank promotion, so trying to concat 0 rank tensors is invalid. I think instead what we would want to do is add `Nx.stack` which just adds a leading dimension and concatenates tensors in that dimension:

```elixir
iex> Nx.stack([0, 1, 2])
#Nx.Tensor<
  s64[3]
  [0, 1, 2]
>
```

This is the same behavior as in NumPy and PyTorch

polvalente:
@seanmor5 @josevalim thanks for the comments. I think I'll close this PR in favor of another one adding `Nx.stack` then. WDYT?

PR #380, opened 2021-04-23
feat: add Nx.stack/2

polvalente:
This PR adds Nx.stack as suggested in #379

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7333,7 +7333,7 @@ defmodule Nx do
   end
 
   @doc """
-  Stacks an `Enumerable` of `n` scalar tensors into an `{n}`-shaped tensor
+  Stacks a list of `n` scalar tensors into an `{n}`-shaped tensor
 
   ### Options
 
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7353,7 +7353,7 @@ defmodule Nx do
         [1, 2]
       >
   """
-  def stack(enumerable, opts \\ []) do
+  def stack(tensors, opts \\ []) when is_list(tensors) do
     opts = keyword!(opts, name: nil)
     name = opts[:name]
 
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7333,6 +7333,36 @@ defmodule Nx do
   end
 
   @doc """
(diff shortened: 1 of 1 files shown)


nx/lib/nx.ex
```diff
@@ -7332,6 +7332,36 @@ defmodule Nx do
...
+        [1, 2, 3]
+      >
+
+      iex> Nx.stack([Nx.tensor(1), Nx.tensor(2)], name: :x)
+      #Nx.Tensor<
+        s64[x: 2]
+        [1, 2]
+      >
+  """
+  def stack(enumerable, opts \\ []) do
```
polvalente, line 7356: Should I validate that is is an enumerable somehow?

Also, should this be delegated to individual backends? It seems to me that this belongs here, but I might have missed something

seanmor5: I would validate it's a list here, I think that makes sense

seanmor5 (approved):
Looks great!

nx/lib/nx.ex
polvalente: Technically this also works for MapSets or any other Enumerable which yields tensors when `Enum.map`ed

polvalente:
@seanmor5 one last experiment I did:

```elixir
iex(7)> Enum.map(1..3, & &1 |> Nx.tensor() |> Nx.reshape({1, 1, 1})) |> Nx.stack()
#Nx.Tensor<
  s64[3][1][1][1]
  [
    [
      [
        [1]
      ]
    ],
    [
      [
        [2]
      ]
    ],
    [
      [
        [3]
      ]
    ]
  ]
>
iex(8)> Enum.map(1..3, & &1 |> Nx.tensor() |> Nx.reshape({1, 1, 1})) |> Nx.concatenate
#Nx.Tensor<
  s64[3][1][1]
  [
    [
      [1]
    ],
    [
      [2]
    ],
    [
      [3]
    ]
  ]
>
```

Is this a desired behavior, or should we only accept `{}`-shaped tensors for `Nx.stack`?

nx/lib/nx.ex
josevalim: Indeed but I think those use cases won't be common, so I would go with something like:

```suggestion
  def stack(tensors, opts \\ []) when is_list(tensors) do
```

josevalim (approved):
Just one small change and ship it!

PR #387, opened 2021-05-04
Fix p-norm throwing ArithmeticError for zero vectors

kshalot:
I found a small thing when fooling around with Nx:

## Issue

Trying to compute the p-norm of a zero vector raises an ArithmeticError.
```elixir
t = Nx.tensor([0,0,0])
Nx.LinAlg.norm(t, ord: 2)

** (ArithmeticError) bad argument in arithmetic expression
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:681: Nx.BinaryBackend.element_divide/3
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:639: Nx.BinaryBackend."-element_wise_bin_op/4-lbc$^17/2-9-"/7
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:638: Nx.BinaryBackend.element_wise_bin_op/4
    (nx 0.1.0-dev) lib/nx/lin_alg.ex:264: Nx.LinAlg.norm_integer/3
```
EDIT: Actually it fails for tensors like `Nx.tensor([0, -2, -3, -1])` as well (0 is the maximum) because abs is not used when computing the stability coefficient.
## Cause

There is this stabilization coefficient that is equal to the maximum absolute value in the tensor (it was non-absolute in the code, I fixed this as well). So for zero vectors the coefficient is zero, which causes a divide-by-zero error in the next step.

Since this is kind of an edge case, I added a regular test for this. It fails on main, works on this branch.

I have some doubts about my code though. I'm not sure whether branching like this and testing for equality with `Nx.tensor(0)` is the way to go. Also typecasting to f32 in the test seems hackish - maybe every norm should return a float? I think that's how numpy handles this, even for 0-norm. Their p-norm is also numerically unstable, so:
```python
a = np.array([1e200])
np.linalg.norm(a)
# inf
```
 Any suggestions/thoughts on this? :smile_cat:

### DIFF
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -248,6 +248,7 @@ defmodule Nx.LinAlg do
 
   defp norm_integer(%{type: type} = t, ord, opts) when is_integer(ord) do
     output_type = Nx.Type.to_floating(type)
+    inv_ord = Nx.tensor(1 / ord, type: output_type)
 
     # We extract this result to a variable because it's used both for
     # getting the normalization coefficient and for the main pipe chain
@@ -259,18 +260,12 @@ defmodule Nx.LinAlg do
     # avoids numerical overflow.
     numerical_stability_coefficient = Nx.reduce_max(abs_t)
 
-    if numerical_stability_coefficient == Nx.tensor(0, type: type) do
-      Nx.tensor(0, type: output_type)
-    else
-      inv_ord = Nx.tensor(1 / ord, type: output_type)
-
-      abs_t
-      |> Nx.divide(numerical_stability_coefficient)
-      |> Nx.power(ord)
-      |> Nx.sum(opts)
-      |> Nx.power(inv_ord)
-      |> Nx.multiply(numerical_stability_coefficient)
-    end
+    abs_t
+    |> Nx.divide(numerical_stability_coefficient)
+    |> Nx.power(ord)
+    |> Nx.sum(opts)
+    |> Nx.power(inv_ord)
+    |> Nx.multiply(numerical_stability_coefficient)
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


josevalim:
Nx.divide always returns a float, so we should always be getting a float from this particular operation. Which norm is not returning a float?

kshalot:
> Nx.divide always returns a float, so we should always be getting a float from this particular operation. Which norm is not returning a float?

The 0-norm:
https://github.com/elixir-nx/nx/blob/1163212bfe71cc6324dc69bf12016c8bdaff2070/nx/lib/nx/lin_alg.ex#L225-L229

and I think it's the only one.

josevalim:
@kshalot can you please send a separate PR with a test for the float issue on 0-norm?

We can likely do this after sum:

    |> Nx.as_type(tensor, Nx.Type.to_floating(tensor.type))

nx/lib/nx/lin_alg.ex
```diff
@@ -258,14 +257,20 @@ defmodule Nx.LinAlg do
...
-
-    abs_t
-    |> Nx.divide(numerical_stability_coefficient)
-    |> Nx.power(ord)
-    |> Nx.sum(opts)
-    |> Nx.power(inv_ord)
-    |> Nx.multiply(numerical_stability_coefficient)
+    numerical_stability_coefficient = Nx.reduce_max(abs_t)
+
+    if numerical_stability_coefficient == Nx.tensor(0, type: type) do
```
josevalim, line 262: Unfortunately this won't work when norm is called from a `defn`. In this case, `Nx.reduce_max` will return an expression, and that will never be equal to 0, and then we will still get an error. The good news is that this will be fixed when we actually support unifity, b

kshalot: Right, right. I had a feeling that something was off with this. I reverted the changes, apart from the abs value issue when calculating `numerical_stability_coefficient` which will fix at least some cases.

kshalot:
> @kshalot can you please send a separate PR with a test for the float issue on 0-norm?
> 
> We can likely do this after sum:
> 
> ```
> |> Nx.as_type(tensor, Nx.Type.to_floating(tensor.type))
> ```

There you go: https://github.com/elixir-nx/nx/pull/396. There were a few more that did not return a float for a int tensor (I saw it before but for some reason wrote here that only 0-norm does this...)

josevalim:

PR #389, opened 2021-05-04
Initial implementation of vmap

### LINKED ISSUE #174: Support vmap
See the tensorflow feature and linked paper: https://www.tensorflow.org/api_docs/python/tf/vectorized_map

seanmor5:
WIP

This is an attempt to resolve #174. I've implemented `vmap` in basically the same style as `grad` except it supports multiple vectorized arguments. We don't include `out_axes` like Jax, but that's easy to add either with an additional `transpose` or to propagate during the expression transformation.

There are still a lot of checks and validations that need to be done:

- Normalize in axes
- Do we support multiple in axes?
- What constraints should be placed on in axes?

I wanted to open this to collect feedback on the foundation before moving forward with implementing some additional batch rules. Based on this PR, we'd then have to revisit some Nx implementations to support batching. Off the top of my head I believe most of the LinAlg operators will need to support batched matrix operations.

I think we can also discuss the possibility of implementing automatic vectorization based on input names (e.g. `:batch` name is always treated as a batch dimension).

### DIFF
--- a/nx/lib/nx/defn/kernel.ex
+++ b/nx/lib/nx/defn/kernel.ex
@@ -234,6 +234,13 @@ defmodule Nx.Defn.Kernel do
   end
 
   @doc """
+  Vectorizes `fun` over given `args`.
+  """
+  def vmap(fun, args, in_axes \\ nil) when is_function(fun) do
+    Nx.Defn.Vectorize.transform(fun, args, in_axes)
+  end
+
+  @doc """
   Computes the value and gradient of the given `var` on `fun`.
 
   It returns a tuple with the value and the gradient.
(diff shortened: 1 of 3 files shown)


nx/lib/nx/defn/kernel.ex
```diff
@@ -233,6 +233,13 @@ defmodule Nx.Defn.Kernel do
     grad
   end
 
+  @doc """
+  Vectorizes `fun` over given `args`.
```
josevalim, line 237: Can we add some examples and more explanation? :) I am struggling to wrap my head around the current API. I assume that `fun` receives the vectorized tensor? Are there any restrictions on `args`? What are the possible values for `in_axes`?

josevalim: Perhaps handling `dot` is a good example of how the code will change before and after with and without vmap. :)

josevalim:
Just so everyone is on the same page, I talked to @seanmor5 and we will try to implement this as tensor axes metadata instead of a transform.

PR #390, opened 2021-05-05
Add Nx.Defn.Kernel.while/3

### LINKED ISSUE #122: Introduce while loops
I would like to add a higher-level looping construct to `defn`. In Elixir, we can use `for+:reduce` but I am afraid it will be too verbose and foreign for new users. Let's imagine we want to translate this python code:

```python
    def _smooth(x):
      out = np.empty_like(x)
      for i in range(1, x.shape[0] - 1):
          for j in range(1, x.shape[1] - 1):
              out[i, j] = x[i + -1, j + -1] + x[i + -1, j + 0] + x[i + -1, j + 1] +
                          x[i +  0, j + -1] + x[i +  0, j + 0] + x[i +  0, j + 1] +
                          x[i +  1, j + -1] + x[i +  1, j + 0] + x[i +  1, j + 1]) // 9
      return out
```

With for+:reduce, we would write it as:

```elixir
    def smooth(x) do
      for i <- 1..elem(x.shape, 0)-1, j <- 1..elem(x.shape, 1)-1, reduce: x do
        x ->
          put_in x[i, j], x[i + -1, j + -1] + x[i + -1, j + 0] + x[i + -1, j + 1] +
                           x[i +  0, j + -1] + x[i +  0, j + 0] + x[i +  0, j + 1] +
                           x[i +  1, j + -1] + x[i +  1, j + 0] + x[i +  1, j + 1]) / 9
      end
    end
```

I propose we introduce a loop construct, inspired by [futhark](https://futhark-lang.org/) that looks like this:

```elixir
loop tuple_or_var [= expr], [pattern <- expr]+ do

end
```

Rewriting the above to this loop construct, we have:

```elixir
    def smooth(x) do
      loop x,
           i <- 1..elem(x.shape, 0)-1,
           j <- 1..elem(x.shape, 1)-1 do
        put_in x[i, j], x[i + -1, j + -1] + x[i + -1, j + 0] + x[i + -1, j + 1] +
                          x[i +  0, j + -1] + x[i +  0, j + 0] + x[i +  0, j + 1] +
                          x[i +  1, j + -1] + x[i +  1, j + 0] + x[i +  1, j + 1]) / 9
      end
    end
```

There is one downside with this approach: the only form of loops we have in XLA are while loops which are sequential. Other languages, such as taichi, can optimize them to run in parallel. For this reason, we may want to introduce higher level constructs for manipulating tensors. In particular, I believe we should introduce functions such as `map`, `map_with_index`, `reduce` and `reduce_with_index`. I have some thoughts in how we can implement said functions so they also work with batching out of the box, but I am waiting for [some feedback on this issue](https://groups.google.com/g/xla-dev/c/yzue7PcI-So) before moving forward.

josevalim:
Closes #122.

### DIFF
--- a/exla/c_src/exla/exla.cc
+++ b/exla/c_src/exla/exla.cc
@@ -2232,7 +2232,7 @@ static ErlNifFunc exla_funcs[] = {
   {"reduce_window", 7, reduce_window},
   {"select_and_scatter", 8, select_and_scatter},
   {"map", 4, map},
-  {"while_loop", 3, while_loop},
+  {"while", 3, while_loop},
   // Shape/Type Manipulation
   {"broadcast_in_dim", 3, broadcast_in_dim},
   {"reshape", 2, reshape},
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -180,7 +180,19 @@ defmodule EXLA.Defn do
     end
   end
 
-  defp cached_recur_operator(:cond, %T{data: %Expr{args: [clauses, last]}} = t, state, cache) do
+  defp cached_recur_operator(:while, %T{data: %Expr{args: args}}, state, cache) do
+    [initial, condition, body] = args
+    {initial, cache} = recur_composite(initial, state, cache)
+    condition = to_computation(condition, {:pred, 8}, state)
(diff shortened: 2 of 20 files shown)


josevalim:

PR #391, opened 2021-05-06
LibTorch Makefile version check

### LINKED ISSUE #368: Makefile LibTorch version check
Torchx's makefile throws an error if the version of LibTorch is not equal to `REQUIRED_LIBTORCH_VERSION` but the error message says it should be "at least" that version.

Inside `torchx/Makefile` on line 29:
```
@ if [ $(LIBTORCH_VERSION) != $(REQUIRED_LIBTORCH_VERSION) ]; then \
    echo "LibTorch version should be at least $(REQUIRED_LIBTORCH_VERSION). Got: $(LIBTORCH_VERSION)."; \
    echo "You can download LibTorch here: https://pytorch.org/get-started/locally/"; \
    exit 1; \
fi
```

So if 1.9 (nightly) version is to be supported, we should change this to `>=` instead of `!=`.

Benjamin-Philip:
Closes #368.

Suggestion/Proposal: Maybe we should also check whether the version installed is beyond the version supported by us (eg: we may not support a nightly version at some point of time). That's why I renamed `REQUIRED_*` to `MINIMUM_*`

### DIFF
--- a/torchx/Makefile
+++ b/torchx/Makefile
@@ -1,5 +1,5 @@
 LIBTORCH_VERSION := $(shell head -c 3 $(LIBTORCH_DIR)/build-version)
-REQUIRED_LIBTORCH_VERSION := 1.8
+MINIMUM_LIBTORCH_VERSION := 1.8
 
 LIBTORCH_INCLUDE_PATH = $(LIBTORCH_DIR)/include/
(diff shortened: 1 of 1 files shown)


josevalim:

Benjamin-Philip:
:smile:

PR #392, opened 2021-05-06
Make argsort expr compatible with both BinaryBackend and EXLA

jonatanklosko:
Currently `argsort/2` doesn't work inside defn unless compiled with EXLA, because the expression is not compatible with BinaryBackend. This unifies the expression for both EXLA and BinaryBackend as @seanmor5 suggested.

### DIFF
--- a/nx/lib/nx/binary_backend.ex
+++ b/nx/lib/nx/binary_backend.ex
@@ -1694,12 +1694,12 @@ defmodule Nx.BinaryBackend do
   def bitcast(out, tensor), do: from_binary(out, to_binary(tensor))
 
   @impl true
-  def sort(output, t, opts), do: do_sort(output, t, opts, false)
+  def sort(output, t, opts, comparator), do: do_sort(output, t, opts, comparator, false)
 
   @impl true
-  def argsort(output, t, opts), do: do_sort(output, t, opts, true)
+  def argsort(output, t, opts, comparator), do: do_sort(output, t, opts, comparator, true)
 
-  defp do_sort(output, t, opts, return_indices) do
+  defp do_sort(output, t, opts, comparator, return_indices) do
     %T{shape: shape, type: type} = t
     last_axis = Nx.rank(t) - 1
 
@@ -1721,11 +1721,11 @@ defmodule Nx.BinaryBackend do
             a <= b
           end
 
-        fun ->
+        _ ->
           fn a, b ->
             a = binary_to_number(a, type)
             b = binary_to_number(b, type)
-            to_scalar(fun.(a, b)) != 0
+            to_scalar(comparator.(a, b)) != 0
           end
       end
 
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -728,7 +728,9 @@ defmodule Nx.Defn.Expr do
       raise "argsort comparator must return a predicate type, got: #{inspect(fun.type)}"
     end
 
-    expr(out, context, :argsort, [tensor, opts, fun])
+    opts = put_in(opts[:comparator_fun], fun)
+
+    expr(out, context, :argsort, [tensor, opts])
   end
 
   defp to_nx_comparator(:asc), do: &Nx.less_equal/2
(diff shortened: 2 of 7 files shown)


nx/lib/nx/defn/expr.ex
```diff
@@ -728,7 +728,9 @@ defmodule Nx.Defn.Expr do
       raise "argsort comparator must return a predicate type, got: #{inspect(fun.type)}"
     end
 
-    expr(out, context, :argsort, [tensor, opts, fun])
+    opts = put_in(opts[:comparator_fun], fun)
```
josevalim, line 731: I think we should call it :comparator (which is the same name as the input to this function).

jonatanklosko: So I thought, but there is already `:comparator` in the options and BinaryBackend uses it:
https://github.com/elixir-nx/nx/blob/bf0293b7495c5b43d86af94e82ba794deb9acba5/nx/lib/nx/binary_backend.ex#L1709

josevalim: I see. So what if we don't send comparator_fun and we let EXLA normalize :asc and :desc like the binarybackend?

jonatanklosko: What about these validations?

https://github.com/elixir-nx/nx/blob/bf0293b7495c5b43d86af94e82ba794deb9acba5/nx/lib/nx/defn/expr.ex#L723-L729

josevalim: I see... so I think EXLA actually got it right. We don't want to hide the `fun`, which is a Nx.Defn.Expr, inside the options. If we do so, some operations like `Nx.Defn.Tree.traverse_args` won't traverse it. So I think it should be `tensor, opts, comparator`, and leave the original comparison function as `:comparator`. Also please mirror the changes to `sort` too.

jonatanklosko: But the problem is that if we pass the comparator as a separate argument, then this would require `BinaryBackend` to accept such argument, and consequently every backend to do so. I certainly miss something though :>

josevalim: Yes, the backend should receive it and call it unless the comparator function is :asc / :desc.

jonatanklosko: Does this mean we have to build this `Nx.Defn.Expr` comparator on the highest level in `Nx.argsort`?

josevalim: No, Nx.Defn.Expr already builds it for argsort. And then, when evaluating, we unwrap the Nx.Defn.Expr for the function into a regular function. The important is to keep the anonymous function at the top level of the args, so the existing code can see it. If you do that, then you don't need to change anything else. :)

jonatanklosko: My point is that if we want to pass `fun` as an argument and not an option, we would need to change `BinaryBackend.argsort/3` to `BinaryBackend.argsort/4`, which doesn't comply with the backend behaviour, isn't that the case?

josevalim: Yes, we should change the backend behaviour for sort and argsort. :)

nx/lib/nx/binary_backend.ex
```diff
@@ -1721,11 +1721,11 @@ defmodule Nx.BinaryBackend do
             a <= b
           end
 
-        fun ->
+        _ ->
           fn a, b ->
             a = binary_to_number(a, type)
             b = binary_to_number(b, type)
-            to_scalar(fun.(a, b)) != 0
+            to_scalar(comparator.(a, b)) != 0
```
jonatanklosko, line 1728: Given we get this normalized `comparator` function, would it make sense to always use that? Or are the specific `:asc`/`:desc` implementations above more efficient here?

nx/lib/nx/defn/expr.ex
jonatanklosko: Thanks! I *think* I got this now, let me know if the updated version makes sense :cat:

nx/lib/nx/binary_backend.ex
josevalim: Correct, they are quite more efficient because they don't create tensors.

josevalim:

PR #394, opened 2021-05-08
Change default sort comparators to use strict inequality

### LINKED ISSUE #393: Argsort results in segmentation fault given tensor of the same values
Given a tensor of repeated value (e.g. 100 zeros) `argsort` results in segfault when compiled with EXLA.

**Reproduction**

```elixir
defmodule N do
  import Nx.Defn

  @default_defn_compiler EXLA

  defn argsort(t), do: Nx.argsort(t)
end

x = Nx.broadcast(0.0, {100})
N.argsort(x)
```

**Debug output**

```
(gdb) bt
#0  0x00007f28824e500f in _Function_0.92076480_1_in_N._-fun.__defn_argsort___1-__--to_operator_4-fun-2--325.3 ()
#1  0x00007f27d83cad7c in void std::__introsort_loop<long long*, long, __gnu_cxx::__ops::_Iter_comp_iter<__xla_cpu_runtime_KeyValueSort::{lambda(long long, long long)#1}> >(long long*, __gnu_cxx::__ops::_Iter_comp_iter<__xla_cpu_runtime_KeyValueSort::{lambda(long long, long long)#1}>, long, __gnu_cxx::__ops::_Iter_comp_iter<__xla_cpu_runtime_KeyValueSort::{lambda(long long, long long)#1}>) ()
   from /home/jonatanklosko/.cache/mix/installs/elixir-1.13.0-dev-erts-12.0/89a051d0b2809c2accd8052dc9cc8a62/_build/dev/lib/exla/priv/libexla.so
#2  0x00007f27d83cc5bd in __xla_cpu_runtime_KeyValueSort ()
   from /home/jonatanklosko/.cache/mix/installs/elixir-1.13.0-dev-erts-12.0/89a051d0b2809c2accd8052dc9cc8a62/_build/dev/lib/exla/priv/libexla.so
#3  0x00007f28824e52cb in _Function_0.92076480_1_in_N._-fun.__defn_argsort___1-__.12 ()
#4  0x00007f27d80a3f22 in xla::cpu::CpuExecutable::ExecuteComputeFunction(xla::ExecutableRunOptions const*, absl::lts_2020_02_25::Span<xla::MaybeOwningDeviceMemory const>, xla::HloExecutionProfile*) ()
   from /home/jonatanklosko/.cache/mix/installs/elixir-1.13.0-dev-erts-12.0/89a051d0b2809c2accd8052dc9cc8a62/_build/dev/lib/exla/priv/libexla.so
#5  0x00007f27d80a4be1 in xla::cpu::CpuExecutable::ExecuteAsyncOnStream(xla::ServiceExecutableRunOptions const*, std::vector<xla::ExecutionInput, std::allocator<xla::ExecutionInput> >, xla::HloExecutionProfile*)::AsyncRunTask::operator()() ()
   from /home/jonatanklosko/.cache/mix/installs/elixir-1.13.0-dev-erts-12.0/89a051d0b2809c2accd8052dc9cc8a62/_build/dev/lib/exla/priv/libexla.so
#6  0x00007f27dca5b7a6 in stream_executor::host::HostStream::WorkLoop() ()
   from /home/jonatanklosko/.cache/mix/installs/elixir-1.13.0-dev-erts-12.0/89a051d0b2809c2accd8052dc9cc8a62/_build/dev/lib/exla/priv/libexla.so
#7  0x00007f27dcbda2ef in tensorflow::(anonymous namespace)::PThread::ThreadFn(void*) ()
   from /home/jonatanklosko/.cache/mix/installs/elixir-1.13.0-dev-erts-12.0/89a051d0b2809c2accd8052dc9cc8a62/_build/dev/lib/exla/priv/libexla.so
#8  0x00007f2882433609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#9  0x00007f2882002293 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

**Observations**

* The issue is not specific to `0`, but rather to tensors of a single unique value
  * Actually it's even the case if some values are the same, for example:\
    `Nx.concatenate([Nx.broadcast(0.0, {10}), Nx.broadcast(1.0, {10})])`
* The issue doesn't affect small tensors (e.g. `{10}`)
* The issue is dimension agnostic (happens for `{20, 20, 20, 20, 20}`, does not happen for `{10, 10, 10, 10, 10}`)

I tried reproducing the issue with Jax with no luck (with`jaxlib` both latest and an older one around the same version of XLA that EXLA uses).

Maybe not related to this issue, but for the smaller tensors for which `argsort` works, the returned indices are descending, while in Jax they are ascending.

jonatanklosko:
Closes #393.

Experimenting with @seanmor5 we found out that even with XLS stable sort, the results don't look stable at all (the order of same elements is reversed). Turns out the comparators need to be strict inequality, and in turn this fixes the segfault problem.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7496,8 +7496,10 @@ defmodule Nx do
 
   ### Options
 
-  * `:axis`: The name or number of the corresponding axis on which the sort should be applied.
-  * `:comparator`: Can be `:asc`, `:desc` or a arity-2 function for comparison between two tensor elements. Defaults to `:asc`
+    * `:axis` - The name or number of the corresponding axis on which the sort
+      should be applied
+    * `:comparator` - Can be `:asc`, `:desc` or an arity-2 function for comparison
+      between two tensor elements (strict inequality). Defaults to `:asc`
 
   ### Examples
 
@@ -7649,8 +7651,8 @@ defmodule Nx do
 
     * `:axis` - The name or number of the corresponding axis on which the sort
       should be applied
-    * `:comparator` - Can be `:asc`, `:desc` or a arity-2 function for comparison
-      between two tensor elements. Defaults to `:asc`
+    * `:comparator` - Can be `:asc`, `:desc` or an arity-2 function for comparison
+      between two tensor elements (strict inequality). Defaults to `:asc`
 
   ## Examples
 
@@ -7765,8 +7767,8 @@ defmodule Nx do
 
   ## Helpers
 
-  defp to_nx_comparator(:asc), do: &Nx.less_equal/2
-  defp to_nx_comparator(:desc), do: &Nx.greater_equal/2
+  defp to_nx_comparator(:asc), do: &Nx.less/2
+  defp to_nx_comparator(:desc), do: &Nx.greater/2
   defp to_nx_comparator(comp) when is_function(comp, 2), do: comp
 
   defp to_nx_comparator(_) do


josevalim:
@jonatanklosko Can we please add a test to both Nx (on nx_test.exs) and EXLA (on defn_test.exs) test suites passing a zero-Ed tensor and asserting on the result?

jonatanklosko:
@josevalim I though so, but as we left the sorting instable I don't think we can reliably have such test?

josevalim:
Oh, I see your point. Ignore me!

PR #396, opened 2021-05-09
Make Nx.LinAlg.norm/2 always return a float

kshalot:
As mentioned here: https://github.com/elixir-nx/nx/pull/387#issuecomment-835773806.

There were a bunch norms that did not return a float: `0`, `1`, `-1`,  `inf`, `-inf`. Alternatively the cast could be moved to `Nx.LinAlg.norm/2` which server as a common 'entrypoint' for all the implementations, then it's only implemented once. It's implicitly tested by the doctests.


josevalim:

PR #397, opened 2021-05-09
Make Nx.stack/2 support stacking along any axis

kshalot:
This is something I needed for personal use, so I figured I ought to make a PR since I have the implementation ready anyway.

I presume the `Nx.stack` function is supposed to deliver similar functionality as `np.stack` which supports stacking along any given axis - [numpy docs](https://numpy.org/doc/stable/reference/generated/numpy.stack.html#numpy-stack).  The current implementation from `main` works only along the first axis (the docs say it only works for scalar tensors, but in reality the implementation supports any rank).

I think this function is a bit lackluster if it doesn't support this. What do you think?

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7489,10 +7489,11 @@ defmodule Nx do
   end
 
   @doc """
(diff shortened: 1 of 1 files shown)


josevalim:

PR #398, opened 2021-05-09
Remove support for user-defined comparator for sort and argsort

jonatanklosko:
Replaces the `:comparator` option of `sort` and `argsort` with `:direction` that is either `:asc` or `:desc`. We decided not to allow arbitrary comparator functions, as some backends may not support it (Torchx) and Jax doesn't support that either.

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -722,25 +722,38 @@ defmodule EXLA.Defn do
     EXLA.Op.select(EXLA.Op.equal(cholesky, tensor), zeros, cholesky)
   end
 
-  defp to_operator(:sort, [tensor, opts, comparator], _ans, state) do
+  defp to_operator(:sort, [tensor, opts], ans, state) do
     dimension = opts[:axis]
-    comp = to_computation(comparator, {:pred, 8}, state)
+
+    fun =
+      case opts[:direction] do
+        :asc -> binary_op_fun(:less)
+        :desc -> binary_op_fun(:greater)
+      end
+
+    args = [%{type: ans.type, shape: {}}, %{type: ans.type, shape: {}}]
+    comp = to_computation(:comparator, args, state, fun)
+
     EXLA.Op.sort(tensor, comp, dimension)
   end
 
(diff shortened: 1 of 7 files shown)


exla/lib/exla/defn.ex
```diff
@@ -722,25 +722,38 @@ defmodule EXLA.Defn do
...
-    arg3 = Expr.parameter(:argsort, ans.type, {}, 3)
+    fun =
+      case opts[:direction] do
+        :asc -> binary_op_fun(:less)
+        :desc -> binary_op_fun(:greater)
+      end
 
-    comparator_4 = Expr.fun([arg0, arg1, arg2, arg3], context, fn x, y, _, _ -> fun.(x, y) end)
+    args = [
+      %{type: op_type(tensor), shape: {}},
```
jonatanklosko, line 750: I wasn't sure how to get the proper type, but this does work, please let me know if it makes sense :>

exla/lib/exla/defn.ex
```diff
@@ -722,25 +722,38 @@ defmodule EXLA.Defn do
...
 
-    # Grow the comparator to arity 4 because argsort uses
-    # variadic_sort underneath
-    %{args: [[arg0, arg1], _expr, fun], context: context} = comparator.data
-
-    arg2 = Expr.parameter(:argsort, ans.type, {}, 2)
-    arg3 = Expr.parameter(:argsort, ans.type, {}, 3)
+    fun =
+      case opts[:direction] do
+        :asc -> binary_op_fun(:less)
```
jonatanklosko, line 745: We use `fun` to build a arity-4 computation, so using binary op is weird, but it does work. If there are better ways for building this computation let me know =)

exla/lib/exla/defn.ex
josevalim: I think you can keep the code as it was before, except that `fun` will be either `Nx.less` or `Nx.greater` based on the direction?

josevalim: In any case, this looks good! :+1:

josevalim:

PR #401, opened 2021-05-11
feat: add more options to triangular_solve

polvalente:
Adds missing opts from [XLA](https://www.tensorflow.org/xla/operation_semantics#triangularsolve) to `Nx.LinAlg.triangular_solve`

### DIFF
--- a/nx/lib/nx/binary_backend/matrix.ex
+++ b/nx/lib/nx/binary_backend/matrix.ex
@@ -3,31 +3,78 @@ defmodule Nx.BinaryBackend.Matrix do
   @default_eps 1.0e-10
   import Nx.Shared
 
-  def ts(a_data, a_type, b_data, b_type, {rows, rows} = shape, output_type, opts) do
-    a_matrix = a_data |> binary_to_matrix(a_type, shape) |> ts_handle_lower_opt(opts, :a)
-    b_matrix = b_data |> binary_to_matrix(b_type, shape) |> ts_handle_lower_opt(opts, :b)
+  def ts(a_data, a_type, b_data, b_type, shape, output_type, opts) do
+    a_shape =
+      case shape do
+        {rows} -> {rows, rows}
+        shape -> shape
+      end
 
+    a_matrix = a_data |> binary_to_matrix(a_type, a_shape) |> ts_handle_lower_opt(opts, :a)
+
+    b_matrix_or_vec =
+      case shape do
+        {rows, rows} ->
+          b_data |> binary_to_matrix(b_type, shape) |> ts_handle_lower_opt(opts, :b)
+
+        {_rows} ->
+          b_data |> binary_to_vector(b_type) |> ts_handle_lower_opt(opts, :b)
+      end
+
+    result =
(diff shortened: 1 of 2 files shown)


nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -3,40 +3,140 @@ defmodule Nx.BinaryBackend.Matrix do
   @default_eps 1.0e-10
   import Nx.Shared
 
-  def ts(a_data, a_type, b_data, b_type, {rows, rows} = shape, output_type, _opts) do
-    a_matrix = binary_to_matrix(a_data, a_type, shape)
-    b_matrix = binary_to_matrix(b_data, b_type, shape)
+  def ts(a_data, a_type, b_data, b_type, shape, output_type, opts) do
```
polvalente, line 6: @josevalim I ended up implementing this "properly" because with a few row-swapping and transposing tricks the base `ts` code could solve all of the new options.

It should be noted that I haven't generated any examples using Jax, so maybe some results should be checked?

polvalente: Also also, this PR doesn't add the new opts to EXLA. We need to add them in a separate pull request. Most are straightforward, but the `transpose` option didn't seem to be a plain boolean.

nx/lib/nx/lin_alg.ex
```diff
@@ -277,11 +277,25 @@ defmodule Nx.LinAlg do
...
+  The following options are defined in order of precedence
+
+  * `:transform_a` - Defines `op(a)`, depending on its value. Can be one of:
+    * `:none` -> `op(a) = a`
+    * `:transpose` -> `op(a) = transpose(a)`
+    Defaults to `:none`
+  * `:lower` - When `true`, defines the `a` matrix as lower triangular. If false, a is upper triangular.
+               Defaults to `true`
+  * `:left_side` - When `true`, solves the system as `op(A).X = B`. Otherwise, solves `X.op(A) = B`. Defaults to `true`.
+  * `:unit_diagonal` - When `true`, the diagonal elements aren't accessed. Defaults to `false`
```
josevalim, line 297: I don't think we need to support this option for now. :)

nx/lib/nx/lin_alg.ex
```diff
@@ -277,11 +277,25 @@ defmodule Nx.LinAlg do
...
   `b` must either be a square matrix with the same dimensions as `a` or a 1-D tensor
   with as many rows as `a`.
 
+  ## Options
+
+  The following options are defined in order of precedence
+
+  * `:transform_a` - Defines `op(a)`, depending on its value. Can be one of:
+    * `:none` -> `op(a) = a`
+    * `:transpose` -> `op(a) = transpose(a)`
```
josevalim, line 292: In my opinion we don't need to support transpose either, but since you added it, we can keep it. :)

josevalim (commented):
Looks great, I have added some very minor comments!

josevalim (approved):

josevalim (approved):

Issue #403, opened 2021-05-12
triangular_solve options for binary backend does not match EXLA behaviour


josevalim:
https://github.com/elixir-nx/nx/blob/main/nx/test/nx/lin_alg_test.exs#L15

PR #405, opened 2021-05-13
fix: use correct contracts for triangular_solve opts

polvalente:
Solves #403 

As reported in said issue, some properties which held in XLA (and also hold in Jax) did not hold for the current implementation in Nx. Upon further inspection, besides some bugs, there was a conceptual error in the `:transform_a` option.
XLA, it interacts with `:lower` for defining the final interpretation of the matrix, and in my implementation, it was a non-informed transform which didn't change `:lower` in any way.

### DIFF
--- a/nx/test/nx/lin_alg_test.exs
+++ b/nx/test/nx/lin_alg_test.exs
@@ -12,16 +12,47 @@ defmodule Nx.LinAlgTest do
       upper = Nx.transpose(a)
       assert Nx.dot(upper, Nx.LinAlg.triangular_solve(upper, b, lower: false)) == b
 
-      # TODO: THese properties hold for EXLA but not here.
-      # assert Nx.dot(
-      #         Nx.LinAlg.triangular_solve(upper, b, left_side: false, lower: false),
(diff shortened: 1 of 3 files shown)


nx/test/nx/lin_alg_test.exs
```diff
@@ -12,16 +12,18 @@ defmodule Nx.LinAlgTest do
...
-      # assert Nx.dot(
-      #          Nx.transpose(a),
-      #          Nx.LinAlg.triangular_solve(a, b, transform_a: :transpose)
-      #        ) == b
+      assert Nx.dot(
+               Nx.LinAlg.triangular_solve(upper, b, left_side: false, lower: false),
+               upper
+             ) == b
+
+      assert Nx.LinAlg.triangular_solve(a, b, transform_a: :transpose) ==
```
polvalente, line 20: Also a sanity check: transposing before passing the tensor has the same effect as letting transposing take hold inside the function

josevalim (approved):
Absolutely awesome!

PR #406, opened 2021-05-14
Maps in defn

### LINKED ISSUE #208: Support maps in defn
The downside of maps though is that you can pass a bag values and only use some of them in the computation, so we will need to do some variable pruning. Therefore, before we support maps, I suggest to implement tree pruning with tuples and effectively remove the restriction that tuples must be pattern matching on the `defn` head. That will prepare us to add maps later on.

Internally, maps can compile to and run as tuples on the GPU. We will stitch them back together on Elixir land. Further consideration about wrapping will be needed for constructs such as `if` and `while`.

josevalim:
This pull request supports maps in defn, including inside block constructs, such as `if` and `map`.

Closes #208.


josevalim:

PR #407, opened 2021-05-14
Extract device id from device assignment

seanmor5:


exla/c_src/exla/exla.cc
```diff
@@ -2045,6 +2045,35 @@ ERL_NIF_TERM run(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[]) {
   return term;
 }
 
+ERL_NIF_TERM device_assignment_to_device_id(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[]) {
+  if (argc != 3) {
+    return exla::nif::error(env, "Bad argument count.");
+  }
+
+  exla::ExlaExecutable** exec;
```
polvalente, line 2053:
```suggestion
  exla::ExlaExecutable* exec;
```

Can't this be just a pointer?
Then on line 2056 you could pass `&exec` and use `exec->` instead of `(*exec)->` everywhere else

exla/lib/exla/executable.ex
```diff
@@ -73,6 +73,8 @@ defmodule EXLA.Executable do
 
     run_id = Keyword.get(options, :run_id, System.unique_integer([:positive, :monotonic]))
     replica = Keyword.get(options, :replica, 1)
+
+    # device_id = device_assignment_to_device_id(executable, {replica, 1})
```
polvalente, line 77: Is this commented on purpose?

PR #410, opened 2021-05-16
fix: LU decomposition

### LINKED ISSUE #409: Function Nx.LinAlg.lu does not work properly
function call
```elixir
Nx.LinAlg.lu(Nx.tensor([[0, 1, 3], [0, -1, -2], [1, 1, 2]]))
```

gives PLU decomposition as given below

```elixir
{#Nx.Tensor<
   s64[3][3]
   [
     [0, 0, 1],
     [0, 1, 0],
     [1, 0, 0]
   ]
>, #Nx.Tensor<
   f32[3][3]
   [
     [1.0, 0.0, 0.0],
     [0.0, 1.0, 0.0],
     [0.0, 0.0, 1.0]
   ]
>, #Nx.Tensor<
   f32[3][3]
   [
     [1.0, 1.0, 2.0],
     [0.0, -1.0, -2.0],
     [0.0, 0.0, 3.0]
   ]
>}
```

But the product of those matrices has different determinant than original matrix (3 instead of 1).

polvalente:
Changes the LU decomposition implementation for the BinaryBackend, based on RosettaCode's Python implementation.
I honestly couldn't find what was the actual error in `main`'s implementation, but this one now works.

@josevalim @seanmor5 I'm thinking that maybe we should try to add some property tests in another PR for these decompositions (since most yield the original matrix after dot-producting the factors -- the current exception is SVD, but even then we can generate some well conditioned checks around it). WDYT?

closes #409


nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -237,44 +237,29 @@ defmodule Nx.BinaryBackend.Matrix do
...
-            else
-              i_range = if j == n - 1, do: [n - 1], else: j..(n - 1)
-
-              for i <- i_range, reduce: l do
-                l ->
-                  u_col_slice = slice_matrix(u, [0, j], [j, 1])
-                  [l_row_i] = get_matrix_rows(l, [i])
-                  s = dot_matrix(u_col_slice, l_row_i)
-                  [a_elem] = get_matrix_elements(a_prime, [[i, j]])
+            for i <- j..(n - 1), i != j, reduce: l do
```
polvalente, line 253: update/facepalm: It seems that the error was that `if u_jj < eps` should be `if abs(u_jj) < eps`
However, the new implementation is cleaner (albeit a little bit less efficient) and I want to keep it for readability.

josevalim (approved):
Awesome job! I agree with adding some poor man’s property based testing where we generate 10 random matrices and decompose then. 👍

PR #414, opened 2021-05-19
feat: add github actions CI workflow

polvalente:
This PR introduces a basic CI workflow for Nx


polvalente:
@josevalim @seanmor5 some configuration is needed in the repo setting so the workflow only runs on pushes and is required for merges

josevalim (approved):
Awesome job! @polvalente, I have added you as an admin, so you should be able to configure what you need!

PR #417, opened 2021-05-21
Update Dockerfiles to use OTP24

seanmor5:
Updates EXLA Dockerfiles to use OTP24 with @wojtekmach beamup.

Also adds Livebook


josevalim (approved):

wojtekmach:
beamup isnt stable, I wouldn’t use it. I’d consider either basing your image off hexpm/elixir from Dockerhub, we have variants for alpine and Ubuntu . Or, use whatever base image you already have but download OTP tarball built by Hex Bob: https://github.com/hexpm/bob#erlang-builds

PR #420, opened 2021-06-06
Truncate float to integer conversion

### LINKED ISSUE #419: Allow as_type to convert from float to integer
I think our current implementation is too restrictive. I have an implementation that ends up with `f32` tensors:

```
rows = Nx.tensor([0.45, 0.52, 0.7])
```

And I need to use them to index:

```
indices = Nx.floor(rows) |> Nx.as_type({:s, 64})
```

But because of the restriction the current implementation is impossible

seanmor5:
Resolves #419

### DIFF
--- a/nx/lib/nx/binary_backend.ex
+++ b/nx/lib/nx/binary_backend.ex
@@ -2,6 +2,7 @@ defmodule Nx.BinaryBackend do
   @moduledoc """
   An opaque backend written in pure Elixir that stores
   the data in Elixir's binaries.
+
   This is the default backend used by the `Nx` module.
   The backend itself (and its data) is private and must
   not be accessed directly.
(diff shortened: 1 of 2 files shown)


seanmor5:
Agh, I'm working on Windows and the formatter does not seem to be working correctly (something with line endings I think?).

nx/lib/nx/binary_backend.ex
```diff
@@ -2,7 +2,7 @@ defmodule Nx.BinaryBackend do
   @moduledoc """
   An opaque backend written in pure Elixir that stores
   the data in Elixir's binaries.
-
+
```
jonatanklosko, line 5: You have a trailing whitespace here, that's what the formatter complains about :)

seanmor5, line 5:
```suggestion

```

josevalim (approved):

nx/lib/nx/binary_backend.ex
```diff
@@ -2,7 +2,6 @@ defmodule Nx.BinaryBackend do
   @moduledoc """
   An opaque backend written in pure Elixir that stores
   the data in Elixir's binaries.
-
   This is the default backend used by the `Nx` module.
```
jonatanklosko, line 5:
```suggestion

  This is the default backend used by the `Nx` module.
```

PR #422, opened 2021-06-06
Add binary comparison operators a special forms

### LINKED ISSUE #361: Discussion: binary comparison operators as special-forms/macros?
There already exist in `Nx.Defn.Kernel` a majority of basic arithmetic operators that are shorthand for functions in the `Nx` module. However, binary comparison operators must be called on `Nx` inside numerical functions:

Operator | Function
------------ | -------------
`==` | `Nx.equal/2`
`!=` | `Nx.not_equal/2`
`<` | `Nx.less/2`
`>` | `Nx.greater/2`
`<=` | `Nx.less_equal/2`
`>=` | `Nx.greater_equal/2`

As a motivating example, consider a Nx2 tensor representing a bounding box, where the first row is the minimum point and the second row is the maximum point. If we define a numerical function that determines whether any given point is inside the bounding box, the current way would require something like the following:

```elixir
  defn inside?(bounds, point) do
    Nx.all?(Nx.greater_equal(point, bounds[0]) and Nx.less_equal(point, bounds[1]))
  end
```

Note that binary `and` expands to/is an alias for `Nx.logical_and/2`. If we had binary comparison operators:

```elixir
  defn inside?(bounds, point) do
    Nx.all?(point >= bounds[0] and point <= bounds[1])
  end
```

### Issues to discuss

* How should booleans (`true`, `false` atoms) behave? Should they be disallowed inside numerical functions? Should they expand to the scalar `1` and `0` tensors? 
* How should comparisons behave inside `if` and `cond` expressions? Should `Nx.all?` or `Nx.any?` be required?
* Does introducing more operators have more value than highlighting the difference between numerical functions and regular Elixir functions?
* Would this introduce too much burden (in implementation, maintenance, performance) on the compilation / evaluation parts of Nx?

See also the [Twitter thread](https://twitter.com/seancribbs/status/1376346097153667079) that started this discussion.

seanmor5:
Resolves #361 

Sorry ahead of time for formatting issues :)

I think this covers the plan outlined in that issue

### DIFF
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -173,13 +173,26 @@ defmodule Nx.Defn.Expr do
   def cond(file, clauses, last) do
     clauses =
       for {meta, {pred, expr}} <- clauses do
-        pred = to_expr(pred)
+        pred =
+          case pred do
+            pred when is_boolean(pred) ->
+              raise CompileError,
+                line: meta[:line],
+                file: file,
+                description: "boolean predicate passed to cond/if, cond/if expects" <>
+                             " scalar tensor predicate, consider using 1 for true" <>
+                             " or 0 for false as an alternative"
+            pred ->
+              to_expr(pred)
+            end
 
         if not match?(%T{shape: {}}, pred) do
           raise CompileError,
             line: meta[:line],
             file: file,
-            description: "condition must be a scalar tensor, got: #{inspect(pred)}"
+            description: "condition must be a scalar tensor, got: #{inspect(pred)}," <>
+                            " consider using Nx.all?/1 or Nx.any?/1 to obtain a scalar" <>
+                            " predicate from tensor"
         end
 
         if not compatible?(last, expr, fn _, _ -> true end) do
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -175,16 +175,6 @@ defmodule Nx.Defn.Expr do
       for {meta, {pred, expr}} <- clauses do
         pred = to_pred(pred, meta[:line], file, :cond)
 
-        if not match?(%T{shape: {}}, pred) do
-          raise CompileError,
-            line: meta[:line],
-            file: file,
-            description:
-              "condition must be a scalar tensor, got: #{inspect(pred)}," <>
-                " consider using Nx.all?/1 or Nx.any?/1 to obtain a scalar" <>
-                " predicate from tensor"
-        end
-
         if not compatible?(last, expr, fn _, _ -> true end) do
           raise CompileError,
             line: meta[:line],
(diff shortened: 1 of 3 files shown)


nx/lib/nx/defn/expr.ex
```diff
@@ -173,13 +173,26 @@ defmodule Nx.Defn.Expr do
...
+            pred when is_boolean(pred) ->
+              raise CompileError,
+                line: meta[:line],
+                file: file,
+                description: "boolean predicate passed to cond/if, cond/if expects" <>
+                             " scalar tensor predicate, consider using 1 for true" <>
+                             " or 0 for false as an alternative"
+            pred ->
+              to_expr(pred)
+            end
```
josevalim, line 187: We have added while now... so maybe we should create a to_pred function and encapsulate this there? Then call it from cond and where.

nx/test/nx/defn_test.exs
```diff
@@ -535,21 +535,42 @@ defmodule Nx.DefnTest do
     end
 
     defn land_two(a, b), do: a and b
+    defn land_true(a) do
```
josevalim, line 538:
```suggestion

    defn land_true(a) do
```

nx/test/nx/defn_test.exs
```diff
@@ -535,21 +535,42 @@ defmodule Nx.DefnTest do
...
     test "and" do
       assert %T{data: %Expr{op: :logical_and, args: [_, _]}} = land_two(1, 2)
+
+      assert_raise ArgumentError, ~r/boolean value passed to Nx.Defn.Kernel.and\/2/, fn ->
+        land_true(2)
+      end
     end
 
     defn lor_two(a, b), do: a or b
+    defn lor_true(a) do
```
josevalim, line 552:
```suggestion

    defn lor_true(a) do
```

nx/lib/nx/defn/kernel.ex
```diff
@@ -635,6 +656,85 @@ defmodule Nx.Defn.Kernel do
...
+  Element-wise greater than operation.
+
+  ## Examples
+
+      defn check_greater_than(a, b) do
+        a > b
+      end
+  """
+  def left > right when Kernel.and(is_number(left), is_number(right)),
+    do: Kernel.>(left, right)
```
josevalim, line 712: We are missing the Nx clause below for `<`, `<=`, and `>=`.

nx/lib/nx/defn/expr.ex
```diff
@@ -206,13 +209,16 @@ defmodule Nx.Defn.Expr do
         {parameter(expr, :while, counter), {counter + 1, merge_context!(expr, acc)}}
       end)
 
-    condition = condition.(arg)
+    condition = to_pred(condition.(arg), line, file, :while)
 
     if not match?(%T{shape: {}}, condition) do
```
josevalim, line 214: I think we can also move this check to `to_pred`. WDYT?

josevalim (approved):
Just one additional comment with another potential refactoring and we can ship it!

PR #424, opened 2021-06-18
feat: add eigen decomposition

polvalente:


josevalim:
Looking great!

floatn:
Btw, why two arguments for `defp identity_matrix`?

polvalente:
> Btw, why two arguments for `defp identity_matrix`?

Somewhere during development for this PR it became useful to have an MxN definition for the identity, instead of a purely square one

PR #429, opened 2021-06-24
First draft of CPU host callbacks

seanmor5:
Implements host callbacks for intermediate values on EXLA CPU. I can mess around with a GPU version later. I'm open to change the API in any way.

### DIFF
--- a/exla/lib/exla/callback_registry.ex
+++ b/exla/lib/exla/callback_registry.ex
@@ -51,12 +51,13 @@ defmodule EXLA.CallbackRegistry do
 
   @impl true
   def handle_info({:run, name, data, shape}, callbacks) do
+    name = List.to_string(name)
     case callbacks[name] do
-      {fun, _} when is_function(fun) ->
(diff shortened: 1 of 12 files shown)


exla/lib/exla/callback_registry.ex
```diff
@@ -0,0 +1,71 @@
...
+
+        fun.(tensor)
+
+      _ ->
+        raise "Callback #{name} not found"
+    end
+
+    {:noreply, callbacks}
+  end
+end
```
josevalim, line 71: We should probably make one process per device. That was my plan for streaming before we added PjRT but I am not sure if it is needed now. But I guess it still is because we can add functionality like this? :)

seanmor5: I think 1 process per device makes sense still, so a single process is not handling potentially many callbacks at once

josevalim:
Closing this! I have a proposal for hooks coming later today!

PR #433, opened 2021-08-01
Implement Nx.take/3 backed by xla::Gather

jonatanklosko:
This PR introduces two bits:

* bindings to [`xla::Gather`](https://www.tensorflow.org/xla/operation_semantics#gather) - a very generic operation that may be used to implement many specific operations
* `Nx.take/3` - one specific operation backed by `xla::Gather`, it corresponds to [np.take](https://numpy.org/doc/stable/reference/generated/numpy.take.html) (Group 1 below). I went with the same naming because gather is too broad, but alternatives are welcome

## Gather overview

Gather is a really broad term and its meaning varies across packages, some variants I found could be grouped like so:

**Group 1: "reorder" tensor along specific axis**

```
tensor: [
  [1, 2],
  [3, 4]
]
indices: [1, 0, 1]

result along axis 0: [
  [3, 4]
  [1, 2],
  [3, 4]
]
```

* [tf.gather](https://www.tensorflow.org/api_docs/python/tf/gather) without batching
* [np.take](https://numpy.org/doc/stable/reference/generated/numpy.take.html)
* [torch.index_select](https://pytorch.org/docs/stable/generated/torch.index_select.html)

**Group 2: zip tensor slices with indices slices and "reorder" accordingly**

```
tensor: [
  [1, 2],
  [3, 4]
]
indices: [
  [1, 0],
  [1, 1]
]

result along axis 1: [
  [2, 1],
  [4, 4]
]
```

* [tf.gather](https://www.tensorflow.org/api_docs/python/tf/gather) with batching (though I think it only works if batching along leading axis)
* [np.take_along_axis](https://numpy.org/doc/stable/reference/generated/numpy.take_along_axis.html)
* [torch.gather](https://pytorch.org/docs/stable/generated/torch.gather.html#torch.gather)

**Group 3: pick elements at specific indices**

```
tensor: [
  [1, 2],
  [3, 4]
]
indices: [[1, 0], [0, 1]]

result: [3, 2]
```

* [tf.gather_nd](https://www.tensorflow.org/api_docs/python/tf/gather_nd)
* `np` and `torch` support this via fancy indexing as far as I understand

### DIFF
--- a/exla/c_src/exla/exla.cc
+++ b/exla/c_src/exla/exla.cc
@@ -511,6 +511,60 @@ ERL_NIF_TERM dynamic_update_slice(ErlNifEnv* env, int argc, const ERL_NIF_TERM a
   return exla::nif::ok(env, exla::nif::make<xla::XlaOp>(env, op));
 }
 
+ERL_NIF_TERM gather(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[]) {
+  if (argc != 7) {
+    return exla::nif::error(env, "Bad argument count.");
+  }
+
+  xla::XlaOp* operand;
+  xla::XlaOp* start_indices;
+  exla::int64 index_vector_dim;
+  std::vector<xla::int64> slice_sizes;
+  std::vector<xla::int64> offset_dims;
+  std::vector<xla::int64> collapsed_slice_dims;
+  std::vector<xla::int64> start_index_map;
+
+  if (!exla::nif::get<xla::XlaOp>(env, argv[0], operand)) {
+    return exla::nif::error(env, "Unable to get operand.");
+  }
+  if (!exla::nif::get<xla::XlaOp>(env, argv[1], start_indices)) {
+    return exla::nif::error(env, "Unable to get start indices.");
+  }
+  if (!exla::nif::get(env, argv[2], &index_vector_dim)) {
+    return exla::nif::error(env, "Unable to get index vector dim.");
+  }
+  if (!exla::nif::get_list(env, argv[3], slice_sizes)) {
+    return exla::nif::error(env, "Unable to get slice sizes.");
+  }
+  if (!exla::nif::get_list(env, argv[4], offset_dims)) {
+    return exla::nif::error(env, "Unable to get offset dims.");
+  }
+  if (!exla::nif::get_list(env, argv[5], collapsed_slice_dims)) {
+    return exla::nif::error(env, "Unable to get collapsed slice dims.");
+  }
+  if (!exla::nif::get_list(env, argv[6], start_index_map)) {
+    return exla::nif::error(env, "Unable to get start index map.");
+  }
+
+  xla::GatherDimensionNumbers dim_numbers;
+
+  dim_numbers.set_index_vector_dim(index_vector_dim);
+
+  for (auto dim : offset_dims) {
+    dim_numbers.add_offset_dims(dim);
+  }
+  for (auto dim : collapsed_slice_dims) {
+    dim_numbers.add_collapsed_slice_dims(dim);
+  }
+  for (auto dim : start_index_map) {
+    dim_numbers.add_start_index_map(dim);
+  }
+
+  xla::XlaOp op = xla::Gather(*operand, *start_indices, dim_numbers, slice_sizes);
+
+  return exla::nif::ok(env, exla::nif::make<xla::XlaOp>(env, op));
+}
+
 // Creation
 
 ERL_NIF_TERM rng_normal(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[]) {
@@ -2101,6 +2155,7 @@ static ErlNifFunc exla_funcs[] = {
   {"slice", 4, slice},
   {"dynamic_slice", 3, dynamic_slice},
   {"dynamic_update_slice", 3, dynamic_update_slice},
+  {"gather", 7, gather},
   // Tensor Creation
   {"rng_normal", 3, rng_normal},
   {"rng_uniform", 3, rng_uniform},
--- a/exla/lib/exla/op.ex
+++ b/exla/lib/exla/op.ex
@@ -236,6 +236,187 @@ defmodule EXLA.Op do
     %Op{builder: builder, ref: ref}
   end
 
+  @doc """
+  The XLA gather operation stitches together several slices
+  of an input array.
+
+  Note that this operation is extremely generic and far from
+  intuitive for regular usage. However, it can be used to implement
+  many specific operations that have to do with combining multiple
+  tensor slices.
+
+  ## Parameteres
+
+  The XLA docs are rather cryptic unless already understood,
+  so here's an attempt of a more intuitive description.
+
+  ### `index_vector_dim`
+
+  Determines which dimension contains index vectors. In most cases
+  we want to set this to the last dimension.
+
+      given
+        start_indices = [[0, 1], [1, 1]]
+      and given
+        index_vector_dim = 1
+      then
+        index vectors are [0, 1] and [1, 1]
+
+  Note that we can set this to `last_dimension + 1`, in which case
+  `start_indices` are implicitly reshaped to have a trailing dimension
+  of 1.
+
+      given
+        start_indices = [[0, 1], [1, 1]]
+      and given
+        index_vector_dim = 2
+      then
+        start_indices <- [[[0], [1]], [[1], [1]]]
+        index vectors are [0], [1], [1], [1]
+
+  ### `start_index_map`
+
+  Note: though given as a list, it can be treated as a map of `list_idx -> value`.
+
+  An index vector may have less elements than the operand tensor shape.
+  For example:
+
+      given
+        operand = [[1, 2], [3, 4]]
+        start_indices = [[1], [0]]
+        index_vector_dim = 1
+
+  As described above, in this case index vectors are `[1]`, `[0]` and they have
+  length 1. However, the operand has rank 2, so we need vectors of the form `[_, _]`
+  to point to a specific element in the operand. The `start_index_map` determines
+  where indices go into this template:
+
+      and given
+        start_index_map = [0] # effectively %{0 => 0}
+      then
+        actual index vectors are [1, _] and [0, _]
+
+      and given
+        start_index_map = [1] # effectively %{0 => 1}
+      then
+        actual index vectors are [_, 1] and [_, 0]
+
+  Finally, the missing elements (`_`) are assumed to be 0.
+
+  Complete examples:
+
+      given
+        operand = [[1, 2], [3, 4]]
+        start_indices = [[0], [1]]
+        index_vector_dim = 1
+      and given
+        start_index_map = [1] # effectively %{0 => 0, 1 => 1}
+      then
+        actual index vectors are [0, 0], [0, 1] (leading 0 is inserted)
+
+      given
+        operand = [[1, 2], [3, 4]]
+        start_indices = [[0, 1], [1, 1]]
+        index_vector_dim = 1
+      and given
+        start_index_map = [0, 1] # effectively %{0 => 0, 1 => 1}
+      then
+        actual index vectors are [0, 1], [1, 1] (as expected)
+
+      given
+        operand = [[1, 2], [3, 4]]
+        start_indices = [[0, 1], [1, 1]]
+        index_vector_dim = 1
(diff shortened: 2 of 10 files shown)


exla/lib/exla/op.ex
```diff
@@ -236,6 +236,187 @@ defmodule EXLA.Op do
...
+
+  Note that this operation is extremely generic and far from
+  intuitive for regular usage. However, it can be used to implement
+  many specific operations that have to do with combining multiple
+  tensor slices.
+
+  ## Parameteres
+
+  The XLA docs are rather cryptic unless already understood,
+  so here's an attempt of a more intuitive description.
```
jonatanklosko, line 251: Reading the docs for the first time gave me pretty much no idea how this operation works, so instead I looked at the usage across tensorflow and jax codebases and tried various combinations myself until I more or less figured out every parameter. After that the docs made more sense, though still not intuitive. Here I tried my best to describe how each parameter works based on examples, hopefully it helps if someone else wants implement another operation on top of gather.

josevalim (approved):
Beautifully done!

exla/c_src/exla/exla.cc
```diff
@@ -511,6 +511,60 @@ ERL_NIF_TERM dynamic_update_slice(ErlNifEnv* env, int argc, const ERL_NIF_TERM a
...
+
+  for (auto dim : offset_dims) {
+    dim_numbers.add_offset_dims(dim);
+  }
+  for (auto dim : collapsed_slice_dims) {
+    dim_numbers.add_collapsed_slice_dims(dim);
+  }
+  for (auto dim : start_index_map) {
+    dim_numbers.add_start_index_map(dim);
+  }
```
seanmor5, line 561: This is just a nit, but generally I've tried to move any XLA Proto object logic like Precision, ConvDimNos, etc. into `exla_nif_util.cc`.

seanmor5 (approved):
Everything looks beautiful!

exla/c_src/exla/exla.cc
jonatanklosko: Ah I see! One concern I have, this consequently changes the API to `Op.gather(op, start_indices, dimension_nums, slice_sizes)` where `dimension_nums` is a tuple and it's not immediately clear how this tuple should look like. Thoughts?

seanmor5: If it introduces unnecessary complexity then don't worry about making the change. I would just ship as is :D

jonatanklosko: To give another example:

https://github.com/elixir-nx/nx/blob/c8353de695a3d70dc6d518e0ab2ec832faa7df68/exla/lib/exla/defn.ex#L306-L320

Here we pass a `{contract_axes1, batch_axes1, contract_axes2, batch_axes2}` tuple, and to figure out what the tuple should be, we probably need to look into `exla::nif::get_dot_dimension_numbers`.

If there's some benefit I'm missing then I can certainly make the change, otherwise feel free to merge =)

nx/lib/nx/binary_backend.ex
```diff
@@ -1679,6 +1679,32 @@ defmodule Nx.BinaryBackend do
...
+    slices =
+      for <<bin::size(idx_size)-bitstring <- to_binary(indices)>> do
+        idx = binary_to_number(bin, indices.type)
+        slice_start = List.replace_at(slice_start, axis, idx)
+        Nx.slice(tensor, slice_start, slice_lengths)
+      end
+
+    slices
+    |> Nx.concatenate(axis: axis)
+    |> Nx.reshape(out.shape, names: out.names)
```
polvalente, line 1705: @josevalim I might be recalling this wrong, but I remember there being an issue with calling `Nx`-functions from binary backend. Is this the case?

@jonatanklosko if this is the case, you could just use `concatenate` directly from the binary backend (the signature is just below on line 1709).

Perhaps you could extract the `for` expression it uses to a defp, so you can avoid converting to and from binaries all the time

jonatanklosko: I was wondering about the usage of `Nx` functions, but we already use `Nx.transpose` and `Nx.pad` quite a lot in `BinaryBackend`, so I assumed it's fine :thinking: @josevalim?

> Perhaps you could extract the for expression it uses to a defp, so you can avoid converting to and from binaries all the time

The for loop in `take` goes over the indices tensor to produce a list of input tensor slices, then `concat` goes over these input tensor slices and combines them. These seem like two separate things, I may be missing something though :)

josevalim: You are correct, I think we changed to not call Nx from it but I don’t remember the reason. This is also related to having optional functions per backend. For example, we shouldn’t need to grad it if it is implemented as part of slice and friends. But likely there is a faster grad for it too.

polvalente: > The for loop in take goes over the indices tensor to produce a list of input tensor slices, then concat goes over these input tensor slices and combines them. These seem like two separate things, I may be missing something though :)

Making myself more clear, I was refering to the `for` in `concatenate`. If it's extracted to a defp, it would be easier to call `do_concatenate` with the binary representation directly, no need to build a tensor just to "unbuild" it  afterwards

jonatanklosko: Hmm, in order to call the backend implementations of `slice` and `concatenate` we would need to compute output shape, names and other arguments, effectively duplicating what we already do in the `Nx` calls.

jonatanklosko, line 1705: This does work, but is certainly less neat. I still wonder what's wrong with using `Nx` calls, we know the tensor backend will be `BinaryBackend` and the call will be dispatched here, so it should be effectively the same.

```suggestion
    slices =
      for <<bin::size(idx_size)-bitstring <- to_binary(indices)>> do
        idx = binary_to_number(bin, indices.type)
        slice_start = List.replace_at(slice_start, axis, idx)
        strides = List.duplicate(1, tensor_rank)
        slice_shape = List.to_tuple(slice_lengths)
        slice(%{tensor | shape: slice_shape}, tensor, slice_start, slice_lengths, strides)
      end

    concat_shape = put_elem(tensor.shape, axis, length(slices))
    result = concatenate(%{tensor | shape: concat_shape}, slices, axis)
    reshape(out, result, out.shape)
```

polvalente: If we're to refactor, I'd drop down to a lower level, manipulating the binary directly, instead of building tensors.
However, neither me nor @josevalim can seem to remember the reasoning to _not_ call Nx.
Since there isn't much in the way of efficiency loss here, I particularly see no problem in keeping your code as is.

jonatanklosko: FWIW there should be no overhead with packing/unpacking tensors and binaries, since they are kept as such.

polvalente: Yeah. I'll mark these comments as resolved :)

josevalim: @jonatanklosko yeah, the goal to drop to a lower level is that we wouldn’t compute the out tensor, we would simply call whatever slice call internally. If there is nothing internal, we should extract it out.

I think the answer to our question will depend on the grad for take. We can see the grad in Jax. If the grad is build on top of concatenate, then this should be left as is and we would try to move it upstream. However, if this is meant to have its own operation, with its own grad, I think making it independent of Nx is a better design.

josevalim: The gradient uses scatter_add, so I would like to see a version of this that doesn't call Nx for us to compare it with. :)

jonatanklosko: Alright, will do! Also, are the jax gradient rules in [lax.py](https://github.com/google/jax/blob/main/jax/_src/lax/lax.py) or somewhere else? The only relevant thing I found was [gather gradient](https://github.com/google/jax/blob/main/jax/_src/lax/lax.py#L4581-L4585), which would suggest gradient is computed on a lower level, but this snippet doesn't look like an actual implementation.

josevalim: @jonatanklosko it is not always straightforward to find but my best guess is that take is implemented on top of gather, yeah.

```
>>> jax.make_jaxpr(jax.grad(lambda x: jnp.sum(jnp.take(x, [0, 1, 2]))))(jnp.array([[1.0, 2, 3], [4, 5, 6]]))
{ lambda a ; b.
  let c = reshape[ dimensions=None
                   new_sizes=(6,) ] b
      d = broadcast_in_dim[ broadcast_dimensions=(0,)
                            shape=(3, 1) ] a
      e = gather[ dimension_numbers=GatherDimensionNumbers(offset_dims=(), collapsed_slice_dims=(0,), start_index_map=(0,))
                  slice_sizes=(1,) ] c d
      _ = reduce_sum[ axes=(0,) ] e
      f = broadcast_in_dim[ broadcast_dimensions=(  )
                            shape=(3,) ] 1.0
      g = broadcast_in_dim[ broadcast_dimensions=(  )
                            shape=(  ) ] 0.0
      h = broadcast_in_dim[ broadcast_dimensions=(  )
                            shape=(6,) ] g
      i = scatter-add[ dimension_numbers=ScatterDimensionNumbers(update_window_dims=(), inserted_window_dims=(0,), scatter_dims_to_operand_dims=(0,))
                       indices_are_sorted=False
                       unique_indices=False
                       update_consts=(  )
                       update_jaxpr={ lambda  ; a b.
                                      let c = add a b
                                      in (c,) } ] h d f
      j = reshape[ dimensions=None
                   new_sizes=(2, 3) ] i
  in (j,) }
```

jonatanklosko: The `take` implementation is [here](https://github.com/google/jax/blob/f7009ed1072bf48ea6ba9d6d2ce573ce0ce13377/jax/_src/numpy/lax_numpy.py#L4779-L4780) and it definitely uses gather, I wasn't sure if the grad is implemented for take separately or rather a generic one for `lax.gather`.

jonatanklosko: @josevalim I refactored `slice` and `concatenate` into `do_slice` and `do_concatenate` respectively and rewrote take in terms of these.

polvalente (approved):

nx/lib/nx/binary_backend.ex
polvalente: Looking good. It ended up not being much more verbose than the Nx-calls version

jonatanklosko: Yup, that was a good call! :tea:

nx/lib/nx.ex
```diff
@@ -7518,6 +7522,150 @@ defmodule Nx do
...
+  @doc """
+  Takes and concatenates slices along an axis.
+
+  Intuitively speaking, `take/3` reorders tensor slices along
+  the given axis based on the given indices, possibly duplicating
+  and removing slices.
+
+  Passing a multi-dimensional indices tensor only affects the
+  resulting shape. Specifically, the given axis in the input shape
+  gets replaced with the indices shape.
```
josevalim, line 7534: What happens if a negative index is given?

nx/lib/nx/binary_backend.ex
```diff
@@ -1679,27 +1681,64 @@ defmodule Nx.BinaryBackend do
...
   @impl true
   def concatenate(out, tensors, axis) do
     %{shape: output_shape, type: {_, size} = output_type} = out
+
+    tensors
+    |> Enum.map(fn %{shape: shape} = t ->
+      t = as_type(%{t | type: output_type}, t)
+      {to_binary(t), shape}
+    end)
+    |> do_concatenate(size, axis, output_shape)
```
josevalim, line 1724: I think the convention in this module is to call them bin_concatenate (and bin_slice).

nx/lib/nx.ex
jonatanklosko: `BinaryBackend` uses slice so it clamps, but apparently so does xla::Gather. This isn't explicitly stated in the docs though. On the other hand `torch.index_select` raises an out of range error. Given the indices are dynamic this is pretty backend specific, do we want to impose some behaviour or leave it as undefined?

josevalim: Let’s raise in the binary backend then. If the other backends support it, it is fine.

jonatanklosko: Sounds good!

PR #439, opened 2021-08-19
Add streaming to EXLA.Defn

josevalim:


exla/lib/exla/defn/stream.ex
```diff
@@ -0,0 +1,56 @@
...
+        #{inspect(send)}
+
+        But got tensor:
+
+        #{inspect(data)}
+        """
+      end
+
+      pred = EXLA.Shape.make_shape({:pred, 8}, {})
+      :ok = EXLA.Client.to_infeed(executable.client, executable.device_id, <<0::8-native>>, pred)
```
polvalente, line 22: shouldn't this hard matches qualify for a `send!` function name instead of `send`? Where is the line drawn?

exla/lib/exla/defn.ex
```diff
@@ -56,69 +56,227 @@ defmodule EXLA.Defn do
...
+    run_options = Keyword.put(run_options, :keep_on_device, true)
+
+    # The input vars should not be converted to buffers as they come from infeed
+    input_vars = Nx.Defn.Tree.flatten_list([input])
+    input_shape = EXLA.Shape.make_tuple_shape(Enum.map(input_vars, &nx_to_shape/1))
+    acc_vars = Nx.Defn.Tree.flatten_list([acc])
+    split_fun = &split_stream(&1, &2, length(input_vars), length(acc_vars))
+
+    comp_fun =
+      &to_stream_computation(key, input_vars, input_shape, acc_vars, &1, &2, compile_options)
```
polvalente, line 72: perhaps `fn ->` syntax here would be more clear on which two arguments this function expects

exla/lib/exla/defn.ex
```diff
@@ -56,69 +56,227 @@ defmodule EXLA.Defn do
...
+    infeed_shape = EXLA.Shape.make_tuple_shape([flag_shape, token_shape])
+
+    arg_shape = EXLA.Shape.make_tuple_shape([infeed_shape, acc_shape, constant_shape])
+
+    cond_b = EXLA.Builder.new(builder, "while-cond-" <> inspected_key)
+    param = EXLA.Op.parameter(cond_b, 0, arg_shape, "arg")
+    infeed = EXLA.Op.get_tuple_element(param, 0)
+    flag = EXLA.Op.get_tuple_element(infeed, 0)
+    cond_op = EXLA.Op.equal(flag, EXLA.Op.constant_r0(cond_b, 1, {:pred, 8}))
+    cond = EXLA.Builder.build(cond_op)
```
polvalente, line 132:
```suggestion
    predicate = EXLA.Builder.build(cond_op)
```

To avoid using an stdlib function name here

nx/lib/nx.ex
```diff
@@ -1217,6 +1200,25 @@ defmodule Nx do
     raise ArgumentError, "expected a %Nx.Tensor{} or a number, got: #{inspect(t)}"
   end
 
+  defp to_tensor(%T{} = tensor, fun) do
```
polvalente, line 1203:
```suggestion
  defp to_tensor(%T{} = tensor, fun) when is_function(fun, 1) do
```

The same would apply to the other clauses

exla/lib/exla/defn/stream.ex
```diff
@@ -0,0 +1,56 @@
...
+    defp nx_to_binary(%Nx.Tensor{} = tensor) do
+      [tensor |> Nx.backend_transfer() |> Nx.to_binary()]
+    end
+
+    defp nx_to_binary(map) when is_map(map) do
+      map |> Enum.sort() |> Enum.flat_map(fn {_, v} -> nx_to_binary(v) end)
+    end
+
+    defp nx_to_binary(tuple) when is_tuple(tuple) do
+      Enum.flat_map(tuple, &nx_to_binary/1)
```
polvalente, line 38:
```suggestion
      tuple |> Tuple.to_list() |> Enum.flat_map(&nx_to_binary/1)
```

exla/lib/exla/defn/stream.ex
josevalim: Most of the time a bang function exists as an alternative to a non bang variant. In this case, they always succeed because when they fail it is usually due to an argument error or something intractable.

nx/lib/nx.ex
josevalim: I think it is fine since it is private API although I would definitely add it if it was public API. :)

exla/lib/exla/defn.ex
josevalim: Done!

exla/lib/exla/defn/stream.ex
polvalente: I see! This makes everything much more clear!

josevalim:

PR #440, opened 2021-08-20
Build XLA tar

seanmor5:
Moving towards shipping precompiled dependencies to drastically lower NIF build times.

With this version, we'd build and release archives for different architectures and accelerators. Compilation of the NIF would take place on the target machine. Currently there are a few issues:

1) The install headers rule includes a LOT of headers, and I'm not really sure how licensing works out if we are to package them all into our repo.

2) The new build process requires an additional `libprotobuf-dev` dependency to be installed. It's available on most package managers, but I'm currently troubleshooting issues with protobuf includes not being found in include paths.

3) I'm not sure exactly how we should differentiate between the dev build process and regular NIF build process

### DIFF
--- a/exla/Makefile
+++ b/exla/Makefile
@@ -28,13 +28,24 @@ TENSORFLOW_DIR = $(EXLA_CACHE)/$(TENSORFLOW_NS)/erts-$(ERTS_VERSION)
 TENSORFLOW_EXLA_NS = tensorflow/compiler/xla/exla
 TENSORFLOW_EXLA_DIR = $(TENSORFLOW_DIR)/$(TENSORFLOW_EXLA_NS)
 
-$(EXLA_SO): $(TENSORFLOW_DIR)/bazel-bin/$(TENSORFLOW_EXLA_NS)/libexla.so
-	mkdir -p priv
-	cp -f $(TENSORFLOW_DIR)/bazel-bin/$(TENSORFLOW_EXLA_NS)/libexla.so $(EXLA_SO)
+XLA_EXTENSION_LIB = /home/sean/projects/xla_extension/lib
+XLA_INCLUDE_PATH = /home/sean/projects/xla_extension/include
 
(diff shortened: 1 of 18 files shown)


seanmor5:
Of note, for contributors, we need a way to:

- Build the `xla_extension` target
- Extract the extension to some directory
- Target the extension `lib` and `include` directories

In my current environment, I extracted `xla_extension` to `/usr/local/` and moved `libxla_extension.so` to `/usr/local/xla_extension/lib`

josevalim:
Exciting work!

> In my current environment, I extracted xla_extension to /usr/local/ and moved libxla_extension.so to /usr/local/xla_extension/lib

I would keep them in a separate directory for now, either create a exla/xla_src directory or a new directory at the root called xla. And then soon we will find a way to get those precompiled and installed via Hex.pm? :) But without needing a system install.

joelberkeley:
Would this tar be suitable for other projects? For example, I'm trying, very slowly, to get a minimal XLA library together for use in my project.

seanmor5:
@joelberkeley Our goal with this PR is to work towards having a separate repository which packages a shared object and the relevant headers into a tarball which can be downloaded and linked to a separate project without needing to compile XLA inside the TensorFlow monorepo - this will hopefully save a lot of time and effort.

As far as your question, we plan to maintain this specifically for our project; however, it will probably include all of the dependencies and headers you need to include in your own project so it would probably be suitable enough for you to fork for your own work. I haven't validated this approach works yet as I've been running in to some dependency issues with GRPC.

If you are having problems building XLA for your project, you can feel free to reach out to me. It took me awhile to get XLA to build and integrate with Elixir because of how tightly it's integrated with TensorFlow, but I can probably point you in the right direction for your project

joelberkeley:
Thanks @seanmor5 that's really helpful. I'll wait to see how this work goes, then may reach out to you after, with the email address on your profile

seanmor5:
Progress has been moved to: https://github.com/elixir-nx/xla/pull/1

PR #447, opened 2021-08-30
feat: add take_along_axis (with binary backend impl)

polvalente:
This PR adds `Nx.take_along_axis/3`, which makes `Nx.argsort/3` more
useful, allows for more flexible indexing of tensors and
enables the grad implementation for `Nx.sort/2`

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7666,6 +7666,19 @@ defmodule Nx do
     impl!(tensor).take(%{tensor | shape: shape, names: names}, tensor, indices, axis)
   end
 
+  @doc """
+
+  ## Examples
+
+      iex> Nx.take_along_axis(Nx.tensor([[1, 2, 3], [4, 5, 6]]), Nx.tensor([[0, 0, 2, 2, 1, 1], [2, 2, 1, 1, 0, 0]]), axis: 1)
+      #Nx.Tensor<
+        s64[2][6]
+        [
+          [1, 1, 3, 3, 2, 2],
+          [6, 6, 5, 5, 4, 4]
+        ]
+      >
+  """
   def take_along_axis(tensor, indices, opts \\ []) when is_list(opts) do
     tensor = to_tensor(tensor)
     indices = to_tensor(indices)
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7702,8 +7702,8 @@ defmodule Nx do
         ]
       >
 
-    The indices returned from `Nx.argsort/2` can be used with `Nx.take_along_axis/3` to
-    produce the sorted tensor (or to sort more tensors according to the same criteria).
+  The indices returned from `Nx.argsort/2` can be used with `Nx.take_along_axis/3` to
+  produce the sorted tensor (or to sort more tensors according to the same criteria).
 
       iex> tensor = Nx.tensor([[[1, 2], [3, 4], [5, 6]]])
       #Nx.Tensor<
@@ -7761,7 +7761,7 @@ defmodule Nx do
         ]
       >
 
-    ### Error cases
+  ### Error cases
 
       iex> tensor = Nx.iota({3, 3})
       iex> idx = Nx.tensor([[2.0], [1.0], [2.0]], type: {:f, 32})
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7678,6 +7678,54 @@ defmodule Nx do
           [6, 6, 5, 5, 4, 4]
         ]
       >
+
+      iex> Nx.take_along_axis(Nx.tensor([[1, 2, 3], [4, 5, 6]]), Nx.tensor([[0, 1, 1], [1, 0, 0], [0, 1, 0]]), axis: 0)
+      #Nx.Tensor<
+        s64[3][3]
+        [
+          [1, 5, 6],
+          [4, 2, 3],
+          [1, 5, 3]
+        ]
+      >
+
+    The indices returned from `Nx.argsort/2` can be used with `Nx.take_along_axis/3` to
+    produce the sorted tensor (or to sort more tensors according to the same criteria).
+
+      iex> tensor = Nx.tensor([[[1, 2], [3, 4], [5, 6]]])
+      #Nx.Tensor<
+        s64[1][3][2]
+        [
+          [
+            [1, 2],
+            [3, 4],
+            [5, 6]
+          ]
+        ]
+      >
+      iex> idx = Nx.argsort(tensor, axis: 2, direction: :desc)
+      #Nx.Tensor<
+        s64[1][3][2]
+        [
+          [
+            [1, 0],
+            [1, 0],
+            [1, 0]
+          ]
+        ]
+      >
+      iex> Nx.take_along_axis(tensor, idx, axis: 2)
+      #Nx.Tensor<
+        s64[1][3][2]
+        [
+          [
+            [2, 1],
+            [4, 3],
+            [6, 5]
+          ]
+        ]
+      >
+
   """
   def take_along_axis(tensor, indices, opts \\ []) when is_list(opts) do
     tensor = to_tensor(tensor)
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7667,6 +7667,19 @@ defmodule Nx do
   end
 
   @doc """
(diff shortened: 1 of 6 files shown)


polvalente:
Still work in progress because I want to add more tests. Otherwise, feedback is welcome already!

nx/lib/nx/shape.ex
```diff
@@ -1192,6 +1192,72 @@ defmodule Nx.Shape do
...
+
+  ### Error cases
+
+      iex> Nx.Shape.take_along_axis({2, 3, 4}, {3, 10, 4}, 1)
+      ** (ArgumentError) non-indexing dimensions must match. Expected {2, *, 4}, got: {3, 10, 4}
+
+      iex> Nx.Shape.take_along_axis({2, 3}, {1, 2, 3}, 0)
+      ** (ArgumentError) shapes must have the same number of dimensions. Expected {*, 3}, got: {1, 2, 3}
+  """
+  def take_along_axis(shape, indices_shape, axis) do
```
polvalente, line 1219: @josevalim I'm thinking that we should just keep the input names in the output, since we keep the same overall dimensions in this implementation. WDYT?

nx/lib/nx.ex
```diff
@@ -7666,6 +7666,96 @@ defmodule Nx do
...
+
+      iex> Nx.take_along_axis(Nx.tensor([[1, 2, 3], [4, 5, 6]]), Nx.tensor([[0, 1, 1], [1, 0, 0], [0, 1, 0]]), axis: 0)
+      #Nx.Tensor<
+        s64[3][3]
+        [
+          [1, 5, 6],
+          [4, 2, 3],
+          [1, 5, 3]
+        ]
+      >
```
josevalim, line 7703:
```suggestion
      >
```

nx/lib/nx/shape.ex
josevalim: Agreed!

nx/lib/nx/shape.ex
```diff
@@ -1192,6 +1192,72 @@ defmodule Nx.Shape do
...
+    if axis >= tuple_size(shape) or axis < 0 do
+      raise ArgumentError, "axis must be a non-negative less than #{tuple_size(shape)}"
+    end
+
+    shape_list = Tuple.to_list(shape)
+
+    shape_template =
+      shape_list
+      |> Enum.with_index()
+      |> Enum.map(fn
```
josevalim, line 1229: You can merge with_index and map into a single operation since Elixir v1.12 by passing a function to with_index.

nx/lib/nx/shape.ex
```diff
@@ -1192,6 +1192,72 @@ defmodule Nx.Shape do
...
+      raise ArgumentError,
+            "shapes must have the same number of dimensions. Expected #{shape_template}, got: #{inspect(indices_shape)}"
+    end
+
+    indices_shape_list = Tuple.to_list(indices_shape)
+
+    [shape_list, indices_shape_list]
+    |> Enum.zip()
+    |> Enum.with_index()
+    |> Enum.each(fn
```
josevalim, line 1247: You can merge with_index and each into a single operation since Elixir v1.12 by passing a function to with_index.

josevalim (commented):
Looking absolutely great so far!

nx/lib/nx/shape.ex
polvalente: :exploding_head:

nx/lib/nx.ex
```diff
@@ -7666,6 +7666,124 @@ defmodule Nx do
...
+    unless Nx.Type.integer?(indices.type) do
+      raise ArgumentError, "indices must be an integer tensor, got #{inspect(indices.type)}"
+    end
+
+    opts = keyword!(opts, axis: 0)
+    axis = Nx.Shape.normalize_axis(tensor.shape, opts[:axis], tensor.names)
+
+    shape = Nx.Shape.take_along_axis(tensor.shape, indices.shape, axis)
+
+    impl!(tensor).take_along_axis(%{tensor | shape: shape}, tensor, indices, axis)
```
seanmor5, line 7784: I saw the earlier discussion with @josevalim, I would dispute simply keeping input names as we are taking on the shape of `indices` - so with the current logic we take the names of `tensor` but the shape of `indices`, which feels inconsistent.

polvalente: The number of axes remains the same. The only thing that changes is the length along the axis on which we're operating.
I suggested keeping the names because they are related to the axes regardless of length

nx/lib/nx.ex
```diff
@@ -7666,6 +7666,124 @@ defmodule Nx do
...
+          [1, 5, 6],
+          [4, 2, 3],
+          [1, 5, 3]
+        ]
+      >
+
+    The indices returned from `Nx.argsort/2` can be used with `Nx.take_along_axis/3` to
+    produce the sorted tensor (or to sort more tensors according to the same criteria).
+
+      iex> tensor = Nx.tensor([[[1, 2], [3, 4], [5, 6]]])
```
josevalim, line 7708: Those examples are not indented four spaces. :)

nx/lib/nx/binary_backend.ex
```diff
@@ -1726,6 +1726,55 @@ defmodule Nx.BinaryBackend do
...
+        %T{shape: t_shape, type: {_, t_size} = t_type} = tensor,
+        %T{shape: idx_shape, type: {_, idx_size} = idx_type} = indices,
+        axis
+      ) do
+    permute = fn t ->
+      permutation =
+        t
+        |> Nx.axes()
+        |> List.delete(axis)
+        |> List.insert_at(Nx.rank(t) - 1, axis)
```
josevalim, line 1741: Maybe you want to use List.update_at?

nx/lib/nx/binary_backend.ex
```diff
@@ -1726,6 +1726,55 @@ defmodule Nx.BinaryBackend do
...
+        |> Enum.sort_by(fn {x, _} -> x end)
+        |> Enum.map(fn {_, i} -> i end)
+
+      shape_list = Tuple.to_list(output.shape)
+      permuted_shape = permutation |> Enum.map(&Enum.at(shape_list, &1)) |> List.to_tuple()
+
+      {permuted_shape, inverse_permutation}
+    end
+
+    {permuted_shape, inverse_permutation} = permute.(tensor)
```
josevalim, line 1755: Does this need to be an anonymous function if you only invoke it once? :D Or is it for code organization?

nx/lib/nx/shape.ex
```diff
@@ -1192,6 +1192,70 @@ defmodule Nx.Shape do
...
+
+    shape_template =
+      shape_list
+      |> Enum.with_index(fn
+        _, ^axis -> "*"
+        l, _ -> "#{l}"
+      end)
+      |> Enum.join(", ")
+
+    shape_template = "{#{shape_template}}"
```
josevalim, line 1234: We want to compute this only if there is an error, so I would move it to a private function and call it inside the conditionals.

josevalim (approved):
I did one last pass and noted some minor details. Barring that, ship it!

nx/lib/nx.ex
polvalente: HAHA

I'm almost making a VSCode extension just to warn me about those :stuck_out_tongue:

nx/lib/nx/binary_backend.ex
polvalente: This effectively shifts the list left, so I don't see how update_at could be applied easily

```elixir
iex(1)> [1, 2, 3, 4] |> List.delete(2) |> List.insert_at(4 - 1, 2)
[1, 3, 4, 2]
```

nx/lib/nx/binary_backend.ex
polvalente: It doesn't!

In the first iteration this would have been called more than once, so it made sense. Now I can just move it to the function body

nx/lib/nx.ex
polvalente: So, they actually are indented 4 spaces, but the paragraph above is indented at 2 spaces. The same with the `### Error cases` header. Should those be at 0 or 4 spaces instead?

josevalim: Ah, the paragragh is the one wrong, yeah. No indentation for paragraphs and section headers.

PR #449, opened 2021-09-06
feat: add take along axis EXLA implementation

polvalente:
This is a continuation of #447, adding the missing EXLA implementation

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -745,13 +745,14 @@ defmodule EXLA.Defn do
     collapsed_slice_dims = Enum.to_list(axes_range)
     start_index_map = Enum.to_list(axes_range)
 
+    indices_exla_shape = EXLA.Op.get_shape(indices)
+
     iotas =
       Enum.map(axes_range, fn axis ->
-        shape = EXLA.Op.get_shape(indices)
-        EXLA.Op.iota(state.builder, shape, axis)
+        EXLA.Op.iota(state.builder, indices_exla_shape, axis)
       end)
 
-    new_axis_shape = Tuple.insert_at(indices_shape, indices_rank, 1)
+    new_axis_shape = Tuple.append(indices_shape, indices_rank)
 
     indices =
       iotas
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -740,7 +740,7 @@ defmodule EXLA.Defn do
     axes_range = 0..(indices_rank-1)
 
     index_vector_dim = indices_rank
-    slice_sizes = 1 |> List.duplicate(indices_rank)
+    slice_sizes = List.duplicate(1, indices_rank)
     offset_dims = []
     collapsed_slice_dims = Enum.to_list(axes_range)
     start_index_map = Enum.to_list(axes_range)
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -733,6 +733,46 @@ defmodule EXLA.Defn do
     )
   end
 
+   defp to_operator(:take_along_axis, [tensor, indices, axis], _ans, state) do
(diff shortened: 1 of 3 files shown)


exla/lib/exla/defn.ex
```diff
@@ -733,6 +733,43 @@ defmodule EXLA.Defn do
...
+
+    index_vector_dim = indices_rank
+    slice_sizes = 1 |> List.duplicate(indices_rank)
+    offset_dims = []
+    collapsed_slice_dims = Enum.to_list(axes_range)
+    start_index_map = Enum.to_list(axes_range)
+
+    iotas =
+      Enum.map(axes_range, fn axis ->
+        shape = EXLA.Op.get_shape(indices)
```
jonatanklosko, line 750: Not sure if this makes a difference, but this could be extracted and used for every iteration.

exla/lib/exla/defn.ex
```diff
@@ -733,6 +733,43 @@ defmodule EXLA.Defn do
...
+        EXLA.Op.iota(state.builder, shape, axis)
+      end)
+
+    new_axis_shape = Tuple.insert_at(indices_shape, indices_rank, 1)
```
jonatanklosko, line 754: Probably makes the intent a bit more clear, but whatever you like best!

```suggestion
    new_axis_shape = Tuple.append(indices_shape, 1)
```

jonatanklosko (commented):
Minor comments, but looks good to me!

exla/lib/exla/defn.ex
polvalente: This is better! I didn't know the function :)

polvalente: @jonatanklosko I've made the suggested changes and I've also added a sanity test with argsort :)

jonatanklosko: Great idea :cowboy_hat_face:

exla/lib/exla/defn.ex
```diff
@@ -733,6 +733,44 @@ defmodule EXLA.Defn do
...
   end
 
+  defp to_operator(:take_along_axis, [tensor, indices, axis], _ans, state) do
+    indices_shape = op_shape(indices)
+    indices_rank = tuple_size(indices_shape)
+
+    axes_range = 0..(indices_rank-1)
+
+    index_vector_dim = indices_rank
+    slice_sizes = 1 |> List.duplicate(indices_rank)
```
josevalim, line 743: You shall never pipe once. :)

```suggestion
    slice_sizes = List.duplicate(1, indices_rank)
```

exla/lib/exla/defn.ex
```diff
@@ -733,6 +733,44 @@ defmodule EXLA.Defn do
     )
   end
 
+  defp to_operator(:take_along_axis, [tensor, indices, axis], _ans, state) do
+    indices_shape = op_shape(indices)
+    indices_rank = tuple_size(indices_shape)
+
+    axes_range = 0..(indices_rank-1)
```
josevalim, line 740: Can the rank be zero? If so, this needs an explicit step:

```suggestion
    axes_range = 0..(indices_rank-1)//1
```

josevalim (approved):
Beautiful, just two minor comments!

exla/lib/exla/defn.ex
polvalente: Haha! There's a credo rule for that :rofl:

polvalente: PS: this was just remaining from a refactor, I would never knowingly commit such an atrocity

exla/lib/exla/defn.ex
polvalente: Good question, let me check

polvalente: Nope, there's a validation for this in Nx.Shape.
Also, it would be just the same as `Nx.concatenate([tensor ...])`, so we don't lose anything by not allowing this

PR #450, opened 2021-09-08
feat: add sort grad

polvalente:
Adds the grad for `Nx.sort`

### DIFF
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -499,6 +499,14 @@ defmodule Nx.Defn.Grad do
     to_grad(input, dl, cache)
   end
 
+  defp grad(:sort, [t, opts], _ans, g, cache) do
+    idx = Nx.argsort(t, opts)
+    take_along_opts = Keyword.take(opts, [:axis])
+    g = Nx.take_along_axis(g, idx, take_along_opts)
+
+    to_grad(t, g, cache)
+  end
+
   defp grad(_op, _args, _ans, _g, _cache) do
     :none
   end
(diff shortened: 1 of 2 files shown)


polvalente:
Actually, I just realized I added no `:axis` option tests. I'll convert this back to a draft :sweat_smile:

josevalim (approved):
So the grad grad of sort is an error or is it zero? Is that in accordance to other tools?

seanmor5 (approved):
Nice! I guess the constant gradient is consistent with both `argmax` and `argmin`

nx/lib/nx/defn/grad.ex
```diff
@@ -766,7 +774,7 @@ defmodule Nx.Defn.Grad do
                [:logical_and, :logical_or, :logical_xor, :logical_not] ++
                [:left_shift, :right_shift, :count_leading_zeros, :population_count] ++
                [:floor, :round, :ceil, :sign] ++
-               [:equal, :greater, :greater_equal, :less, :less_equal, :not_equal]
+               [:equal, :greater, :greater_equal, :less, :less_equal, :not_equal, :argsort]
```
polvalente, line 777: @josevalim argsort returns a new tensor of constants (the indices corresponding to the sorted positions), which is why it's grad is zero. This is in accordance with Jax.

polvalente:
> So the grad grad of sort is an error or is it zero? Is that in accordance to other tools?

Oh, I think I misread your question.
For now, rhe second order grad raises due to missing grad for take along axis, but this function does have a grad definition which we can implement

PR #451, opened 2021-09-11
Update build process to use installed XLA extension

seanmor5:
This shouldn't be merged until we have everything worked out in elixir-nx/xla obviously, but for the impatient and for making sure this new build process is sound this PR updates everything to use an installed XLA extension - assuming it is installed in `priv`.

For Linux users you can basically pull from this branch and do:

```
$ cd priv
$ wget <XLA URL>
$ tar -xvf <XLA Tar>
$ cd ..
$ mix test
```

Where XLA URL is one of the tarballs from here: https://github.com/jonatanklosko/xla/releases/tag/v0.1.0-dev

And it should build EXLA in about 30-45 seconds :D


exla/mix.exs
```diff
@@ -33,7 +33,9 @@ defmodule EXLA.MixProject do
   defp deps do
     [
       {:nx, path: "../nx"},
-      {:elixir_make, "~> 0.6"},
+      # TODO: adjust
+      {:xla, "~> 0.1.1-dev", runtime: false, github: "jonatanklosko/xla", branch: "jk-build"},
```
jonatanklosko, line 37: Note: once the build succeeds we can update this reference and should be ready to go :)

jonatanklosko:
I did some adjustments in #454, let me know if it makes sense :)

seanmor5:
I am going to merge this to the main branch

PR #454, opened 2021-09-16
Finalize XLA integration

jonatanklosko:
Ref #451.

### DIFF
--- a/exla/Makefile
+++ b/exla/Makefile
@@ -8,19 +8,25 @@ XLA_EXTENSION_LIB = $(XLA_EXTENSION_DIR)/lib
 XLA_INCLUDE_PATH = $(XLA_EXTENSION_DIR)/include
 
 # Private configuration
+EXLA_DIR = c_src/exla
 PRIV_DIR = priv
 EXLA_SO = $(PRIV_DIR)/libexla.so
-EXLA_DIR = c_src/exla
+EXLA_LIB_DIR = $(PRIV_DIR)/lib
 
 # Build Flags
 CFLAGS = -fPIC -I$(ERTS_INCLUDE_DIR) -isystem $(XLA_INCLUDE_PATH) -O3 -Wall -Wextra \
 				 -Wno-unused-parameter -Wno-missing-field-initializers -Wno-comment \
 				 -shared -std=c++14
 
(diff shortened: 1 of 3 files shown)


exla/mix.exs
```diff
@@ -33,8 +33,7 @@ defmodule EXLA.MixProject do
   defp deps do
     [
       {:nx, path: "../nx"},
-      # TODO: adjust
-      {:xla, "~> 0.1.1-dev", runtime: false, github: "jonatanklosko/xla", branch: "jk-build"},
+      {:xla, "~> 0.1.0", runtime: false, github: "elixir-nx/xla"},
```
jonatanklosko, line 36: I think it actually makes sense to point to GitHub for now, so if we need any fixes to the Elixir code, we can do so quickly. We can publish to hex at any time.

exla/Makefile
```diff
@@ -8,19 +8,25 @@ XLA_EXTENSION_LIB = $(XLA_EXTENSION_DIR)/lib
...
 CFLAGS = -fPIC -I$(ERTS_INCLUDE_DIR) -isystem $(XLA_INCLUDE_PATH) -O3 -Wall -Wextra \
 				 -Wno-unused-parameter -Wno-missing-field-initializers -Wno-comment \
 				 -shared -std=c++14
 
-LDFLAGS = -L$(XLA_EXTENSION_LIB) -Wl,-rpath,$(XLA_EXTENSION_LIB) -lxla_extension
+# Use a relative RPATH, so at runtime libexla.so looks for libxla_extension.so
+# in ./lib regardless of the absolute location. This way priv can be safely
+# packed into an Elixir release. Also, we use $$ to escape Makefile variable
+# and single quotes to escape shell variable
+LDFLAGS = -L$(EXLA_LIB_DIR) -Wl,-rpath,'$$ORIGIN/lib' -lxla_extension
```
jonatanklosko, line 25: I realised that we need `libxla_extension.so` in the `priv` directory too and also that we shouldn't use an absolute RPATH, because then it wouldn't work when moved to a release archive. So now we link `priv/lib` to `cache/xla_extension/lib` and use a relative RPATH. Note: we currently link the lib directory to save on some space and it should be copied into the release normally anyway. If we run into any issues with that we can just copy directly.

seanmor5 (approved):
This makes sense and looks good to me!

jonatanklosko:
(Feel free to merge since I can't :))

PR #455, opened 2021-09-16
Point XLA to Hex

jonatanklosko:
As long as we point `:xla` to GitHub it needs to be a specific version, so that we can build a new XLA release before it gets "public".

### DIFF
--- a/exla/mix.exs
+++ b/exla/mix.exs
@@ -33,7 +33,7 @@ defmodule EXLA.MixProject do
   defp deps do
     [
       {:nx, path: "../nx"},
-      {:xla, "0.1.0", runtime: false, github: "elixir-nx/xla"},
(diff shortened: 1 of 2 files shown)


jonatanklosko:
Alternatively we can already publish xla to hex, so that we don't need to manually bump the version, WDYT?

seanmor5 (approved):

seanmor5:
I don't see any issues with publishing to Hex, we can keep republishing as issues arise :)

jonatanklosko:
Yeah, the only difference is that with github we can push changes to `mix.exs` without actually publishing a new version. That said, this part seems to work just fine, so just gonna publish to hex and update this PR :)

jonatanklosko:
Updated, feel free to merge :)

josevalim:

PR #456, opened 2021-09-16
Implement gather

jonatanklosko:
Implements "Group 3" outlined in #433. Essentially this allows for building a new tensor from arbitrary individual elements from the original tensor.

As pointed out in #433, *gather* is not exactly an unambiguous term, but seems like the most appropriate name for this operation, especially that this should pair with how scatter works.

### DIFF
--- a/exla/test/exla/defn_expr_test.exs
+++ b/exla/test/exla/defn_expr_test.exs
@@ -2569,7 +2569,7 @@ defmodule EXLA.DefnExprTest do
   end
 
   describe "gather" do
-    defn(gather(t, idx), do: Nx.gather(t, idx))
+    defn gather(t, idx), do: Nx.gather(t, idx)
 
     test "1d result" do
       assert gather(Nx.tensor([[1, 2], [3, 4]]), Nx.tensor([[1, 1], [0, 1], [1, 0]])) ==
--- a/exla/test/exla/defn_expr_test.exs
+++ b/exla/test/exla/defn_expr_test.exs
@@ -2568,6 +2568,26 @@ defmodule EXLA.DefnExprTest do
     end
   end
 
+  describe "gather" do
+    defn(gather(t, idx), do: Nx.gather(t, idx))
+
+    test "1d result" do
+      assert gather(Nx.tensor([[1, 2], [3, 4]]), Nx.tensor([[1, 1], [0, 1], [1, 0]])) ==
+               Nx.tensor([4, 2, 3])
+
+      assert gather(
+               Nx.tensor([[[1, 2], [11, 12]], [[101, 102], [111, 112]]]),
+               Nx.tensor([[0, 0, 0], [0, 1, 1], [1, 1, 1]])
+             ) ==
+               Nx.tensor([1, 12, 112])
+    end
+
+    test "2d indices" do
+      assert gather(Nx.tensor([[1, 2], [3, 4]]), Nx.tensor([[[1, 1], [0, 0]], [[1, 0], [0, 1]]])) ==
+               Nx.tensor([[4, 1], [3, 2]])
+    end
+  end
+
   describe "reverse" do
     defn reverse(t), do: Nx.reverse(t)
     defn reverse1(t), do: Nx.reverse(t, axes: [1])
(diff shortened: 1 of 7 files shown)


nx/lib/nx.ex
```diff
@@ -7784,6 +7784,54 @@ defmodule Nx do
     impl!(tensor).take_along_axis(%{tensor | shape: shape}, tensor, indices, axis)
   end
 
+  @doc """
```
polvalente, line 7787: This looks good! I'm not too sure if this is an API we want to expose just yet, but I can see this being useful for a lot of use-cases.

@josevalim WDYT?

polvalente: This could be the underlying function for:

```elixir
t = Nx.tensor([[1, 2], [3, 4]])
i = Nx.tensor([[0, 1], [0, 0], [1, 1]])

t[i] == Nx.tensor([2, 1, 4])
```

polvalente: Yeah, if you change:

```elixir
def fetch(tensor, %Nx.Tensor{} = index), do: {:ok, fetch_axes(tensor, [{0, index}])}
```

to 

```elixir
def fetch(%Nx.Tensor{} = tensor,%Nx.Tensor{} = index), do: {:ok, Nx.gather(tensor, index)}
```

We can have the behaviour as I described above. Of course, with arbitrary shapes in the output:

```elixir
iex(1)> t = Nx.iota({2, 3, 4})
iex(10)> t[Nx.tensor([[[[[0, 0, 0], [0, 0, 1], [0, 1, 0], [1, 2, 3]]]]])]
#Nx.Tensor<
  s64[1][1][1][4]
  [
    [
      [
        [0, 1, 4, 23]
      ]
    ]
  ]
>
```

josevalim (commented):
This is awesome @jonatanklosko! 🎉 

I have pushed some changes to the docs, as I was reviewing and trying to understand the differences between `take`, `take_along_axis`, and `slice`. They all seem very close to each other but still distinct. At this point, I don't think `take_along_axis` is particularly a good name, but I certainly expect people coming from Numpy to look for it, so I think no action is necessary.

Overall, this looks good to merge, I think we only need to:

1. Document if negative indexes are supported. If so, what happens in EXLA and what should happen in Elixir land? I believe we clamp them for slices (i.e. if negative, it becomes zero, if more than length, it becomes length). But I can't  say the same will happen for gather. Once we figure this out, please retroactively document and update take/take_along_axis

2. Regarding the acess proposal from @polvalente, I think we can accept some variation of it but it needs to be done with care. Currently, the `foo[bar]` syntax does not allow the shape of `foo` to "grow", and I don't think we should allow that quite yet. This can be done in a separate PR.

3. The other thing to keep in mind is what is the derivative of gather. We should be careful with adding an operation that we won't be able to cleanly compute a derivative for.

polvalente:
@josevalim if I'm not mistaken, the derivative of gather is some form of scatter-add. We can experiment on this after I have a working implementation

josevalim:
Cool. To be clear, I think we should merge this once step 1 is done. If we can't find a derivative later, we can simply remove this (with the benefit we will have the commit in our history if needed in the feature).

jonatanklosko:
@josevalim as for negative indices we discussed it briefly in https://github.com/elixir-nx/nx/pull/433#discussion_r681239629, the same applies here. So currently I made binary backend raise an error, wdyt?

seanmor5:
Fantastic work @jonatanklosko :D

exla/test/exla/defn_expr_test.exs
```diff
@@ -2568,6 +2568,26 @@ defmodule EXLA.DefnExprTest do
     end
   end
 
+  describe "gather" do
+    defn(gather(t, idx), do: Nx.gather(t, idx))
```
polvalente, line 2572:
```suggestion
    defn gather(t, idx), do: Nx.gather(t, idx)
```

Are the parens needed?

jonatanklosko: Nope, probably formatted within the editor and it didn't pick up the formatter config :)

josevalim:

PR #457, opened 2021-09-16
Include EXLA in the CI workflow

jonatanklosko:
Having the precompiled XLA avilable we can run EXLA tests on the CI just fine. However, we need to resolve the failing tests first :) [Here](https://github.com/jonatanklosko/nx/runs/3626433592?check_suite_focus=true) is an example build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,41 +1,41 @@
 name: Test
-
 on:
   push:
     branches:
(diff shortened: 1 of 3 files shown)


seanmor5:
@jonatanklosko I will look into the failing test

jonatanklosko:
@seanmor5 nice solve!!

PR #458, opened 2021-09-17
feat: scatter-add

polvalente:
Work in progress for Nx.scatter_add, which will be used as the grad for all gather-like operations (gather, take, take_along_axis)


exla/c_src/exla/exla.cc
```diff
@@ -2149,149 +2230,149 @@ ERL_NIF_TERM start_log_sink(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
-  {"triangular_solve", 6, triangular_solve},
-  {"svd", 2, svd},
-  // Infeed/Outfeed
-  {"infeed", 2, infeed},
-  {"outfeed", 3, outfeed},
-  {"create_token", 1, create_token},
-  // Log Sink
-  {"start_log_sink", 1, start_log_sink}
-};
+    // XlaBuilder
```
polvalente, line 2233: I'm not sure what happened here. I'll revert this soon

exla/test/exla/defn_expr_test.exs
```diff
@@ -1134,6 +1134,50 @@ defmodule EXLA.DefnExprTest do
...
+  describe "scatter_add" do
+    defn scatter_add(t, i, u) do
+      Nx.scatter_add(t, i, u)
+    end
+
+    test "scatter_add works for multi-dim tensor" do
+      target =
+        Nx.broadcast(0, {2, 3, 4})
+
+      indices = Nx.tensor([[[
```
polvalente, line 1146: @josevalim @jonatanklosko at first I was going to support indices like the ones resulting from argsort, which are along an specific axis. However, this would hinder any progress with using this as the grad for gather.

Do you think this interface is fine? The shape of the indices doesn't matter much, we can think of it as something like iodata being converted to a binary, as the second test shows.

The only restriction is that the indices rank must be the same as the target rank + 1, since the last dim is filled with a tensor of shape {rank}.

polvalente: If this is ok, I'll move on to implementing the binary backend for this function

nx/lib/nx.ex
```diff
@@ -4280,6 +4280,35 @@ defmodule Nx do
...
+    #   num_el = Enum.reduce(shape_l, &*/2)
+
+    #   flat_dim =
+    #     List.duplicate(1, tuple_size(shape)) |> List.replace_at(0, num_el) |> List.to_tuple()
+
+    #   Nx.take_along_axis(t, i |> Nx.reshape(flat_dim) |> Nx.tile(List.replace_at(shape_l, 0, 1)),
+    #     axis: 0
+    #   )
+    #   |> Nx.reshape(List.replace_at(shape_l, 0, shape_l) |> List.flatten() |> List.to_tuple())
+    # end
```
polvalente, line 4307: @josevalim this will be removed, but just out of curiosity, this is how we can implement Nx.take through Nx.take_along_axis.

I'm actually thinking that it'll be simpler if we just implement both as Nx.gather after @jonatanklosko's PR and then just add the grad for gather like Jax does. WDYT?

jonatanklosko: @polvalente in order to implement `take*` in terms of `gather`, we would need to generate `indices` tensor with all the individual indices though? If so, the generated tensor could be really large.

polvalente: @jonatanklosko at the EXLA level we can use some tricks with the update and indices axis options. We shouldn't need to expand the tensor explicitly. This was just a proof of concept that I was toying with earlier to convince myself that take is basically an special case of take_along_axis :)

polvalente: At second thought, needing to drop down to EXLA would mean that we have to implement the grads separately anyway, so the refactor to Nx.gather isn't desirable (at least with the current exposed API, which I wouldn't change)

exla/test/exla/defn_expr_test.exs
jonatanklosko: I'd consider restricting the shape to be 2d, since logically any other shape isn't relevant.

jonatanklosko: Similarly to gather, scatter could have versions working with larger slices, thought whatever route we go I think our `scatter` and `gather` should do *the same* ^^

nx/lib/nx.ex
jonatanklosko: Yeah, as far as I understand we need separate backend operations to distinguish on EXLA level, which in turn implies separate grads.

exla/test/exla/defn_expr_test.exs
polvalente: I see, so we expose, for instance, indices and updates like: `[[0, 1, 1], [0, 2, 3], [1, 1, 2]]` and `[1, 2, 3, 4]` and just reshape them accordingly at the EXLA level

polvalente: WDYT?

jonatanklosko: Yeah, that's what I thought, but could definitely use another opinion :) @seanmor5, @josevalim

nx/lib/nx/binary_backend.ex
```diff
@@ -1571,6 +1571,65 @@ defmodule Nx.BinaryBackend do
...
+          idx,
+          Nx.tensor(upd) |> Nx.reshape(Tuple.duplicate(1, length(idx))),
+          fn x, y ->
+            x = from_binary(%T{names: [], shape: {}, type: t.type}, x)
+            y = from_binary(%T{names: [], shape: {}, type: updates.type}, y)
+
+            Nx.add(x, y) |> Nx.as_type(out.type) |> Nx.to_binary()
+          end
+        )
+    end
```
polvalente, line 1630: @seanmor5 @josevalim @jonatanklosko does anybody know of a better way to do this "traverse and update" for the tensor?

I thought of recursevely using aggregate_axes until the last dim is reached, having the accumulated indices so far, but perhaps there's a better solution.

Although this implementation works, I think we can do better.

nx/lib/nx/binary_backend.ex
```diff
@@ -1648,7 +1707,7 @@ defmodule Nx.BinaryBackend do
   defp top_dimension_slice?(_, _, _), do: false
 
   @impl true
-  def put_slice(out, tensor, start_indices, slice) do
+  def put_slice(out, tensor, start_indices, slice, combine_fn \\ fn _prev, new -> new end) do
```
polvalente, line 1710: I plan on reverting this after the better implementation for scatter_add is reached

nx/lib/nx/binary_backend.ex
polvalente: Nevermind! Looking at gather code I had an idea which might work :)
https://github.com/elixir-nx/nx/pull/456/files#diff-15095458a1b2fb983edb557f9d4b0664e25425d06d053665f48867c8df015307R1784-R1804

I plan on producing a list of {offset, increment} for each of the index-update pairs, and then just travelling across the binary jumping between offsets. This can be done in O(indices) instead of O(binary_elements), I think.

Thanks for the inspiration @jonatanklosko

jonatanklosko: Yup, I think you can iteratively use `binary_part` for the unchanged areas and put the updates inbetween, then just binary-concat everything. That's probably what you meant :D

polvalente: Yeah! There'll be some type conversion needed along the way, but that's the idea :D

nx/lib/nx.ex
```diff
@@ -4280,6 +4280,90 @@ defmodule Nx do
     )
   end
 
+  @doc """
```
polvalente, line 4283: @jonatanklosko @josevalim @seanmor5 PR is ready for review. Suggestions for better error messages are more than welcome.

Also, the second example in the docs is kinda useful, but a bit verbose. Should I replace it with something more obvious? Perhaps move that into an unit test of sorts (heh)?

nx/lib/nx.ex
```diff
@@ -4280,6 +4280,116 @@ defmodule Nx do
...
+      iex> indices = Nx.concatenate(iotas, axis: 1)
+      iex> Nx.scatter_add(Nx.broadcast(0, Nx.shape(t)), indices, Nx.reshape(t, {num_elements}))
+      #Nx.Tensor<
+        s64[3][2]
+        [
+          [5, 4],
+          [3, 2],
+          [1, 0]
+        ]
+      >
```
josevalim, line 4345: Yeah, this should definitely be a unit test. :)

nx/lib/nx.ex
```diff
@@ -4280,6 +4280,116 @@ defmodule Nx do
...
+      ** (ArgumentError) expected indices to have shape {*, 2}, got: {1, 3}
+
+      iex> Nx.scatter_add(Nx.tensor([[1], [2]]), Nx.tensor([[1, 2]]), Nx.tensor([0, 1]))
+      ** (ArgumentError) expected updates tensor to match the first axis of indices tensor with shape {1, 2}, got {2}
+  """
+  def scatter_add(target, indices, updates, opts \\ []) do
+    %T{} = target = to_tensor(target)
+    %T{} = indices = to_tensor(indices)
+    %T{} = updates = to_tensor(updates)
+
```
josevalim, line 4385: They will be tensors, no need to double check. :)

```suggestion
    target = to_tensor(target)
    indices = to_tensor(indices)
    updates = to_tensor(updates)

```

nx/lib/nx.ex
```diff
@@ -4280,6 +4280,116 @@ defmodule Nx do
...
 
+  @doc """
+  Performs a scatter-add operation on the `target` tensor,
+  adding the `updates` into the corresponding `indices` positions.
+
+  This operation is the grad for `gather/2` and gather-like operations such as
+  `take/3` and `take_along_axis/3`.
+
+  `indices` must be a fully qualified tensor of shape `{n, Nx.rank(target)}`, with `n`
+  being an arbitrary number of indices, while `updates` must have a compatible `{n}` shape.
```
josevalim, line 4291: What happens for negative indexes? Are we raising for the binary backend?

nx/lib/nx/binary_backend.ex
```diff
@@ -1571,6 +1571,89 @@ defmodule Nx.BinaryBackend do
...
+              to_traverse,
+              before_slice_size + target_byte_size,
+              total_size - (before_slice_size + target_byte_size)
+            )
+
+          updated_element =
+            scalar_to_binary(binary_to_number(element, target.type) + update, out.type)
+
+          {traversed <> before_offset <> updated_element, to_traverse}
+      end
```
josevalim, line 1652: I think this overall function looks solid but this last part of updating the binary seems a bit expensive. I will write pseudo-codeish but hopefully it shows you how I would tackle it using recursion:

    def scatter_add([{previous, current, update} | rest], binary) do
      match_types [type] do
        <<previous_bin::size(previous), match!(x, 0), rest_bin::binary>> = binary
        <<previous_bin::size(current - previous), write!(read!(x, 0) + update, 0), scatter_add(rest, rest_bin)::binary>>
      end
    end

    def scatter_add([], binary), do: binary

I think the compiler will be smart enough to optimize it too. :)

josevalim (commented):
Absolutely beautiful! 😍 I have added some small suggestions.

nx/lib/nx/binary_backend.ex
polvalente: Oh, I see! Maybe I could just accumulate an iodata list and the iodata_to_binary in the end? This would change almost nothing in my code.

nx/lib/nx.ex
polvalente: I'm not checking for them!

We can either raise or transform them into the corresponding positive index as we go.

EXLA will probably just break

nx/lib/nx/binary_backend.ex
josevalim: @polvalente yes, that would also work! You would need a revert in your code and in mine it could be list recursive, but it all turns out to be the same.

polvalente: Wouldn't `[acc | [iter_result]]` bypass the Enum.reverse?

josevalim: true!!!

nx/lib/nx.ex
polvalente: Actually, since I've copied the offset calculation from @jonatanklosko's gather implementation, it actually raises on negative indices :)

nx/lib/nx/binary_backend.ex
```diff
@@ -1571,6 +1571,89 @@ defmodule Nx.BinaryBackend do
...
+         }, next_offset}
+      end)
+
+    target_binary = to_binary(target)
+
+    {result, tail} =
+      for {previous, current, update} <- offsets_with_updates, reduce: {<<>>, target_binary} do
+        {traversed, to_traverse} ->
+          before_slice_size = max(current - previous, 0)
+          before_offset = binary_part(to_traverse, 0, before_slice_size)
```
polvalente, line 1635: @josevalim are all our types 8-byte aligned? I've just realized that this line should be followed by a match-read-write combo to cast from one type to another... The updated elements are being treated properly.

polvalente: Well, anyway, I fixed this in the latest commit, although I couldn't reproduce any problems with f32 + s64 or f32 + f64

josevalim: They are _8-bit_ aligned, yes. Btw, instead of binary_part, consider using the binary syntax I showed, it is going to be more concise and more efficient:

    <<pre::size(pre), match!(...), post::binary>>

:)

polvalente: I was facing some issues with this syntax, but I'll work into a refactor for this now :)

polvalente: https://github.com/elixir-nx/nx/pull/464

PR #460, opened 2021-09-18
feat: take along axis grad

polvalente:


nx/test/nx/defn/grad_test.exs
```diff
@@ -2503,6 +2503,101 @@ defmodule Nx.Defn.GradTest do
...
+                 Nx.tensor([
+                   [0, 0, 0],
+                   [1, 1, 0],
+                   [0, 1, 1]
+                 ])
+               )
+
+      assert Nx.tensor([
+               [-0.0, -0.0],
+               [4.370079, 0.5701862],
```
polvalente, line 2583: @josevalim This assertion is failing in a rather bizarre way. The actual tensor returned is:

```elixir
#Nx.Tensor<
  f32[3][2]
  [
    [-0.0, -0.0],
    [1.8369784355163574, 0.570186197757721],
    [-2.3156425952911377, 13.522059440612793]
  ]
>
```

which only differs from the current tensor on the index `[1, 0]`, with actual value `1.8` instead of `4.3`...

Any ideas? Also, this might or might not be related to me needing to call `Nx.gather` to actually propagate the chain rule from the outermost functions, although it seemed to make sense.

polvalente: The other two assertions are working as intended :)

nx/lib/nx/defn/grad.ex
```diff
@@ -507,6 +507,29 @@ defmodule Nx.Defn.Grad do
...
+    num_elements = i |> Nx.shape() |> Tuple.product()
+
+    iotas =
+      Enum.map(0..(Nx.rank(g) - 1)//1, fn axis ->
+        i |> Nx.iota(axis: axis, backend: Nx.Defn.Expr) |> Nx.reshape({num_elements, 1})
+      end)
+
+    iotas = List.replace_at(iotas, axis, Nx.reshape(i, {num_elements, 1}))
+
+    indices = Nx.concatenate(iotas, axis: 1)
```
jonatanklosko, line 520: Instead of replacing the list afterwards, we can handle the specific axis while iterating, something like:

```suggestion
indices =
  g
  |> Nx.axes()
  |> Enum.map(fn
    ^axis -> Nx.reshape(i, {num_elements, 1})
    axis -> i |> Nx.iota(axis: axis, backend: Nx.Defn.Expr) |> Nx.reshape({num_elements, 1})
  end)
  |> Nx.concatenate(axis: 1)
```

jonatanklosko:
What happens if an element is taken twice?

nx/test/nx/defn/grad_test.exs
```diff
@@ -2503,6 +2503,101 @@ defmodule Nx.Defn.GradTest do
...
+          |> Nx.take_along_axis(i, axis: 1)
+          |> Nx.power(2)
+          |> Nx.log()
+          |> Nx.sum()
+        end
+      )
+    end
+
+    test "computes gradient" do
+      assert Nx.tensor([
```
polvalente, line 2545: @jonatanklosko this assert might shed some light onto your question!

Basically, we scatter_add the corresponding grads to each index taken onto a zero-tensor

jonatanklosko: Ah, so we do sum up grads for the same element, makes sense!

nx/test/nx/defn/grad_test.exs
jonatanklosko: Why do we do gather actually? Shouldn't we use `Nx.reshape(g, {num_elements})` as the updates? I.e. we take every gradient element and put it into wherever element it originated from.

polvalente: Huh, this could work! Gather was just more instinctive, but they might be equivalent here :)

jonatanklosko: Well, actually the scatter indices are valid indices into the original tensor, not into the grad tensor. To give an example, the original tensor may very well shrink after `take_along_axis`, in which case the grad tensor is smaller than the original tensor and the indexing is different :)

polvalente: Right, so maybe this is what's causing the strange error I'm getting

josevalim: Just for curiosity, those values came from JAX, right?

polvalente: @josevalim yes

nx/lib/nx/defn/grad.ex
```diff
@@ -507,6 +507,29 @@ defmodule Nx.Defn.Grad do
...
+
+    g =
+      t
+      |> Expr.broadcast(0, Nx.shape(t), Nx.axes(t))
+      |> Nx.scatter_add(
+        indices,
+        Nx.reshape(g, {num_elements})
```
polvalente: @jonatanklosko reshaping `g` worked!

nx/test/nx/defn/grad_test.exs
polvalente: As commented above, reshaping `g` to a `{num_elements}`-shaped tensor did it

josevalim (approved):
AMAZING!

PR #463, opened 2021-09-19
Fix exla and libtorch for Mac OS

josevalim:
Currently, EXLA does not link on Mac OS. It cannot find the NIF symbols. There are [two known answers to this](https://stackoverflow.com/questions/8288358/erlang-nif-test-os-x-lion) which I tried. But both fail next with the following reason:

```
10:20:23.468 [warn]   The on_load function for module Elixir.EXLA.NIF returned:
{:error,
 {:load_failed,
  'Failed to load NIF library: \'dlopen(/Users/jose/OSS/nx/exla/_build/test/lib/exla/priv/libexla.so, 2): Library not loaded: bazel-out/darwin-opt/bin/tensorflow/compiler/xla/extension/libxla_extension.so\n  Referenced from: /Users/jose/OSS/nx/exla/priv/libexla.so\n  Reason: image not found\''}}
```

I am wondering, instead of linking to libxla_extension.so, could we instead build a fat binary with libxla_extension and simplify stuff like rpath handling and so on?

### DIFF
--- a/exla/Makefile
+++ b/exla/Makefile
@@ -18,21 +18,25 @@ CFLAGS = -fPIC -I$(ERTS_INCLUDE_DIR) -isystem $(XLA_INCLUDE_PATH) -O3 -Wall -Wex
 				 -Wno-unused-parameter -Wno-missing-field-initializers -Wno-comment \
 				 -shared -std=c++14
 
-# Use a relative RPATH, so at runtime libexla.so looks for libxla_extension.so
-# in ./lib regardless of the absolute location. This way priv can be safely
-# packed into an Elixir release. Also, we use $$ to escape Makefile variable
-# and single quotes to escape shell variable
-LDFLAGS = -L$(EXLA_LIB_DIR) -Wl,-rpath,'$$ORIGIN/lib' -lxla_extension
+LDFLAGS = -L$(EXLA_LIB_DIR) -lxla_extension
 
 ifeq ($(shell uname -s), Darwin)
-	# LDFLAGS += -undefined dynamic_lookup -dynamiclib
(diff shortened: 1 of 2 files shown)


josevalim:
Btw, if I add a bazel-out/darwin-opt/bin/tensorflow/compiler/xla/extension/libxla_extension.so to the current working dir, then it works.

josevalim:
@jonatanklosko @seanmor5 this works for me but I am not sure if I broke something for Linux. Can you try this branch, run `rm -rf _build` inside EXLA and let me know if it compiles?

exla/Makefile
```diff
@@ -1,15 +1,15 @@
...
 # XLA Extension Installation Location
 XLA_EXTENSION_DIR ?= cache/xla_extension
-
 XLA_EXTENSION_LIB = $(XLA_EXTENSION_DIR)/lib
 XLA_INCLUDE_PATH = $(XLA_EXTENSION_DIR)/include
 
 # Private configuration
 EXLA_DIR = c_src/exla
-PRIV_DIR = priv
+PRIV_DIR = $(MIX_APP_PATH)/priv
```
jonatanklosko, line 12: By using environment specific priv we need to compile EXLA separately for dev and test, while currently we compile into top-level priv that gets linked in all environments. Is this the desired behavoiur?

jonatanklosko:
Compiles fine for me :)

exla/Makefile
josevalim: @jonatanklosko yes, it was on purpose because nuking _build should recompile this stuff and we should avoid doing things in priv.

josevalim:

PR #466, opened 2021-09-20
feat: add take grad

polvalente:
Similar to #460


nx/test/nx/defn/grad_test.exs
```diff
@@ -2503,6 +2503,141 @@ defmodule Nx.Defn.GradTest do
...
+                 Nx.tensor([
+                   [
+                     [[0], [1], [2]],
+                     [[2], [1], [0]],
+                     [[0], [1], [2]]
+                   ]
+                 ])
+               )
+
+      assert Nx.tensor([
```
polvalente, line 2618: This is mostly working, but this last assertion fails

nx/lib/nx/defn/grad.ex
```diff
@@ -507,6 +507,62 @@ defmodule Nx.Defn.Grad do
...
+  defp grad(:take, [t, i, axis], _ans, g, cache) do
+    axes_range = 0..(tuple_size(t.shape) - 1)//1
+
+    indices_shape =
+      axes_range
+      |> Enum.map(fn
+        ^axis -> Tuple.to_list(i.shape)
+        _ -> 1
+      end)
+      |> List.flatten()
```
josevalim, line 519:
```suggestion
      |> Enum.flat_map(fn
        ^axis -> Tuple.to_list(i.shape)
        _ -> [1]
      end)
```

A similar technique can be employed below if you want.

nx/lib/nx/defn/grad.ex
```diff
@@ -507,6 +507,62 @@ defmodule Nx.Defn.Grad do
     to_grad(t, g, cache)
   end
 
+  defp grad(:take, [t, i, axis], _ans, g, cache) do
+    axes_range = 0..(tuple_size(t.shape) - 1)//1
```
josevalim, line 511:
```suggestion
    axes_range = 0..(tuple_size(t.shape) - 1)//1
```

Do we use `Nx.rank(t)` or `tuple_size(t.shape)` in this file? make sure we are following one "standard" or the other. :)

josevalim (approved):
Looking gorgeous, just some small comments added!

nx/lib/nx/defn/grad.ex
polvalente: Nx.rank :)

I'll change this and possibly other occurences

josevalim (approved):

PR #471, opened 2021-09-20
Add note on CUDA stack size and lower default memory allocation

### LINKED ISSUE #197: Convolution causes segmentation fault with cuDNN 8
Some convolutions will lead to a segmentation fault when using cuDNN8. The segmentation fault seems to be an issue with how the dirty scheduler is interacting with CUDA. The issue was passed on to NVIDIA, and they are debugging.

Issue is reproducible from within a NIF as seen here: https://github.com/seanmor5/conv_seg_fault

And the related issue opened in TensorFlow: https://github.com/tensorflow/tensorflow/issues/46616

If working with convolutions is necessary, the best option is to use the CUDA dockerfile with CUDA 10.1/cuDNN7.

seanmor5:
Closes #197 

Also sets a lower `:memory_allocation` in runtime configuration so all tests can pass. The old default may cause issues with certain GPUs and CUDA versions which leads to CUDA OOM. The recommended default of 0.9 is still what JAX uses, but 0.8 is what I needed to set to get my GPU to pass all of our tests.

### DIFF
--- a/exla/lib/exla.ex
+++ b/exla/lib/exla.ex
@@ -92,17 +92,15 @@ defmodule EXLA do
 
   ### GPU Runtime Issues
 
-  Some versions of CUDA and cuDNN may interact poorly with the Erlang
-  VM leading to strange Segmentation Faults. GPU Executions run in dirty
-  IO threads, which have a considerable smaller stack size than regular
-  scheduler threads. This may lead to problems with certain CUDA or cuDNN
-  programs. In a development environment, it is suggested to set:
+  GPU Executions run in dirty IO threads, which have a considerable smaller
+  stack size than regular scheduler threads. This may lead to problems with
+  certain CUDA or cuDNN versions, leading to segmentation fails. In a development
+  environment, it is suggested to set:
 
-  `ELIXIR_ERL_OPTIONS=+sssdio 128`
+      ELIXIR_ERL_OPTIONS="+sssdio 128"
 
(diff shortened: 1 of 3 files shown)


polvalente (approved):
Nice job!

exla/lib/exla.ex
```diff
@@ -90,6 +90,20 @@ defmodule EXLA do
...
+  VM leading to strange Segmentation Faults. GPU Executions run in dirty
+  IO threads, which have a considerable smaller stack size than regular
+  scheduler threads. This may lead to problems with certain CUDA or cuDNN
+  programs. In a development environment, it is suggested to set:
+
+  `ELIXIR_ERL_OPTIONS=+sssdio 128`
+
+  To increase the stack size of dirty IO threads from 40 kilowords to
+  128 kilowords in order to avoid poor interactions between CUDA, cuDNN,
+  and the dirty IO scheduler. In a release, you can set this flag in `vm.args`.
```
josevalim, line 105:
```suggestion
  GPU Executions run in dirty IO threads, which have a considerable smaller
  stack size than regular scheduler threads. This may lead to problems with
  certain CUDA or cuDNN versions, leading to segmentation fails. In a development
  environment, it is suggested to set:

      ELIXIR_ERL_OPTIONS=+sssdio 128

  To increase the stack size of dirty IO threads from 40 kilowords to
  128 kilowords. In a release, you can set this flag in your `vm.args`.
```

josevalim (approved):

PR #474, opened 2021-09-20
fix: handle tall/wide b matrix on tri solve

### LINKED ISSUE #473: Nx.LinAlg, Nx.BinaryBackend.triangular_solve B-argument shape mismatch
In `Nx.BinaryBackend.triangular_solve`, and `Nx.BinaryBackend.Matrix.ts` the shape checks seem out of whack.

It seems like the valid dimensions for A should be `{n,n}`, and B should be `{n,m}`, including `m = 1` or `{n}` for one-dimensional tensors.

The shape check in ~`Nx.BinaryBackend.triangular_solve`~`Nx.LinAlg.triangular_solve` does not reflect this, and even when fixing that, it seems like `Nx.BinaryBackend.Matrix.ts` makes assumptions about B being one-dimensional, not even accepting `{n,1}` tensors.

To check I'm not completely off my rocker, using the `EXLA` backend seems to work in [this test](https://gist.github.com/dognotdog/01e6e2a38efbdef06d9a545f04618b63), with the results of the attempted regression matching the inputs, but just doing it with the `BinaryBackend` runs into shape mismatch related errors.

My naive attempt at fixing it has not been successful, as it hasn't been immediately obvious in `do_ts` how to make it do the right thing for `{n,m}` B arguments. 

EDIT: see 4628f5a74d9d8846738acf19c560b0f65dd45bfe to make snippet work with EXLA, without this fix neither backend works!

polvalente:
This is an attempt to fix the `Nx.LinAlg.triangular_solve` BinaryBackend implementation, as reported on #473

closes #473

### DIFF
--- a/nx/lib/nx/binary_backend.ex
+++ b/nx/lib/nx/binary_backend.ex
@@ -1200,14 +1200,14 @@ defmodule Nx.BinaryBackend do
   @impl true
   def triangular_solve(
         %{type: output_type} = out,
-        %{type: a_type, shape: {rows, rows}} = a,
+        %{type: a_type, shape: {rows, rows} = a_shape} = a,
         %{type: b_type, shape: b_shape} = b,
         opts
       )
-      when b_shape == {rows, rows} or b_shape == {rows} do
+      when tuple_size(b_shape) == 2 or b_shape == {rows} do
     a_data = to_binary(a)
     b_data = to_binary(b)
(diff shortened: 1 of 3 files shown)


polvalente:
@dognotdog could you test this branch to see if it actually solves your problem? If it doesn't I think I'll rework the implementation later this week.

I've added tests for both tall and wide matrices with different params, but I might have missed some combos

josevalim (approved):

PR #478, opened 2021-09-22
chore: organize torchx tests

polvalente:
First step in the effort of making Torchx useable again. All tests now pass and can be added to the CI workflow.
Afterwards, some more tests will be added to ensure all functions are tested

### DIFF
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -287,9 +287,9 @@ defmodule Nx.Defn.Expr do
   end
 
   @impl true
-  def bitcast(out, tensor, type) do
+  def bitcast(out, tensor) do
     tensor = to_expr(tensor)
-    expr(out, tensor.data.context, :bitcast, [tensor, type])
+    expr(out, tensor.data.context, :bitcast, [tensor])
   end
 
   @impl true
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -276,7 +276,7 @@ defmodule Nx.Defn.Expr do
     [:exp, :expm1, :log, :log1p, :logistic, :cos, :sin, :tan, :cosh, :sinh, :tanh] ++
       [:acosh, :asinh, :atanh, :sqrt, :rsqrt, :cbrt, :negate, :sign, :abs, :bitwise_not] ++
       [:population_count, :count_leading_zeros, :floor, :ceil, :round] ++
-      [:erf, :erfc, :erf_inv, :acos, :asin, :atan]
+      [:erf, :erfc, :erf_inv, :acos, :asin, :atan, :bitcast]
 
   for op <- unary_ops do
     @impl true
@@ -284,12 +284,6 @@ defmodule Nx.Defn.Expr do
       tensor = to_expr(tensor)
       expr(out, tensor.data.context, unquote(op), [tensor])
     end
-  end
-
-  @impl true
-  def bitcast(out, tensor) do
-    tensor = to_expr(tensor)
-    expr(out, tensor.data.context, :bitcast, [tensor])
   end
 
   @impl true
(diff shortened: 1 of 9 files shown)


torchx/test/torchx/nx_doctest_test.exs
```diff
@@ -69,32 +20,64 @@ defmodule Torchx.NxDoctestTest do
       Torchx.Backend.__unimplemented__()
       |> Enum.map(fn {fun, arity} -> {fun, arity - 1} end)
       |> Kernel.++(
+        # to_binary - not supported, must call backend_transfer before
         to_binary: 2,
+        # to_batched_list - Shape mismatch
```
josevalim, line 25:
```suggestion
        # to_batched_list - Shape mismatch
```

I don't think it is shape mismatch but rather the fact we don't support all options. In the Torchx.Backend we have a functionality to raise on unsupported options, we should probably do the same here (and update this comment). :)

polvalente: I ran all tests which weren't commented. For this one, I got:

 shape mismatch in Torchx: expected {2, 2}, got: {1, 2}. Please report this bug

polvalente: Ah, I see. There are 4 tests and only 2 fail.

polvalente: I plan on merging this PR without any changes to functionality or tests, and then adding proper tests for each function in subsequent PRs, after we add Torchx to the CI workflow

torchx/test/torchx/nx_doctest_test.exs
```diff
@@ -69,32 +20,64 @@ defmodule Torchx.NxDoctestTest do
       Torchx.Backend.__unimplemented__()
       |> Enum.map(fn {fun, arity} -> {fun, arity - 1} end)
       |> Kernel.++(
```
josevalim, line 22: Two additional suggestions:

1. What if we break this down in two categories: a) can't be fixed because it is an inherent limitation, like round or b) it is temporarily broken or needs more investigation, like dot, slice_axis, slice, all_close?/2, etc 

2. Then inside each category we order alphabetically

torchx/test/torchx/nx_doctest_test.exs
```diff
@@ -69,32 +20,64 @@ defmodule Torchx.NxDoctestTest do
...
+        # bitcast - not supported
         bitcast: 2,
+        # cbrt - not supported
         cbrt: 1,
+        # cos - removed due to rounding error
         cos: 1,
-        erfc: 1
+        # atanh - removed due to rounding error
+        atanh: 1,
+        # concatenate - unsupported on torchx
```
josevalim, line 67: It is likely supported but we haven't implemented it yet. Theoretically we shouldn't need to list it here because concatenate/3 should be returned as part of `Torchx.Backend.__unimplemented__()`. :)

torchx/test/torchx/nx_doctest_test.exs
polvalente: That's a good idea :)

torchx/test/torchx/nx_doctest_test.exs
josevalim: Sounds good. We can also just add something like "needs investigation". I think the main point is to separate the things we need to look into and the things that are inherently different. :)

torchx/test/torchx/nx_doctest_test.exs
polvalente: Oh, sure! I completely missed this function :)

nx/lib/nx/defn/expr.ex
```diff
@@ -276,7 +276,7 @@ defmodule Nx.Defn.Expr do
     [:exp, :expm1, :log, :log1p, :logistic, :cos, :sin, :tan, :cosh, :sinh, :tanh] ++
       [:acosh, :asinh, :atanh, :sqrt, :rsqrt, :cbrt, :negate, :sign, :abs, :bitwise_not] ++
       [:population_count, :count_leading_zeros, :floor, :ceil, :round] ++
-      [:erf, :erfc, :erf_inv, :acos, :asin, :atan, :bitcast]
+      [:erf, :erfc, :erf_inv, :acos, :asin, :atan]
```
josevalim, line 279:
```suggestion
      [:erf, :erfc, :erf_inv, :acos, :asin, :atan, :bitcast]
```

nx/lib/nx/defn/expr.ex
```diff
@@ -286,6 +286,12 @@ defmodule Nx.Defn.Expr do
     end
   end
 
+  @impl true
+  def bitcast(out, tensor) do
+    tensor = to_expr(tensor)
+    expr(out, tensor.data.context, :bitcast, [tensor])
+  end
+
```
josevalim, line 294:
```suggestion
```

josevalim (approved):
Two small comments and ship it. :D

josevalim:

PR #479, opened 2021-09-22
feat: add torchx to CI

polvalente:


.github/workflows/ci.yml
```diff
@@ -30,6 +30,14 @@ jobs:
...
           path: deps
           key: ${{ matrix.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-mix-${{ hashFiles(format('{0}{1}', github.workspace, '/mix.lock')) }}
+      - name: Download libtorch
+        if: ${{ matrix.working_directory == 'torchx' }}
+        run: wget https://download.pytorch.org/libtorch/lts/1.8/cpu/libtorch-shared-with-deps-1.8.2%2Bcpu.zip -q -O /tmp/libtorch.zip
+      - name: Install libtorch
+        if: ${{ matrix.working_directory == 'torchx' }}
+        uses: TonyBogdanov/zip@1.0
+        with:
+          args: unzip -qq /tmp/libtorch.zip -d . && export LIBTORCH_DIR=/tmp/libtorch
```
jonatanklosko, line 40: The `export`-ed variables are not visible to further steps, we can however write to special files to set these ([ref](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-an-environment-variable)).

Also, `unzip` should be available on the default image.

```suggestion
      - name: Download libtorch
        if: ${{ matrix.working_directory == 'torchx' }}
        run: |
          wget -q -O libtorch.zip https://download.pytorch.org/libtorch/lts/1.8/cpu/libtorch-shared-with-deps-1.8.2+cpu.zip
          unzip -qq libtorch.zip -d .
          echo "LIBTORCH_DIR=$(pwd)/libtorch" >> $GITHUB_ENV
```

polvalente: Ok, so I'm really confused, because now it worked haha

Oh well, onwards we go

jonatanklosko: Hmm, according to [installed packages](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#installed-apt-packages) it should :thinking:

polvalente: Yeah, it is. I probably had a typo somewhere

jonatanklosko: Nit: I still suggest using multiline `run`, and setting the env from there :)

polvalente: I'm trying to understand why the formatter is complaining even though locally it doesn't change anything...

polvalente: After I get the first green check, I'll apply your suggestion and figure out how to cache libtorch

jonatanklosko: The formatting error is about trailing comma here:

https://github.com/elixir-nx/nx/blob/37d612fb0e92f9cedc05f14e36b7a73fa5ec34a6/torchx/test/torchx/nx_doctest_test.exs#L75

jonatanklosko: As for caching, I think you can download/extract to `~/libtorch` or similar, so there's a clear location for caching.

.github/workflows/ci.yml
```diff
@@ -30,10 +31,24 @@ jobs:
...
+        if: ${{ matrix.working_directory == 'torchx' }}
+        uses: actions/cache@v1
+        id: libtorch-cache
+        with:
+          path: libtorch
+          key: libtorch.zip
+      - name: Download and install libtorch
+        if: ${{ matrix.working_directory == 'torchx' && steps.libtorch-cache.outputs.cache-hit != 'true' }}
+        run: |
+          if [ ! -d libtorch ]; then
```
jonatanklosko, line 44: If the cache doesn't hit we don't expect the dir there right? And if for some reason it's there it's probably not what we want :)

polvalente: yeah :)

jonatanklosko (approved):
Looks good to me :cat:

josevalim (approved):

PR #480, opened 2021-09-22
test: add tests for rounding error doctests

polvalente:
Adds specific tests for some of the functions which don't have exactly the same output as on Nx.BinaryBackend (float rounding errors, different rounding strategies, etc)

### DIFF
--- a/torchx/test/support/torchx_case.ex
+++ b/torchx/test/support/torchx_case.ex
@@ -1,0 +1,20 @@
+defmodule Torchx.Case do
+  @moduledoc """
+  Test case for tensor assertions
+  """
+
+  use ExUnit.CaseTemplate
+
+  using do
+    quote do
+      import Torchx.Case
+    end
+  end
+
+  defmacro assert_tensor({:==, _, [left, right]}) do
+    quote do
+      assert Nx.backend_transfer(unquote(left), Nx.BinaryBackend) ==
+               Nx.backend_transfer(unquote(right), Nx.BinaryBackend)
+    end
+  end
+end
--- a/torchx/test/support/torchx_case.ex
+++ b/torchx/test/support/torchx_case.ex
@@ -11,10 +11,13 @@ defmodule Torchx.Case do
     end
   end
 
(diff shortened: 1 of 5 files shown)


torchx/test/support/torchx_case.ex
```diff
@@ -0,0 +1,20 @@
...
+  using do
+    quote do
+      import Torchx.Case
+    end
+  end
+
+  defmacro assert_tensor({:==, _, [left, right]}) do
+    quote do
+      assert Nx.backend_transfer(unquote(left), Nx.BinaryBackend) ==
+               Nx.backend_transfer(unquote(right), Nx.BinaryBackend)
```
seanmor5, line 17: Wouldn't an `allclose?` or `isclose?` fit better here? See: https://pytorch.org/docs/stable/testing.html

polvalente: Yeah, but we don't have it implemented yet! Should I withhold this PR and implement `all?` + `allclose?` instead?

torchx/test/torchx/nx_doctest_test.exs
```diff
@@ -80,6 +75,57 @@ defmodule Torchx.NxDoctestTest do
       Torchx.Backend.__unimplemented__()
       |> Enum.map(fn {fun, arity} -> {fun, arity - 1} end)
       |> Kernel.++(@temporarily_broken_doctests)
+      |> Kernel.++(@rounding_error_doctests)
       |> Kernel.++(@inherently_unsupported_doctests)
       |> Kernel.++([:moduledoc])
+
+  describe "rounding error tests" do
```
josevalim, line 82: Let's move those to backend_test.exs instead? And keep only doctests here, since they are already a lot. We may also need to break `backend_test.exs` itself into multiple modules (as we discussed) but it can be done another day.

torchx/test/support/torchx_case.ex
polvalente: @seanmor5 I've refactored this to use Nx.all_close as suggested :)

josevalim (approved):
Beautiful!

PR #481, opened 2021-09-23
fix: remove Torchx compilation warnings

polvalente:
Two minor fixes which silence all warnings which were being output by Torchx on my machine during compilation

### DIFF
--- a/torchx/Makefile
+++ b/torchx/Makefile
@@ -3,7 +3,7 @@ LIBTORCH_API_INCLUDE_PATH = $(LIBTORCH_DIR)/include/torch/csrc/api/include
 LIBTORCH_LINK_PATH = $(LIBTORCH_DIR)/lib/
 CFLAGS = -fPIC -I$(ERTS_INCLUDE_DIR) -I$(LIBTORCH_INCLUDE_PATH) -I$(LIBTORCH_API_INCLUDE_PATH) \
 	 -O3 -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -shared \
-	 -std=c++14 -D_GLIBCXX_USE_CXX11_ABI=0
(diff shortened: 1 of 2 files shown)


torchx/Makefile
```diff
@@ -3,7 +3,7 @@ LIBTORCH_API_INCLUDE_PATH = $(LIBTORCH_DIR)/include/torch/csrc/api/include
 LIBTORCH_LINK_PATH = $(LIBTORCH_DIR)/lib/
 CFLAGS = -fPIC -I$(ERTS_INCLUDE_DIR) -I$(LIBTORCH_INCLUDE_PATH) -I$(LIBTORCH_API_INCLUDE_PATH) \
 	 -O3 -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -shared \
-	 -std=c++14 -D_GLIBCXX_USE_CXX11_ABI=0
+	 -std=c++14 -D_GLIBCXX_USE_CXX11_ABI=0 -fopenmp
```
polvalente, line 6: This is to handle a `pragma` directive

torchx/c_src/torchx.cpp
```diff
@@ -58,7 +58,7 @@ inline std::string type2string(const torch::ScalarType type)
   }
 
 #define CATCH()                                              \
-  catch (c10::Error error)                                   \
+  catch (c10::Error &error)                                   \
```
polvalente, line 61: This one silences the vast majority of warnings, which all stemmed from a "catch by value" warning here.

josevalim (approved):

torchx/Makefile
josevalim: Actually, this option is not available on Mac. Should we move it to the Linux branch?

josevalim (changes requested):
One minor comment.

torchx/Makefile
polvalente: Sure!

PR #482, opened 2021-09-23
feat: add all? implementation for Torchx

polvalente:


torchx/lib/torchx/nif.ex
```diff
@@ -2,8 +2,11 @@ defmodule Torchx.NIF do
   @moduledoc false
   @on_load :__on_load__
 
+  @nif_path Path.join([:code.priv_dir(:torchx), "torchx"])
```
josevalim, line 5: The path at compile time may not be the same at runtime (a release for example may move things around, so we need to keep this inside `__on_load__`. Generally speaking, avoid doing things at compile time. :)

josevalim (changes requested):
Just one minor comment and ship it!

torchx/lib/torchx/nif.ex
```diff
@@ -3,7 +3,7 @@ defmodule Torchx.NIF do
   @on_load :__on_load__
 
   def __on_load__ do
-    path = :filename.join(:code.priv_dir(:torchx), 'torchx')
+    path = Path.join([:code.priv_dir(:torchx), 'torchx'])
```
josevalim, line 6: load_nif prefers a char list, I think we should keep this code as is rather. :)

polvalente: Forgot to rollback this part :sweat_smile:

PR #484, opened 2021-09-24
feat: add bitcast Torchx implementation

polvalente:
Altough this isn't necessarily a "no-read" conversion, this makes for a more complete API implementation for `Nx.Backend` on Torchx

### DIFF
--- a/torchx/lib/torchx/backend.ex
+++ b/torchx/lib/torchx/backend.ex
@@ -175,19 +175,8 @@ defmodule Torchx.Backend do
     do: Torchx.to_type(from_nx(t), to_torch_type(type)) |> to_nx(out)
 
   @impl true
-  def bitcast(%T{} = out, %T{} = t) do
-    case {out.type, t.type} do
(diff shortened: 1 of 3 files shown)


torchx/lib/torchx/backend.ex
```diff
@@ -174,6 +174,25 @@ defmodule Torchx.Backend do
...
+  def bitcast(%T{} = out, %T{} = t) do
+    out_type = to_torch_type(out.type)
+
+    case {out.type, t.type} do
+      {{_, s}, {_, s}} -> :ok
+      {out_type, in_type} ->
+        raise ArgumentError,
+            "input type width must match new type width," <>
+              " got input type #{inspect(in_type)} and" <>
+              " output type #{inspect(out_type)}"
```
polvalente, line 187: This is just a copy-paste from the same raise in Nx. I wonder if we should expose this as a validation `Nx.Type.bitcast` like we have with `Nx.Shape`

PR #488, opened 2021-09-26
Add assert_shape/2 and assert_shape_pattern/2 to Nx.Defn.Kernel

josevalim:


nx/lib/nx/defn/kernel.ex
```diff
@@ -1013,4 +1013,129 @@ defmodule Nx.Defn.Kernel do
...
+
+      iex> assert_shape_pattern Nx.tensor([1, 2]), {x, y} when x > y
+      ** (ArgumentError) expected tensor to match shape {x, y} when x > y, got tensor with shape {2}
+
+  """
+  defmacro assert_shape_pattern(tensor, shape) do
+    shape_pattern_string = shape_pattern_to_string(shape)
+
+    quote do
+      tensor = unquote(tensor)
```
jonatanklosko, line 1115: Any reason not to use `:bind_quoted`?

jonatanklosko (commented):
Awesome!!

nx/lib/nx/defn/kernel.ex
josevalim: bind_quoted disables unquoting and we need to unquote the shape inside the pattern position. So I would need to pass both bind_quoted and enable unquoting and that would be more verbose. :)

jonatanklosko: Ah right, thanks!

nx/lib/nx/defn/kernel.ex
```diff
@@ -1013,4 +1013,129 @@ defmodule Nx.Defn.Kernel do
...
+          [1],
+          [2]
+        ]
+      >
+
+      iex> assert_shape_pattern Nx.tensor([1, 2]), {x, y} when x > y
+      ** (ArgumentError) expected tensor to match shape {x, y} when x > y, got tensor with shape {2}
+
+  """
+  defmacro assert_shape_pattern(tensor, shape) do
```
seanmor5, line 1111: Is it possible to extend this such that we can support multiple patterns? e.g. a tensor can be `{_, _} or {_, _, _}`, etc.

seanmor5 (commented):
Nice! Just the one comment on extending to multiple patterns.

nx/lib/nx/defn/kernel.ex
seanmor5: Or perhaps we can support a wildcard pattern for arbitrary ranks, e.g. for `conv` the shape of the `input` must be:

```elixir
{batch, features, ...any spatial ranks...}
```

And the kernel needs:

```elixir
{out_features, features, ...kernel size...}
```

josevalim: @seanmor5 you would be able to do this:

    assert_shape_pattern tensor, shape when tuple_size(shape) in [2, 3]

But I w7ld say at this point we are way too much into the weeds. Probably `assert_rank` is necessary in those cases?

polvalente: I think what @seanmor5 is getting into is that sometimes we want to assert that the shape ends in a certain shape, like when we work with tensor batches.

In these cases a simple assertion on the rank wouldn't suffice. For lists this is easy through `[a, b | _]` matching, though I'm not sure if we can (or want to) have this kind of ellipsis matching

josevalim: We could introduce a new notation, such as `{..., a, b}` or something of sorts, but then we are straying a bit from Elixir. For now, you can do something like `tuple when elem(tuple, 0) == batch and elem(tuple, 1) == features` but I don't think that's much better?

polvalente: Yeah... Perhaps we could have assert_shape also accept a list, which would allow for at least cons matching.
Not sure about that, though.

I do think that the best way for more specific cases is using elem.

Perhaps allowing a %{index => axis_size} map to be passed could be a compromise. 

For maps it would be a sequence of assert elem(shape, index) == axis_size

josevalim: The list doesn't allow suffix match, so it would help in some cases but not others. If we are going to add special semantics, then I think `...` is probably the route I would go. The big question for now is if we think this function is useful as is or if we want to wait until we can match on those more complex patterns?

seanmor5: I have done more thinking about this and I think the need for these kind of special assertions becomes less necessary with named Tensors because we already have rules for how dimensions should interact based on matching names. 

So I think we is the current implementation is fine :)

josevalim:

josevalim:
Btw @seanmor5, the heisenbug I am seeing on CI also happens locally for me on CPU. Maybe the binary is being deallocated/changed too early?

zacky1972:
Thank you for implementing.

The following code works well:

```
  defn add_tensor_and_scalar(t, s) do
    assert_shape(s, {})
    Nx.add(t, s)
  end
```

However, the following code doesn't works well:

```
  defn add_and_matrix(m1, m2) do
    assert_shape_pattern(m1, {_, _})
    assert_shape_pattern(m2, {_, _})
    Nx.add(m1, m2)
  end
```

I got the following error:

```
== Compilation error in file lib/defn_test.ex ==
** (CompileError) lib/defn_test.ex:26: invalid numerical expression:

    case(Nx.shape(tensor)) do
      {_, _} -> 
        tensor
      shape ->
        Nx.Defn.Kernel.__assert_shape_pattern__!("match shape {_, _}", shape)
    end

    (nx 0.1.0-dev) lib/nx/defn/compiler.ex:735: Nx.Defn.Compiler.compile_error!/3
    (elixir 1.12.2) lib/enum.ex:1704: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (elixir 1.12.2) lib/enum.ex:1704: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (nx 0.1.0-dev) lib/nx/defn/compiler.ex:462: Nx.Defn.Compiler.normalize/2
    (nx 0.1.0-dev) lib/nx/defn/compiler.ex:485: anonymous fn/3 in Nx.Defn.Compiler.normalize/2
    (elixir 1.12.2) lib/enum.ex:1704: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (nx 0.1.0-dev) lib/nx/defn/compiler.ex:483: Nx.Defn.Compiler.normalize/2
    (nx 0.1.0-dev) lib/nx/defn/compiler.ex:541: Nx.Defn.Compiler.normalize/2
** (exit) shutdown: 1
    (mix 1.12.2) lib/mix/tasks/compile.all.ex:78: Mix.Tasks.Compile.All.compile/4
    (mix 1.12.2) lib/mix/tasks/compile.all.ex:59: Mix.Tasks.Compile.All.with_logger_app/2
    (mix 1.12.2) lib/mix/tasks/compile.all.ex:36: Mix.Tasks.Compile.All.run/1
    (mix 1.12.2) lib/mix/task.ex:394: anonymous fn/3 in Mix.Task.run_task/3
    (mix 1.12.2) lib/mix/tasks/compile.ex:128: Mix.Tasks.Compile.run/1
    (mix 1.12.2) lib/mix/task.ex:394: anonymous fn/3 in Mix.Task.run_task/3
    (iex 1.12.2) lib/iex/helpers.ex:106: IEx.Helpers.recompile/1
```

zacky1972:
@josevalim fixes the above-mentioned issue. Thanks!

PR #489, opened 2021-09-27
feat(torchx): add to_batched_list and concatenate

polvalente:


torchx/lib/torchx/backend.ex
```diff
@@ -125,11 +125,56 @@ defmodule Torchx.Backend do
...
+
+        true ->
+          t
+      end
+
+    batched =
+      to_batch
+      |> from_nx()
+      |> Torchx.split(batch_size)
+      |> Enum.take(num_batches)
```
polvalente, line 169: @josevalim I'm not sure what's happening, but the `slice` and `concatenate` above are effectively concatenating two full copies of the tensor. This `Enum.take` ends up being necessary because otherwise I get more batches than needed. Is this "soft slicing" expected?

This seems to be a behavior with slice itself, which can be checked by manually extending the shape of the resulting tensor.

torchx/lib/torchx.ex
```diff
@@ -336,6 +353,12 @@ defmodule Torchx do
       {dev, ref}, other_dev when is_tensor(dev, ref) ->
         raise ArgumentError, "cannot perform operation across devices #{dev} and #{other_dev}"
 
+      [{dev, ref} | _] = tensors, nil when is_tensor(dev, ref) ->
+        prepare_tensors_list!(tensors, dev)
+
+      tensors, dev when is_list(tensors) ->
+        prepare_tensors_list!(tensors, dev)
+
```
josevalim, line 361: One alternative is to make `prepare_tensors!` recursive:

```elixir
  defp prepare_tensors!(tensors, dev \\ nil) do
    Enum.map_reduce(tensors, dev, fn
      {dev, ref}, nil when is_tensor(dev, ref) ->
        {ref, dev}

      {dev, ref}, dev when is_tensor(dev, ref) ->
        {ref, dev}

      {dev, ref}, other_dev when is_tensor(dev, ref) ->
        raise ArgumentError, "cannot perform operation across devices #{dev} and #{other_dev}"

      tensors, dev when is_list(tensors) ->
        prepare_tensors!(tensors, dev)

      bad_tensor, _dev ->
        raise ArgumentError, "expected a Torchx tensor, got: #{inspect(bad_tensor)}"
    end)
  end
```

This way you don't need to introduce `prepare_tensors_list!`. It is your call on which approach you prefer.

torchx/lib/torchx/backend.ex
```diff
@@ -125,11 +125,56 @@ defmodule Torchx.Backend do
...
+        remainder != 0 and leftover == :repeat ->
+          concat_shape =
+            t.shape |> Tuple.delete_at(0) |> Tuple.insert_at(0, num_batches * batch_size)
+
+          slice_length = t_axis_0 - remainder
+          slice_shape = t.shape |> Tuple.delete_at(0) |> Tuple.insert_at(0, slice_length)
+
+          slice = slice(%{t | shape: slice_shape}, t, [0], [slice_length], [1])
+
+          concatenate(%{t | shape: concat_shape}, [t, slice], 0)
```
josevalim, line 159: My suggestion: avoid calling the backend itself because we need to do all of these shapes manipulation. Instead, call the Torchx.xyz function directly. If necessary, encapsulate the logic behind `Torchx.Backend.slice` into a private `torchx_slice` so you can reuse it here. This will hopefully make things easier to debug too. :)

torchx/lib/torchx.ex
polvalente: The issue is that `prepare_tensors` returns a list of `{ref, dev}`, and `prepare_tensors_list` returns `{list(ref), dev}`

I'll try to make things work after all tests pass :)

torchx/lib/torchx/backend.ex
polvalente:
```elixir
iex(9)> t = Nx.tensor([0, 0, 0, 1, 1, 1])
#Nx.Tensor<
  s64[6]
  Torchx.Backend(cpu)
  [0, 0, 0, 1, 1, 1]
>
iex(10)> s = Nx.slice(t, [0], [3])
#Nx.Tensor<
  s64[3]
  Torchx.Backend(cpu)
  [0, 0, 0]
>
iex(11)> Nx.concatenate([t, s, t])
#Nx.Tensor<
  s64[15]
  Torchx.Backend(cpu)
  [0, 0, 0, 1, 1, 1, 0, 0, 0, 1, 1, 1, 0, 0, 0]
>
iex(12)> Nx.concatenate([t, t, s])
#Nx.Tensor<
  s64[15]
  Torchx.Backend(cpu)
  [0, 0, 0, 1, 1, 1, 0, 0, 0, 1, 1, 1, 0, 0, 0]
>
```

The underlying issue is that `slice` is not actually getting rid of the underlying data of the tensor, and this is somehow leaking through `concatenate`, but can also be observed with `Torchx.to_blob/1`:

```elixir
iex(14)> Torchx.to_blob(Torchx.from_nx(s))
<<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0,
  0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0>>
iex(15)> s
#Nx.Tensor<
  s64[3]
  Torchx.Backend(cpu)
  [0, 0, 0]
>
```

torchx/lib/torchx.ex
josevalim: I see. So ignore me. We also don’t want to support lists of lists, so the current approach is better.

torchx/lib/torchx/backend.ex
josevalim: Maybe slice is returning a view only? And then when you concatenate it goes against the actual tensor?

polvalente: Yeah, I think that's the case. Even if I apply narrow_copy or .detach().clone().resize_(..) this same behavior persists, so probably there's another function for this use case.

polvalente: In the examples I found, narrow was always used for in-place updates

josevalim: Do we still need Enum.take after the slice fix?

torchx/lib/torchx/backend.ex
```diff
@@ -195,14 +232,25 @@ defmodule Torchx.Backend do
...
-    |> stride(shape, lengths, strides)
+    |> do_slice(input_shape, output_shape, start_indices, lengths, strides)
     |> to_nx(out)
   end
 
+  defp do_slice(t, input_shape, output_shape, start_indices, lengths, strides) do
+    t
+    |> narrow(start_indices, lengths, 0, input_shape)
+    |> stride(output_shape, lengths, strides)
+  end
```
josevalim, line 252: I recommend avoiding the `do_` whenever possible cause it is not very clear. My suggestion:

```suggestion
  defp torchx_slice(ref, input_shape, output_shape, start_indices, lengths, strides) do
    ref
    |> narrow(start_indices, lengths, 0, input_shape)
    |> stride(output_shape, lengths, strides)
  end
```

PR #492, opened 2021-09-28
fix(nx): slice strides along first dim

polvalente:


nx/lib/nx/binary_backend.ex
```diff
@@ -1687,23 +1687,56 @@ defmodule Nx.BinaryBackend do
...
-      # with respect to the stride for each dimension
-      weighted_shape =
-        Enum.zip_with(weighted_shape, strides, fn {d, dim_size}, s ->
-          {d, dim_size + (s - 1) * dim_size}
-        end)
-
-      IO.iodata_to_binary(weighted_traverse(weighted_shape, data, size, offset))
+    is_top_dim_slice = top_dimension_slice?(tuple_size(shape), shape, output_shape)
+
+    cond do
```
polvalente, line 1692: Turns out we can also just revert the top-level optimization as well. The `else`-block code works properly for all axes.
Another possibility is that we can apply this optimization only if the top-level stride is 1, as the first clause below.

I'm leaning towards this second possibility, because the "optimized" code for the strided case doesn't seem optimized at all

seanmor5: Yeah the strided case makes things difficult, I'm okay with getting rid of the optimization in favor of the simpler general case. WDYT @josevalim ?

josevalim: Ditching the optimization sounds good to me. It is your call!

josevalim: Actually, let's go with the idea of only applying the optimization if the top level stride is 1. We likely want to keep this fast anyway because of the access syntax.

nx/lib/nx.ex
```diff
@@ -7385,45 +7385,40 @@ defmodule Nx do
...
-        [
-          [
-            [182, 183, 184]
-          ],
-          [
-            [632, 633, 634]
-          ]
-        ]
+        s64[3]
+        [0, 1, 2]
```
josevalim, line 7391:
```suggestion
        [1, 2, 3]
```

josevalim (approved):

PR #494, opened 2021-09-30
feat: auto-download libtorch and better compilation for torchx

polvalente:
As per the PyTorch docs, the recommended way to install libtorch is with CMake (https://pytorch.org/cppdocs/installing.html).

I tried using plain Make for this, but I kept running into linking issues.
The only downside I see for the time being is that we now require CMake 3+ for Torchx compilation.


polvalente:
This is mostly working, but there's some adjustments needed to the CI workflow which I'll look into tomorrow

josevalim:
I would prefer to not request additional dependencies. Where is the need for CMake coming from? If you simply bump to 1.9.1, without any change to the Makefile, it doesn't compile at all?

Also, I think the downloading code can be written in Elixir. You can change mix.exs to something like this:

    compilers: [:libtorch, :makefile] ++ ...,
    aliases: ["compile.libtorch": &download_libtorch/1]

polvalente:
> I would prefer to not request additional dependencies. Where is the need for CMake coming from? If you simply bump to 1.9.1, without any change to the Makefile, it doesn't compile at all?
> 
> Also, I think the downloading code can be written in Elixir. You can change mix.exs to something like this:
> 
>     compilers: [:libtorch, :makefile] ++ ...,
>     aliases: ["compile.libtorch": &download_libtorch/1]
> 
> 

Without cmake I couldn't get things to link properly outside of libtorch 1.8.2 pre CXX11 ABI. With cmake both 1.9.1 and 1.8.2 with or without CXX11 ABI work fine

polvalente:
Closing this in favor of #495

PR #495, opened 2021-09-30
feat: auto download libtorch

polvalente:
Alternative implementation to #494.
This one doesn't require CMake, but is more verbose and more restrictive on which versions are accepted


torchx/Makefile
```diff
@@ -14,25 +13,28 @@ TORCHX_SO = $(PRIV_DIR)/torchx.so
 ifeq ($(shell uname -s), Darwin)
 	LDFLAGS += -flat_namespace -undefined suppress
 	POST_INSTALL = install_name_tool -add_rpath @loader_path/lib $(TORCHX_SO)
+	LIBTORCH_DOWNLOAD_URL ?= https://download.pytorch.org/libtorch/cpu/libtorch-macos-1.9.1.zip
 else  # Linux
-    CFLAGS += -fopenmp
-	LDFLAGS += -Wl,-rpath,'$$ORIGIN/lib'
+    CFLAGS += -isystem $(LIBTORCH_INCLUDE_PATH) -isystem $(LIBTORCH_API_INCLUDE_PATH) -isystem $(ERTS_INCLUDE_DIR)
```
polvalente, line 18: I think all of these can just be moved to the CFLAGS at L6

torchx/Makefile
```diff
@@ -14,25 +13,28 @@ TORCHX_SO = $(PRIV_DIR)/torchx.so
...
 
-build: check_libtorch $(TORCHX_SO)
+build: libtorch $(TORCHX_SO)
 
-check_libtorch:
-	@ if [ -z "$(LIBTORCH_DIR)" ]; then \
-		echo "LIBTORCH_DIR env var is not set. It should point to the directory, where LibTorch is installed."; \
-		echo "You can download LibTorch here: https://pytorch.org/get-started/locally/"; \
-		exit 1; \
+libtorch:
```
polvalente, line 28: After we get this makefile working, we can move this to download via Elixir as suggested in #494

.github/workflows/ci.yml
```diff
@@ -19,6 +19,7 @@ jobs:
     env:
       MIX_ENV: test
       LIBTORCH_DIR: libtorch
+      LIBTORCH_ZIP: libtorch.zip
```
josevalim, line 22: In theory we don't need this var? :)

polvalente: Yeah, although I just realized the cache isn't working... Perhaps @jonatanklosko can help here?

torchx/Makefile
```diff
@@ -1,9 +1,10 @@
+LIBTORCH_DIR ?= cache/libtorch
 LIBTORCH_INCLUDE_PATH = $(LIBTORCH_DIR)/include/
 LIBTORCH_API_INCLUDE_PATH = $(LIBTORCH_DIR)/include/torch/csrc/api/include
 LIBTORCH_LINK_PATH = $(LIBTORCH_DIR)/lib/
 CFLAGS = -fPIC -I$(ERTS_INCLUDE_DIR) -I$(LIBTORCH_INCLUDE_PATH) -I$(LIBTORCH_API_INCLUDE_PATH) \
 	 -O3 -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -shared \
-	 -std=c++14 -D_GLIBCXX_USE_CXX11_ABI=0
+	 -std=c++14 -D_GLIBCXX_USE_CXX11_ABI=1
```
polvalente, line 7: @josevalim please check if this change affects the MacOS build. If it does we can move this to the if-else below

torchx/mix.exs
```diff
@@ -43,4 +44,145 @@ defmodule Torchx.MixProject do
...
+  defp aliases do
+    [
+      "compile.torchx": &compile/1
+    ]
+  end
+
+  defp compile(args) do
+    cache_dir = Path.join(__DIR__, "cache")
+
+    force_download = "--force" in args
```
polvalente, line 57: @josevalim should we ever react to `--force`? I don't know if this will actually be passed if a user runs `mix compile --force` or just `mix deps.compile --force`  having torchx as a dependency

.github/workflows/ci.yml
jonatanklosko: You need to update the cache action `with.path` to the new cache directory path, or did I miss it?

torchx/mix.exs
```diff
@@ -43,4 +44,145 @@ defmodule Torchx.MixProject do
...
+
+    # Unpack libtorch to libtorch_dir
+
+    {_, 0} =
+      System.cmd("unzip", ~w(-d #{cache_dir} #{libtorch_zip})) |> IO.inspect(label: "unzip")
+
+    unpack_dir = Path.join(cache_dir, "libtorch") |> IO.inspect(label: "unpack dir")
+
+    unless unpack_dir == libtorch_dir do
+      {_, 0} = System.cmd("mv", ~w(#{unpack_dir} #{libtorch_dir}))
```
jonatanklosko, line 142: You can use `File.mv!` :)

torchx/mix.exs
```diff
@@ -43,4 +44,145 @@ defmodule Torchx.MixProject do
...
+            "https://download.pytorch.org/libtorch/cpu/libtorch-macos-1.9.1.zip"
+        end
+
+      download!(url, libtorch_zip)
+    end
+
+    # Unpack libtorch to libtorch_dir
+
+    {_, 0} =
+      System.cmd("unzip", ~w(-d #{cache_dir} #{libtorch_zip})) |> IO.inspect(label: "unzip")
```
jonatanklosko, line 137: Erlang has `:zip.unzip` that we can use.

.github/workflows/ci.yml
polvalente: Let me try this :)
I tested with the updated path, however, just forgot to keep this change in the "final" commit

torchx/mix.exs
polvalente: Huh, I didn't find it in the autocomplete

polvalente:
```elixir
iex(1)> File.     
CopyError        Error            LinkError        RenameError      
Stat             Stream           cd!/1            cd!/2            
cd/1             chgrp!/2         chgrp/2          chmod!/2         
chmod/2          chown!/2         chown/2          close/1          
copy!/2          copy!/3          copy/2           copy/3           
cp!/2            cp!/3            cp/2             cp/3             
cp_r!/2          cp_r!/3          cp_r/2           cp_r/3           
cwd!/0           cwd/0            dir?/1           dir?/2           
exists?/1        exists?/2        ln!/2            ln/2             
ln_s!/2          ln_s/2           ls!/0            ls!/1            
ls/0             ls/1             lstat!/1         lstat!/2         
lstat/1          lstat/2          mkdir!/1         mkdir/1          
mkdir_p!/1       mkdir_p/1        open!/1          open!/2          
open!/3          open/1           open/2           open/3           
read!/1          read/1           read_link!/1     read_link/1      
regular?/1       regular?/2       rename!/2        rename/2         
rm!/1            rm/1             rm_rf!/1         rm_rf/1          
rmdir!/1         rmdir/1          stat!/1          stat!/2          
stat/1           stat/2           stream!/1        stream!/2        
stream!/3        touch!/1         touch!/2         touch/1          
touch/2          write!/2         write!/3         write/2          
write/3          write_stat!/2    write_stat!/3    write_stat/2     
write_stat/3  
```

This is on Elixir 1.12.2

jonatanklosko: Ah sorry, it's `rename`!

polvalente: Nice!

.github/workflows/ci.yml
```diff
@@ -36,18 +35,17 @@ jobs:
...
         with:
-          path: libtorch
-          key: libtorch.zip
-      - name: Download and install libtorch
-        if: ${{ matrix.working_directory == 'torchx' && steps.libtorch-cache.outputs.cache-hit != 'true' }}
-        run: |
-          wget https://download.pytorch.org/libtorch/lts/1.8/cpu/libtorch-shared-with-deps-1.8.2%2Bcpu.zip -q -O libtorch.zip
-          unzip -qq libtorch.zip
+          path: ./torchx/cache/libtorch
+          key: ./torchx/cache/libtorch.zip
```
polvalente, line 39: @jonatanklosko the cache action isn't reacting to `matrix.working-directory` properly! Now I got a cache hit :)

polvalente: (and we didn't re-download libtorch because of it)

jonatanklosko: Oh, turns out `defaults.run` doesn't apply to actions (`uses`). So we need to update the mix cache action too by manually building path relative to matrix.working_directory

torchx/mix.exs
josevalim: Let's keep this but I don't think it can really be called from a dep. It may be useful for local dev though.

polvalente: Nice. I was worried deps.compile --force or something would be passed into the torchx.compile task. But I just checked and deps.compile --force just removes the dep directory beforehand, so this `--force` won't do any harm outside of torchx development

torchx/Makefile
josevalim: It works with both values and without, so it probably makes sense to move it to the Linux one anyway?

torchx/mix.exs
```diff
@@ -12,8 +12,9 @@ defmodule Torchx.MixProject do
       elixir: "~> 1.12-dev",
       deps: deps(),
       docs: docs(),
-      compilers: [:elixir_make] ++ Mix.compilers(),
-      elixirc_paths: elixirc_paths(Mix.env())
+      compilers: [:torchx, :elixir_make] ++ Mix.compilers(),
```
josevalim, line 15: Here is what I have been thinking:

```suggestion
      compilers: compilers() ++ Mix.compilers(),
```

where:

```elixir
  defp compilers do
    if System.get_env("LIBTORCH_DIR"), do: [:elixir_make], else: [:torchx, :elixir_make]
  end

  defp aliases do
    [
      "compile.torchx": &compile/1
    ]
  end

  defp compile(args) do
    cache_dir = Path.join(__DIR__, "cache")
    # TODO: This directory name must also include the libtorch version.
    # Otherwise if someone depends on torchx as a git dependency, they
    # can upgrade the version and we will think it is cached but it is actually old.
    libtorch_dir = Path.join(cache_dir, "libtorch")

    if "--force" in args do
      File.rm_rf!(cache_dir)
    end

    unless File.dir?(libtorch_dir) do
      install_libtorch(cache_dir, libtorch_dir)
    end
  end
```

If people setup a bad LIBTORCH_DIR, then it is on them. :)

torchx/mix.exs
```diff
@@ -43,4 +44,145 @@ defmodule Torchx.MixProject do
...
+    end
+  end
+
+  defp install_libtorch(cache_dir, libtorch_dir) do
+    File.mkdir_p!(cache_dir)
+
+    libtorch_zip = System.get_env("LIBTORCH_ZIP", Path.join(cache_dir, "libtorch.zip"))
+
+    os =
+      case System.cmd("uname", ~w(-s)) do
```
josevalim, line 105: Please use :os.type.

torchx/mix.exs
polvalente: @josevalim I've commited this + a file which indicates the default libtorch version, to be used for the default zip and dir by both the Makefile and mix.exs. Let me know if I should revert this last addition :)

torchx/Makefile
```diff
@@ -1,9 +1,9 @@
+LIBTORCH_DIR ?= cache/libtorch_$(shell cat .default_libtorch_version)
 LIBTORCH_INCLUDE_PATH = $(LIBTORCH_DIR)/include/
 LIBTORCH_API_INCLUDE_PATH = $(LIBTORCH_DIR)/include/torch/csrc/api/include
 LIBTORCH_LINK_PATH = $(LIBTORCH_DIR)/lib/
 CFLAGS = -fPIC -I$(ERTS_INCLUDE_DIR) -I$(LIBTORCH_INCLUDE_PATH) -I$(LIBTORCH_API_INCLUDE_PATH) \
-	 -O3 -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -shared \
-	 -std=c++14 -D_GLIBCXX_USE_CXX11_ABI=0
```
polvalente, line 6: turns out thix CXX11_ABI thing wasn't needed after all

torchx/mix.exs
```diff
@@ -3,6 +3,7 @@ defmodule Torchx.MixProject do
 
   @source_url "https://github.com/elixir-nx/nx"
   @version "0.1.0-dev"
+  @default_libtorch_version File.read!(".default_libtorch_version")
```
josevalim, line 6: Why a file? I don't think Elixir makes it easy to configure a file inside a dependency. Instead it uses environment variables: `{:torchx, ..., system_env: %{"LIBTORCH_TARGET" => "cpu"}}`.

polvalente: So we can have the same source of truth for the makefile as well. This isn't supposed to be configurable anyway, it's just for better defaults on Torchx's side

polvalente: I added this because I ran into the exact issue which this prevents, in which mix.exs is updated but the makefile default value is not

josevalim: Oh, it is a file because you want to share it with the makefile. This is not a problem! It is possible to pass environment variables to the makefile. See https://hexdocs.pm/elixir_make. We should pass the LIBTORCH_DIR downstream and assume it is set in the `Makefile`.

josevalim: Something like this:

```elixir
@libtorch_version "1.9.1"
@libtorch_target "cpu"
@libtorch_cache Path.expand(__DIR__, "cache/libtorch-#{@libtorch_version}-#{@libtorch_target}")

def project do
  [
    ...,
    make_env: %{"LIBTORCH_DIR" => System.get_env("LIBTORCH_DIR") || @libtorch_cache}
  ]
```

josevalim: Then cache_dir is `Path.dirname(@libtorch_cache)`. :)

polvalente: Oh, nice! I'll change this as soon as I get back on the PC. Besides this, is eveything ok with this PR? I want to kerged it asap because it blocks further development a little bit

josevalim: Yes, everything else looks good. If you want to wait for this, then just revert this change. The important is that we change the filename BEFORE we bump the version/target again.

torchx/mix.exs
```diff
@@ -43,4 +45,119 @@ defmodule Torchx.MixProject do
...
+
+  defp compile(args) do
+    cache_dir = Path.join(__DIR__, "cache")
+    # TODO: This directory name must also include the libtorch version.
+    # Otherwise if someone depends on torchx as a git dependency, they
+    # can upgrade the version and we will think it is cached but it is actually old.
+    libtorch_dir = Path.join(cache_dir, "libtorch_#{@default_libtorch_version}")
+
+    if "--force" in args do
+      File.rm_rf!(cache_dir)
```
josevalim, line 67:
```suggestion
      File.rm_rf!(libtorch_dir)
```

torchx/mix.exs
```diff
@@ -43,4 +45,119 @@ defmodule Torchx.MixProject do
...
+    end
+
+    # Unpack libtorch to libtorch_dir
+
+    {:ok, _} =
+      libtorch_zip |> String.to_charlist() |> :zip.unzip(cwd: String.to_charlist(cache_dir))
+
+    unpack_dir = Path.join(cache_dir, "libtorch")
+
+    unless unpack_dir == libtorch_dir do
```
josevalim, line 116: This can never be true, can it?

polvalente: This made more sense when we always ran torchx.compile. Basically, we would always download and unpack to cache/libtorch, and this File.rename would then unpack to the specified directory if the env var wasn't a valid directory originally.

I'll remove this (and the rename) because now we either download and unpack specifically to cache/libtorch, or we just skip the whole download/unpacking thing

torchx/mix.exs
polvalente: I think we want to remove cache dir itself. Otherwise libtorch won't ever be downloaded again without manually erasing cache/libtorch.zip

torchx/mix.exs
polvalente: No worries, we can merge this when it's actually ready :)

torchx/mix.exs
josevalim: Good call!

torchx/mix.exs
```diff
@@ -43,4 +49,113 @@ defmodule Torchx.MixProject do
...
+  end
+
+  defp install_libtorch(cache_dir) do
+    File.mkdir_p!(cache_dir)
+
+    libtorch_zip =
+      System.get_env(
+        "LIBTORCH_ZIP",
+        Path.join(cache_dir, "libtorch-#{@libtorch_version}-#{@libtorch_target}.zip")
+      )
```
josevalim, line 84:
```suggestion
  defp compile(args) do
    libtorch_cache = @libtorch_cache
    cache_dir = Path.dirname(libtorch_cache)

    if "--force" in args do
      File.rm_rf!(cache_dir)
    end

    if File.dir?(libtorch_cache) do
      {:ok, []}
    else
      install_libtorch(cache_dir, libtorch_cache)
    end
  end

  defp install_libtorch(cache_dir, libtorch_cache) do
    File.mkdir_p!(cache_dir)
    libtorch_zip = libtorch_cache <> ".zip"
```

torchx/mix.exs
```diff
@@ -43,4 +49,113 @@ defmodule Torchx.MixProject do
...
+      download!(url, libtorch_zip)
+    end
+
+    # Unpack libtorch and move to the target cache dir
+
+    {:ok, _} =
+      libtorch_zip |> String.to_charlist() |> :zip.unzip(cwd: String.to_charlist(cache_dir))
+
+    # Keep libtorch cache scoped by version and target
+    File.rename!(Path.join(cache_dir, "libtorch"), @libtorch_cache)
```
josevalim, line 116: If you merge the above, then do this too:

```suggestion
    File.rename!(Path.join(cache_dir, "libtorch"), libtorch_cache)
```

torchx/mix.exs
josevalim: In particular, note that setting LIBTORCH_ZIP doesn't quite work if you set it to anything outside of the cache dir, so we should go ahead and remove it for now. :)

josevalim:
Btw, I slightly messed up. I wanted to push a README update to this branch saying we automatically download stuff but I pushed it to main instead. :D So no worries about updating the README, it has already been done. :D

josevalim (approved):

Issue #358, opened 2021-03-29
Sigils for tensors


dannote:
Some time ago, I decided to implement a simple library for operations on tensors, and I found that curly braces and commas are too verbose, especially in tests, for instance:

```elixir
assert Nx.tensor([
  [1.0, -1.0, 4.0],
  [1.0, 4.0, -2.0],
  [1.0, 4.0, 2.0],
  [0.0, 0.0, 0.0]
]) == q
```

So I came up with this simple macro:

```elixir
defmacro sigil_m({:<<>>, _meta, [string]}, _) do
  string
  |> split_numbers()
  |> Nx.tensor()
end

def split_numbers(string) do
  for row <- String.split(string, "\n", trim: true) do
    row
    |> String.split(" ", trim: true)
    |> Enum.map(&Code.eval_string(&1) |> elem(0))
  end
end
```

Now you can write instead:

```elixir
~m(1.0 -1.0  4.0
   1.0  4.0 -2.0
   1.0  4.0  2.0
   0.0  0.0  0.0)
```
We can make it even better by adding a modifier for transposing tensors:

```elixir
defmacro sigil_m({:<<>>, _meta, [string]}, [?t]) do
  string
  |> split_numbers()
  |> Nx.tensor()
  |> Nx.transpose()
end

iex(1)> assert ~m(-1 2 3)t == ~m(-1
...(1)>                           2
...(1)>                           3)
true
```

If you like the idea, I can create a PR.

josevalim:
I actually like this idea, even if restricted for matrices, I can see how it is useful. I would love to hear other people's thoughts.

We should probably go only with `~M` though, especially if we don't support interpolation.

seanmor5:
+1 to what @josevalim said. I would also be interested to see what other modifiers besides transpose could be added to extend this specifically for matrices. E.g. determinants, inverses, etc.

cc @polvalente for his thoughts and LinAlg expertise

polvalente:
Sigils are a great idea! The things I currently miss the most for translating math formulas into Nx code in a "less programatic/more descriptive" way are basically the ones @seanmor5 cited:
  - Matrix exponent notation (e.g. `A^2 === dot(A, A)`, `A^-1 === inverse(A)`, `A^-2 === inverse(dot(A, A))`);
  - Conjugate tensor (this will be **really** helpful when we support complex numbers/expressions;
[...]

PR #498, opened 2021-10-03
Add sigil for tensors

dannote:
This PR adds `~M[]` sigil that evaluates to `#Nx.Tensor<>` as discussed in #358, but with type in the modifier.

I've added conditional compilation since support for digits in sigils was just merged to Elixir (elixir-lang/elixir#11283).

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -8483,7 +8483,17 @@ defmodule Nx do
       for row <- String.split(string, "\n", trim: true) do
         row
         |> String.split(" ", trim: true)
-        |> Enum.map(&elem(Code.eval_string(&1), 0))
+        |> Enum.map(fn str ->
+          module = if String.contains?(str, "."), do: Float, else: Integer
+
+          case module.parse(str) do
+            {number, ""} ->
+              number
+
+            _ ->
+              raise ArgumentError, "expected a numerical value for tensor, got #{str}"
+          end
+        end)
       end
     end
   end
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -8534,10 +8534,13 @@ defmodule Nx do
 
   """
   defmacro sigil_V({:<<>>, _meta, [string]}, modifiers) do
-    string
-    |> binary_to_numbers
-    |> hd()
-    |> numbers_to_tensor(modifiers)
+    case binary_to_numbers(string) do
+      [numbers] ->
+        numbers_to_tensor(numbers, modifiers)
+
+      _ ->
+        raise ArgumentError, "must be one-dimensional"
+    end
   end
 
   defp numbers_to_tensor(numbers, modifiers) do
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -8454,18 +8454,18 @@ defmodule Nx do
   ## Sigils
 
   @doc """
-  Provides `~M` sigil to build a tensor.
-
-  > Note: `~M` requires Elixir >= `1.13.0` in order to suppport type specifiers like `u8`
+  A convenient `~M` sigil for building matrices.
 
   ## Examples
 
       import Nx
 
-      ~M[-1 0 0 1
-          0 2 0 0
-          0 0 3 0
-          0 0 0 4]
+      ~M<
+        -1 0 0 1
+        0 2 0 0
+        0 0 3 0
+        0 0 0 4
+      >
 
   Is equivalent to:
 
@@ -8476,7 +8476,9 @@ defmodule Nx do
         [0, 0, 0, 4]
       ])
 
-  You can specify the tensor type:
+  If the tensor has any float type, it defaults to f32.
+  Otherwise, it is s64. If you are using Elixir 1.13+,
+  you can specify the tensor type as a sigil modifier:
 
       iex> import Nx
       iex> ~M[0.1 0.2 0.3 0.4]f16
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -8453,48 +8453,46 @@ defmodule Nx do
 
   ## Sigils
 
-  if Version.match?(System.version(), ">= 1.13.0-dev") do
-    defmacro sigil_M({:<<>>, _meta, [string]}, modifiers) do
-      numbers =
-        case binary_to_numbers(string) do
-          [vector] -> vector
-          matrix -> matrix
-        end
-
-      type =
-        case modifiers do
-          [unit | size] ->
-            Nx.Type.normalize!({List.to_atom([unit]), List.to_integer(size)})
-
-          [] ->
-            Nx.Type.infer(numbers)
-        end
-
-      {shape, binary} = flatten(numbers, type)
-
-      quote do
-        unquote(binary)
-        |> Nx.from_binary(unquote(type))
-        |> Nx.reshape(unquote(Macro.escape(shape)))
+  defmacro sigil_M({:<<>>, _meta, [string]}, modifiers) do
+    numbers =
+      case binary_to_numbers(string) do
+        [vector] -> vector
+        matrix -> matrix
       end
+
+    type =
+      case modifiers do
+        [unit | size] ->
+          Nx.Type.normalize!({List.to_atom([unit]), List.to_integer(size)})
+
+        [] ->
+          Nx.Type.infer(numbers)
+      end
+
+    {shape, binary} = flatten(numbers, type)
+
+    quote do
+      unquote(binary)
+      |> Nx.from_binary(unquote(type))
+      |> Nx.reshape(unquote(Macro.escape(shape)))
     end
+  end
 
-    defp binary_to_numbers(string) do
-      for row <- String.split(string, "\n", trim: true) do
-        row
-        |> String.split(" ", trim: true)
-        |> Enum.map(fn str ->
-          module = if String.contains?(str, "."), do: Float, else: Integer
+  defp binary_to_numbers(string) do
+    for row <- String.split(string, "\n", trim: true) do
+      row
+      |> String.split(" ", trim: true)
+      |> Enum.map(fn str ->
+        module = if String.contains?(str, "."), do: Float, else: Integer
 
-          case module.parse(str) do
-            {number, ""} ->
-              number
+        case module.parse(str) do
+          {number, ""} ->
+            number
 
-            _ ->
-              raise ArgumentError, "expected a numerical value for tensor, got #{str}"
-          end
-        end)
-      end
+          _ ->
+            raise ArgumentError, "expected a numerical value for tensor, got #{str}"
+        end
+      end)
     end
   end
 
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -8453,6 +8453,39 @@ defmodule Nx do
 
   ## Sigils
 
+  @doc """
+  Provides `~M` sigil to build a tensor.
+
+  > Note: `~M` requires Elixir >= `1.13.0` in order to suppport type specifiers like `u8`
+
+  ## Examples
+
+      import Nx
+
+      ~M[-1 0 0 1
+          0 2 0 0
+          0 0 3 0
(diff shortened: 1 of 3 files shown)


nx/lib/nx.ex
```diff
@@ -8451,6 +8451,43 @@ defmodule Nx do
     )
   end
 
+  ## Sigils
+
+  if Version.match?(System.version(), ">= 1.13.0-dev") do
+    defmacro sigil_M({:<<>>, _meta, [string]}, modifiers) do
```
polvalente, line 8457: Nice work! 

This implementation only supports a literal tensor, without resolving interpolations. Is this intentional?
Perhaps a better error could be raised in such case. For now the behavior is:

```elixir
iex(7)> x = 1       
1
iex(8)> ~M[#{x} 2]
** (ArgumentError) cannot infer the numerical type of nil
    (nx 0.1.0-dev) lib/nx/type.ex:110: Nx.Type.infer/2
    (elixir 1.13.0-dev) lib/enum.ex:2394: Enum."-reduce/3-lists^foldl/2-0-"/3
    (nx 0.1.0-dev) lib/nx/type.ex:98: Nx.Type.infer/1
    (nx 0.1.0-dev) expanding macro: Nx.sigil_M/2
    iex:8: (file)
```

dannote: Yes, this is intentional. I followed the convention in [built-in sigils](https://hexdocs.pm/elixir/1.12/Kernel.html#sigil_C/2) and used uppercase "M" to indicate this behaviour.

polvalente: I would try to raise a more semantic message then. Otherwise, this seems good :)

polvalente: It seems that for non-numerical values the current message is fine:

```elixir
iex(11)> ~M[1 2 "a" 3]
** (ArgumentError) cannot infer the numerical type of "a"
    (nx 0.1.0-dev) lib/nx/type.ex:110: Nx.Type.infer/2
    (elixir 1.13.0-dev) lib/enum.ex:2394: Enum."-reduce/3-lists^foldl/2-0-"/3
    (nx 0.1.0-dev) lib/nx/type.ex:98: Nx.Type.infer/1
    (nx 0.1.0-dev) expanding macro: Nx.sigil_M/2
    iex:11: (file)
```

polvalente: I'm not sure if this is also a predicted edge case:

```elixir
iex(11)> x = 1
1
iex(12)> ~M[1 2 x 3]  
warning: variable "x" does not exist and is being expanded to "x()", please use parentheses to remove the ambiguity or change the variable name
  nofile:1

** (CompileError) nofile:1: undefined function x/0 (there is no such import)
    (elixir 1.13.0-dev) lib/code.ex:404: Code.validated_eval_string/3
    (nx 0.1.0-dev) lib/nx.ex:8486: anonymous fn/1 in Nx.binary_to_numbers/1
    (elixir 1.13.0-dev) lib/enum.ex:1591: Enum."-map/2-lists^map/1-0-"/2
    (elixir 1.13.0-dev) lib/enum.ex:1591: Enum."-map/2-lists^map/1-0-"/2
    (nx 0.1.0-dev) lib/nx.ex:8486: anonymous fn/2 in Nx.binary_to_numbers/1
    (nx 0.1.0-dev) expanding macro: Nx.sigil_M/2
```

It seems that the sigil is trying to expand the variable on a "raw" env instead of the caller env

polvalente: Perhaps instead of using Code.eval_string, you could try to parse it with Integer.parse and Float.parse like so:

```elixir
defp binary_to_numbers(string) do
      for row <- String.split(string, "\n", trim: true) do
        row
        |> String.split(" ", trim: true)
        |> Enum.map(fn str ->
          case Integer.parse(str) do
            {int, ""} ->
              int

            _ ->
              case Float.parse(str) do
                {float, ""} -> float
                _ -> raise ArgumentError, "expected a numerical value for tensor, got #{str}"
              end
          end
        end)
      end
    end
```

This conditional structure is perhaps not ideal, but you can take this as a starting point. Maybe a `with` statement could be of use here.

dannote: I tried to avoid it:
```elixir
defp binary_to_numbers(string) do
  for row <- String.split(string, "\n", trim: true) do
    row
    |> String.split(" ", trim: true)
    |> Stream.map(&Code.eval_string/1)
    |> Enum.map(fn
      {number, _} when is_number(number) -> number
      _ -> raise ArgumentError, "non-numerics and interpolations are not allowed"
    end)
  end
end
```

But there was another bad side effect:
```
iex(3)> ~M[x 2]
warning: variable "x" does not exist and is being expanded to "x()", please use parentheses to remove the ambiguity or change the variable name
  nofile:1
```

So I'll follow your advice, thanks :)

dannote: I think I found a more concise way

nx/lib/nx.ex
```diff
@@ -8451,6 +8451,53 @@ defmodule Nx do
     )
   end
 
+  ## Sigils
+
+  if Version.match?(System.version(), ">= 1.13.0-dev") do
```
josevalim, line 8456: We don't need this check. We can always define the sigil but we will note in the docs the type modifiers require 1.13.

nx/test/nx_test.exs
```diff
@@ -1628,4 +1628,69 @@ defmodule NxTest do
...
+        assert_raise(
+          ArgumentError,
+          "expected a numerical value for tensor, got x",
+          fn ->
+            defmodule Test3 do
+              import Nx
+
+              def case do
+                unquote(Code.string_to_quoted!("~M[1 2 x 4]u8"))
+              end
```
josevalim, line 1689:
```suggestion
              ~M[1 2 x 4]u8
```

This should be enough to test it!

nx/test/nx_test.exs
```diff
@@ -1628,4 +1628,69 @@ defmodule NxTest do
...
+        assert ~M[1.0 2  3
+                  11  12 13] == Nx.tensor([[1.0, 2, 3], [11, 12, 13]])
+      end
+
+      test "evaluates with proper type" do
+        defmodule Test do
+          import Nx
+
+          def case1 do
+            unquote(Code.string_to_quoted!("~M[1 2 3 4]f32"))
```
josevalim, line 1651: The unquote should not be needed here.

nx/test/nx_test.exs
```diff
@@ -1628,4 +1628,69 @@ defmodule NxTest do
...
+      test "evaluates to tensor" do
+        import Nx
+
+        assert ~M[-1 2 3 4] == Nx.tensor([-1, 2, 3, 4])
+        assert ~M[1
+                  2
+                  3
+                  4] == Nx.tensor([[1], [2], [3], [4]])
+        assert ~M[1.0 2  3
+                  11  12 13] == Nx.tensor([[1.0, 2, 3], [11, 12, 13]])
```
josevalim, line 1643: Please don't forget to format this file. :)

josevalim (commented):
The implementation looks great! I have dropped only some minor comments. :)

nx/test/nx_test.exs
dannote: It won't compile on `1.12` then:

```elixir
== Compilation error in file test/nx_test.exs ==
** (SyntaxError) test/nx_test.exs:1688:29: syntax error before: "8"
    (elixir 1.12.0) lib/kernel/parallel_compiler.ex:428: Kernel.ParallelCompiler.require_file/2
    (elixir 1.12.0) lib/kernel/parallel_compiler.ex:321: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/7
```

josevalim: I see. So the whole test can be `Code.eval_string("import Nx; ~M[1 2 3 4]f32")`. You don't need the wrapper module. :)

nx/test/nx_test.exs
dannote: That's how it looks after running `mix formatter`, I double-checked :)

nx/test/nx_test.exs
```diff
@@ -1628,4 +1628,51 @@ defmodule NxTest do
                )
     end
   end
+
+  if Version.match?(System.version(), ">= 1.13.0-dev") do
```
josevalim, line 1632: Move this conditional to only wrap around the tests that use eval and therefore requires the modifiers.

nx/test/nx_test.exs
josevalim: Interesting. It may be a formatter bug. :)

nx/lib/nx.ex
```diff
@@ -8451,6 +8451,51 @@ defmodule Nx do
     )
   end
 
+  ## Sigils
+
+  defmacro sigil_M({:<<>>, _meta, [string]}, modifiers) do
```
josevalim, line 8456: Please don’t forget documentation. :)

nx/test/nx_test.exs
```diff
@@ -1628,4 +1628,51 @@ defmodule NxTest do
                )
     end
   end
+
+  describe "sigil" do
+    test "evaluates to tensor" do
+      import Nx
+
+      assert ~M[-1 2 3 4] == Nx.tensor([-1, 2, 3, 4])
```
josevalim, line 1636: I have been thinking more about this and it seems to be impossible to specify a tensor that is a single row with multiple columns. Given the sigil is called `~M`, maybe this should always be a matrix? And we can introduce `~V` if we want to support vectors later on?

nx/lib/nx.ex
```diff
@@ -8451,6 +8451,86 @@ defmodule Nx do
...
   end
 
+  ## Sigils
+
+  @doc """
+  A convenient `~M` sigil for building matrices.
+
+  ## Examples
+
+      import Nx
```
josevalim, line 8461:
```suggestion
  Before using sigils, you must first import them:

      import Nx, only: [sigil_M: 2]

  If you are using Elixir v1.13, then you can also write:

      import Nx, only: :sigils

  Then you use the sigil to create matrices:
```

nx/test/nx_test.exs
dannote: Good point! Can I add `~V[-1 2 3 4]` as equivalent of `Nx.tensor([-1, 2, 3, 4])` now?

josevalim: Sure!

nx/lib/nx.ex
```diff
@@ -8451,6 +8451,128 @@ defmodule Nx do
     )
   end
 
+  ## Sigils
+
+  @doc """
+  A convenient `~M` sigil for building matrices.
```
josevalim, line 8457:
```suggestion
  A convenient `~M` sigil for building matrices (two-dimensional tensors).
```

nx/lib/nx.ex
```diff
@@ -8451,6 +8451,128 @@ defmodule Nx do
...
+  You can specify the vector type:
+
+      iex> import Nx
+      iex> ~V[0.1 0.2 0.3 0.4]f16
+      #Nx.Tensor<
+        f16[4]
+        [0.0999755859375, 0.199951171875, 0.300048828125, 0.39990234375]
+      >
+
+  """
```
josevalim, line 8531: I have pushed some improvements to `~M` docs, can you please mirror them here? :)

nx/lib/nx.ex
```diff
@@ -8451,6 +8451,128 @@ defmodule Nx do
...
+  defmacro sigil_V({:<<>>, _meta, [string]}, modifiers) do
+    string
+    |> binary_to_numbers
+    |> hd()
```
josevalim, line 8535: We should assert it is only a single row and raise otherwise.

nx/lib/nx.ex
dannote: Ooops

josevalim (approved):

josevalim:

PR #500, opened 2021-10-03
Add rem/2 to Nx.Defn.Kernel

dannote:


nx/lib/nx/defn/kernel.ex
```diff
@@ -403,6 +403,27 @@ defmodule Nx.Defn.Kernel do
...
+  It delegates to `Nx.remainder/2` (supports broadcasting).
+
+  ## Examples
+
+      defn divides_by_5?(a) do
+        rem(a, 5)
+        |> Nx.any?
+        |> Nx.equal(Nx.tensor(1))
+      end
+
```
polvalente, line 419:
```suggestion
```

polvalente: @josevalim I just noticed that the other docs in this file tend to end on a blank line. Does this make any difference?

nx/lib/nx/defn/kernel.ex
```diff
@@ -403,6 +403,26 @@ defmodule Nx.Defn.Kernel do
...
+  """
+
```
josevalim, line 420:
```suggestion
```

nx/lib/nx/defn/kernel.ex
josevalim: No, it does not matter, but that's the "convention" we follow in Elixir (and apparently Nx).

josevalim (approved):

nx/lib/nx/defn/kernel.ex
polvalente: I see!

josevalim:

PR #502, opened 2021-10-04
feat(torchx): add take

polvalente:


torchx/c_src/torchx.cpp
```diff
@@ -351,6 +351,19 @@ NIF(permute)
...
+  TENSOR_PARAM(0, t);
+  TENSOR_PARAM(1, i);
+  PARAM(2, int64_t, axis);
+
+  if (axis == 0)
+  {
+    TENSOR(t->index({*i}));
+  } else {
+    TENSOR(t->transpose(0, axis).index({*i}).transpose(0, axis));
+  }
```
polvalente, line 365: @josevalim I started writing this code, but ended up ditching it because I would have to basically rewrite the code which I moved into `Nx.Shape` here to make things work for axis != 0.

We have some options:

1. Ditch this C++ code altogether in favor of using `gather` (which is the current code, minus erasing this file)
2. Call this C++ version only for axis = 0 as a kind of optimization
3. Rewrite the "fully-qualified" code here + moving the "index offsets" thing which we use in gather to C++ so we can use here as well.

nx/lib/nx/shape.ex
```diff
@@ -1218,6 +1218,85 @@ defmodule Nx.Shape do
     {List.to_tuple(result_shape), result_names}
   end
 
+  @doc """
```
polvalente, line 1221: I'm not sure if this file is the best place for this function. Also, name suggestions are more than welcome :)

torchx/lib/torchx/backend.ex
```diff
@@ -296,6 +296,11 @@ defmodule Torchx.Backend do
     |> to_nx(out)
   end
 
+  @impl true
+  def take(out, t, i, axis) do
+    gather(out, t, Nx.Shape.take_fully_qualified_indices(t, i, axis))
```
josevalim, line 301: I am very sorry to inform you that the module `Nx.Shape` is private and we should not access it from `Torchx`. I think we will have to copy and paste the functionality here.

torchx/c_src/torchx.cpp
josevalim: Generally speaking, I would prefer to keep the C++ code to a thin wrapper around Libtorch API. So if Libtorch does not expose a `take`, I wouldn't expose one here either.

The only exception is if we can use several destructive operations (which we don't expose to Elixir for obvious reasons).

torchx/lib/torchx/backend.ex
josevalim: Is `take` in EXLA implemented as a gather? Is it performing the same logic as in here? Because if it is, we should consider removing `take` from Nx.Backend and moving this logic to `Nx` itself. This way we get the gradient and this function for free.

polvalente: No problems. In EXLA take is also a gather operation, but it doesn't perform exactly the same logic. We have this logic for the grad implementation of take, though. However, we can implement it through gather + reshape in Nx

torchx/c_src/torchx.cpp
polvalente: I see! Perhaps I should remove take_along_axis then, in a future PR?

josevalim: Yeah, you probably want to rename Torchx.take_along_axis to Torchx.gather and then implement the backend gather using Torchx APIs. :)

torchx/lib/torchx/backend.ex
josevalim: I see. So what I would do for now is to duplicate the logic and ship this.

torchx/c_src/torchx.cpp
josevalim: In general, our Torchx module should really mimic Libtorch actual functions. :)

torchx/c_src/torchx.cpp
```diff
@@ -313,19 +313,6 @@ NIF(concatenate)
 }
 
 NIF(gather)
-{
```
polvalente, line 316: I've removed `take_along_axis` to expose the actual torch gather semantics instead

torchx/lib/torchx/backend.ex
```diff
@@ -296,18 +296,86 @@ defmodule Torchx.Backend do
...
+      |> from_nx()
+      |> Torchx.tensordot(linear_indices_offsets, [tuple_size(idx.shape) - 1], [0])
+      |> Torchx.reshape({lin_idx_num_elements})
 
     tensor
     |> from_nx()
-    |> Torchx.gather(from_nx(idx), from_nx(linear_indices_offsets), out.shape)
+    |> Torchx.reshape({Tuple.product(tensor.shape)})
+    |> Torchx.gather(linear_indices, 0)
+    |> Torchx.reshape(out.shape)
```
polvalente, line 378: The logic which was inside the C++ code has been moved here, using `Torchx` directly for the operations

torchx/lib/torchx/backend.ex
polvalente: I've duplicated the logic as discussed, and also took the opportunity to move logic from C++ to Elixir

josevalim (approved):
Beautiful!

PR #505, opened 2021-10-05
feat(torchx): population count

polvalente:


torchx/lib/torchx/backend.ex
```diff
@@ -470,6 +470,27 @@ defmodule Torchx.Backend do
     end
   end
 
+  @impl true
+  def population_count(out, %{type: {_, num_bits}, shape: shape} = tensor) do
```
polvalente, line 474: @josevalim I wrote this one more as a brain teaser than anything. It works and we could merge as is, but I think it would probably be best if we dropped down to C++ and iterated over the tensor there, accumulation on an output tensor. Dropping to C++ would consume a lot less memory and would be an O(n) operation.

WDYT?

josevalim: > I didn't go directly to C++ because of what we talked about on keeping C++ logic to a minimum, but I think this should be an exception.

I agree. I think if we need to traverse the whole tensor, then there is a reason to implement it in C++. The big question though is if we can implement this traversal in a way that also works in the GPU. If not, then it is probably best to add this function to the list of not supported for now.

polvalente: If I understood the docs correctly, we can. Torch provided tensor accessors which work like iterators for them.

polvalente:
Closing this for now. The indexing API changes depending on wether you're using CUDA or not, so I think we can leave this implementation for later.

https://pytorch.org/cppdocs/notes/tensor_indexing.html

PR #508, opened 2021-10-06
feat(torchx): support cuda

polvalente:


torchx/Makefile
```diff
@@ -9,6 +9,9 @@ PRIV_DIR = $(MIX_APP_PATH)/priv
 LIBTORCH_LIB_DIR = $(PRIV_DIR)/lib
 TORCHX_SO = $(PRIV_DIR)/torchx.so
 
+CMAKE_BUILD_DIR ?= $(shell pwd)/_build/cmake
```
polvalente, line 12: should we build to `cache` instead of `_build`? Or if we build to `_build`, should it be the mix-env-scoped directory? `_build/dev/torchx/cmake` or something like that?

torchx/Makefile
```diff
@@ -34,6 +37,10 @@ check_libtorch:
 
 $(TORCHX_SO): c_src/torchx.cpp c_src/nx_nif_utils.hpp
 	mkdir -p $(PRIV_DIR)
-	ln -sf $(abspath $(LIBTORCH_LINK_PATH)) $(LIBTORCH_LIB_DIR)
-	$(CXX) $(CFLAGS) c_src/torchx.cpp -o $(TORCHX_SO) $(LDFLAGS)
-	$(POST_INSTALL)
+	mkdir -p $(CMAKE_BUILD_DIR)
+	cd $(CMAKE_BUILD_DIR) && \
+		cmake -DCMAKE_PREFIX_PATH=$(LIBTORCH_DIR) -DC_SRC=$(C_SRC) \
```
polvalente, line 42: @josevalim please check if this works on mac, or if we need to add the `ln -s` back

PR #512, opened 2021-10-10
feat: add eigen decomposition (Symmetric matrix)

kenken-neko:
### Implemented function
The function to calculate the eigenvalues and eigenvectors of real symmetric matries for Issue #157 .
The eigenvalues of real symmetric matrices are always real in a mathematic, thus guaranteeing the current state of nx where input and output are limited to real numbers.
The function name `eigh` is the reference to [`numpy.linalg.eigh`](https://numpy.org/doc/stable/reference/generated/numpy.linalg.eigh.html).

### Logic
The numerical procedure followed LAPACK([DGEEV]( https://www.nag.com/numeric/fl/nagdoc_fl24/pdf/f08/f08naf.pdf)). DGEEV uses the QR method as its core logic. 
In addition to the QR method, there are other methods for solving eigenvalue equations of real symmetric matrices: the Jacobi method and the divide-and-conquer method.
In this case, I chose the QR method for the following reasons.

- The Jacobi method has poor convergence for high dimensional matrices.
- The divide-and-conquer method is complicated to implement to ensure convergence stability.

### Test
I checked several cases to make sure that the correct eigenvalues and eigenvectors were obtained by satisfying the eigen equations [[example](https://github.com/elixir-nx/nx/blob/31f64335c892ecfcf4192f5227cf311fabb2aa66/nx/test/nx/lin_alg_test.exs#L151-L160)].

### Reference

- Orthogonal Bases and the QR Algorithm : https://www-users.cse.umn.edu/~olver/ln_/qr.pdf
- ITERATIVE METHODS FOR COMPUTING EIGENVALUES AND EIGENVECTORS：https://arxiv.org/abs/1105.1185

### DIFF
--- a/exla/lib/exla/client.ex
+++ b/exla/lib/exla/client.ex
@@ -4,7 +4,7 @@ defmodule EXLA.Client do
 
   See `EXLA` module docs for a general introduction.
   """
-
+  require Logger
   use GenServer
   @name __MODULE__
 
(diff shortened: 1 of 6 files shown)


kenken-neko:
@seanmor5 
> Does the current behavior currently match that of `np.linalg.eig` or `np.linalg.eigh`? 

If the input is real symmetric matrix, it match  `np.linalg.eig` and `np.linalg.eigh`.
If the input is a Hermitian as complex symmetric matrix, the logic of this PR may not be able to obtain the proper eigenvalues and eigenvectors. 
By the way, this PR implementation does not support asymmetric matrices, so the scope of adaptation is different from `np.linalg.eig`, which supports asymmetric matrices.

> I think it's best not to change an API only to change it back when more general cases are supported. For now, you can ensure the method raises in unsupported cases, with references to opening an issue when needed.

Yes, it is.
I mistakenly thought that this logic could be easily extended to Herimitian, but it wasn't, so I'll think of a good way to extend it or another logic and then resubmit the PR again.

seanmor5:
@kenken-neko To be clear, I still think this is a valuable contribution. In my opinion we can get this to a point to merge for `eig` with symmetric matrices. Raising for asymmetric matrices is okay for now, we can extend it later. What I meant to say was that we shouldn't rename to `eigen_sym` just because it would change later. It's okay to merge this for specific cases for now and then iterate/extend it later!

kenken-neko:
@seanmor5 @josevalim 
Sorry. I was in a hurry to make a decision and left out the point of view of creating a starting point for the implementation of eigenvalues.
As for the name of the API, if you consider that this implementation is a special case where the imaginary part of Hermitian is zero, there is no mistake in using `eigh`. Therefore, there was no need to do something inappropriate like changing the API name to match the implementation.
  

I will make the PR open again, so please review.

polvalente:
@josevalim EXLA formatting is failing on CI, but locally I get no issues. Any ideas?

PR #515, opened 2021-10-12
Add container protocol

seanmor5:
This defines a protocol and implements it for Tuples and Maps which can be integrated into Nx defn so we can return pretty much any container from defn. All you have to do is define a function `decompose` which returns `{leaves, fun}`, where `leaves` are a flat list of leaf (tensor) values and `fun` is a function which rebuilds the container structure from leaves.

Opening now for feedback on the API. Will integrate once the API is ironed out


josevalim:
I will have a new PR coming soon.

PR #518, opened 2021-10-14
Improve eigh test and default value of max_iter

kenken-neko:
The test code is Improved for the cases where the eigenvalues are different from each other, i.e., easy to converge, and for the cases where the eigenvalues are close to each other, i.e., difficult to converge.


kenken-neko:
I shared the draft test code just for you to try out `max_iter` and `eps`, but if you want to review it, I will share it again with a better appearance.

PR #540, opened 2021-11-01
Nx.diag implementation

### LINKED ISSUE #517: Add Nx.diag implementation
This function should mimic [numpy diag](https://numpy.org/doc/stable/reference/generated/numpy.diag.html), returning the diagonal from a 2D tensor, or building a 2D diagonal tensor from a 1D tensor

Benjamin-Philip:
This pr provides a implementation for Nx.diag.

Closes #517

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -1121,6 +1121,90 @@ defmodule Nx do
   end
 
   @doc """
+  Extract a diagonal or create a diagonal tensor.
+
+  Extracts a diagonal if given a 2d tensor, or constructs a tensor given a 1d one.
+
+  ## Examples
+
+  When given a 1d tensor:
+
+      iex> Nx.iota({4}) |> Nx.diag
+      #Nx.Tensor<
+        s64[4][4]
+        [
+          [0, 0, 0, 0],
+          [0, 1, 0, 0],
+          [0, 0, 2, 0],
+          [0, 0, 0, 3]
+        ]
+      >
+
+  Given a 2d tensor:
+
+      iex> Nx.iota({3, 3}) |> Nx.diag()
+      #Nx.Tensor<
+        s64[3]
+        [0, 4, 8]
+      >
+
+  And if given a 2d tensor along with an offset:
+
+      iex> Nx.iota({3, 3}) |> Nx.diag(offset: 1)
+      #Nx.Tensor<
+        s64[2]
+        [1, 5]
+      >
+
+      iex> Nx.iota({3, 3}) |> Nx.diag(offset: -1)
+      #Nx.Tensor<
+        s64[2]
+        [3, 7]
+      >
+
+  ## Options
+
+    * `:offset` - offset used for extracting the diagonal.
+      Use offset > 0 for diagonals above the main diagonal,
+      and offset < 0 for diagonals below the main diagonal.
+
+  """
+  @doc type: :creation
+  def diag(tensor, opts \\ []) do
+    opts = keyword!(opts, [:offset])
+
+    case shape(tensor) do
+      {len, breadth} ->
+        offset = opts[:offset] || 0
+
+        indices =
+          case offset do
+            i when i >= 0 ->
+              Enum.zip_with(0..(len - 1), i..(breadth - 1), fn x, y -> [x, y] end)
+
+            i when i < 0 ->
+              Enum.zip_with(-i..(len - 1), 0..(breadth - 1), fn x, y -> [x, y] end)
+          end
+
+        indices =
+          indices
+          |> Nx.tensor()
+
+        Nx.gather(tensor, indices)
+
+      {len} ->
+        identity =
+          {len, len}
+          |> Nx.eye()
+
+        Nx.multiply(tensor, identity)
+
+      _bad_rank ->
+        raise ArgumentError, "diag/2 expects tensor of rank 1 or 2, got: #{inspect(tensor)}"
+    end
+  end
+
+  @doc """
   Creates a one-dimensional tensor from a `binary` with the given `type`.
 
   If the binary size does not match its type, an error is raised.
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -1170,38 +1170,39 @@ defmodule Nx do
 
   """
   @doc type: :creation
-  def diag(tensor, opts \\ []) do
+  def diag(tensor, opts \\ [])
+
+  def diag(tensor = %Nx.Tensor{shape: {len, breadth}}, opts) do
     opts = keyword!(opts, [:offset])
 
-    case shape(tensor) do
-      {len, breadth} ->
-        offset = opts[:offset] || 0
+    offset = opts[:offset] || 0
 
-        indices =
-          case offset do
-            i when i >= 0 ->
-              Enum.zip_with(0..(len - 1), i..(breadth - 1), fn x, y -> [x, y] end)
+    indices =
+      case offset do
+        i when i >= 0 ->
+          Enum.zip_with(0..(len - 1), i..(breadth - 1), fn x, y -> [x, y] end)
 
-            i when i < 0 ->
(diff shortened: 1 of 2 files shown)


nx/lib/nx.ex
```diff
@@ -1120,6 +1120,91 @@ defmodule Nx do
...
+        i when i >= 0 ->
+          Enum.zip_with(0..(len - 1), i..(breadth - 1), fn x, y -> [x, y] end)
+
+        i when i < 0 ->
+          Enum.zip_with(-i..(len - 1), 0..(breadth - 1), fn x, y -> [x, y] end)
+      end
+
+    indices =
+      indices
+      |> Nx.tensor()
```
polvalente, line 1191:
```suggestion
    indices_tensor = Nx.tensor(indices)
```

nx/lib/nx.ex
```diff
@@ -1120,6 +1120,91 @@ defmodule Nx do
...
+
+    Nx.gather(tensor, indices)
```
polvalente, line 1193:
```suggestion
    Nx.gather(tensor, indices_tensor)
```

nx/lib/nx.ex
polvalente: I'd rather avoid variable overriding, and also single-function pipes

nx/lib/nx.ex
```diff
@@ -1120,6 +1120,91 @@ defmodule Nx do
...
+    identity =
+      {len, len}
+      |> Nx.eye()
+
+    Nx.multiply(tensor, identity)
+  end
+
+  def diag(tensor, _opts) do
+    raise ArgumentError, "diag/2 expects tensor of rank 1 or 2, got: #{inspect(tensor)}"
+  end
```
polvalente, line 1206: we normally do shape checks with `Nx.Shape.<function_name>`. Perhaps you could move the separate clauses to a `defp extract_diag` and `defp build_diag` or something like that to keep with the pattern

nx/lib/nx.ex
```diff
@@ -1120,6 +1120,91 @@ defmodule Nx do
...
+      indices
+      |> Nx.tensor()
+
+    Nx.gather(tensor, indices)
+  end
+
+  def diag(tensor = %Nx.Tensor{shape: {len}}, _opts) do
+    identity =
+      {len, len}
+      |> Nx.eye()
```
polvalente, line 1199:
```suggestion
    identity = Nx.eye(len)
```

nx/lib/nx.ex
```diff
@@ -1120,6 +1120,91 @@ defmodule Nx do
...
+
+    Nx.multiply(tensor, identity)
```
polvalente, line 1201:
```suggestion
    Nx.select(identity, tensor, Nx.broadcast(0, tensor))
```

@josevalim I'm not sure which is better for backends like EXLA and Torchx... WDYT?

nx/test/nx_test.exs
```diff
@@ -1684,6 +1684,42 @@ defmodule NxTest do
     end
   end
 
+  describe "diag/2" do
+    test "extracts valid diagonal given no offset" do
+      diag =
+        Nx.iota({3, 3})
+        |> Nx.diag()
```
polvalente, line 1691:
```suggestion
        {3, 3}
        |> Nx.iota()
        |> Nx.diag()
```

nx/test/nx_test.exs
```diff
@@ -1684,6 +1684,42 @@ defmodule NxTest do
...
+
+      assert diag == Nx.tensor([0, 4, 8])
+    end
+
+    test "extracts valid diagonal given offset" do
+      diag =
+        Nx.iota({3, 3})
```
polvalente, line 1698:
```suggestion
        {3, 3}
        |> Nx.iota()
```

nx/test/nx_test.exs
```diff
@@ -1684,6 +1684,42 @@ defmodule NxTest do
...
+        |> Nx.diag(offset: 1)
+
+      assert diag == Nx.tensor([1, 5])
+    end
+
+    test "constructs diagonal given 1-d tensor" do
+      diag =
+        Nx.iota({3})
```
polvalente, line 1706:
```suggestion
        {3}
        |> Nx.iota()
```

nx/test/nx_test.exs
```diff
@@ -1684,6 +1684,42 @@ defmodule NxTest do
     end
   end
 
+  describe "diag/2" do
+    test "extracts valid diagonal given no offset" do
```
polvalente, line 1688: I think we need a test for the shape {m, n} when m > n and another for when m < n

polvalente (commented):
Thanks for the contribution! I've left some (mostly minor) comments, but barring that the overral idea is great!

nx/lib/nx.ex
seanmor5: @polvalente I would go with multiply, but the performance difference on EXLA seems to be negligible. Here's a benchmark in JAX:

```python
%%timeit -n 25
jax.jit(lambda identity, tensor: jax.lax.select(identity, tensor, jax.lax.broadcast(0., tensor.shape)))(identity, tensor)
```

```
25 loops, best of 5: 25.5 ms per loop
```

```python
%%timeit -n 25
jax.jit(lambda identity, tensor: jax.lax.mul(identity, tensor))(identity, tensor)
```

```
25 loops, best of 5: 24.2 ms per loop
```

polvalente: Since in EXLA the difference is negligible, let's go with multiply, which will tend to be more memory-efficient for non-compiled backends

nx/lib/nx.ex
```diff
@@ -1120,6 +1120,91 @@ defmodule Nx do
...
+          [0, 0, 0, 0],
+          [0, 1, 0, 0],
+          [0, 0, 2, 0],
+          [0, 0, 0, 3]
+        ]
+      >
+
+  Given a 2d tensor:
+
+      iex> Nx.iota({3, 3}) |> Nx.diag()
```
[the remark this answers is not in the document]
Benjamin-Philip: @polvalente think something like this might make the docs feel very repetitive. At the same time, I also think that it will make the docs easier to reason about. How about showing the literal tensor once and the rest using iota?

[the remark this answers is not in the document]
polvalente: Makes sense. You could merge all tests in two, one for 2D tensors and one for 1D tensors. 

Then you could write the literal tensor and then some examples. For each case

nx/lib/nx.ex
```diff
@@ -1120,6 +1120,91 @@ defmodule Nx do
...
+      #Nx.Tensor<
+        s64[2]
+        [1, 5]
+      >
+
+      iex> Nx.iota({3, 3}) |> Nx.diag(offset: -1)
+      #Nx.Tensor<
+        s64[2]
+        [3, 7]
+      >
```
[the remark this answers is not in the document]
Benjamin-Philip: That's what I initially did, but I felt calling iota would have the effects of separating docstests and personally felt a little more uniform to the other docs as `iota` calls were more common than variable assignment...

nx/lib/nx.ex
Benjamin-Philip: @polvalente,

Could you specify what you mean? I don't quite understand.. Is there a `Nx.Shape.shape!` function that you expect me to use, or am I supposed to do something else...

Benjamin-Philip:
@polvalente, @seanmor5, and @josevalim please and have a look.

nx/lib/nx.ex
polvalente: Instead of hard-matching on the tensor shapes you would delegate the check for 2D vs 1D (and this raise) do a `Nx.Shape.diag` function. Don't worry about this, though. We can change this later if need comes :)

polvalente (approved):
Thanks for the contribution!

Issue #542, opened 2021-11-04
Generalize and split Nx.diag in two


josevalim:
The current `Nx.diag/2` is a great starting point but, if we want to support vectorization, such as vmap, we need to allow extra a diagonal from any `rank >= 1`  tensor and build diagonal matrices from any `rank > 1` tensor.

We may not need to generalize immediately but we should probably split sooner than later. The big question is what we should call these operations. Given we are increasing and decreasing the rank, my suggestion would be to call it `gather_diagonal` (to go from a matrix to a vector) and `scatter_diagonal` (to go from a vector to a matrix).

Thoughts @polvalente and @seanmor5?

polvalente:
@josevalim I think that `gather` and `scatter` are good prefixes, although I might be biased because I already have the mental image of what a `gather` and what a `scatter` is.

As we had discussed, `to_diagonal` and `from_diagonal` could be confusing.
WDYT about `build_diagonal` and `extract_diagonal` for `scatter` and `gather` respectively? build/extract seem more layman's terms than scatter/gather

josevalim:
Here are other suggestions:

  * `as_diagonal` - Returns a vector as a diagonal matrix
  * `take_diagonal` - Takes the diagonal of a matrix

I think those should not be ambiguous. Thoughts?

seanmor5:
Just a quick question: if we we're to implement batching on only primitives like with gradients, how would the diag implementation look (assuming we changed some of the shape logic to support batched dims)?

I think I am still +1 on generalizing/splitting the diagonal operation and differentiating between taking a diagonal from an input or creating a diagonal matrix/tensor. +1 for `as_diagonal` and `take_diagonal`.

polvalente:
@seanmor5 I'm not sure if I quite follow the first part of the question

From what I was thinking, `take_diagonal` would receive a tensor (for simplicity, square in the inner dims) in the shape {..., N, N} and yield a tensor {..., N}. The grad would be akin to the grad for `gather`, in which the grads corresponding to each element would be positioned in the corresponding indices.

And for `as_diagonal`, it would be akin to `scatter-add`, in which the resulting {..., N, N} tensor's grad would be zero except for the diagonal elements.

Ideally, `take_diagonal` could be implemented in terms of `gather` and `as_diagonal`, in terms of `scatter_add`.

---

As for naming, I'm comfortable with `take_diagonal` and `as_diagonal`

seanmor5:
[...]

PR #544, opened 2021-11-08
Split Nx.diag into Nx.take_diagonal and Nx.make_diagonal

Benjamin-Philip:
This pr provides alternate implementations to Nx.diag.

See #542


Benjamin-Philip:
I'm pretty sure I am breaking some convention or the other. Hence, submitting a draft pr with only `take_diagonal` implemented.

Please have a look and let me know what to fix. Once `take_diagonal` is okay, I'll submit `make_diagonal` written in a similar fashion as `take_diagonal` at that point of time in this PR.

nx/lib/nx.ex
```diff
@@ -1228,6 +1217,85 @@ defmodule Nx do
     raise ArgumentError, "diag/2 expects tensor of rank 1 or 2, got: #{inspect(tensor)}"
   end
 
+  @doc """
+  Extract a diagonal from a 2D tensor.
```
josevalim, line 1221:
```suggestion
  Extracts the diagonal from a 2D tensor.
```

josevalim:
This looks great to me! :+1: Just one tiny comment.

nx/lib/nx.ex
Benjamin-Philip: How about this: 
```suggestion
  Extracts the diagonal of a 2D tensor.
```

josevalim: :+1:

Benjamin-Philip:
@josevalim, Have a look at how I am managing checking the tensor shape and offset. I am not sure if it's the right way to go about it..

Benjamin-Philip:
> This looks great to me! +1 Just one tiny comment.

Great! I'll submit `make_diagonal` soon then.

Benjamin-Philip:
@josevalim Will we remove `Nx.diag` after `take` and `make` are defined?

josevalim:
> @josevalim Will we remove Nx.diag after take and make are defined?

Yes!

nx/lib/nx/shape.ex
```diff
@@ -1218,6 +1218,57 @@ defmodule Nx.Shape do
...
+
+  ## Error cases
+
+    iex> Nx.Shape.validate_offset!({3, 4}, 3)
+    ** (ArgumentError) offset must be less than length when positive, got: 3
+
+    iex> Nx.Shape.validate_offset!({3, 4}, -4)
+    ** (ArgumentError) absolute value of offset must be less than breadth when negative, got: -4
+  """
+  def validate_offset!(shape, offset) do
```
polvalente, line 1253: I think this should have something to indicate that it pertains to *_diagonal functions

Benjamin-Philip: In the docs or in the function name?

nx/lib/nx/shape.ex
```diff
@@ -1218,6 +1218,57 @@ defmodule Nx.Shape do
...
+      i when i >= 0 ->
+        raise ArgumentError, "offset must be less than length when positive, got: #{inspect(i)}"
+
+      i when i < 0 and -i < breadth ->
+        :ok
+
+      i when i < 0 ->
+        raise ArgumentError,
+              "absolute value of offset must be less than breadth when negative, got: #{inspect(i)}"
+    end
```
polvalente, line 1269: Shouldn't this be either a cond statement or a series of function clauses with guards instead?

nx/lib/nx/shape.ex
```diff
@@ -1218,6 +1218,57 @@ defmodule Nx.Shape do
...
+  ## Error cases
+
+    iex> Nx.Shape.validate_offset!({3, 4}, 3)
+    ** (ArgumentError) offset must be less than length when positive, got: 3
+
+    iex> Nx.Shape.validate_offset!({3, 4}, -4)
+    ** (ArgumentError) absolute value of offset must be less than breadth when negative, got: -4
+  """
+  def validate_offset!(shape, offset) do
+    {len, breadth} = shape
```
polvalente, line 1254: This should be matched in the function head to avoid raising a RuntimeError upon invalid shape

nx/lib/nx/shape.ex
polvalente: Both, if possible!

Benjamin-Philip: Ok will do.

nx/lib/nx/shape.ex
Benjamin-Philip: At first thought, function clauses and guards seem more appropriate. However, when we generalise, `cond` would be more appropriate.

WDYT?

nx/lib/nx/shape.ex
```diff
@@ -1218,6 +1218,57 @@ defmodule Nx.Shape do
...
+
+    case offset do
+      i when i >= 0 and i < len ->
+        :ok
+
+      i when i >= 0 ->
+        raise ArgumentError, "offset must be less than length when positive, got: #{inspect(i)}"
+
+      i when i < 0 and -i < breadth ->
```
polvalente, line 1263: Shouldn't the checks be using div(len, 2) and div(breadth, 2) instead?

From what I gathered, the main diagonal is offset 0, and you add 1 to the offset for each diagonal above the main and subtract 1 for the ones below, right?

If so, for a 3x4 matrix, for instance, you would have offsets -2, -1, 0 and 1

nx/lib/nx/shape.ex
polvalente: Both would be fine to be. It's just that the case statement adds a new variable without need

nx/lib/nx/shape.ex
Benjamin-Philip: Okay. I am assuming that the shape is valid, but a check won't hurt.

nx/lib/nx/shape.ex
Benjamin-Philip: > From what I gathered, the main diagonal is offset 0, and you add 1 to the offset for each diagonal above the main and subtract 1 for the ones below, right?

Right. In essence, you give a positive value to "move" the starting indices horizontally and a negative value to move it vertically.

So if I take your example of a 3 x 4 matrix:

```elixir
iex(2)> t = Nx.iota({3, 4})
#Nx.Tensor<
  s64[3][4]
  [
    [0, 1, 2, 3],
    [4, 5, 6, 7],
    [8, 9, 10, 11]
  ]
>
```

This matrix has 6 diagonals:

```elixir
iex(4)> Nx.take_diagonal(t, offset: -2)
#Nx.Tensor<
  s64[1]
  [8]
>
iex(5)> Nx.take_diagonal(t, offset: -1)
#Nx.Tensor<
  s64[2]
  [4, 9]
>
iex(6)> Nx.take_diagonal(t, offset: 0) 
#Nx.Tensor<
  s64[3]
  [0, 5, 10]
>
iex(7)> Nx.take_diagonal(t, offset: 1)
#Nx.Tensor<
  s64[3]
  [1, 6, 11]
>
iex(8)> Nx.take_diagonal(t, offset: 2)
#Nx.Tensor<
  s64[2]
  [2, 7]
>
```

So my current check makes sense.

polvalente: Yeah, the current implementation looks correct :)

nx/lib/nx/shape.ex
Benjamin-Philip: @polvalente, 

On second thoughts, is the extra variable really a problem? Using `case` allows me to pattern match within a function. If I use `cond`, I will have to give up the readability of pattern matching, and if I use function clauses, I will have to give up the readability of having everything in one function.

josevalim: >  If I use cond, I will have to give up the readability of pattern matching

The issue here is that you are not really pattern matching on anything. The `i when` is redundant on all clauses and that's exactly what `cond` will help address. :)

nx/lib/nx.ex
```diff
@@ -1228,6 +1217,199 @@ defmodule Nx do
...
+  end
+
+  @doc """
+  Creates a diagonal tensor from a 1D tensor.
+
+  Converse of `take_diagonal/2`.
+
+  If given an offset, the diagonal tensor will be a square matrix
+  with length `size(tensor)` + `Kernel.abs(:offset)`, with the 1D
+  tensor as the diagonal of offset `:offset`.
```
Benjamin-Philip, line 1293: I feel this description is can be a bit unclear. Any suggestions?

nx/lib/nx/shape.ex
```diff
@@ -1218,6 +1218,69 @@ defmodule Nx.Shape do
...
+      _bad_shape ->
+        raise ArgumentError, "take_diagonal/2 expects tensor of rank 2, got:\n#{inspect(tensor)}"
+    end
+  end
+
+  @doc """
+  Returns shape if valid and raises error if not.
+  """
+  def make_diagonal(tensor) do
+    case tensor.shape do
```
Benjamin-Philip, line 1238: Should I call `shape` instead? If a person passes a list instead of a tensor, it will fail with an `apply/3` error.

```suggestion
    case Nx.shape(tensor) do
```

nx/lib/nx.ex
```diff
@@ -1228,6 +1217,199 @@ defmodule Nx do
...
+        [0, 1, 2]
+      >
+
+      iex> Nx.take_diagonal(Nx.iota({3, 3}), offset: 3)
+      ** (ArgumentError) offset must be less than breadth when positive, got: 3
+
+      iex> Nx.take_diagonal(Nx.iota({3, 3}), offset: -4)
+      ** (ArgumentError) absolute value of offset must be less than length when negative, got: -4
+  """
+  @doc type: :creation
```
josevalim, line 1274: Is `:creation` the correct tag here and below? I am thinking it is rather `:shape`, because we are mostly keeping the same data but under a different shape? What do you think @polvalente?

josevalim (approved):
LGTM! Just one tiny comment.

nx/lib/nx/shape.ex
josevalim: The functions in `Nx.Shape` do not receive the tensor, they always receive the shape, so you should change it do the same.

Back in the Nx module, you should do:

```elixir
def make_diagonal(tensor, opts \\ []) do
  tensor = to_tensor(tensor)
```

and then you can call `tensor.shape`, as it is guaranteed to be normalized.

Benjamin-Philip: Will update `take_diagonal` and `make_diagonal`

Benjamin-Philip:
The `Torchx` test are failing for some reason. When I had a look, it was calling some functions in `Nx.BinaryBackend` which doesn't look correct. However, previous commits passed without any issues.

Any clue why?

josevalim:
@Benjamin-Philip it is a bug which I fixed it in master. In any case, Torchx does not support scatter_add yet, so you will need to list make_diagonal here: https://github.com/elixir-nx/nx/blob/main/torchx/test/torchx/nx_doctest_test.exs#L40

nx/lib/nx.ex
```diff
@@ -1195,15 +1182,141 @@ defmodule Nx do
...
+  ## Error cases
+
+      iex> Nx.take_diagonal(Nx.tensor([0, 1, 2]))
+      ** (ArgumentError) take_diagonal/2 expects tensor of rank 2, got tensor of rank: 1
+
+      iex> Nx.take_diagonal(Nx.iota({3, 3}), offset: 3)
+      ** (ArgumentError) offset must be less than breadth when positive, got: 3
+
+      iex> Nx.take_diagonal(Nx.iota({3, 3}), offset: -4)
+      ** (ArgumentError) absolute value of offset must be less than length when negative, got: -4
```
polvalente, line 1194: I don't think length and breadth are names we use "publicly". WDYT about using "must be less then length of axis 0" and "must be less than length of axis 1" instead?

nx/lib/nx.ex
polvalente: For `make_diagonal` I'd use `:creation`, since we can think of the operation as imposing certain elements of an already existing zero-matrix to have specific values. For me `:shape` would make sense if there weren't any other (zero) elements appearing in the resulting tensor.

nx/lib/nx.ex
polvalente:
```suggestion
  If given an offset, the resulting tensor will have the corresponding diagonal
  filled with the input tensor's values. The main diagonal as `offset = 0`, whereas
  diagonals above it can be filled with `offset = 1`, `offset = 2` and so on, while the
  diagonals below it can be filled with `offset = -1`, `offset = -2` and so on.
```

polvalente: WDYT about using a less "mathematical" language like this?

nx/lib/nx.ex
```diff
@@ -1195,15 +1182,141 @@ defmodule Nx do
...
+          [0, 0, 0, 0, 0, 0, 0],
+          [0, 0, 0, 0, 0, 0, 0]
+        ]
+      >
+
+  ## Options
+
+    * `:offset` - offset used for extracting the diagonal.
+      Use offset > 0 for diagonals above the main diagonal,
+      and offset < 0 for diagonals below the main diagonal.
```
polvalente, line 1292:
```suggestion
    * `:offset` - offset used for extracting the diagonal.
      Use offset > 0 for diagonals above the main diagonal,
      and offset < 0 for diagonals below the main diagonal.
      Defaults to `0`
```

nx/lib/nx.ex
```diff
@@ -1195,15 +1182,141 @@ defmodule Nx do
...
   @doc type: :creation
-  def diag(tensor, opts \\ [])
+  def make_diagonal(tensor, opts \\ []) do
+    tensor = to_tensor(tensor)
 
-  def diag(tensor = %Nx.Tensor{shape: {len, breadth}}, opts) do
     opts = keyword!(opts, [:offset])
 
+    {len} = Nx.Shape.make_diagonal(tensor.shape)
     offset = opts[:offset] || 0
```
polvalente, line 1306:
```suggestion
    opts = keyword!(opts, [offset: 0])

    {len} = Nx.Shape.make_diagonal(tensor.shape)
    offset = opts[:offset]
```

nx/lib/nx.ex
```diff
@@ -1195,15 +1182,141 @@ defmodule Nx do
...
+      ** (ArgumentError) absolute value of offset must be less than length when negative, got: -4
+  """
+  @doc type: :creation
+  def take_diagonal(tensor, opts \\ []) do
+    tensor = to_tensor(tensor)
+
+    opts = keyword!(opts, [:offset])
+
+    shape = Nx.Shape.take_diagonal(tensor.shape)
+    offset = opts[:offset] || 0
```
polvalente, line 1203:
```suggestion
    opts = keyword!(opts, [offset: 0])

    shape = Nx.Shape.take_diagonal(tensor.shape)
    offset = opts[:offset]
```

nx/lib/nx/shape.ex
```diff
@@ -1218,6 +1218,71 @@ defmodule Nx.Shape do
...
+
+      offset < 0 and -offset < len ->
+        :ok
+
+      offset < 0 ->
+        raise ArgumentError,
+              "absolute value of offset must be less than length when negative, got: #{inspect(offset)}"
+    end
+  end
+
```
polvalente, line 1285:
```suggestion
  end

  def validate_diag_offset!(shape, _offset) when is_tuple(shape) do
    raise ArgumentError, "expected shape of rank 2 to be given, got shape of rank #{tuple_size(shape)}"
  end
```

nx/lib/nx.ex
Benjamin-Philip: > For make_diagonal I'd use :creation, since we can think of the operation as imposing certain elements of an already existing zero-matrix to have specific values. For me :shape would make sense if there weren't any other (zero) elements appearing in the resulting tensor.

@polvalente, What about for `take_diagonal`?

nx/lib/nx.ex
Benjamin-Philip: > I don't think length and breadth are names we use "publicly". 

You're right!

> WDYT about using "must be less then length of axis 0" and "must be less than length of axis 1" instead?

I think this is probably the best we can do.

nx/lib/nx.ex
Benjamin-Philip: I should have mentioned the default...

Benjamin-Philip: > WDYT about using a less "mathematical" language like this?

I can see that my "mathematical"  description can be a bit difficult to wrap the head around at first.
But at the same time it is quite concise, precise, and unambiguous.

>  If given an offset, the resulting tensor will have the corresponding diagonal
  filled with the input tensor's values. The main diagonal as `offset = 0`, whereas
  diagonals above it can be filled with `offset = 1`, `offset = 2` and so on, while the
  diagonals below it can be filled with `offset = -1`, `offset = -2` and so on.

I get the basic idea over here, but don't I explain how to use `:offset` in the options docs?

josevalim: Yeah, I think the one in the docs is ok. I will send a small suggestion to the "mathematical" one.

nx/lib/nx.ex
josevalim:
```suggestion
  The returned tensor will be a square matrix of dimensions equal
  to the size of the tensor. If an offset is given, the absolute value
  of the offset is added to the matrix dimensions sizes.
```

?

nx/lib/nx.ex
josevalim: To be clear, @polvalente was commenting on the previous description as the mathematical one, not this one. :)

Benjamin-Philip: @josevalim, So is this resolved then?

Benjamin-Philip:
@josevalim, and @polvalente,

I think we should be ready to ship after we have fixed the `make_diagonal` docs.

nx/lib/nx.ex
josevalim: @Benjamin-Philip please see the suggestion above. :)

nx/lib/nx.ex
polvalente: I think we can use the same tag for both. Although `take_diagonal` is more of "destruction" than "creation", because we eliminate elements from the matrix, but we can also think in terms of "fill this vector with elements from this matrix"

nx/lib/nx.ex
```diff
@@ -1144,28 +1144,15 @@ defmodule Nx do
   end
 
   @doc """
-  Extract a diagonal or create a diagonal tensor.
+  Extracts the diagonal of a 2D tensor.
 
-  Extracts a diagonal if given a 2D tensor, or constructs a tensor given a 1D one.
+  Converse of `make_diagonal/2`.
```
polvalente, line 1149: I think we could also explain `:offset` here, like in `make_diagonal`

Benjamin-Philip:
@josevalim, I have fixed the docs.

josevalim (approved):

nx/lib/nx.ex
josevalim: So far all of our creation operations do not receive a tensor as input, only shape/parameters. So to me this is more of a transformation on the shape, but we can always revisit it later.

torchx/test/torchx/nx_doctest_test.exs
```diff
@@ -24,6 +24,8 @@ defmodule Torchx.NxDoctestTest do
     broadcast: 3,
     # dot - Batching not supported
     dot: 6,
+    # make_diagonal - depends on scatter_add
+    make_diagonal: 4,
```
josevalim, line 28:
```suggestion
    make_diagonal: 2,
```

josevalim:

PR #545, opened 2021-11-09
feat(torchx): triangular solve

polvalente:


torchx/test/support/torchx_case.ex
```diff
@@ -15,9 +15,17 @@ defmodule Torchx.Case do
...
+      |> Nx.all_close(right, atol: atol, rtol: rtol)
+      |> Nx.backend_transfer(Nx.BinaryBackend)
+
+    if equals != Nx.tensor(1, type: {:u, 8}, backend: Nx.BinaryBackend) do
+      flunk("""
+      Tensor assertion failed.
+      left: #{inspect(left)}
+      right: #{inspect(right)}
+      """)
+    end
```
polvalente, line 29: This makes for a better error message in which we can actually see the compared tensors :)

josevalim (approved):

PR #547, opened 2021-11-10
Return 0/1 from comparison operators in Kernel

josevalim:
Otherwise someone can write this code:

```elixir
defn example(matrix) do
  {m, n} = Nx.shape(matrix)
  if m == n do
    ...
  else
    ...
  end
end
```

And a boolean would be given to `if`, making it raise. Alternatively we could allow booleans in `if`, `and`, etc. But I don't think we should go down this slippery slope yet. :D

### DIFF
--- a/nx/lib/nx/binary_backend.ex
+++ b/nx/lib/nx/binary_backend.ex
@@ -22,8 +22,8 @@ defmodule Nx.BinaryBackend do
   ## Creation
 
   @impl true
-  def scalar(%{type: type, shape: shape} = out, scalar, _backend_options) do
-    data = :binary.copy(number_to_binary(scalar, type), Nx.size(shape))
+  def constant(%{type: type, shape: shape} = out, constant, _backend_options) do
+    data = :binary.copy(number_to_binary(constant, type), Nx.size(shape))
     from_binary(out, data)
   end
 
--- a/nx/lib/nx/defn/kernel.ex
+++ b/nx/lib/nx/defn/kernel.ex
@@ -536,23 +536,19 @@ defmodule Nx.Defn.Kernel do
   def not tensor when is_number(tensor), do: logical_not(tensor)
   def not tensor, do: Nx.logical_not(tensor)
 
-  defp logical_and(l, r) when Kernel.==(l, 0), do: zero(l, r)
-  defp logical_and(l, r) when Kernel.==(r, 0), do: zero(l, r)
-  defp logical_and(l, r), do: one(l, r)
+  defp logical_and(l, _) when Kernel.==(l, 0), do: zero()
+  defp logical_and(_, r) when Kernel.==(r, 0), do: zero()
+  defp logical_and(_, _), do: one()
 
-  defp logical_or(l, r) when Kernel.and(Kernel.==(l, 0), Kernel.==(r, 0)), do: zero(l, r)
-  defp logical_or(l, r), do: one(l, r)
+  defp logical_or(l, r) when Kernel.and(Kernel.==(l, 0), Kernel.==(r, 0)), do: zero()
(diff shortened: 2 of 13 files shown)


nx/test/nx/defn/kernel_test.exs
```diff
@@ -23,6 +23,15 @@ defmodule Nx.Defn.KernelTest do
       assert Nx.Defn.Kernel./(1, 2) == 0.5
     end
 
+    test "comparison" do
+      assert Nx.Defn.Kernel.==(0, 0) == 1
+      assert Nx.Defn.Kernel.!=(0, 0) == 0
+      assert Nx.Defn.Kernel.>(0, 0) == 0
+      assert Nx.Defn.Kernel.>=(0, 0) == 1
+      assert Nx.Defn.Kernel.<(0, 0) == 0
+      assert Nx.Defn.Kernel.<=(0, 0) == 1
```
polvalente, line 32: Shouldn't we return Nx.tensor(0, type: {:u, 8}) and the equivalent for 1, instead?
I think it would make for more uniform return types in defn, even though the overrides would make eveything work fine...

Approving this nevertheless

polvalente (approved):

nx/test/nx/defn/kernel_test.exs
josevalim: You are correct. I will tackle this.

nx/lib/nx/binary_backend.ex
```diff
@@ -22,8 +22,8 @@ defmodule Nx.BinaryBackend do
   ## Creation
 
   @impl true
-  def scalar(%{type: type, shape: shape} = out, scalar, _backend_options) do
-    data = :binary.copy(number_to_binary(scalar, type), Nx.size(shape))
+  def constant(%{type: type, shape: shape} = out, constant, _backend_options) do
```
polvalente, line 25: Why the change from scalar to constant?

polvalente: I think that scalar reflects better the mathematical sense, whereas constant seems to be the opposite of variable (which isn't the case)

josevalim: @polvalente it is a constant because it can have any shape. :)

polvalente: Nice, I missed that :)

josevalim: To be clear, it was previously incorrect, it was always a constant because we use it for optimizations, but I noticed it only now when exposing the API.

nx/lib/nx/defn/kernel.ex
```diff
@@ -536,23 +536,19 @@ defmodule Nx.Defn.Kernel do
   def not tensor when is_number(tensor), do: logical_not(tensor)
   def not tensor, do: Nx.logical_not(tensor)
 
-  defp logical_and(l, r) when Kernel.==(l, 0), do: zero(l, r)
-  defp logical_and(l, r) when Kernel.==(r, 0), do: zero(l, r)
-  defp logical_and(l, r), do: one(l, r)
+  defp logical_and(l, _) when Kernel.==(l, 0), do: zero()
```
polvalente, line 539:
```suggestion
  defp logical_and(0, _), do: zero()
```
Why not hard-match like this?

polvalente (approved):
Other than the last stylistic question, the code looks fine

josevalim:

nx/lib/nx/defn/kernel.ex
josevalim: We need to handle both floats and integers and a match is type specific. :)

polvalente: Nice. == vs ===

PR #549, opened 2021-11-18
Add Nx.Container protocol

josevalim:
The protocol gives users control over which fields of a struct are tensors, which ones must be reset (the default), and which ones must be sent to `defn` and therefore be part of the cache key. This PR has two breaking changes:

1. A tuple of functions is no longer valid argument to defn, you need to wrap it in a struct instead. @seanmor5, is this ok? It is very easy to add this back if you want (like 10 LOC). UPDATE: it is currently deprecated.

2. Passing structs to `defn` is no longer supported, unless they implement the protocol.

There was a massive amount of code that was cleaned up because of this, to the point we have more lines of code removed than added, even after adding more docs and tests.


josevalim:

PR #551, opened 2021-11-19
Fix race condition on preloaded buffers

josevalim:
We tagged the buffers as zero copy but the VM only
guarantees the binary won't be relocated during the
NIF call. So we set it to immutable until transfer
and block until ready.

### DIFF
--- a/exla/c_src/exla/exla_client.cc
+++ b/exla/c_src/exla/exla_client.cc
@@ -8,7 +8,9 @@
 
 namespace exla {
 
-ExlaBuffer::ExlaBuffer(std::unique_ptr<xla::PjRtBuffer> buffer): buffer_(std::move(buffer)) {}
+ExlaBuffer::ExlaBuffer(std::unique_ptr<xla::PjRtBuffer> buffer,
+                       bool can_be_released_after_run): buffer_(std::move(buffer)),
+                                                        can_be_released_after_run_(can_be_released_after_run) {}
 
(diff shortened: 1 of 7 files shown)


exla/c_src/exla/exla_client.cc
```diff
@@ -136,14 +138,8 @@ xla::StatusOr<ERL_NIF_TERM> ExlaExecutable::Run(ErlNifEnv* env,
...
   for (auto buf : input_buffers) {
+    EXLA_EFFECT_OR_RETURN_NIF(buf->BlockHostUntilReady(), env);
     pjrt_buffers.push_back(buf->buffer());
-    // If the buffer was not received as a resource (e.g. we converted
-    // it from a binary to a buffer), we need to make it a resource so
-    // it's tracked and GC'ed along with other buffers that are no longer
-    // in use
-    if (buf->release_after_run()) {
-      terms.push_back(nif::make<ExlaBuffer*>(env, buf));
-    }
```
josevalim, line 146: @seanmor5 please check if we indeed can remove this. I think the goal of this code was to make sure the binaries are not relocated but, as we mentioned, there is no such guarantee.

seanmor5: I don't think we can remove this, the purpose of this is to avoid memory leaks. If we move a binary to the GPU, then it needs to be cleaned up immediately after execution OR we take this approach which allows the GC to do it for us.

josevalim:

PR #552, opened 2021-11-19
Remove unused device api

josevalim:


polvalente (approved):
Aren't there any tests to remove?

josevalim:
@polvalente nope.

seanmor5:
@josevalim So I think we are going to end up needing these in a multi-device setting when we get there. We can remove for now, but it will probably end up getting reintroduced later

josevalim:
@seanmor5 I think last time we talked it seems we could stick with using the device_id exclusively. But yeah, we can add this back if necessary. :)

josevalim:

PR #555, opened 2021-11-22
Add Nx.Defn.default_options, closes #506

### LINKED ISSUE #506: Add default_compiler functions in Nx.Defn and relevant application config
The only question is if and how it should interact with `Nx.Defn.default_compiler`. Should we remove the module configuration? Should we keep them distinct?

josevalim:


exla/lib/exla.ex
```diff
@@ -184,6 +176,45 @@ defmodule EXLA do
...
+  preferred client based on the available platforms.
+
+  This function is typically invoked at the top of scripts
+  and code notebooks which might be potentially executed from
+  multiple platforms. Do not invoke this function during runtime,
+  as it changes the `Nx.Defn` options globally. If you have a
+  specific client that you want to use throughout your project,
+  use configuration files instead:
+
+      config :nx, :default_defn_options, [compiler: EXLA, client: :cuda]
```
polvalente, line 190: I'm thinking here if maybe : global_defn_options or just :defn_options could work here, since Application.get_env is global by definition

nx/test/nx/defn/tree_test.exs
```diff
@@ -6,7 +6,11 @@ defmodule Nx.Defn.TreeTest do
   doctest Nx.Defn.Tree
 
   import Nx.Defn
-  @default_defn_compiler Nx.Defn.Identity
+
+  setup do
+    Nx.Defn.default_options(compiler: Nx.Defn.Identity)
```
polvalente, line 11: In the same vein, instead of default_options, global_options or just options could work here

josevalim: The reason I went with `default` is because they can still be overridden on `Nx.Defn.jit/3` and friends.

exla/lib/exla.ex
seanmor5: +1

seanmor5 (commented):
Beautiful  as always :)

exla/lib/exla.ex
```diff
@@ -184,6 +176,45 @@ defmodule EXLA do
...
+  ## Examples
+
+      EXLA.set_preferred_defn_options([:tpu, :cuda, :rocm, :host])
+
+  The above will try to find the first client available and set
+  the `EXLA` compiler with the client as the compilers for `Nx.Defn`.
+  If no client is found, `EXLA` is not set as compiler at all,
+  therefore it is common to add `:host` as the last option.
+  """
+  def set_preferred_defn_options(clients) do
```
jonatanklosko, line 201: Would it make sense to also accept additional options like `:keep_on_device`?

exla/lib/exla.ex
jonatanklosko: Actually it's nice that it corresponds to to `Nx.Defn.global_default_options`, more precisely it's `global_default_defn_options` and we strip the global part exactly as you said :)

exla/lib/exla.ex
```diff
@@ -184,6 +176,45 @@ defmodule EXLA do
...
+
+    chosen =
+      Enum.find(clients, fn client ->
+        client_config = all_clients[client]
+        client_platform = client_config[:platform] || :host
+        client_config && Map.has_key?(supported_platforms, client_platform)
+      end)
+
+    if chosen do
+      Nx.Defn.global_default_options(compiler: EXLA, client: chosen)
```
jonatanklosko, line 213: `Nx.Defn.set_global_default_options` would be more consistent, but maybe it's too much?

jonatanklosko (commented):
Absolutely fantastic! :cat:

exla/lib/exla.ex
josevalim: Generally speaking, we should be pushing towards using configuration files to change the application environment, which is why we have warnings in the docs to avoid calling the global functions unless scripting.

exla/lib/exla.ex
```diff
@@ -184,6 +176,45 @@ defmodule EXLA do
 
   @behaviour Nx.Defn.Compiler
 
+  @doc """
+  Sets the global defn options to the EXLA compiler with the
+  preferred client based on the available platforms.
```
josevalim:
```suggestion
  preferred client based on their availability.
```

exla/lib/exla.ex
```diff
@@ -184,6 +175,44 @@ defmodule EXLA do
 
   @behaviour Nx.Defn.Compiler
 
+  @doc """
+  Sets the global defn options to the EXLA compiler with the preferred
+  client based on the available platforms.
```
josevalim, line 180:
```suggestion
  client based on their availability.
```

nx/lib/nx.ex
```diff
@@ -2745,6 +2728,11 @@ defmodule Nx do
   This means if you start a separate process, such as `Task`,
   the default backend must be set on the new process too.
 
+  This function is mostly used for scripting and testing. In your
+  applications, you typically set the backend in your config files:
+
+        config :nx, :default_backend, {Lib.CustomBackend, device: :cuda}
```
josevalim, line 2734:
```suggestion
      config :nx, :default_backend, {Lib.CustomBackend, device: :cuda}
```

nx/lib/nx.ex
```diff
@@ -2759,6 +2747,24 @@ defmodule Nx do
...
 
+  @doc """
+  Sets the default backend globally.
+
+  You must avoid calling this function at runtime. It is mostly
+  useful during scripts or code notebooks to set a default.
+  If you need to configure a global default backend in your
+  applications, you can do so in your `config/*.exs` files:
+
+      config :nx, default_backend: {Lib.CustomBackend, []}
```
josevalim, line 2758:
```suggestion
      config :nx, :default_backend, {Lib.CustomBackend, []}
```

nx/lib/nx/defn.ex
```diff
@@ -201,6 +197,60 @@ defmodule Nx.Defn do
...
+  @doc """
+  Sets the default options globally.
+
+  The options defined here apply to all future invocations of
+  `defn`. It also applies to calls to the `jit/3` and `stream/3`
+  functions in this module.
+
+  You must avoid calling this function at runtime. It is mostly
+  useful during scripts or code notebooks to set a default.
+  If you need to configure a global default compiler in your
```
josevalim, line 235:
```suggestion
  If you need to configure a global default options in your
```

josevalim:

PR #559, opened 2021-11-25
Implement hooks for EXLA

### LINKED ISSUE #556: Add defn hooks
This is a proposal for adding hooks to defn. Hooks allow custom
functions to run with values from the device.

In order to explain this functionality, please read the docs for the
upcoming `hook/3` function below:

---

    def hook(tensor_expression, name, function)

Defines a hook.

Hooks are a mechanism to execute an anonymous function for
side-effects with runtime tensor values.

Let's see an example:

    defmodule Hooks do
      import Nx.Defn

      defn add_and_mult(a, b) do
        add = hook(a + b, fn tensor -> IO.inspect({:add, tensor}) end)
        mult = hook(a * b, fn tensor -> IO.inspect({:mult, tensor}) end)
        {add, mult}
      end
    end

The `defn` above defines two hooks, one is called with the
value of `a + b` and another with `a * b`. Once you invoke
the function above, you should see this printed:

    Hooks.add_and_mult(2, 3)
    {:add, #Nx.Tensor<
       s64
       5
    >}
    {:mult, #Nx.Tensor<
       s64
       6
    >}

In other words, the `hook` function accepts a tensor
expression as argument and it will invoke a custom
function with a tensor value at runtime. `hook` returns
the result of the given expression.

Note **you must return the result of the `hook` call**.
For example, the code below won't inspect the `:add`
tuple, because the hook is not returned from `defn`:

    defn add_and_mult(a, b) do
      _add = hook(a + b, fn tensor -> IO.inspect({:add, tensor}) end)
      mult = hook(a * b, fn tensor -> IO.inspect({:mult, tensor}) end)
      mult
    end

We will learn how to hook into a value that is not part
of the result in the "Hooks and tokens" section.

## Named hooks

It is possible to give names to the hooks. This allows them
to be defined or overridden by calling `Nx.Defn.jit/3` or
`Nx.Defn.stream/3`. Let's see an example:

    defmodule Hooks do
      import Nx.Defn

      defn add_and_mult(a, b) do
        add = hook(a + b, :hooks_add)
        mult = hook(a * b, :hooks_mult)
        {add, mult}
      end
    end

Now you can pass the hook as argument as follows:
    
    hooks = %{
      hooks_add: fn tensor ->
        IO.inspect {:add, tensor}
      end
    }

    args = [Nx.tensor(2), Nx.tensor(3)]
    Nx.Defn.jit(&Hooks.add_and_mult/2, args, hooks: hooks)

> **Important!** We recommend to prefix your hook names
> by the name of your library and the custom function,
> to avoid conflicts.

If a named hook is not given, compilers can optimize
that away and not transfer the tensor from the device
in the first place.

You can also mix named hooks with callbacks:

    defn add_and_mult(a, b) do
      add = hook(a + b, :hooks_add, fn tensor -> IO.inspect({:add, tensor}) end)
      mult = hook(a * b, :hooks_mult, fn tensor -> IO.inspect({:mult, tensor}) end)
      {add, mult}
    end

If a hook with the same name is given to `Nx.Defn.jit/3`
or `Nx.Defn.stream/3`, then it will override the default
callback.

## Hooks and tokens

So far, we have always return the result of the `hook`
call. However, what happens if the values we want to
hook are not part of the return value, such as below?

    defn add_and_mult(a, b) do
      _add = hook(a + b, :hooks_add, &IO.inspect({:add, &1}))
      mult = hook(a * b, :hooks_mult, &IO.inspect({:mult, &1}))
      mult
    end

In such cases, you must use tokens. Tokens are used to
create an ordering over hooks, ensuring hooks execute
in a certain order:

    defn add_and_mult(a, b) do
      token = create_token()
      {token, _add} = hook_token(token, a + b, :hooks_add, &IO.inspect({:add, &1}))
      {token, mult} = hook_token(token, a * b, :hooks_mult, &IO.inspect({:mult, &1}))
      attach_token(token, mult)
    end    

The example above creates a token and uses `hook_token/4`
to create hooks attached to said token. By using a token,
we guarantee that those hooks will be invoked in the given
order. Then, at the end of the function, we attach the token
(and the hooks associated with it) to the result `mult`.

Note you must attach the token at the end, otherwise the hooks
will be lost. In fact, the `hook/3` function is implemented like
this:

    def hook(tensor_expr, name, function) do
      {token, result} = hook_token(create_token(), tensor_expr, name, function)
      attach_token(token, result)
    end

josevalim:
Closes #556.

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -34,20 +34,19 @@ defmodule EXLA.Defn do
     lock = EXLA.Defn.Lock.lock(run_key(executable))
     buffers = EXLA.Defn.Buffer.from_nx!(inputs)
 
-    # Now that we have transferred to device, we spawn a task to
-    # execute the stream. We keep this with regular async/await
-    # because this task will never really exist beyond the scope
-    # of the current process.
+    # Now that we have transferred to device, we spawn a runner process
+    # to execute the stream. We use a runner instead of a task to avoid
+    # leaking messages in the inbox. We also don't use a supervisor
+    # to keep them linked, which is safe because the agent is not used
+    # outside the scope of the current process.
     #
-    # Finally, note the task cannot start immediately, we need to
+    # Finally, note the runner cannot start immediately, we need to
     # setup the outfeed reader and register the on_unlock callback
     # that cancels the stream atomically. This is done inside
     # EXLA.Defn.Stream.run.
-    task =
-      Task.async(fn ->
-        receive do
-          ^lock -> EXLA.Executable.run(executable, buffers, run_options)
-        end
+    {:ok, runner} =
+      EXLA.Defn.Runner.start_link(lock, fn ->
+        EXLA.Executable.run(executable, buffers, run_options)
       end)
 
     # The outfeed reader will redirect all outputs with flag 1 to the current
@@ -58,7 +57,7 @@ defmodule EXLA.Defn do
     EXLA.Defn.Stream.run(
       executable,
       lock,
-      task,
+      runner,
       outfeed,
       input,
       input_shape,
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -40,7 +40,9 @@ defmodule EXLA.Defn do
     # of the current process.
     #
     # Finally, note the task cannot start immediately, we need to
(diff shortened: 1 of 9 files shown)


josevalim:

PR #560, opened 2021-11-26
feat(torchx): Nx.product

polvalente:


torchx/lib/torchx/backend.ex
```diff
@@ -428,6 +428,62 @@ defmodule Torchx.Backend do
...
+    check_type!(out_type)
+
+    axes = opts[:axes] || []
+    keep_axes = opts[:keep_axes] || false
+
+    result =
+      if axes == [] do
+        product_whole_tensor(t, keep_axes)
+      else
+        product_over_axes(t, axes, keep_axes)
```
josevalim, line 442: Quick question: why don't we need to perform these checks for `sum` above? :)

polvalente: libtorch accepts a list of axes as a parameter for torch::sum, but for torch::prod we can only pass a single axis per call. Fortunately, we can implement the multi-axis version on top of the single-axis one

polvalente: https://pytorch.org/cppdocs/api/function_namespaceat_1af3361ed6a99b54e317d3b5da8bac4b4b.html

vs

https://pytorch.org/cppdocs/api/function_namespaceat_1abf7768a24e21f76f6d0c2177d526581d.html

polvalente: IntArrayRef is just a fancy version of an int array, for all intents and purposes here

josevalim (approved):

PR #562, opened 2021-11-27
Add custom_grad/3

josevalim:


josevalim:
This is not what we need for now :( Closing.

PR #563, opened 2021-11-27
Poor man's CSE

josevalim:
This is a very inneficient implementation of common
subexpression elimination. This should be enough for
us to assess if implementing a more robust of this
can be useful in the long term.


josevalim:
Too slow to be useful.

PR #564, opened 2021-11-28
Add fan-out to grads

josevalim:
Our grad was incorrect when the same expression appeared
more than once. To address it, we need to collect parents (all
nodes that receive a node as argument) and add their g together.
You can find a discussion of how this should be solved conceptually
in this article:

http://h2.jaguarpaw.co.uk/posts/automatic-differentiation-worked-examples/

In practice, we traverse the nodes, building the child => parent
graphs, and then we make it so the parents fill in the gs of the
children, until all parents are traversed recursively.


nx/lib/nx/defn/grad.ex
```diff
@@ -73,198 +67,198 @@ defmodule Nx.Defn.Grad do
...
   end
 
-  defp zerofy_each(t, cache, ids) do
-    {args, {cache, tainted?}} =
-      Tree.apply_args(t, {cache, false}, fn arg, {cache, acc_tainted?} ->
-        {arg, cache, tainted?} = zerofy_ids(arg, cache, ids)
-        {arg, {cache, tainted? or acc_tainted?}}
-      end)
+  # For some functions, only a subset of the args participate in the grad,
+  # so we handle them accordingly here.
```
polvalente, line 118: How can we ensure new grad implementations pass through here as needed?

Perhaps we explicitly add all functions to either an specific clause or to a clause guard `when fun in ~w(a b c d...)a` instead of a catch-all clause.

This would make it so a new grad raises here and pulls the developers attention to ensure the implementation takes this part into account.

nx/lib/nx/defn/grad.ex
```diff
@@ -870,33 +845,21 @@ defmodule Nx.Defn.Grad do
...
+  @error [:map, :window_product]
 
-  defp no_g_grad(op, _, _) when op in @error do
+  defp grad(op, args, _, _) when op in @error do
     raise ArgumentError, """
-    cannot compute gradient for Nx.#{op}.
+    cannot compute gradient for Nx.#{op}/#{length(args)}.
 
     Consider using stop_grad/1 to make the gradient equivalent to \
     the identify function or use custom_grad/2 to define a proper \
```
polvalente, line 873:
```suggestion
    the identity function or use custom_grad/2 to define a proper \
``

nx/lib/nx/defn/grad.ex
josevalim: I added an automatic check that runs in our tests only. :)

josevalim:
:green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #568, opened 2021-11-29
Remove brackets from tensor expressions

josevalim:
They are not necessary and therefore become noise.


josevalim:
Merged manually due to conflicts.

PR #569, opened 2021-11-30
Use macro escape to work with empty map

seanmor5:
This fixed a compilation error for me for "reset" values in containers

### DIFF
--- a/nx/lib/nx/container.ex
+++ b/nx/lib/nx/container.ex
@@ -125,7 +125,7 @@ defimpl Nx.Container, for: Any do
       defimpl Nx.Container, for: unquote(module) do
         def traverse(%{unquote_splicing(full_pattern)} = struct, var!(acc), var!(fun)) do
           unquote_splicing(updates)
-          {%{unquote_splicing(return)}, var!(acc)}
+          {%{unquote_splicing(Macro.escape(return))}, var!(acc)}
         end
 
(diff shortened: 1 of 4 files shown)


seanmor5:
Oops, I broke many things. I've updated the Test support container module with an additional field / default value that triggers a compilation error to illustrate the issue

polvalente (approved):

polvalente:
> Oops, I broke many things. I've updated the Test support container module with an additional field / default value that triggers a compilation error to illustrate the issue

Lol I approved before reading this comment ¯\_(ツ)_/¯

seanmor5:
Okay that last one did it, not sure if it's the best way to go about it though :)

nx/lib/nx/container.ex
```diff
@@ -119,7 +119,17 @@ defimpl Nx.Container, for: Any do
         end
       end
 
-    return = struct |> Map.to_list() |> Keyword.merge(full_pattern)
+    return =
+      struct
+      |> Map.to_list()
+      |> Keyword.merge(full_pattern)
+      |> Enum.map(fn {k, v} = x ->
+        if k in keep or k in containers do
```
polvalente, line 127: What happens if k in keep but v is a map?

seanmor5: I think in those cases it will get normalized properly? I've updated the test to have default values for both a keep and a reset field.

josevalim:

PR #570, opened 2021-11-30
Gradient for while loops

josevalim:
It is currently inefficient as the loop is executed twice, one for the value and one for the grad. I am working on cleaning it up next.


josevalim:

PR #571, opened 2021-12-03
feat: LU decompostion grad

polvalente:
Implementation for #161 
The base case works, but something is off in the accumulation for the more complex case

definition taken from: https://sethaxen.com/blog/2021/02/differentiating-the-lu-decomposition/


josevalim:
LGTM so far! Make sure to include a link to the article in the source. Perhaps we should do the same for QR. :)

polvalente:
> LGTM so far! Make sure to include a link to the article in the source. Perhaps we should do the same for QR. :)

Good idea!

josevalim (approved):

PR #574, opened 2021-12-04
Add load/save to Nx

### LINKED ISSUE #550: Add option to persist tensors to disk
Currently we can use :erlang.term_to_binary/binary_to_term to do bulk of the work. The only issue is endianness. So we would also need to store the endianness and if the endianness differ, we traverse the tensors remapping them.

seanmor5:
Resolves #550.

Basically just as described in the issue

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -9224,6 +9224,7 @@ defmodule Nx do
   def serialize(tensor_or_container, opts \\ []) do
     data_term = to_term(tensor_or_container)
     term = {@file_version, System.endianness(), data_term}
+
     term
     |> :erlang.term_to_binary(opts)
   end
@@ -9242,13 +9243,15 @@ defmodule Nx do
           Nx.Container.traverse(container, :ok, fn container_elem, :ok ->
             {to_term(container_elem), :ok}
           end)
+
         {:container, serialized}
 
       value ->
-        raise ArgumentError, "unable to serialize #{inspect(value)} as a tensor" <>
-                                " or container. Only tuples and maps are supported" <>
-                                " If you are attempting to serialize a custom container," <>
-                                " you will need to serialize fields in the container manually"
+        raise ArgumentError,
+              "unable to serialize #{inspect(value)} as a tensor" <>
+                " or container. Only tuples and maps are supported" <>
+                " If you are attempting to serialize a custom container," <>
+                " you will need to serialize fields in the container manually"
     end
   end
 
@@ -9307,6 +9310,7 @@ defmodule Nx do
           Nx.Container.traverse(container, :ok, fn container_elem, :ok ->
             {from_term({1, endianness, container_elem}), :ok}
           end)
+
         deserialized
 
       _ ->
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -9185,65 +9185,137 @@ defmodule Nx do
   ## Utilities
 
   @doc """
-  Saves a tensor or container of tensors to the given `path`.
+  Serializes the given tensor or container of tensors to a binary.
 
-  If `path` does not contain `.nx` file extension, it will be appended
-  to the file path.
+  You may pass a tensor, tuple, or map to serialize.
 
   `opts` controls the serialization options. For example, you can choose
   to compress the given tensor or container of tensors by passing a
   compression level:
 
-      Nx.save(tensor, "a.nx", compressed: 9)
+      Nx.serialize(tensor, compressed: 9)
 
-  Compression level corresponds to compression options in `:erlang.term_to_binary/2`. 
+  Compression level corresponds to compression options in `:erlang.term_to_binary/2`.
+
+  ## Examples
+
+      iex> a = Nx.tensor([1, 2, 3])
+      iex> serialized_a = Nx.serialize(a)
+      iex> Nx.deserialize(serialized_a)
+      #Nx.Tensor<
+        s64[3]
+        [1, 2, 3]
+      >
+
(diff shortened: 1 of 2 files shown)


polvalente (approved):

nx/lib/nx.ex
```diff
@@ -8738,6 +8740,68 @@ defmodule Nx do
 
   ## Utilities
 
+  @doc """
+  Saves a tensor or container of tensors to the given `path`.
+
+  If `path` does not contain `.nx` file extension, it will be appended
+  to the file path.
```
josevalim, line 8747: We don't seem to be doing this?

nx/lib/nx.ex
```diff
@@ -8738,6 +8740,68 @@ defmodule Nx do
...
+  """
+  def save(tensor_or_container, path, opts \\ []) do
+    term = {@file_version, System.endianness(), tensor_or_container}
+    term
+    |> :erlang.term_to_binary(opts)
+    |> then(&File.write!(path, &1))
+  end
+
+  @doc """
+  Loads a tensor or container of tensors from the given `path`.
```
josevalim, line 8765:
```suggestion
  Loads a tensor or container of tensors from the given `path`.
  
  It is the opposite of `save/3`.
```

nx/lib/nx.ex
```diff
@@ -8738,6 +8740,68 @@ defmodule Nx do
...
+    case value do
+      %T{shape: shape, type: {_, size} = type} = value ->
+        data =
+          value
+          |> to_binary()
+          |> new_byte_order(size, System.endianness())
+
+        data
+        |> from_binary(type)
+        |> reshape(shape)
```
josevalim, line 8798: Should this actually be an operation in the backend? 🤔

josevalim:
@seanmor5 I have been thinking about this... we cannot simply save any tensor. Should we automatically convert them to binaries before?

There is another issue, if we add new fields to either Nx.Tensor or Nx.BinaryBackend, then the serialized struct is wrong. So I think for saving, we really want to convert them into `{:tensor, shape, type, names, binary}` tuples and then build them back.

For containers, we have a similar issue, the struct can change, so we would need a special way to encode them. One option is `{:struct, fields, containers}`. Another option is to augment the protocol or have a separate protocol for this.

Finally, we may need in the future an API that loads/saves it without writing to a file. What would this API be? What do you think about `Nx.serialize`/`Nx.deserialize` and `Nx.serialize_to_file`/`Nx.deserialize_from_file`. Or alternatively we don't provide the `to_file`/`from_file` versions, since those are one-liners anyway (not providing one is my vote).

seanmor5:
I see what you're saying about the API change - I agree then that we should simply convert each to a tuple with enough info to reconstruct the tensor. I don't think the problem applies to containers though? Ideally even if a container changes the work is actually traversing the container and converting leaves (tensors) to a serializable term that's easy to reconstruct. Then we'd just need to reconstruct them when we deserialize. For our base containers (maps and tuples), I think this is sufficient? 

I can support not providing the IO operations and only providing serialization. Tangentially, I would like to propose a `from_text` utility similar to https://numpy.org/doc/stable/reference/generated/numpy.loadtxt.html

We can shelve that for another day, but it would definitely make AoC parsing easier :D

josevalim:
The issue with container is that the struct in disk may no longer match the struct in memory. For example, if you serialize Axon struct and then you rename a field, the serialized field is invalid.

we could start by supporting only tensors, but I assume the basic use cases will require at least more than one tensor?

josevalim:
Oh, I guess we can just handles tuples and maps and not the other containers. That would be fine for me. We would need to replace the traverse by manual recursion but I think this would be necessary anyway if we want to serialize to {:tensor, …} tuples. So yeah, we can go forward.

seanmor5:
I see the concern, but isn't that also a problem if we convert it to a tuple and then attempt to convert it back to a struct with different field names then what's present in what we already serialized? My thought process is that handling of compatibility between changing structs should fall to library maintainers and not on us. For example in Axon if I needed to freeze training state I would probably go the route you are suggesting with serializing to a standard container and then handling conversion back to training state myself. Does that make sense?

seanmor5:
Along these lines, I think we could add a note in the docs that if you are using a non-standard container, then deserialization may fail if the fields of your struct have changed and we suggest using a standard container and handling conversion internally?

josevalim:
@seanmor5 my main concern is that Elixir does not make it easy or obvious for you to match on a struct with the same name but different fields (i.e. to match on an old struct). So I think an explicit conversion from the container to a {version, a, b, c, ...} tuple is clearer for everyone.

josevalim:
> Along these lines, I think we could add a note in the docs that if you are using a non-standard container, then deserialization may fail if the fields of your struct have changed and we suggest using a standard container and handling conversion internally?

This is also a good route, but I would honestly impose this upfront. :)

nx/lib/nx.ex
```diff
@@ -9182,6 +9184,144 @@ defmodule Nx do
...
+        s64[3]
+        [4, 5, 6]
+      >
+  """
+  def serialize(tensor_or_container, opts \\ []) do
+    data_term = to_term(tensor_or_container)
+    term = {@file_version, System.endianness(), data_term}
+
+    term
+    |> :erlang.term_to_binary(opts)
```
polvalente, line 9229:
```suggestion
    term
    |> :erlang.term_to_binary(opts)
```
```suggestion
    :erlang.term_to_binary(term, opts)
```

polvalente (approved):

josevalim (approved):

PR #578, opened 2021-12-06
Make `:torchx` compatible with Nerves

cocoa-xu:
## Motivations

To integrate `:torchx` into a nerves firmware, we need to make the following changes because `:torchx` currently 
1. only creates a symbolic link to the LibTorch library. And these symbolic links are preserved when generating the firmware. 
   
   And we can't let `fwup` follow symbolic links as that will lead to many duplicates. Because it's common to see 3rd party creates their library files using the following pattern:

   ```
   libname.so -> libname.1.so
   libname.1.so -> libname.1.10.so
   libname.1.10.so
   ```

   If we tell `fwup` to follow symbolic links, then it will largely increase the footprint.

2. `:torchx` only automatically download LibTorch build for `{:unix, :darwin}` and `{:unix, :linux}` at the moment.
   
   It's possible to create minimal LibTorch build and fetch them for nerves builds. Meanwhile, I also made it possible for users to specify their own LibTorch URL/file path.

## Main Changes

- `@libtorch_cache` is now a function (`get_libtorch_cache/0`) instead of a constant. 

   Because nerves-project uses environmental variable `MIX_TARGET` to identify the build target. It allows building for different targets within the same Mix project directory. 

   To avoid conflicts, the name of the cache directory will be using the following template
   ```elixir
   Path.join(__DIR__, "cache/libtorch-#{@libtorch_version}-#{@libtorch_target}-#{libtorch_build_target_name()}")
   ```
- `libtorch_build_target_name/0` returns `$MIX_TARGET` if this is a nerves build, otherwise, it returns "host".

- To make it even more flexible, users can set an environment variable `LIBTORCH_DOWNLOAD_URL` which points to their own build of LibTorch (requires to be a valid .zip file).

   `$LIBTORCH_DOWNLOAD_URL` will be checked in `install_libtorch/2`. And if it exists on the local filesystem, we will skip `download/2` and just copy it to `#{libtorch_zip}`.

- `@libtorch_nerves_url_base` specifies the LibTorch library download URL base.

   In this draft, I'm using my GitHub repo for it, but if it's okay with you, I'd love to transfer that repo to elixir-nx.

   I've tested it on my machine, and I'm currently waiting for the GitHub action to be done.

Cheers,
Cocoa

### DIFF
--- a/exla/c_src/exla/exla_client.cc
+++ b/exla/c_src/exla/exla_client.cc
@@ -92,7 +92,7 @@ xla::StatusOr<std::vector<ExlaBuffer*>> UnpackRunArguments(ErlNifEnv* env,
     arguments = tail;
   }
 
-  return std::move(arg_buffers);
+  return arg_buffers;
 }
 
 xla::StatusOr<ERL_NIF_TERM> UnpackResult(ErlNifEnv* env, std::vector<std::unique_ptr<xla::PjRtBuffer>> result, bool keep_on_device) {


torchx/Makefile
```diff
@@ -5,11 +5,29 @@ TORCHX_SO = $(PRIV_DIR)/torchx.so
 # CMAKE_BUILD_FLAGS = --verbose
 CMAKE_BUILD_DIR ?= $(MIX_APP_PATH)/cmake
 C_SRC = $(shell pwd)/c_src
+NERVES_BUILD ?= nerves_build_no
```
polvalente, line 8: I'm not sure if we should add nerves-specific terminology to the Nx build process. @josevalim @seanmor5  thoughts?

torchx/Makefile
```diff
@@ -5,11 +5,29 @@ TORCHX_SO = $(PRIV_DIR)/torchx.so
...
+
+# this .cmake file makes this project compatible with nerves and it should have no effect on normal build
+TOOLCHAIN_FILE ?= $(shell pwd)/nerves/toolchain.cmake
```
polvalente, line 11: This seems to be nerves-specific as well, so perhaps this default can be removed?

torchx/Makefile
```diff
@@ -5,11 +5,29 @@ TORCHX_SO = $(PRIV_DIR)/torchx.so
...
 build: check_libtorch $(TORCHX_SO)
 
+nerves_build_yes:
+	@ mkdir -p $(PRIV_DIR)
+	@ if [ "$(PRIV_DIR)/libtorch" != "$(abspath $(PRIV_DIR)/libtorch)" ]; then \
+		rm -rf "$(PRIV_DIR)/libtorch" ; \
+		cp -a "$(abspath $(LIBTORCH_DIR)/lib)" "$(PRIV_DIR)/libtorch" ; \
+	fi
+
+nerves_build_no:
```
polvalente, line 24: Instead of `nerves_build_yes/no`, we could go with a more general `generate_symlinks_yes/no` terminology.
Also, instead of having 2 make rules, perhaps we could improve the internal `if` to take the `NERVES_BUILD` variable (which could be called `USE_SYMLINKS`) into account

torchx/Makefile
polvalente: From what I can tell, if this isn't present, everything should work normally, right? So we can just remove this default.

It seems that a good compromise would be adding a section on the Torchx README regarding Nerves build

torchx/mix.exs
```diff
@@ -75,6 +85,33 @@ defmodule Torchx.MixProject do
...
+    case Kernel.function_exported?(Mix, :target, 0) do
+      true ->
+        case Mix.target() do
+          :host -> :os.type()
+          target -> {:nerves, target}
+        end
+
+      _ ->
+        :os.type()
+    end
```
polvalente, line 98:
```suggestion
    if Kernel.function_exported?(Mix, :target, 0) do
      case Mix.target() do
        :host -> :os.type()
        target -> {:nerves, target}
      end
    else
      :os.type()
    end
```

torchx/mix.exs
```diff
@@ -6,7 +6,7 @@ defmodule Torchx.MixProject do
 
   @libtorch_version "1.9.1"
   @libtorch_target "cpu"
-  @libtorch_cache Path.join(__DIR__, "cache/libtorch-#{@libtorch_version}-#{@libtorch_target}")
+  @libtorch_nerves_url_base "https://github.com/cocoa-xu/elixir-nx-libtorch/releases/download/v#{@libtorch_version}"
```
polvalente, line 9: In the same vein as the Makefile comments, this seems to be too nerves-specific to be here.
Perhaps the logic can be reworked to just decide on whether to copy libtorch?

I also don't see why someone couldn't just download this nerves libtorch version and point LIBTORCH_DIR to ir directly, which bypasses the download process altogether.

polvalente: I think you can just revert the changes on mix.exs and point LIBTORCH_DIR to the unpacked .zip you created and things should work as expected (with the changes in the makefile)

torchx/Makefile
cocoa-xu: I'm also worried about this. But the way of building an Erlang/Elixir application leaves us (almost?) no choice...

I wonder if it's possible to move all these compatibility related code to another Elixir package...?

torchx/nerves/toolchain.cmake
```diff
@@ -0,0 +1,28 @@
+# maybe the easiest way to identify a nerves build?
+if(NOT "$ENV{TARGET_CPU}" STREQUAL "" AND NOT "$ENV{TARGET_ARCH}" STREQUAL "" AND NOT "$ENV{TARGET_OS}" STREQUAL "" AND NOT "$ENV{TARGET_ABI}" STREQUAL "")
```
polvalente, line 2: I'll wait on the others to respond on whether we should include Nerves-specific things in Nx. Perhaps we're gonna need another solution for people to have this available without actually incorporating this in the codebase.

polvalente: My main concern is: if we're willing to add Nerves-specific code to Nx, why wouldn't we support X or Y library  with more specific code? Also, changes on the Nerves side could break this code section as well.

polvalente:
Thanks for the contribution @cocoa-xu! There are some points we need to discuss on how we can add these improvements in a more general way, but this looks great overall!

torchx/mix.exs
cocoa-xu: I'm totally okay with either way. The reason I'm doing this is just to simplify the install process. It's not necessary to do this in Nx. But since there is code for downloading LibTorch on `{:unix, :darwin}` and `{:unix, :linux}`, I thought it might be more complete if we add this.

polvalente: The main difference for me is that these are platforms directly supported by libtorch and by the BEAM directly. All other platforms can be supported by pointing directly to the correct LIBTORCH_DIR

cocoa-xu:
> Thanks for the contribution @cocoa-xu! There are some points we need to discuss on how we can add these improvements in a more general way, but this looks great overall!

Thank you for spending time reviewing this PR! 

To be honest, I wish I can do all of this in a standalone Elixir package, instead of submitting it to either `:torchx` or `:nerves`. It just doesn't quite fit on each side, yet it is necessary for people who would like to use both in their projects.

torchx/Makefile
seanmor5: I wouldn't add anything Nerves specific to this repo, but perhaps we can have an `nx_nerves` or similar which keeps these changes as a diff or something and applies them specifically when necessary

polvalente:
> > Thanks for the contribution @cocoa-xu! There are some points we need to discuss on how we can add these improvements in a more general way, but this looks great overall!
> 
> Thank you for spending time reviewing this PR!
> 
> To be honest, I wish I can do all of this in a standalone Elixir package, instead of submitting it to either `:torchx` or `:nerves`. It just doesn't quite fit on each side, yet it is necessary for people who would like to use both in their projects.

I'm sure we'll find a good compromise which makes it this work with Nerves. From what I gathered, if Nerves was to provide the toolchain.cmake somehow, the workflow here could be adapted with only the Makefile changes and some README instructions.

In any case, we could just add an empty default toolchain.cmake which could be overriden by the user. This + the "copy vs creat symlinks" variable + LIBTORCH_DIR being set accordingly seem to be the core of the changes in this PR. Did I miss something?

cocoa-xu:
I'll just revert everything for now until I find a good solution! 

Thanks for everyone spending time on this!

cocoa-xu:
I think it's not easy to make it fully universal, it needs both sides to work on this. For example, Nerves is a good project for running Elixir on (almost) bare-metal hardware. But `TARGET_*` variables are still somewhat specific to Nerves. 

I wonder if we could do something like, well, hard to find the analogy, maybe like CMake or C/C++. C/C++ projects usually are easier to composite together. However, when you have multiple Erlang/Elixir project which uses/compiles other C/C++ libraries inside them, it can be quite tricky.

cocoa-xu:
Is it possible to standardize this for Elixir packages that use/compile C/C++ libraries internally (if their maintainers want to accept a standard)? On [this Nerves webpage](https://hexdocs.pm/nerves/compiling-non-beam-code.html#target-cpu-arch-os-and-abi), they also mentioned that

> The TARGET_* variables are optionally set by the Nerves system. All official Nerves systems set them, but it is not mandatory for forks. These variables are useful for guiding compilation of LLVM-based tools.

If we could have something like CMake, where if you want to do a cross-compile, then you can provide a single toolchain file, and all variables set there will automatically apply to all sub-projects (unless they use hardcoded values).

josevalim:
The reason I am comfortable with using `TARGET_` is because Nerves sets them, they are not specific to Nerves.

So removing `MIX_TARGET` is one step forward. Let’s give this step and then we figure out the next step next :)

cocoa-xu:
> Is it possible to standardize this for Elixir packages that use/compile C/C++ libraries internally (if their maintainers want to accept a standard)? On [this Nerves webpage](https://hexdocs.pm/nerves/compiling-non-beam-code.html#target-cpu-arch-os-and-abi), they also mentioned that
> 
> > The TARGET_* variables are optionally set by the Nerves system. All official Nerves systems set them, but it is not mandatory for forks. These variables are useful for guiding compilation of LLVM-based tools.
> 
> If we could have something like CMake, where if you want to do a cross-compile, then you can provide a single toolchain file, and all variables set there will automatically apply to all sub-projects (unless they use hardcoded values).

And one more point (or making this general issue more clear), I found that it's not clear for an Elixir (or perhaps I should say Mix) project to know which toolchain (not particular means `.cmake` toolchain) should they use when they have one or more internal C/C++ projects.

fhunleth:
@cocoa-xu I was just pointed to this thread. I don't see the code any more, but the comments that I'm reading so far seem good - especially moving to the `TARGET_` and other variables. To date, cmake hasn't been used much with Nerves projects, but I can expose a `toolchainfile.cmake`. I just noticed that the one we use for cmake projects in Buildroot isn't published. It might need some tweaks, though. For regular `Makefile` projects, the exposed environment variables cover most needs and often "just work" since they match Makefile conventions.

cocoa-xu:
> @cocoa-xu I was just pointed to this thread. I don't see the code any more, but the comments that I'm reading so far seem good - especially moving to the `TARGET_` and other variables. To date, cmake hasn't been used much with Nerves projects, but I can expose a `toolchainfile.cmake`. I just noticed that the one we use for cmake projects in Buildroot isn't published. It might need some tweaks, though. For regular `Makefile` projects, the exposed environment variables cover most needs and often "just work" since they match Makefile conventions.

Hi @fhunleth, sorry I was focusing on replying to this thread, I'll push these changes to my fork (it's kinda messy so maybe no PR for now until we figure out how to make it more like a universal approach).

Issue #579, opened 2021-12-06
"Scalar" term confusion


floatn:
I've just been confused by finding out that Nx.to_scalar is used to convert a zero-dimension tensor to scalar. Since, technically, 0-dim tensor is already considered a scalar. I suggest we fix the terms by, for example, calling tensor a 'tensor', 0-dim tensor a 'scalar', and non-tensor scalar a 'number'. Since it sticks to `Kernel.is_number`. WDYT?

josevalim:
I like it. Please do send a PR to rename `Nx.to_scalar` to `Nx.to_number`. :)

floatn:
Ok.

PR #580, opened 2021-12-06
Fix some inconsistency with terminology

floatn:
#579

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -1584,7 +1584,7 @@ defmodule Nx do
       1
 
       iex> Nx.to_number(Nx.tensor([1.0, 2.0, 3.0]))
-      ** (ArgumentError) cannot convert tensor of shape {3} to a number
+      ** (ArgumentError) cannot convert tensor of shape {3} to number
 
   """
   @doc type: :conversion
@@ -1612,11 +1612,11 @@ defmodule Nx do
 
   ## Examples
 
-      iex> Nx.to_scalar(1)
+      iex> Nx.to_number(1)
       1
 
-      iex> Nx.to_scalar(Nx.tensor([1.0, 2.0, 3.0]))
-      ** (ArgumentError) cannot convert tensor of shape {3} to scalar
+      iex> Nx.to_number(Nx.tensor([1.0, 2.0, 3.0]))
+      ** (ArgumentError) cannot convert tensor of shape {3} to number
 
   """
   @doc type: :conversion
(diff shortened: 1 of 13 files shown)


nx/lib/nx.ex
```diff
@@ -1573,21 +1573,54 @@ defmodule Nx do
...
       1
 
-      iex> Nx.to_scalar(Nx.tensor([1.0, 2.0, 3.0]))
-      ** (ArgumentError) cannot convert tensor of shape {3} to scalar
+      iex> Nx.to_number(Nx.tensor([1.0, 2.0, 3.0]))
+      ** (ArgumentError) cannot convert tensor of shape {3} to number
 
   """
   @doc type: :conversion
+  @deprecated "Use Nx.to_number/1 instead"
```
josevalim, line 1623: We don't need to deprecate it, you can go ahead and remove it. :)

nx/lib/nx/defn/compiler.ex
```diff
@@ -48,7 +48,7 @@ defmodule Nx.Defn.Compiler do
 
   # These operations do not have valid meaning for Nx.Defn.Expr
   @forbidden_ops [:backend_copy, :backend_deallocate, :backend_transfer] ++
-                   [:to_binary, :to_scalar, :to_flat_list, :to_heatmap, :to_batched_list] ++
+                   [:to_binary, :to_number, :to_scalar, :to_flat_list, :to_heatmap, :to_batched_list] ++
```
josevalim, line 51:
```suggestion
                   [:to_binary, :to_number, :to_flat_list, :to_heatmap, :to_batched_list] ++
```

josevalim:
Thank you! I have dropped two minor comments. Also please update exla/torchx accordingly, the amount of changes there should be minimal. :)

nx/lib/nx.ex
floatn: Axon will not be happy.

josevalim: Ok, let's keep this for a week and then we can change it. :)

floatn: :) Alright.

nx/lib/nx/defn/compiler.ex
floatn: In a week.

josevalim, line 51: This is annoying the formatter and without this the code will fail to compile anyway, so we are fine.

```suggestion
                   [:to_binary, :to_number, :to_flat_list, :to_heatmap, :to_batched_list] ++
```

josevalim (approved):

josevalim:

Issue #582, opened 2021-12-07
Implement `Nx.LinAlg.matrix_power`


rhbvkleef:
As a consequence of [Advent of Code 2021, day 6](https://adventofcode.com/2021/day/7), we discovered that Nx doesn't have a matrix_power function:

`@spec matrix_power(Nx.t, positive_integer) :: Nx.t`

This function would result in the power of the matrix by dot-product instead of by multiplication. We can solve this problem in linear time, using (source: https://github.com/josevalim/aoc/blob/main/2021/day-06.livemd)

```elixir
def matrix_power(matrix, power) do
  Enum.reduce(1..(power - 1), matrix, fn _, acc ->
    Nx.dot(matrix, acc)
  end
end
```

but, this is `O(n)` where n is the power. It is possible to do this in `O(log n)`, namely with repeated squaring:

```elixir
def matrix_power(matrix, 0) do
  Nx.subtract(matrix, matrix)
end
def matrix_power(matrix, power) do
  Integer.digits(power, 2)
  |> Enum.reverse()
  |> Enum.reduce({nil, matrix}, fn bit, {result_matrix, exp_matrix} ->
    cond do
      result_matrix == nil && bit == 1 -> {exp_matrix, Nx.dot(exp_matrix, exp_matrix)}
      result_matrix == nil && bit == 0 -> {nil, Nx.dot(exp_matrix, exp_matrix)}
      bit == 1 -> {Nx.dot(result_matrix, exp_matrix), Nx.dot(exp_matrix, exp_matrix)}
      bit == 0 -> {result_matrix, Nx.dot(exp_matrix, exp_matrix)}
    end
  end)
  |> elem(0)
end
```

I'm sure that this code is quite ugly, and should be refactored, but the logic should be clear. For external sources, reference can be made to the NumPy implementation: https://github.com/numpy/numpy/blob/main/numpy/linalg/linalg.py#L554. To see this implementation used, refer to [my Advent of Code 2021 day 6 solution](https://github.com/rhbvkleef/aoc2021/blob/master/lib/solutions/year2021/day6.ex)

polvalente:
Wouldn't `matrix_power(matrix, 0), do: eye(matrix)` be the correct implementation instead of `subtract(matrix, matrix)`?

Besides that, I think a PR is welcome, as this kind of operation is common in numerical calculus.

One question that this would raise is that we need to choose between `power` and `matrix_power` for the implementation of the `**` operator in `defn`.

[...]

PR #583, opened 2021-12-08
Implement Nx.LinAlg.dot_power/2 (#582)

rhbvkleef:
This PR implements the matrix_power function as requested in #582, using repeated squaring to calculate the power in logarithmic time.

Two doctests are provided, one to make sure that identity matrices don't change, and one to test a simple exponent case. This should be enoguh to catch serious implementation issues, and enough to provide documentation on what this function actually does.

Note that technically, if we were to ever support arbitrarily large numbers, this algorithm would take more in the neighborhood of linear time, as the multiplication of very large integers becomes slower with the width of such integers.

### DIFF
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -967,4 +967,48 @@ defmodule Nx.LinAlg do
       opts
     )
   end
+
+  @doc """
+  Produces the tensor taken to the given power by dot-product.
+
+  The input is always a square tensor and a non-negative integer,
+  and the output is a square tensor of the same dimensions as the input tensor.
+
+  # Examples
+
+      iex> Nx.LinAlg.matrix_power(Nx.tensor([[1, 2], [3, 4]]), 6)
+      #Nx.Tensor<
+        s64[2][2]
+        [
+          [5743, 8370],
+          [12555, 18298]
+        ]
+      >
+
+      iex> Nx.LinAlg.matrix_power(Nx.eye(3), floor(:math.pow(2, 32))-1)
+      #Nx.Tensor<
+        s64[3][3]
+        [
+          [1, 0, 0],
+          [0, 1, 0],
+          [0, 0, 1]
+        ]
+      >
+
+  """
+  @doc from_backend: false
+  def matrix_power(matrix, 0) do
+    Nx.eye(matrix)
+  end
+  def matrix_power(matrix, power) do
+    Integer.digits(power, 2)
+    |> Enum.reverse()
+    |> Enum.reduce({Nx.eye(matrix), matrix}, fn bit, {result_matrix, exp_matrix} ->
+      case bit do
+        1 -> {Nx.dot(result_matrix, exp_matrix), Nx.dot(exp_matrix, exp_matrix)}
+        0 -> {result_matrix, Nx.dot(exp_matrix, exp_matrix)}
+      end
+    end)
+    |> elem(0)
+  end
 end
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -1026,37 +1026,31 @@ defmodule Nx.LinAlg do
   def dot_power(tensor, 0) do
     # We need a special-case for 0 since the code below
     # is optimized to not compute an initial eye.
-    Nx.eye(tensor)
-  end
-
-  def dot_power(tensor, 1) do
-    # We also need a special case for 1, because the code
-    # below is optimized for calculating powers for
-    # integers where `length(Integer.digits(n, 2)) > 1`.
     Nx.Defn.Kernel.assert_shape_pattern(tensor, {x, x})
 
-    tensor
+    Nx.eye(tensor)
   end
 
   def dot_power(tensor, power) when is_integer(power) do
     Nx.Defn.Kernel.assert_shape_pattern(tensor, {x, x})
 
-    {result, exp_tensor} =
-      power
-      |> Integer.digits(2)
-      |> tl()
-      |> Enum.reverse()
-      |> Enum.reduce({nil, tensor}, fn
-        1, {nil, exp_tensor} ->
-          {exp_tensor, Nx.dot(exp_tensor, exp_tensor)}
+    power
+    |> Integer.digits(2)
+    |> tl()
+    |> Enum.reverse()
+    |> Enum.reduce({nil, tensor}, fn
+      1, {nil, exp_tensor} ->
+        {exp_tensor, Nx.dot(exp_tensor, exp_tensor)}
 
-        1, {result_tensor, exp_tensor} ->
-          {Nx.dot(result_tensor, exp_tensor), Nx.dot(exp_tensor, exp_tensor)}
+      1, {result_tensor, exp_tensor} ->
+        {Nx.dot(result_tensor, exp_tensor), Nx.dot(exp_tensor, exp_tensor)}
 
-        0, {result_tensor, exp_tensor} ->
-          {result_tensor, Nx.dot(exp_tensor, exp_tensor)}
-      end)
-
-    Nx.dot(result, exp_tensor)
+      0, {result_tensor, exp_tensor} ->
+        {result_tensor, Nx.dot(exp_tensor, exp_tensor)}
+    end)
+    |> then(fn
+      {nil, exp_tensor} -> exp_tensor
+      {result, exp_tensor} -> Nx.dot(result, exp_tensor)
+    end)
   end
 end
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -1014,6 +1014,9 @@ defmodule Nx.LinAlg do
           [1.5, -0.5]
         ]
       >
+
+      iex> Nx.LinAlg.dot_power(Nx.tensor([[1, 2], [3, 4], [5, 6]]), 1)
+      ** (ArgumentError) expected tensor to match shape {x, x}, got tensor with shape {3, 2}
   """
   @doc from_backend: false
   def dot_power(tensor, power) when is_integer(power) and power < 0 do
@@ -1030,14 +1033,18 @@ defmodule Nx.LinAlg do
     # We also need a special case for 1, because the code
     # below is optimized for calculating powers for
     # integers where `length(Integer.digits(n, 2)) > 1`.
+    Nx.Defn.Kernel.assert_shape_pattern(tensor, {x, x})
+
     tensor
   end
 
   def dot_power(tensor, power) when is_integer(power) do
+    Nx.Defn.Kernel.assert_shape_pattern(tensor, {x, x})
+
     {result, exp_tensor} =
       power
       |> Integer.digits(2)
-      |> tl
+      |> tl()
       |> Enum.reverse()
       |> Enum.reduce({nil, tensor}, fn
         1, {nil, exp_tensor} ->
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -1006,15 +1006,20 @@ defmodule Nx.LinAlg do
 
   """
   @doc from_backend: false
-  def matrix_power(matrix, power) when is_integer(power) and power >= 0 do
-    Integer.digits(power, 2)
+  def matrix_power(matrix, power) when is_integer(power) and power < 0 do
+    matrix_power(Nx.LinAlg.invert(matrix), -power)
+  end
+  def matrix_power(matrix, power) when is_integer(power) do
+    {result, _exp_matrix} = power
+    |> Integer.digits(2)
     |> Enum.reverse()
-    |> Enum.reduce({Nx.eye(matrix), matrix}, fn bit, {result_matrix, exp_matrix} ->
-      case bit do
-        1 -> {Nx.dot(result_matrix, exp_matrix), Nx.dot(exp_matrix, exp_matrix)}
-        0 -> {result_matrix, Nx.dot(exp_matrix, exp_matrix)}
-      end
+    |> Enum.reduce({Nx.eye(matrix), matrix}, fn
+      1, {result_matrix, exp_matrix} ->
+        {Nx.dot(result_matrix, exp_matrix), Nx.dot(exp_matrix, exp_matrix)}
+      0, {result_matrix, exp_matrix} ->
+        {result_matrix,                     Nx.dot(exp_matrix, exp_matrix)}
     end)
-    |> elem(0)
+
+    result
   end
 end
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -974,9 +974,11 @@ defmodule Nx.LinAlg do
   The input is always a square tensor and a non-negative integer,
   and the output is a square tensor of the same dimensions as the input tensor.
 
+  The dot-products are unrolled inside `defn`.
+
   ## Examples
 
-      iex> Nx.LinAlg.matrix_power(Nx.tensor([[1, 2], [3, 4]]), 0)
+      iex> Nx.LinAlg.dot_power(Nx.tensor([[1, 2], [3, 4]]), 0)
       #Nx.Tensor<
         s64[2][2]
         [
@@ -985,7 +987,7 @@ defmodule Nx.LinAlg do
         ]
       >
 
-      iex> Nx.LinAlg.matrix_power(Nx.tensor([[1, 2], [3, 4]]), 6)
+      iex> Nx.LinAlg.dot_power(Nx.tensor([[1, 2], [3, 4]]), 6)
       #Nx.Tensor<
         s64[2][2]
         [
@@ -994,7 +996,7 @@ defmodule Nx.LinAlg do
         ]
       >
 
-      iex> Nx.LinAlg.matrix_power(Nx.eye(3), 65535)
+      iex> Nx.LinAlg.dot_power(Nx.eye(3), 65535)
       #Nx.Tensor<
         s64[3][3]
         [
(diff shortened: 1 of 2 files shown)


nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,48 @@ defmodule Nx.LinAlg do
...
+          [0, 0, 1]
+        ]
+      >
+
+  """
+  @doc from_backend: false
+  def matrix_power(matrix, 0) do
+    Nx.eye(matrix)
+  end
+  def matrix_power(matrix, power) do
```
rhbvkleef, line 1003: It might be a good idea to guard `power` with `is_integer/1`.

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,48 @@ defmodule Nx.LinAlg do
...
+    Integer.digits(power, 2)
+    |> Enum.reverse()
+    |> Enum.reduce({Nx.eye(matrix), matrix}, fn bit, {result_matrix, exp_matrix} ->
```
rhbvkleef, line 1006: So, using Nx.eye(matrix) as the start of the accumulator for result makes the solution nicer (and I'm pretty sure it makes the special-case for the power 0 redundant), but it does technically make the solution slightly slower, because we need to do one extra dot-product because of it. I chose the elegant solution, as I think that's a price we're willing to pay for code that reads nicer.

rhbvkleef: After testing, yes: this makes the special-case for the power of 0 unnecessary.

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,48 @@ defmodule Nx.LinAlg do
...
+        [
+          [1, 0, 0],
+          [0, 1, 0],
+          [0, 0, 1]
+        ]
+      >
+
+  """
+  @doc from_backend: false
+  def matrix_power(matrix, 0) do
```
rhbvkleef, line 1000: Todo: remove this special case.

rhbvkleef: But! Do insert a test for the 0-case then!

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,53 @@ defmodule Nx.LinAlg do
...
     )
   end
+
+  @doc """
+  Produces the tensor taken to the given power by dot-product.
+
+  The input is always a square tensor and a non-negative integer,
+  and the output is a square tensor of the same dimensions as the input tensor.
+
+  # Examples
```
seanmor5, line 977:
```suggestion
  ## Examples
  
```

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,53 @@ defmodule Nx.LinAlg do
...
+      iex> Nx.LinAlg.matrix_power(Nx.tensor([[1, 2], [3, 4]]), 6)
+      #Nx.Tensor<
+        s64[2][2]
+        [
+          [5743, 8370],
+          [12555, 18298]
+        ]
+      >
+
+      iex> Nx.LinAlg.matrix_power(Nx.eye(3), floor(:math.pow(2, 32))-1)
```
seanmor5, line 996: I think this might be a bit much to include in a doctest. I would instead put this in `nx_test`.

seanmor5 (commented):
Nice work! Just two minor comments :)

nx/lib/nx/lin_alg.ex
rhbvkleef: It's not very slow though. It just takes 64-ish dot-products of an identity matrix. I'll go time it for you.

rhbvkleef: On my box it takes 511µs.

seanmor5: Sorry, I did not mean too much in terms of performance. What I meant was as an example in the docs it's a little noisy when compared to the other 2 examples. While it is a simple test, on the surface it may seem confusing or complex for someone who is just trying to understand what this function does. Does that make sense?

rhbvkleef: Ah, yes. I can put `65535` into it or so. That way the noise is removed, while much of the spirit of the test remains?

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,54 @@ defmodule Nx.LinAlg do
...
+          [1, 0, 0],
+          [0, 1, 0],
+          [0, 0, 1]
+        ]
+      >
+
+  """
+  @doc from_backend: false
+  def matrix_power(matrix, power) when is_integer(power) and power >= 0 do
+    Integer.digits(power, 2)
```
polvalente, line 1010:
```suggestion
    power
    |> Integer.digits(2)
```

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,54 @@ defmodule Nx.LinAlg do
...
+    |> Enum.reverse()
+    |> Enum.reduce({Nx.eye(matrix), matrix}, fn bit, {result_matrix, exp_matrix} ->
+      case bit do
+        1 -> {Nx.dot(result_matrix, exp_matrix), Nx.dot(exp_matrix, exp_matrix)}
+        0 -> {result_matrix, Nx.dot(exp_matrix, exp_matrix)}
+      end
+    end)
```
polvalente, line 1017: I would remove this case in favor of using a multiclause fn for the reduce

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,54 @@ defmodule Nx.LinAlg do
...
+    |> elem(0)
```
polvalente, line 1018: Instead of `elem(0)` I think hard matching on the tuple can be more explict on what's happening

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,54 @@ defmodule Nx.LinAlg do
...
+        [
+          [1, 0, 0],
+          [0, 1, 0],
+          [0, 0, 1]
+        ]
+      >
+
+  """
+  @doc from_backend: false
+  def matrix_power(matrix, power) when is_integer(power) and power >= 0 do
```
polvalente, line 1009: You can use Nx.LinAlg.invert to support power < 0 here

if power < 0, use matrix = invert(matrix)

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,59 @@ defmodule Nx.LinAlg do
...
+      iex> Nx.LinAlg.matrix_power(Nx.eye(3), 65535)
+      #Nx.Tensor<
+        s64[3][3]
+        [
+          [1, 0, 0],
+          [0, 1, 0],
+          [0, 0, 1]
+        ]
+      >
+
```
polvalente, line 1006: We need an example for when power < 0.

I'd add one that shows that matrix_power(t, -1) == invert(t)

And another which has a bigger absolute value for the power.

polvalente: Btw, @josevalim, this makes me think that `**` should become matrix_power for when the left operand is a tensor. WDYT?

rhbvkleef: The behaviour could change depending on what the right operator is: an integer could cause this function, and another equally-sized tensor could cause pairwise exponentiation.

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,59 @@ defmodule Nx.LinAlg do
...
+  def matrix_power(matrix, power) when is_integer(power) and power < 0 do
+    matrix_power(Nx.LinAlg.invert(matrix), -power)
+  end
+  def matrix_power(matrix, power) when is_integer(power) do
+    {result, _exp_matrix} = power
+    |> Integer.digits(2)
+    |> Enum.reverse()
+    |> Enum.reduce({Nx.eye(matrix), matrix}, fn
+      1, {result_matrix, exp_matrix} ->
+        {Nx.dot(result_matrix, exp_matrix), Nx.dot(exp_matrix, exp_matrix)}
```
rhbvkleef, line 1018: I did come to the realization that this algorithm potentially does one dot-product too many, namely the dot product to calculate the last (discarded) exponential matrix. This is also the most costly matrix to calculate, though again, with the maximum number width of 64 bits, this doesn't matter much.

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,59 @@ defmodule Nx.LinAlg do
...
+
+  """
+  @doc from_backend: false
+  def matrix_power(matrix, power) when is_integer(power) and power < 0 do
+    matrix_power(Nx.LinAlg.invert(matrix), -power)
+  end
+  def matrix_power(matrix, power) when is_integer(power) do
+    {result, _exp_matrix} = power
+    |> Integer.digits(2)
+    |> Enum.reverse()
```
polvalente, line 1015: If you extract this list to a variable, you can get it's length, and then iterate with the `Enum.reduce` by also accumulating the current index. Then you can skip the last superfluous dot-product by just returning the result matrix without calculating the next `exp_matrix`. This will also simplify things because you don't need to return a tuple in the last iteration, which enables you to remove the pattern match on `{result, _exp_matrix}`

nx/lib/nx/lin_alg.ex
polvalente: Suggestion above :)

polvalente: Also, you're probably missing a `mix format` in the code

josevalim:
Hi everyone,

Sorry for the disappearance but I have been thinking about this. Here is my collection of notes so far:

1. I am not very comfortable with the name `matrix_power`. If we ever support vmap in the future, then this function needs to support `rank(tensor) >= 2 `, where the outer dimensions work as a batch. So is it worth having the word `matrix` in the name? Would `dot_power` be a better name? I.e. we compute the power using dots instead of multiplication.

2. I think `**` should work as pairwise exponential because `*` is pairwise multiplication. However, there was a proposal to introduce `<*>` as the dot operator, and maybe we should introduce `<**>` as dot power?

3. I think we should at least check the shape of the input. We may be able to call `Nx.Defn.Kernel.assert_shape` here.

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,88 @@ defmodule Nx.LinAlg do
...
+      |> Enum.reduce({Nx.eye(matrix), matrix}, fn
+        1, {result_matrix, exp_matrix} ->
+          {Nx.dot(result_matrix, exp_matrix), Nx.dot(exp_matrix, exp_matrix)}
+
+        0, {result_matrix, exp_matrix} ->
+          {result_matrix, Nx.dot(exp_matrix, exp_matrix)}
+      end)
+
+    Nx.dot(result, exp_matrix)
+  end
```
josevalim, line 1053: If you want to remove the special case for 0:

```suggestion
  def matrix_power(matrix, power) when is_integer(power) do
    power
    |> Integer.digits(2)
    |> Enum.reverse()
    |> Enum.reduce({Nx.eye(matrix), matrix}, fn
      1, {result_matrix, exp_matrix} ->
        {Nx.dot(result_matrix, exp_matrix), Nx.dot(exp_matrix, exp_matrix)}

      0, {result_matrix, exp_matrix} ->
        {result_matrix, Nx.dot(exp_matrix, exp_matrix)}
    end)
    |> elem(0)
  end
```

However, there is a small issue where we are always starting from with an eye multiplication, which we can actually optimize out. So I would keep the special case and then do the other suggestion I will post next.

josevalim:
```suggestion
  def matrix_power(matrix, 0) do
    # We need a special-case for 0 since the code below
    # is optimized to not compute an initial eye.
    Nx.eye(matrix)
  end

  def matrix_power(matrix, power) when is_integer(power) do
    power
    |> Integer.digits(2)
    |> Enum.reverse()
    |> Enum.reduce({nil, matrix}, fn
      # It is guaranteed that we will have at least one digit, when that happens,
      # we start with exp_matrix to avoid multiplying by eye.
      1, {nil, exp_matrix} ->
        {exp_matrix, Nx.dot(exp_matrix, exp_matrix)}

      1, {result_matrix, exp_matrix} ->
        {Nx.dot(result_matrix, exp_matrix), Nx.dot(exp_matrix, exp_matrix)}

      0, {result_matrix, exp_matrix} ->
        {result_matrix, Nx.dot(exp_matrix, exp_matrix)}
    end)
    |> elem(0)
  end
```

josevalim: Theoretically we can optimize the eye operation away, but only inside `defn`, so I would ship with the optimized form.

rhbvkleef: I specifically changed from an implementation similar to yours to the special 0-case because the implementation you proposed computes one extra fairly large dot-product. Specifically that one might (under certain circumstances) take about 33% of the full execution time of this function.

rhbvkleef:
> I am not very comfortable with the name matrix_power. If we ever support vmap in the future, then this function needs to support rank(tensor) >= 2 , where the higher dimensions work as a batch. So is it worth having the word matrix in the name? Would dot_power be a better name? I.e. we compute the power using dots instead of multiplication.

I think dot-power is completely acceptable. I named it `matrix_power` because that was the term used in NumPy, though I suppose we need to attempt to generalize to tensors here.

> I think ** should work as pairwise exponential because * is pairwise multiplication. However, there was a proposal to introduce <*> as the dot operator, and maybe we should introduce <**> as dot power?

I agree. After reading that particular issue, I thought of the same.

> I think we should at least check the shape of the input. We may be able to call Nx.Defn.Kernel.assert_shape here.

That's a good suggestion. I'll take a look at that.

rhbvkleef:
> I think we should at least check the shape of the input. We may be able to call `Nx.Defn.Kernel.assert_shape` here.

Though, how would I go about that. As we want it to be generalized for tensors, even `Nx.Defn.Kernel.assert_shape_pattern` is not suited to this purpose. The specific test we would need here is that all elements in the shape tuple are equal.

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,89 @@ defmodule Nx.LinAlg do
...
+    # below is optimized for calculating powers for
+    # integers where `length(Integer.digits(n, 2)) > 1`.
+    tensor
+  end
+
+  def dot_power(tensor, power) when is_integer(power) do
+    {result, exp_tensor} =
+      power
+      |> Integer.digits(2)
+      |> tl
```
josevalim, line 1040:
```suggestion
      |> tl()
```

josevalim:
> Though, how would I go about that. As we want it to be generalized for tensors, even Nx.Defn.Kernel.assert_shape_pattern is not suited to this purpose. 

I think right now we can keep it working only for matrices and we can generalize it later, especially because nothing in Nx.LinAlg works with batching dimensions.

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,96 @@ defmodule Nx.LinAlg do
...
+          [1.5, -0.5]
+        ]
+      >
+
+      iex> Nx.LinAlg.dot_power(Nx.tensor([[1, 2], [3, 4], [5, 6]]), 1)
+      ** (ArgumentError) expected tensor to match shape {x, x}, got tensor with shape {3, 2}
+  """
+  @doc from_backend: false
+  def dot_power(tensor, power) when is_integer(power) and power < 0 do
+    dot_power(Nx.LinAlg.invert(tensor), -power)
```
polvalente, line 1023:
```suggestion
    tensor |> invert() |> dot_power(abs(power))
```

```suggestion
    dot_power(invert(tensor), abs(power))
```

I know `- power when power < 0` is equivalent to `abs`, but I think `abs` reduces the cognitive overload for the code here. I don't have a strong opinion on whether to use the pipe here, though.

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,96 @@ defmodule Nx.LinAlg do
...
+  end
+
+  def dot_power(tensor, 0) do
```
polvalente, line 1026: Don't we need `assert_shape_pattern(tensor, {x, x})` here as well?

nx/lib/nx/lin_alg.ex
```diff
@@ -967,4 +967,96 @@ defmodule Nx.LinAlg do
...
+          {exp_tensor, Nx.dot(exp_tensor, exp_tensor)}
+
+        1, {result_tensor, exp_tensor} ->
+          {Nx.dot(result_tensor, exp_tensor), Nx.dot(exp_tensor, exp_tensor)}
+
+        0, {result_tensor, exp_tensor} ->
+          {result_tensor, Nx.dot(exp_tensor, exp_tensor)}
+      end)
+
+    Nx.dot(result, exp_tensor)
```
polvalente, line 1060:
```suggestion
      power
      |> Integer.digits(2)
      |> tl()
      |> Enum.reverse()
      |> Enum.reduce({nil, tensor}, fn
        1, {nil, exp_tensor} ->
          {exp_tensor, Nx.dot(exp_tensor, exp_tensor)}

        1, {result_tensor, exp_tensor} ->
          {Nx.dot(result_tensor, exp_tensor), Nx.dot(exp_tensor, exp_tensor)}

        0, {result_tensor, exp_tensor} ->
          {result_tensor, Nx.dot(exp_tensor, exp_tensor)}
      end)
      |> then(fn {result, exp_tensor} -> 
        Nx.dot(result, exp_tensor)
      end)
```

WDYT?

polvalente: The diff ended up with low readability. I basically removed the hard-match in favor of using `Kernel.then` to keep everything in a pipe

nx/lib/nx/lin_alg.ex
rhbvkleef: Nx.eye does an assertion, but we could.

nx/lib/nx/lin_alg.ex
rhbvkleef: I came to the realization that I made a mistake with the extra 1-case (specifically, it fails for all `power=2^n`. I'll take your suggestion, and alter it slightly to fix that mistake.

josevalim:
I pushed one commit to solve Torchx suite and we will be good to go!

polvalente (approved):

seanmor5:
FWIW - I am not in favor of changing from matrix_power. I don't think this will generalize to higher dimensions, and for vmap, the operation would still be on a "batch" of matrices, so matrix_power is still technically correct in that sense.

josevalim:
Alright, I will merge this then and rollback on the name, so @rhbvkleef  doesn't have to rename once more.

josevalim:

PR #584, opened 2021-12-08
Allow users to build torchx with embedded libtorch lib directory

cocoa-xu:
Following [this](https://github.com/elixir-lang/elixir_make/blob/84655d2643796bf35489fd2b6936ddc5e904b913/lib/mix/tasks/compile.make.ex#L88) comment/suggestion/guideline from `elixir_make`.

> However, note that Mix projects supports the `:build_embedded`
  configuration, which controls if assets in the `_build` directory
  are symlinked (when `false`, the default) or copied (`true`).
  In order to support both options for `:build_embedded`, it is
  important to follow the given guidelines:

Users can use the environment variable `BUILD_EMBEDDED` to control how the libtorch lib directory will be processed.
- "false" (default), symlink it
- otherwise, copy the libtorch lib directory. Symbolic links inside the libtorch lib directory are preserved.


torchx/Makefile
```diff
@@ -22,7 +22,11 @@ check_libtorch:
 $(TORCHX_SO): c_src/torchx.cpp c_src/nx_nif_utils.hpp
 	@ mkdir -p $(PRIV_DIR)
 	@ mkdir -p $(CMAKE_BUILD_DIR)
-	@ ln -sf $(abspath $(LIBTORCH_DIR)/lib) $(PRIV_DIR)/libtorch
+	@ if [ "${BUILD_EMBEDDED}" = "false" ]; then \
+		ln -sf $(abspath $(LIBTORCH_DIR)/lib) $(PRIV_DIR)/libtorch ; \
+	else \
+		cp -a $(abspath $(LIBTORCH_DIR)/lib) $(PRIV_DIR)/libtorch ; \
+	fi
 	@ cd $(CMAKE_BUILD_DIR) && \
```
polvalente, line 30: Nice addition!

I think what remains to be solved is the optional toolchain cmake file, right?
Perhaps @fhunleth can help with this?

We could provide an empty toolchain file which is the default for the env var, and then it can be overriden as needed with the one provided by, say, Nerves

torchx/mix.exs
```diff
@@ -19,7 +19,14 @@ defmodule Torchx.MixProject do
       compilers: compilers() ++ Mix.compilers(),
       elixirc_paths: elixirc_paths(Mix.env()),
       aliases: aliases(),
-      make_env: %{"LIBTORCH_DIR" => System.get_env("LIBTORCH_DIR", @libtorch_cache)}
+      make_env: %{
+        "LIBTORCH_DIR" => System.get_env("LIBTORCH_DIR", @libtorch_cache),
+        "BUILD_EMBEDDED" =>
+          System.get_env(
+            "BUILD_EMBEDDED",
+            "#{Keyword.get(Mix.Project.config(), :build_embedded)}"
```
polvalente, line 27: Doesn't this render "nil" if `:build_embedded` is not defined in `Mix.Project.config()`?

torchx/Makefile
```diff
@@ -22,7 +22,11 @@ check_libtorch:
 $(TORCHX_SO): c_src/torchx.cpp c_src/nx_nif_utils.hpp
 	@ mkdir -p $(PRIV_DIR)
 	@ mkdir -p $(CMAKE_BUILD_DIR)
-	@ ln -sf $(abspath $(LIBTORCH_DIR)/lib) $(PRIV_DIR)/libtorch
+	@ if [ "${BUILD_EMBEDDED}" = "false" ]; then \
+		ln -sf $(abspath $(LIBTORCH_DIR)/lib) $(PRIV_DIR)/libtorch ; \
+	else \
+		cp -a $(abspath $(LIBTORCH_DIR)/lib) $(PRIV_DIR)/libtorch ; \
+	fi
```
polvalente, line 29: I'd invert the test to validate if `"${BUILD_EMBEDDED}" = "true"`, which makes my comment on `"nil"` below kinda superfluous

torchx/mix.exs
cocoa-xu: Hi @polvalente, this default value of this variable is set to `false` by `elixir_make`. At least, it is said so in its inline docs [here](https://github.com/elixir-lang/elixir_make/blob/84655d2643796bf35489fd2b6936ddc5e904b913/lib/mix/tasks/compile.make.ex#L90).

polvalente: Cool. I'd still invert the test to check for "true" because this will make the behavior closer to Elixir, in which "false" and "nil" will test as falsy

torchx/Makefile
cocoa-xu: OK, I've inverted the test

torchx/mix.exs
josevalim: There is another option here, which is to extract the archive directly to priv, instead of the cache directory. This way we get rid of the symlink altogether. WDYT @polvalente?

polvalente (approved):

torchx/Makefile
cocoa-xu: I'm not entirely sure about how `elixir_make` works, but I think we can do it in `elixir_make` by adding a `toolchain`(or `cmake_toolchain`, or whatever suits best) option to `elixir_make`. And then, we can set the location of the optional toolchain CMake file in the top project (where torchx is a dependency), and fetch and use it here.

torchx/mix.exs
polvalente: Oh, I didn't see this comment in time for the merge.
This only solves the issue when we're using autodownload, though. The solution in this PR will also work when we're dealing with user-provided LIBTORCH_DIR values as well (in which we don't use the cache dir)

torchx/Makefile
polvalente: This sounds nice, but I think that elixir_make is not aware of cmake at all, which is why we need to have a makefile on the top-level that calls cmake.

polvalente: Perhaps @josevalim can chip in on how we can use elixir_make with the cmake toolchain without actually having to make elixir_make itself aware of this.

It might be helpful to clarify how a given cmake toolchain file would interact with torchx, for instance.

cocoa-xu: I changed my mind. I think maybe we should not make any changes to elixir_make, instead, we set an env var, say `CMAKE_TOOLCHAIN_FILE`, in the top project. And, as you said, we provide an empty toolchain file here by default and add `-D CMAKE_TOOLCHAIN_FILE=$(CMAKE_TOOLCHAIN_FILE)` to the Makefile.

torchx/mix.exs
```diff
@@ -19,7 +19,14 @@ defmodule Torchx.MixProject do
       compilers: compilers() ++ Mix.compilers(),
       elixirc_paths: elixirc_paths(Mix.env()),
       aliases: aliases(),
-      make_env: %{"LIBTORCH_DIR" => System.get_env("LIBTORCH_DIR", @libtorch_cache)}
+      make_env: %{
+        "LIBTORCH_DIR" => System.get_env("LIBTORCH_DIR", @libtorch_cache),
+        "BUILD_EMBEDDED" =>
+          System.get_env(
+            "BUILD_EMBEDDED",
```
josevalim, line 26: Do we need this initial `System.get_env("BUILD_EMBEDDED")`? This is typically set through Mix, not through env vars. :)

cocoa-xu: No, I think we don't need to read it from env. It's easier to change it from a single source.

PR #586, opened 2021-12-08
Allow users to use their own toolchain

cocoa-xu:
Allow users to use their own toolchain by setting env var "CMAKE_TOOLCHAIN_FILE". 

But I'm not sure if this is where you would like to put this empty CMake toolchain file (`torchx/toolchain.cmake`). If not, could you please suggest the best location for this file? Thanks!

Also, I had a thought that if we want to avoid adding an empty file, we can use

```
CMAKE_TOOLCHAIN_FILE ?= /dev/null
```

but that's probably not gonna work on systems that do not have a `/dev/null` (e.g., Windows), or its `/dev/null` behaves differently than what we would have expected.

### DIFF
--- a/torchx/Makefile
+++ b/torchx/Makefile
@@ -5,6 +5,9 @@ TORCHX_SO = $(PRIV_DIR)/torchx.so
 # CMAKE_BUILD_FLAGS = --verbose
 CMAKE_BUILD_DIR ?= $(MIX_APP_PATH)/cmake
 C_SRC = $(shell pwd)/c_src
+ifdef CMAKE_TOOLCHAIN_FILE
+	CMAKE_CONFIGURE_FLAGS=-DCMAKE_TOOLCHAIN_FILE="$(CMAKE_TOOLCHAIN_FILE)"
+endif
 
 .DEFAULT_GLOBAL := build
 
@@ -27,9 +30,6 @@ $(TORCHX_SO): c_src/torchx.cpp c_src/nx_nif_utils.hpp
 	else \
 		ln -sf $(abspath $(LIBTORCH_DIR)/lib) $(PRIV_DIR)/libtorch ; \
 	fi
-ifneq ($(CMAKE_TOOLCHAIN_FILE),)
-	CMAKE_CONFIGURE_FLAGS=-DCMAKE_TOOLCHAIN_FILE="$(CMAKE_TOOLCHAIN_FILE)"
-endif
 	@ cd $(CMAKE_BUILD_DIR) && \
 		cmake -DCMAKE_PREFIX_PATH=$(LIBTORCH_DIR) -DC_SRC=$(C_SRC) \
 			-DERTS_INCLUDE_DIR=$(ERTS_INCLUDE_DIR) -S $(shell pwd) \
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @fhunleth, sorry for the ping, but is this how you would usually approach this?

torchx/Makefile
```diff
@@ -4,6 +4,7 @@ TORCHX_SO = $(PRIV_DIR)/torchx.so
 # If you want verbose output for debugging
 # CMAKE_BUILD_FLAGS = --verbose
 CMAKE_BUILD_DIR ?= $(MIX_APP_PATH)/cmake
+CMAKE_TOOLCHAIN_FILE ?= $(shell pwd)/toolchain.cmake
```
polvalente, line 7: This makefile is already Unix-specific, so it's probably fine to use /dev/null as the default

cocoa-xu: Oh, that's true. Let's use /dev/null instead of adding an empty file.

polvalente:
> If you don't like the `/dev/null`, you can do:
> 
> ```
> ifneq ($(CMAKE_TOOLCHAIN_FILE),)
> CMAKE_CONFIGURE_FLAGS=-DCMAKE_TOOLCHAIN_FILE=$(CMAKE_TOOLCHAIN_FILE)
> endif
> ```
> 
> and then use `$(CMAKE_CONFIGURE_FLAGS)` below.

@cocoa-xu I like this idea of having a CMAKE_CONFIGURE_FLAGS variable instead of resorting do /dev/null. It seems a little bit cleaner

cocoa-xu:
Yep, I'm in favour of having a `CMAKE_CONFIGURE_FLAGS` var too!

cocoa-xu:
@fhunleth Thanks for putting the toolchainfile.cmake here! I just had a try, and all changes we made in this repo worked as expected. We only need to adjust some lines in the toolchainfile.cmake to make everything works. Let's address relevant issues in the nerves toolchain repo. :)

Lastly, thanks for everyone's time and effort on this! ❤️

torchx/Makefile
```diff
@@ -27,8 +27,12 @@ $(TORCHX_SO): c_src/torchx.cpp c_src/nx_nif_utils.hpp
 	else \
 		ln -sf $(abspath $(LIBTORCH_DIR)/lib) $(PRIV_DIR)/libtorch ; \
 	fi
+ifneq ($(CMAKE_TOOLCHAIN_FILE),)
+	CMAKE_CONFIGURE_FLAGS=-DCMAKE_TOOLCHAIN_FILE="$(CMAKE_TOOLCHAIN_FILE)"
+endif
```
polvalente, line 32: I think this should be at the top of the file, right after `C_SRC`. Also, I think this will work and be more clear:
```suggestion
ifdef $(CMAKE_TOOLCHAIN_FILE)
	CMAKE_CONFIGURE_FLAGS=-DCMAKE_TOOLCHAIN_FILE="$(CMAKE_TOOLCHAIN_FILE)"
endif
```

cocoa-xu: Putting CMAKE_* together indeed looks better and more clear. I'll commit these changes!

josevalim (approved):
😍

cocoa-xu:
Solved conflicts in the Makefile ;)

polvalente (approved):

Issue #362, opened 2021-03-31
Add mechanism for optional backend functions


josevalim:
While we provide built-in implementations for inverse, tile, etc. This has the benefit that we get `grad` of these functions for free. However, we also want to allow backends to provide their own implementation for performance reasons.

This may seem straight-forward at first but the biggest issue is related to how optional backend function ties in to Expr/grad. We will need Expr to implement all optional callbacks while keeping the default implementation for grad.

seanmor5:
I propose adding a `custom_call/2` metadata node or similar which looks something like:

`custom_call(:target, fallback_impl_fn)`

Then compilers and backends can register custom calls for specific targets, during evaluation and compilation those special targets are dispatched to the registered custom call implementation. In the backward pass it is just metadata so the gradient is still valid.

This would be useful in `defn` as well. For example in Axon some functions or layers might have specialized implementations. BatchNorm for example has an XLA specialized implementation which is faster than the one that is generated from the code. Torchx would benefit as well because things like softmax and other activations would not need to fall back on the evaluator

PR #591, opened 2021-12-14
Adding the Upsample layer support to AxonOnnx.Deserialize

stefkohub:
In order to start working on adding upsample layer type support for AxonOnnx, I started adding the interpolate Libtorch function (which is the current function superseding the upsample one).
So I added some new code trying to follow as much as possible the code style and function conventions.
Next steps will be:
1. add this somewhere into Nx (Maybe into Nx.LinAlg?) 
2. create the function into Axon.Deserialize

Are you interested into this functionality?

### DIFF
--- a/torchx/c_src/nx_nif_utils.hpp
+++ b/torchx/c_src/nx_nif_utils.hpp
@@ -272,6 +272,25 @@ namespace nx
       return 1;
     }
 
+    int get_list(ErlNifEnv *env, ERL_NIF_TERM list, std::vector<double> &var)
+    {
+      unsigned int length;
+      if (!enif_get_list_length(env, list, &length))
+        return 0;
+      var.reserve(length);
+      ERL_NIF_TERM head, tail;
+
+      while (enif_get_list_cell(env, list, &head, &tail))
+      {
+        double elem;
+        if (!get(env, head, &elem))
+          return 0;
+        var.push_back(elem);
+        list = tail;
+      }
+      return 1;
+    }
+
     int get_list(ErlNifEnv *env, ERL_NIF_TERM list, std::vector<torch::Tensor> &var)
     {
       unsigned int length;
--- a/torchx/c_src/torchx.cpp
+++ b/torchx/c_src/torchx.cpp
@@ -395,7 +395,7 @@ NIF(interpolate)
   torch::Tensor result;
   if (argc == 2) 
   {
-    // The default mode is kNearest
+    // In case we receive (tensor, size) the default mode is kNearest
     LIST_PARAM(1, std::vector<int64_t>, size);
     result = F::interpolate(*t, 
 		    F::InterpolateFuncOptions()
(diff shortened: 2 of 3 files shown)


josevalim:
Hi @stefkohub! We are definitely interested in this, but our approach to adding new functions has been slightly different:

1. First we must implement those functions with pure Nx on top of the existing primitives
2. This means non-optimized versions will be available for all backends/compilers
3. Then we can add backend specific optimizations, but this requires #362 which I am to prioritize

Does that make sense?

That said, we would really appreciate help with step 1. Perhaps we can leave this open and revisit once step 1 and #362 are complete?

seanmor5:
My opinion is that this is too specific to Axon to include here. A libtorch implementation would be useful for those wanting to use libtorch but that would need to be included as a custom call. I have a draft upsample PR that partially implements this functionality. It is possible purely in defn which is the route I would go so it's not backend specific.

josevalim:
I am wondering how we can navigate these waters in the long term though. If we want to optimize it for Torchx, then we need to have specified as part of the backend? Unless we let Axon tell the Torchx specific implementation, but then Torchx concerns start leaking into Axon.

Perhaps we should start by benchmarking the pure Nx with the Torchx specific implementation? Maybe upscale is implemented on top of existing primitives in Torchx and there won't be such a large difference after all...

stefkohub:
Hi all, and thank you for your answers!
Yes that's a Torchx specific implementation since that's the quickest solution I found to obtain a tensor upsample/downsample functionality. I am still learning the great job you all have done since now, and now I am better understanding the approach and priorities.
Since @seanmor5 is already working on an upsample layer in Nx I think I'll wait for the PR and I'll try to help improving its solution.
So I agree with you @josevalim, please leave this open until the #362 and the native upsample PR are available.
The benchmarking proposal is quite interesting. Maybe some functionalities (pure Nx or coming from any other library) have different performances than the expected and will drive some choices.

loretoparisi:
> I am wondering how we can navigate these waters in the long term though. If we want to optimize it for Torchx, then we need to have specified as part of the backend? Unless we let Axon tell the Torchx specific implementation, but then Torchx concerns start leaking into Axon.
> 
> 
> 
> Perhaps we should start by benchmarking the pure Nx with the Torchx specific implementation? Maybe upscale is implemented on top of existing primitives in Torchx and there won't be such a large difference after all...

At the end of the day, `libtorch` is the best option for deep learning. From keras back and forth within Tensorflow we have learned a lot of things...
I would see a fully fledged axon Torch backend, so I vote for a specific Torchx implementation and optimization for Nx!

seanmor5:
@stefkohub It seems that ONNX Upsample is deprecated in favor of Resize. I've started implementing `Axon.Layers.resize` to support the Resize operator in AxonOnnx. Will that work for your use case?

stefkohub:
Hi @seanmor5, that's new for me too. I just read about it now and it seems the Resize operator is a superset of the Upsample. So I just need to understand how this new operator woks and modify my darknet_to_onnx code to create an Onnx model using Resize layer instead of Upsample ones... And then wait for your new code to import it into an AxonOnnx model. Thank you!

seanmor5:
I believe this has been addressed now as `resize` is apart of Axon and Upsample has been added to AxonOnnx. We can reevaluate adding a `resize` to the Nx API later.

PR #595, opened 2021-12-24
feat(torchx): add svd

polvalente:


torchx/test/torchx/nx_linalg_doctest_test.exs
```diff
@@ -22,6 +22,8 @@ defmodule Torchx.NxLinAlgDoctestTest do
     norm: 2,
     # qr - Torchx: "geqrf_cpu" not implemented for 'Long' in NIF.qr/2
     qr: 2,
+    # svd - Torchx: "svd_cpu" not implemented for 'Long' in NIF.svd/2
+    svd: 2,
```
josevalim, line 26: Maybe we should remove those long tests from doctests and move them to unit tests in Nx so we can keep those doctests?

josevalim (approved):

torchx/test/torchx/nx_linalg_doctest_test.exs
polvalente: I tried to edit them, but then I got -0.0 here where Nx has 0.0, and SVD is also not unique, so the result is slightly different.

I'll copy the tests to a linalg test module for Torchx

PR #596, opened 2021-12-30
to avoid divided by 0.0 when all the elements of the tensor data are the same value.

kikuyuta:
The `to_heatmap` function causes exceptions by zero division when all the elements of the given tensor data are the same. In that case, the values of both `max` and `min` are the same so `base` becomes zero that will used as the divisor.
I made a quick hack to use 1 as the divisor instead of `base` to avoid zero-division. In this case, all the cell will be displayed as black. I use the integer 1 as the divisor because I expect the compiler might optimize to limit divide operations, but I am not sure about that.

### DIFF
--- a/nx/lib/nx/heatmap.ex
+++ b/nx/lib/nx/heatmap.ex
@@ -75,7 +75,7 @@ defmodule Nx.Heatmap do
     defp render(%{shape: {size}} = tensor, _opts, _doc, entry_fun, line_fun) do
       data = Nx.to_flat_list(tensor)
       {data, [], min, max} = take_min_max(data, size)
-      base = max - min
+      base = if max == min, do: 1, else: max - min
 
       data
       |> Enum.map(fn elem -> entry_fun.((elem - min) / base) end)
@@ -104,7 +104,7 @@ defmodule Nx.Heatmap do
 
     defp chunk([], acc, limit, {rows, cols, entry_fun, line_fun}, _docs) do
       {acc, rest, min, max} = take_min_max(acc, rows * cols)
-      base = max - min
+      base = if max == min, do: 1, else: max - min
 
       {[], doc} =
         Enum.reduce(1..rows, {acc, empty()}, fn _, {acc, doc} ->


nx/lib/nx/heatmap.ex
```diff
@@ -75,7 +75,7 @@ defmodule Nx.Heatmap do
     defp render(%{shape: {size}} = tensor, _opts, _doc, entry_fun, line_fun) do
       data = Nx.to_flat_list(tensor)
       {data, [], min, max} = take_min_max(data, size)
-      base = max - min
+      base = if max == min, do: 1, else: max - min
```
polvalente, line 78: What happens when the tensor is composed of all zeros, but on of the positions is 1.0e-10 or some low value like that?

It would be interesting to see how this fix behaves in a test for the all-zeros and for my proposed case.

Also, for all zeros, show the heatmap render as if it was all max values or if it was all min values?

kikuyuta: In the current version,  for example, the following cases cause exceptions.

- `[0, 0] |> Nx.tensor |> Nx.to_heatmap`
- `[[-1, -1, -1], [-1, -1, -1]]  |> Nx.tensor |> Nx.to_heatmap` 

The second one shows the following.
```
%Inspect.Error{
  message: "got ArithmeticError with message \"bad argument in arithmetic expression\" while inspecting %{__struct__: Nx.Heatmap, opts: [], tensor: #Nx.Tensor<\n    s64[2][3]\n    [\n      [-1, -1, -1],\n      [-1, -1, -1]\n    ]\n  >}"
}
```

I found this when I used `to_heatmap` for my `Conway's game of life` program, not for any real tensor computation. When all the cell are dead, my program crashes.
https://twitter.com/kikuyuta/status/1476205911886929921
I do not know that we should consider those situation in natural science, but I feel it is so convenient for us to make quick hack for visualizing data in programs even if they are not for tensor computation of natural science.

For example, assuming all the celles are zeros, min and max are zero and base (= max - min) is zero, so `data |> Enum.map(fn elem -> entry_fun.((elem - min) / base) end)` causes a division by zero in the current version. The `render` function of my version will make all the cells black in those cases.

polvalente: Just to be clear, I do think we need to fix this edge case!

I think that rendering black is the more natural result, so since your version does exatcly that, it should be good to go.

Could you add those cases as test cases for the feature? And if there isn't one like the one I suggested, with mostly zeros and an almost zero value, already, that one too, just to ensure we're handling all of these.

kikuyuta: Ah, extreme cases as you said causes arithmetic exceptions.
`[1.0e200, 1.0e-200] |> Nx.tensor |> Nx.to_heatmap`
is a kind.

Anyway, it becomes midnight in Japan. I will think about these cases tomorrow.

kikuyuta: It seems we should consider more to avoid crash from any arbitrary tensor data, especially from extreme data. I could not find a quick hack. I think it is possible to use the bit patterns of the number representation, instead of `(elem - min) / base` expression, but just an idea now.
I will leave my console for a week due to my vacation, so I want to know any ideas about `to_heatmap` before leaving. If there is some pointers or links, please let me know.

1. Do you think `to_heatmap` must not crash by any tensor data? I am not familiar with machine learning and other natural science using tensor, so I am not sure that my worry is meaningful or meaningless.
2. Do you think it is acceptable to change the *from tensor to heat* mapping? In other words, can we break the backward compatibility around the heatmap functions?
3. What should the heat map of [1, 10, 100] be? Currently [black, black, white], but [black, gray, white] might be better.

josevalim: Other implementations of heatmap use special values to deal with NaN and Infinity. However, we can't model infinity/nan in the Erlang VM yet, which is something I hope we can change by OTP 25 (OTP 26 latest). So my suggestion is the following:

1. If the value is NaN/Infinity, then we do nothing for now, we will handle this in the future

2. If the value is truly an exception (division by zero), then we should special handle it

So my proposal is to merge this PR and tackle the remaining issues later. :) WDYT?

Thank you @kikuyuta, happy new year and enjoy your holidays!

josevalim:

nx/lib/nx/heatmap.ex
kikuyuta: Okay, I understand the situation. I agree.
Thanks @polvalente and @josevalim, happy new year!

josevalim: For completion, we are now handling non-finite numbers in heatmap since 2eeb6178d2669a1b3f2bae83058a57c252f110b9.

Issue #587, opened 2021-12-09
Nx.window_reduce Compilation error case


floatn:
This
```elixir
defmodule NxTest do
import Nx.Defn

  defn test() do
    Nx.window_reduce(Nx.tensor([1, 2, 3]),
      0, {2}, [padding: :same],
      fn x, acc -> x + acc end
    )
  end
end

NxTest.test() |> IO.inspect()
```
gives me:

== Compilation error in file lib/nx_test.ex ==
** (MatchError) no match of right hand side value: 1
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:623: Nx.BinaryBackend.element_wise_bin_op/4
    (nx 0.1.0-dev) lib/nx/defn/evaluator.ex:147: Nx.Defn.Evaluator.eval_apply/4
    (nx 0.1.0-dev) lib/nx/defn/evaluator.ex:63: Nx.Defn.Evaluator.eval/3
    (nx 0.1.0-dev) lib/nx/defn/evaluator.ex:83: anonymous fn/4 in Nx.Defn.Evaluator.eval_apply/4
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:1353: Nx.BinaryBackend."-window_reduce/6-_@5/2-2-"/4
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:1351: anonymous fn/9 in Nx.BinaryBackend.window_reduce/6
    (elixir 1.12.2) lib/enum.ex:2385: Enum."-reduce/3-lists^foldl/2-0-"/3
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:1345: Nx.BinaryBackend.window_reduce/6
```

seanmor5:
This seems to go wrong in here: https://github.com/elixir-nx/nx/blob/main/nx/lib/nx/defn/evaluator.ex#L81 where the input arguments don't get cast to tensors correctly and the binary backend implementation of add doesn't like that

seanmor5:
Note this doesn't fail outside of `defn`

PR #598, opened 2022-01-03
Add: elementwise ops for numbers + Fix: window_reduce + Fix: typo

floatn:
#587 
There's a case when an element_wise operation is appled to a number. Maybe it's ok to convert it to scalar just before the operation.

### DIFF
--- a/nx/test/nx/defn/evaluator_test.exs
+++ b/nx/test/nx/defn/evaluator_test.exs
@@ -32,6 +32,13 @@ defmodule Nx.Defn.EvaluatorTest do
     assert %T{shape: {3, 2}, type: {:s, 64}} = reshape(Nx.iota({2, 3}))
   end
 
+  defn reduce_window(t1, acc),
+    do: Nx.window_reduce(t1, acc, {2}, [padding: :valid], fn x, acc -> x + acc end)
+
+  test "window reduce" do
+    assert reduce_window(Nx.tensor([1, 2, 3]), 0) == Nx.tensor([3, 5])
+  end
+
   describe "decompositions" do
     defn lu(t), do: Nx.LinAlg.lu(t)
 
--- a/nx/lib/nx/binary_backend.ex
+++ b/nx/lib/nx/binary_backend.ex
@@ -593,6 +593,12 @@ defmodule Nx.BinaryBackend do
     end
   end
 
+  defp element_wise_bin_op(out, left, right, fun) when is_number(left) or is_number(right) do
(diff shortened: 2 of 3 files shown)


seanmor5:
@floatn Nice! Can you add a test which verifies #587 is fixed? For the additional scalar clause on element-wise ops, is there a failure case you're specifically addressing? If so, can you add a test for that as well?

polvalente (commented):
Can you provide a test case on which this change is needed?

Would the same test work in EXLA as well? (Assuming EXLA does implement window_reduce, I don't actually remember this)

floatn:
@seanmor5
I was thinking about adding a test. It requires a defn call during test. It's possible, right?
Number-clause on element-wise op is what partially was a cause of #587

@polvalente 
There's a simple test case, provided in #587 description. I'll work on adding it to defn tests.

nx/test/nx/defn/evaluator_test.exs
```diff
@@ -32,6 +32,13 @@ defmodule Nx.Defn.EvaluatorTest do
     assert %T{shape: {3, 2}, type: {:s, 64}} = reshape(Nx.iota({2, 3}))
   end
 
+  defn reduce_window(t1, acc),
+    do: Nx.window_reduce(t1, acc, {2}, [padding: :valid], fn x, acc -> x + acc end)
+
+  test "window reduce" do
+    assert reduce_window(Nx.tensor([1, 2, 3]), 0) == Nx.tensor([3, 5])
```
polvalente, line 39: So the problem is actually on the accumulator being passed as a number, right?

It seems to me that this should also break outside of BinaryBackend.

Perhaps the "to_tensor" call should actually be in the Nx module itself instead of the backend implementation.

polvalente: Also, if this is the case and my suggestion does fix the problem, the BinaryBackend changes should be rolled back

floatn: > So the problem is actually on the accumulator being passed as a number, right?

`x` value also is passed as a number. And pre-converting acc to scalar before actually starting reduce process
https://github.com/elixir-nx/nx/blob/ce396667e54ee7096d9ab05506df083949fbda84/nx/lib/nx/binary_backend.ex#L1361
will break some non-defn cases.

polvalente: Thanks for the explanation!

polvalente: And for the PR :D

PR #599, opened 2022-01-03
test: fix svd property test

### LINKED ISSUE #526: Nx.LinAlg.svd doesn't work in some tests for the BinaryBackend
#519 showed that for the "property" tests, SVD doesn't work in the BinaryBackend. It seems that the problem lies in the right eigenvector matrix (Vt where `A = U.Sigma.Vt`)

polvalente:
closes #526 

My attempt was to show that SVD does converge as expected, but depending on the matrix conditioning, it can be slow.
Should we remove these tests instead, since we have fixed-matrix ones?


nx/test/nx/lin_alg_test.exs
```diff
@@ -339,52 +339,83 @@ defmodule Nx.LinAlgTest do
...
-
-        # TODO: SVD does not work for wide matrices and
-        # raises a non-semantic error
-
-        #  wide = Nx.random_uniform({3, 4})
-
-        # assert {u, d, vt} = Nx.LinAlg.svd(wide)
-        # m = u |> Nx.shape() |> elem(1)
-        # n = vt |> Nx.shape() |> elem(0)
+      random_offset = fn -> :rand.uniform() * 0.5 + 0.5 end
```
polvalente, line 343: @josevalim as per the description on the PR, I'm thinking that we should just remove this test entirely instead. WDYT?
We could open an issue for a better implementation of SVD instead, since it seems ours has convergence problems

josevalim: I would remove the test and call it a day, especially as I plan to work on optional callbacks soon. We have a barebones implementation. If you want performance and accuracy, use one of the backends.

PR #601, opened 2022-01-04
feat(torchx): Nx.any?

polvalente:


torchx/lib/torchx/backend.ex
```diff
@@ -438,13 +441,26 @@ defmodule Torchx.Backend do
   ## Aggregators
 
   @impl true
-  def sum(%T{type: out_type} = out, %T{} = t, opts) do
-    check_type!(out_type)
+  def sum(%T{} = out, %T{} = t, opts) do
+    out =
```
polvalente, line 445: @josevalim I ended up having to make sum do the conversion automatically, because otherwise Nx.LinAlg.determinant would have to be aware of this quirk. WDYT about this? I think the same could be done to Nx.product.

polvalente: Also, it seems that `any?`, `all` and `product` all need the same "aggregate over axes" logic. Should I do this in this PR or should I leave it to a follow-up one?

polvalente: Also also, shouldn't we rename `any?` to `any`?

nx/lib/nx/lin_alg.ex
```diff
@@ -1148,9 +1148,13 @@ defmodule Nx.LinAlg do
 
     is_zero = diag |> Nx.equal(0) |> Nx.any?() |> Nx.to_number() |> Kernel.==(1)
 
-    iota = Nx.iota({n})
+    iota = Nx.iota({n}, type: type, backend: backend)
```
josevalim, line 1151: If `determinant` was a defn, then this would not be necessary. Should we convert it to a `defn`?

polvalente: I'll leave this to a follow-up PR

torchx/lib/torchx/backend.ex
```diff
@@ -9,10 +9,13 @@ defmodule Torchx.Backend do
         iex> Nx.tensor([1, 2, 3], type: {:u, 16}, backend: Torchx.Backend)
         ** (ArgumentError) Torchx does not support unsigned 16 bit integer
 
-    2. Torchx doesn't support u8 on sums, you should convert input to signed integer.
+    2. Torchx doesn't support usigned integers on sums and will convert them to signed integers.
```
josevalim, line 12:
```suggestion
    2. Torchx doesn't support unsigned integers on sums and will convert them to signed integers.
```

josevalim (approved):
One small comment and ship it!

PR #603, opened 2022-01-04
refactor: determinant as defn

polvalente:


nx/lib/nx/lin_alg.ex
```diff
@@ -1117,38 +1117,59 @@ defmodule Nx.LinAlg do
...
-    end)
+    # This is an unrwapped Enum.reduce(0..2, Nx.tensor(0), fn offset, acc -> .. end)
+
+    Nx.tensor(0)
+    |> Nx.add(diagonal_product(pos_t, 0))
+    |> Nx.subtract(diagonal_product(neg_t, 0))
+    |> Nx.add(diagonal_product(pos_t, 1))
+    |> Nx.subtract(diagonal_product(neg_t, 1))
+    |> Nx.add(diagonal_product(pos_t, 2))
+    |> Nx.subtract(diagonal_product(neg_t, 2))
```
polvalente, line 1159: The compiler complained about Enum.reduce not being defined as defn, so I ended up unwrapping it since there were few iterations. Should I use while instead?

nx/lib/nx/type.ex
```diff
@@ -363,7 +365,7 @@ defmodule Nx.Type do
   end
 
   def merge_number({:u, size}, integer) when is_integer(integer) do
-    merge_number({:s, size * 2}, integer)
+    merge_number({:s, min(size * 2, 64)}, integer)
```
polvalente, line 368: This was yielding `{:s, 128}` for args `{:u, 64}` and any negative integer

nx/lib/nx/lin_alg.ex
```diff
@@ -1160,7 +1181,7 @@ defmodule Nx.LinAlg do
```
polvalente, line 1181: The Nx.sum calls here are facing the same problem as before, in which Torchx can't return {:u, 64}, and the defn compiler enforces that check. A simple solution is converting the boolean tensors to {:s, 8}, but this would imply that Nx-core is aware of Torchx's limitations somehow.

Any ideas?

cc @seanmor5 @josevalim

nx/lib/nx/lin_alg.ex
```diff
@@ -1117,38 +1117,59 @@ defmodule Nx.LinAlg do
...
+    cond do
+      Nx.equal(n, 2) ->
+        determinant_2by2(tensor)
+
+      Nx.equal(n, 3) ->
+        determinant_3by3(tensor)
+
+      :otherwise ->
+        determinant_NbyN(tensor)
+    end
```
josevalim, line 1134: I would do this inside a transform because we don't want to the comparison and the branching inside `defn` (i.e. inside the gpu). Something like:

```suggestion
    transform(n, fn
      2 -> determinant_2by2(tensor)
      3 -> determinant_3by3(tensor)
      _ -> determinant_NbyN(tensor)
    end)
```

nx/lib/nx/lin_alg.ex
josevalim: Nah, this is fine as is. You could also use a transform.

nx/lib/nx/lin_alg.ex
josevalim: > the defn compiler enforces that check

Do you mean the evaluator is raising when we get a different value from Torchx? Ping me later and we can pair ono it!

nx/lib/nx/lin_alg.ex
```diff
@@ -1117,60 +1117,79 @@ defmodule Nx.LinAlg do
...
+    transform(n, fn
+      2 ->
+        determinant_2by2(tensor)
+
+      3 ->
+        determinant_3by3(tensor)
+
+      _ ->
+        determinant_NbyN(tensor)
+    end)
```
polvalente, line 1134: I've rewritten all of these defnps to a more "defn-like" syntax, instead of "elixir-like". Let me know if I should roll this back

torchx/test/torchx/backend_test.exs
```diff
@@ -559,6 +559,7 @@ defmodule Torchx.BackendTest do
       )
     end
 
+    @tag :skip
```
josevalim, line 562: Should we add a TODO?

polvalente, line 562:
```suggestion
    # TO-DO: deal with `Nx.sum` returning s64 on Torchx instead of u64
    @tag :skip
```

PR #604, opened 2022-01-04
Propagate device_id option to EXLA computation

jonatanklosko:
Currently the `device_id` configured in global defn options is ignored by EXLA.

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -318,7 +318,7 @@ defmodule EXLA.Defn do
         {computation, extra, hooks} =
           to_computation.(expr || fun.(vars), inputs_and_shapes, used_hooks)
 
-        executable = EXLA.Computation.compile(computation, client, shapes)
+        executable = EXLA.Computation.compile(computation, client, shapes, options)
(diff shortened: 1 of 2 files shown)


exla/lib/exla/defn.ex
```diff
@@ -318,7 +318,7 @@ defmodule EXLA.Defn do
         {computation, extra, hooks} =
           to_computation.(expr || fun.(vars), inputs_and_shapes, used_hooks)
 
-        executable = EXLA.Computation.compile(computation, client, shapes)
```
polvalente, line 321: Is there a way to write a test in which this makes a difference?

josevalim: Yeah, we can write a test that sets the option to an invalid device_id and see if it raises accordingly. :)

jonatanklosko: Done!

josevalim (approved):

josevalim:

grzegorz-wcislo:
Thank you @jonatanklosko, I'll definitely use this :heart:

Issue #504, opened 2021-10-05
Add missing implementations in Torchx.Backend


polvalente:
There are some missing operations in Torchx.Backend. For some of the following there are suggestions on where to look in the Libtorch API. It is best to look for the function in the [C++ API](https://pytorch.org/cppdocs/api/library_root.html) and then to see how it works in the [Python docs](https://pytorch.org/docs/stable/index.html).

- [x] window_scatter_min
- [x] window_scatter_max
- [x] window_sum
- [x] window_product
- [x] window_min
- [x] window_max
- [x] triangular_solve (torch::triangular_solve)
- [x] svd (torch::linalg::svd)
- [x] sort (torch::sort)
- [x] solve (torch::solve)
- [x] select (torch::select)
- [x] reduce_min
- [x] reduce_max
- [x] reverse
- [x] put_slice
- [x] product (torch::prod)
- [x] population_count
- [x] pad
- [x] lu (torch::_lu_with_info + torch::lu_unpack)
- [x] indexed_add (torch::scatter_add)
- [x] count_leading_zeros
- [x] conv (torch::conv1d/2d and 3d)
- [x] clip (torch::clip)
- [x] cbrt
- [x] bitcast
- [x] backend_copy
- [x] any? (torch::any)

seanmor5:
@polvalente Just dropping in to point out that `Nx.conv` parity will be a bit more in depth than just `torch::convolution` since LibTorch provides a separate implementation for each type of conv (1d, 2d, 3d).

stefkohub:
If I start adding the convolutional layer, which are the functions I will need to integrate? I cannot find torch::convolution but torch::conv1d, 2d and 3d here:
[https://pytorch.org/cppdocs/api/unabridged_api.html?highlight=torch%20convolution](https://pytorch.org/cppdocs/api/unabridged_api.html?highlight=torch%20convolution)

Thank you.

polvalente:
> If I start adding the convolutional layer, which are the functions I will need to integrate? I cannot find torch::convolution but torch::conv1d, 2d and 3d here:
> [https://pytorch.org/cppdocs/api/unabridged_api.html?highlight=torch%20convolution](https://pytorch.org/cppdocs/api/unabridged_api.html?highlight=torch%20convolution)
> 
> Thank you.

Edited the issue accordingly!

PR #606, opened 2022-01-09
Feat(torchx): Add support for clip

goodhamgupta:
Hi everyone,

This PR aims to add support for the `clip` method in TorchX(#504).

### DIFF
--- a/torchx/c_src/torchx.cpp
+++ b/torchx/c_src/torchx.cpp
@@ -602,10 +602,10 @@ NIF(sort)
 NIF(clip)
 {
   TENSOR_PARAM(0, t);
-  SCALAR_PARAM(1, min);
-  SCALAR_PARAM(2, max);
+  TENSOR_PARAM(1, min);
+  TENSOR_PARAM(2, max);
 
-  TENSOR(torch::clip(*t, min, max));
+  TENSOR(torch::clip(*t, *min, *max));
 }
 
 /* Aggregates */
--- a/torchx/c_src/torchx.cpp
+++ b/torchx/c_src/torchx.cpp
@@ -599,6 +599,15 @@ NIF(sort)
   TENSOR(std::get<0>(result));
 }
 
+NIF(clip)
+{
+  TENSOR_PARAM(0, t);
+  PARAM(1, double, min);
+  PARAM(2, double, max);
+
+  TENSOR(torch::clip(*t, min, max));
+}
+
 /* Aggregates */
 
 NIF(sum)
@@ -898,6 +907,7 @@ static ErlNifFunc nif_functions[] = {
     DF(triangular_solve, 4),
     DF(determinant, 1),
     DF(sort, 3),
+    DF(clip, 3),
 
     F(cuda_is_available, 0),
     F(cuda_device_count, 0),
--- a/torchx/c_src/torchx.cpp
+++ b/torchx/c_src/torchx.cpp
@@ -602,8 +602,8 @@ NIF(sort)
 NIF(clip)
 {
   TENSOR_PARAM(0, t);
-  PARAM(1, double, min);
-  PARAM(2, double, max);
+  SCALAR_PARAM(1, min);
+  SCALAR_PARAM(2, max);
 
   TENSOR(torch::clip(*t, min, max));
 }
--- a/torchx/lib/torchx.ex
+++ b/torchx/lib/torchx.ex
@@ -256,7 +256,7 @@ defmodule Torchx do
   deftensor triangular_solve(tensor_a, tensor_b, transpose, upper)
   deftensor determinant(tensor)
   deftensor sort(tensor, axis, descending)
-  deftensor clip(tensor, min, max)
+  deftensor clip(tensor, tensor_min, tensor_max)
 
   ## Dirty non-tensor return values
 
--- a/torchx/lib/torchx.ex
+++ b/torchx/lib/torchx.ex
@@ -256,6 +256,7 @@ defmodule Torchx do
   deftensor triangular_solve(tensor_a, tensor_b, transpose, upper)
   deftensor determinant(tensor)
   deftensor sort(tensor, axis, descending)
+  deftensor clip(tensor, min, max)
 
   ## Dirty non-tensor return values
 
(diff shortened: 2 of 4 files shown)


torchx/c_src/torchx.cpp
```diff
@@ -599,6 +599,15 @@ NIF(sort)
   TENSOR(std::get<0>(result));
 }
 
+NIF(clip)
+{
+  TENSOR_PARAM(0, t);
+  SCALAR_PARAM(1, min);
+  SCALAR_PARAM(2, max);
+
+  TENSOR(torch::clip(*t, min, max));
```
goodhamgupta, line 608: This NIF doesn't work for mixed types for some reason. For example, the output of using integers is:

```ex
iex> Nx.clip(Nx.tensor([[1, 2, 3], [4, 5, 6]], names: [:x, :y]), Nx.tensor(2), Nx.tensor(4))
#Nx.Tensor<
  s64[x: 2][y: 3]
  Torchx.Backend(cpu)
  [
    [2, 2, 3],
    [4, 4, 4]
  ]
>
```
But, when we use floating type numbers(scalars or tensors)

```ex
iex> Nx.clip(Nx.tensor([[1, 2, 3], [4, 5, 6]], names: [:x, :y]), Nx.tensor(2.1), Nx.tensor(4))
#Nx.Tensor<
  f32[x: 2][y: 3]
  Torchx.Backend(cpu)
  [
    [2.802596928649634e-45, 0.0, 2.802596928649634e-45],
    [0.0, 4.203895392974451e-45, 0.0]
  ]
>
```

I noticed that the `min` and `max` variables are passed as tensors from the `backend.ex` file, but when I replace `SCALAR_PARAM` with `TENSOR_PARAM` I get the error:

```ex
iex(1)>   Nx.clip(Nx.tensor([[1, 2, 3], [4, 5, 6]], names: [:x, :y]), Nx.tensor(2), Nx.tensor(4))
** (RuntimeError) Torchx: Unable to get min tensor param in NIF.clip/3
    (torchx 0.1.0-dev) lib/torchx.ex:328: Torchx.unwrap!/1
    (torchx 0.1.0-dev) lib/torchx.ex:331: Torchx.unwrap_tensor!/2
    (torchx 0.1.0-dev) lib/torchx/backend.ex:763: Torchx.Backend.clip/4
```

This errors occurs both when I use `to_number` and without it in the `backend.ex` file. Could someone please help me fix this issue? 😄

polvalente: If you replace `SCALAR_PARAM` with `TENSOR_PARAM` in the `.cpp` file, you need to remove the `to_number` call in the  backend.ex file. Also, you need to rename the params in `deftensor clip(tensor, tensor_min, tensor_max)`, to ensure they contain tensors.

However, I suggest keeping the `to_number` calls as the code is now. You'll probably need to promote the args to float whenever one of them is a float already, which should get rid of the type mismatch issue

polvalente: Oh, I just read the "floating" result (on mobile it didn't show up). What's probably happening is that the `4` is being passed as an integer but is being bitcast as a float due to `2.1` coming first.

polvalente: Looking at the docs it seems that `torch::clip` does accept tensors for `min` and `max`, so perhaps you could try that? Just change the `deftensor` as I pointed out and change `to_number` to `from_nx` in Torchx.Backend

torchx/lib/torchx/backend.ex
```diff
@@ -756,6 +756,14 @@ defmodule Torchx.Backend do
     |> to_nx(out)
   end
 
+  @impl true
+  def clip(%T{} = out, %T{} = t, %T{} = min, %T{} = max) do
+    t
+    |> from_nx()
+    |> Torchx.clip(to_number(min), to_number(max))
+    |> to_nx(out)
```
polvalente, line 764:
```suggestion
    t
    |> from_nx()
    |> Torchx.clip(to_number(min), to_number(max))
    |> to_nx(out)
```
```suggestion
    t
    |> Nx.as_type(out.type)
    |> from_nx()
    |> Torchx.clip(from_nx(min), from_nx(max))
    |> to_nx(out)
```

I've applied my suggestions locally and they work fine, but this conversion is needed beforehand :)

torchx/test/torchx/backend_test.exs
```diff
@@ -588,4 +588,27 @@ defmodule Torchx.BackendTest do
...
+      )
+    end
+
+    test "currently fails for floats" do
+      assert_all_close(
+        Nx.clip(Nx.tensor([[1, 2, 3], [4, 5, 6]], names: [:x, :y]), 2.0, 4.0),
+        Nx.tensor([[2.0, 2.0, 3.0], [4.0, 4.0, 4.0]])
+      )
+    end
+  end
```
polvalente, line 613:
```suggestion
```

You don't need to add these because the Nx doctests are run with the torchx backend (you can check that if you comment the `clip` definition from Torchx.Backend, the doctest count decreases).
We normally add these "manual" tests only when there's some unfixable issue such as rounding errors

torchx/lib/torchx.ex
```diff
@@ -256,6 +256,7 @@ defmodule Torchx do
   deftensor triangular_solve(tensor_a, tensor_b, transpose, upper)
   deftensor determinant(tensor)
   deftensor sort(tensor, axis, descending)
+  deftensor clip(tensor, min, max)
```
polvalente, line 259:
```suggestion
  deftensor clip(tensor, tensor_min, tensor_max)
```

torchx/c_src/torchx.cpp
polvalente, line 608:
```suggestion
  TENSOR_PARAM(1, min);
  TENSOR_PARAM(2, max);

  TENSOR(torch::clip(*t, *min, *max));
```

goodhamgupta: Thanks for your review @polvalente! Indeed I did try to use tensors directly and use `TENSOR_PARAM`, and I remember seeing the following trace:

```ex
** (SyntaxError) iex:3:18: unexpected ( after alias Backend. Function names and identifiers in Elixir start with lowercase characters or underscore. For example:

    hello_world()
    _starting_with_underscore()
    numb3rs_are_allowed()
    may_finish_with_question_mark?()
    may_finish_with_exclamation_mark!()

Unexpected token: (
``` 

I think I missed the `from_nx` change for the min and max tensors. It does work as expected after applying the changes you've suggested! 😄

polvalente (approved):

polvalente:
Thanks for the contribution! Nice work :)

PR #611, opened 2022-01-25
Add Nx.Constants

josevalim:


exla/lib/exla/lib.ex
```diff
@@ -123,31 +123,31 @@ defmodule EXLA.Lib do
...
 
-  defp min_value_binary({:pred, 8}), do: <<0>>
-  defp min_value_binary(type), do: Nx.Type.min_value_binary(type)
+  defp min_finite_binary({:pred, 8}), do: <<0>>
+  defp min_finite_binary(type), do: Nx.Type.min_finite_binary(type)
 
-  defp max_value_binary({:pred, 8}), do: <<1>>
-  defp max_value_binary(type), do: Nx.Type.max_value_binary(type)
+  defp max_finite_binary({:pred, 8}), do: <<1>>
+  defp max_finite_binary(type), do: Nx.Type.max_finite_binary(type)
```
josevalim, line 150: @polvalente @seanmor5 I would like your thoughts on this. We have a `Nx.Type.max_finite_binary(type)` that returns the underlying binary, not a tensor. `Nx.Constants.max_finite(type)` is simply a wrapper around the `Nx.Type` function.

We could remove the `Nx.Type` function and instead do something like this here:

    Nx.Constants.max_finite(type, backend: Nx.BinaryBackend).data.binary

So the question is: do we want to keep the `Nx.Type.*_binary` functions for low-level access? Or should we remove them?

polvalente: I think that keeping things under `*_binary` allows for future `tensor.data` changes, where we don't necessarily need to keep the representation under `tensor.data.binary`. Other than that, removing those would mean a less bloated API.

Is the `Nx.Type` module public? If so, I would remove the `*_binary` functions and use the `.data.binary` syntax instead. Otherwise, I don't have a clear preference on that

PR #616, opened 2022-01-28
add utility functions to NX

tiagodavi:
This PR adds some utility functions to NX.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -9176,23 +9176,25 @@ defmodule Nx do
 
   ## Examples
 
-  iex> Nx.variance(Nx.tensor([[1, 2], [3, 4]]))
+      iex> Nx.variance(Nx.tensor([[1, 2], [3, 4]]))
+      #Nx.Tensor<
+        f32
+        1.25
+      >
 
-  #Nx.Tensor<f32 1.25>
-
-  iex> Nx.variance(Nx.tensor([[1, 2], [3, 4]]), 1)
-
-  #Nx.Tensor<f32 1.6666666269302368>
+      iex> Nx.variance(Nx.tensor([[1, 2], [3, 4]]), ddof: 1)
+      #Nx.Tensor<
+        f32
+        1.6666666269302368
+      >
   """
   @doc type: :aggregation
-  @spec variance(tensor :: Nx.Tensor.t(), ddof :: number()) :: Nx.Tensor.t()
-  def variance(tensor, ddof \\ 0)
-
-  def variance(tensor, ddof) do
+  @spec variance(tensor :: Nx.Tensor.t(), opts :: Keyword.t()) :: Nx.Tensor.t()
+  def variance(tensor, opts \\ []) do
     %T{shape: shape} = tensor = to_tensor(tensor)
 
     total = Tuple.product(shape)
-
+    ddof = Keyword.get(opts, :ddof, 0)
     mean = mean(tensor)
 
     tensor
@@ -9209,49 +9211,24 @@ defmodule Nx do
 
   ## Examples
 
-  iex> Nx.standard_deviation(Nx.tensor([[1, 2], [3, 4]]))
+      iex> Nx.standard_deviation(Nx.tensor([[1, 2], [3, 4]]))
+      #Nx.Tensor<
+        f32
+        1.1180340051651
+      >
 
-  #Nx.Tensor<f32 1.1180340051651>
+      iex> Nx.standard_deviation(Nx.tensor([[1, 2], [3, 4]]), ddof: 1)
+      #Nx.Tensor<
+        f32
+        1.29099440574646
+      >
   """
   @doc type: :aggregation
-  @spec standard_deviation(tensor :: Nx.Tensor.t(), ddof :: number()) :: Nx.Tensor.t()
-  def standard_deviation(tensor, ddof \\ 0)
+  @spec standard_deviation(tensor :: Nx.Tensor.t(), ddof :: Keyword.t()) :: Nx.Tensor.t()
+  def standard_deviation(tensor, opts \\ []) do
+    tensor = to_tensor(tensor)
 
-  def standard_deviation(tensor, ddof) do
-    %T{} = tensor = to_tensor(tensor)
-
-    sqrt(variance(tensor, ddof))
-  end
-
-  @doc """
-  Normalizes the tensor by using standard scale.
-
-  if ddof is specified, the divisor N - ddof is used to calculate the variance.
-
-  ## Examples
-
-  iex> Nx.standard_scale(Nx.tensor([[1, 2], [3, 4]]))
-
-  #Nx.Tensor<
-  f32[2][2]
-  [
-    [-1.3416407108306885, -0.4472135901451111],
-    [0.4472135901451111, 1.3416407108306885]
-  ]
-  >
-  """
-  @doc type: :aggregation
-  @spec standard_scale(tensor :: Nx.Tensor.t(), ddof :: number()) :: Nx.Tensor.t()
-  def standard_scale(tensor, ddof \\ 0)
-
-  def standard_scale(tensor, ddof) do
-    %T{} = tensor = to_tensor(tensor)
-
-    mean = mean(tensor)
-
-    tensor
-    |> Nx.subtract(mean)
-    |> Nx.divide(standard_deviation(tensor, ddof))
+    sqrt(variance(tensor, opts))
   end
 
   ## Sigils
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -9171,8 +9171,9 @@ defmodule Nx do
   Finds the variance of a tensor.
 
   The variance is the average of the squared deviations from the mean.
-  The mean is typically calculated as sum(x) / N, where N = total of elements.
-  If, however, ddof is specified, the divisor N - ddof is used instead.
+  The mean is typically calculated as `sum(tensor) / n`, where `n` is the total
+  of elements. If, however, `:ddof` (delta degrees of freedom) is specified, the
+  divisor `n - ddof` is used instead.
 
   ## Examples
 
@@ -9208,7 +9209,9 @@ defmodule Nx do
   @doc """
   Finds the standard deviation of a tensor.
 
-  if ddof is specified, the divisor N - ddof is used to calculate the variance.
+  The standard deviation is taken as the square root of the variance.
+  If the `:ddof` (delta degrees of freedom) option is given, the divisor
+  `n - ddof` is used to calculate the variance. See `variance/2`.
 
   ## Examples
 
@@ -9225,7 +9228,7 @@ defmodule Nx do
       >
   """
   @doc type: :aggregation
-  @spec standard_deviation(tensor :: Nx.Tensor.t(), ddof :: Keyword.t()) :: Nx.Tensor.t()
+  @spec standard_deviation(tensor :: Nx.Tensor.t(), opts :: Keyword.t()) :: Nx.Tensor.t()
   def standard_deviation(tensor, opts \\ []) do
     sqrt(variance(tensor, opts))
   end
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -9193,8 +9193,9 @@ defmodule Nx do
   def variance(tensor, opts \\ []) do
     %T{shape: shape} = tensor = to_tensor(tensor)
 
+    opts = keyword!(opts, ddof: 0)
     total = Tuple.product(shape)
-    ddof = Keyword.get(opts, :ddof, 0)
+    ddof = Keyword.fetch!(opts, :ddof)
     mean = mean(tensor)
 
     tensor
@@ -9226,8 +9227,6 @@ defmodule Nx do
   @doc type: :aggregation
   @spec standard_deviation(tensor :: Nx.Tensor.t(), ddof :: Keyword.t()) :: Nx.Tensor.t()
   def standard_deviation(tensor, opts \\ []) do
-    tensor = to_tensor(tensor)
-
     sqrt(variance(tensor, opts))
   end
 
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -9167,6 +9167,84 @@ defmodule Nx do
     end
   end
 
+  @doc """
+  Finds the variance of a tensor.
+
+  ## Examples
+
+  iex> Nx.var(Nx.tensor([[1, 2], [3, 4]]))
+
+  #Nx.Tensor<
+  f32
+  1.25
+  >
+  """
+  @doc type: :element
+  @spec var(tensor :: Nx.Tensor.t()) :: Nx.Tensor.t()
+  def var(%Nx.Tensor{shape: shape} = tensor) do
+    total =
+      shape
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you @tiagodavi! @polvalente / @seanmor5, do you think this belongs in Nx? Or some statistics oriented package?

polvalente:
> Thank you @tiagodavi! @polvalente / @seanmor5, do you think this belongs in Nx? Or some statistics oriented package?

@josevalim I think std/var/avg are common and basic enough to belong in Nx. For instance, random variables are often normalized through the standard deviation, and those are Nx vectors.

However, perhaps we could think about moving them to another module.

Speaking of avg (or mean), do we have it yet? Could be a good companion function here.

josevalim:
We do have `mean`. @tiagodavi, can you please rename those functions to they have the full name instead of abbreviations? We typically avoid abbreviations in Nx.

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,84 @@ defmodule Nx do
...
+  Finds the variance of a tensor.
+
+  ## Examples
+
+  iex> Nx.var(Nx.tensor([[1, 2], [3, 4]]))
+
+  #Nx.Tensor<
+  f32
+  1.25
+  >
```
josevalim, line 9180:
```suggestion
      iex> Nx.var(Nx.tensor([[1, 2], [3, 4]]))
      #Nx.Tensor<
        f32
        1.25
      >
```

Please fix the other ones accordingly. :)

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,84 @@ defmodule Nx do
...
+  """
+  @doc type: :element
```
josevalim, line 9182: I think the type is aggregation or similar. Pls check what `Nx.mean` does :)

nx/lib/nx.ex
tiagodavi: sure.

nx/lib/nx.ex
tiagodavi: I'll take a look. thank you

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,84 @@ defmodule Nx do
...
+      |> Tuple.to_list()
+      |> Enum.reduce(fn val, acc -> acc * val end)
+
+    mean = mean(tensor)
+
+    tensor
+    |> subtract(mean)
+    |> power(2)
+    |> sum()
+    |> divide(total)
```
jonatanklosko, line 9196: We should document the exact formula, in particular that we divide by `N`. In statistical use cases it's often desired to divide by `N - 1` when calculating sample variance/stdev, since that gives an unbiased estimate.

For the record [numpy.var](https://numpy.org/doc/stable/reference/generated/numpy.var.html) has `ddof` option and divides by `N - ddof`, [torch.var](https://pytorch.org/docs/stable/generated/torch.var.html) has a more explicit `unbiased` option, while [tfp.stats.variance](https://www.tensorflow.org/probability/api_docs/python/tfp/stats/variance) always uses `N`.

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,84 @@ defmodule Nx do
...
+  >
+  """
+  @doc type: :element
+  @spec normalize(tensor :: Nx.Tensor.t()) :: Nx.Tensor.t()
+  def normalize(%Nx.Tensor{} = tensor) do
+    mean = mean(tensor)
+
+    tensor
+    |> Nx.subtract(mean)
+    |> Nx.divide(std(tensor))
```
floatn, line 9243: What about zero `std` value here?

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,84 @@ defmodule Nx do
...
+  1.25
+  >
+  """
+  @doc type: :element
+  @spec var(tensor :: Nx.Tensor.t()) :: Nx.Tensor.t()
+  def var(%Nx.Tensor{shape: shape} = tensor) do
+    total =
+      shape
+      |> Tuple.to_list()
+      |> Enum.reduce(fn val, acc -> acc * val end)
```
polvalente, line 9188:
```suggestion
    total = Tuple.product(shape)
```

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,84 @@ defmodule Nx do
...
+  iex> Nx.var(Nx.tensor([[1, 2], [3, 4]]))
+
+  #Nx.Tensor<
+  f32
+  1.25
+  >
+  """
+  @doc type: :element
+  @spec var(tensor :: Nx.Tensor.t()) :: Nx.Tensor.t()
+  def var(%Nx.Tensor{shape: shape} = tensor) do
```
polvalente, line 9184: What is the definition for the variance of a 0-dimensional tensor? Perhaps we should have a check against that

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,84 @@ defmodule Nx do
...
+    mean = mean(tensor)
+
+    tensor
+    |> subtract(mean)
+    |> power(2)
+    |> sum()
+    |> divide(total)
+  end
+
+  def var(_), do: raise(ArgumentError, "must be a tensor")
```
polvalente, line 9199: I think we should use `to_tensor` to make errors more uniform. Take a look at the implementation for `reshape` for instance, which uses it.

polvalente: Same goes for `std`

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,84 @@ defmodule Nx do
...
+  f32[2][2]
+  [
+    [-1.3416407108306885, -0.4472135901451111],
+    [0.4472135901451111, 1.3416407108306885]
+  ]
+  >
+  """
+  @doc type: :element
+  @spec normalize(tensor :: Nx.Tensor.t()) :: Nx.Tensor.t()
+  def normalize(%Nx.Tensor{} = tensor) do
```
polvalente, line 9238: I think we should either move this to another module (such as `Nx.Statistics`) or rename this function, because without a specific context, one could expect normalize to be `def normalize(t), do: Nx.divide(t, Nx.LinAlg.norm(t))`

polvalente: For all of these functions, we could also have a "per axis" version, like the ones we have for `sum`.
Especially in the case where we're dealing with a set of random variables, this could be really useful.

WDYT of having a follow-up PR for this? If you can't tackle this (lack of time, etc), let us know and we can add it to the issues tracker.

nx/lib/nx.ex
tiagodavi: done. thank you

nx/lib/nx.ex
tiagodavi:
```
iex(3)> Nx.variance Nx.tensor(5)
#Nx.Tensor<
  f32
  0.0
>
```

nx/lib/nx.ex
tiagodavi: done thank you.

nx/lib/nx.ex
tiagodavi: added. good call.

nx/lib/nx.ex
tiagodavi: I can definitely add a follow-up PR next week to address this.

nx/lib/nx.ex
tiagodavi: what do you mean by zero ?

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,93 @@ defmodule Nx do
...
+  @doc """
+  Finds the variance of a tensor.
+
+  The variance is the average of the squared deviations from the mean.
+  The mean is typically calculated as sum(x) / N, where N = total of elements.
+  If, however, ddof is specified, the divisor N - ddof is used instead.
+
+  ## Examples
+
+  iex> Nx.variance(Nx.tensor([[1, 2], [3, 4]]))
```
josevalim, line 9179: Please indent all code examples by four spaces and remove the new line between the expression and the result. :)

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,93 @@ defmodule Nx do
...
+
+  ## Examples
+
+  iex> Nx.standard_deviation(Nx.tensor([[1, 2], [3, 4]]))
+
+  #Nx.Tensor<f32 1.1180340051651>
+  """
+  @doc type: :aggregation
+  @spec standard_deviation(tensor :: Nx.Tensor.t(), ddof :: number()) :: Nx.Tensor.t()
+  def standard_deviation(tensor, ddof \\ 0)
```
josevalim, line 9218: Since we may want to add an `axes` option to those functions, perhaps `ddof` must be an option too? See how `mean` is done, we want to accept the same options. :)

nx/lib/nx.ex
josevalim: I believe the standard_deviation can return 0 (for example if all values are equal). In this case, this function will error. Are there other formulas to implement `standard_scale`?

seanmor5:
I am in favor of standard deviation and variance, but I'm not sure standard scale belongs in Nx. It feels more like the responsibility of another library related to data preprocessing

nx/lib/nx.ex
msluszniak: Look that std is 0 only and only if all the values are equal. In this case, there is no possibility to achieve unit variance, so the data should be left as-is. And actually, I've searched how it is calculated in Sci-kit learn. They based on this article http://www.cs.yale.edu/publications/techreports/tr222.pdf.

msluszniak: But this problematic case is still handled particularly. What is more, they catch the cases when the std is non-zero, but values are almost constant for stability issues. https://github.com/scikit-learn/scikit-learn/blob/7e1e6d09bcc2eaeba98f7e737aac2ac782f0e5f1/sklearn/preprocessing/_data.py#L84

josevalim: Yeah, let's give up on the standard_scale/normalize for now then and revisit later.

tiagodavi:
I am trying to do something like this to respect the guidelines, but not sure how to access the tensor over its axes to compute the variance at the end:

```
elixir
def variance(tensor, opts \\ []) do
    %T{shape: shape, names: names} = tensor = to_tensor(tensor)

    mean_den =
      if axes = opts[:axes] do
        mean_den(shape, Nx.Shape.normalize_axes(shape, axes, names))
      else
        mean_den(shape, nil)
      end

    ddof = Keyword.get(opts, :ddof, 0)
    mean = mean(tensor, Keyword.take(opts, [:axes, keep_axes: false]))

    axes = axes(tensor)

    IO.inspect axes

    tensor[axes] # I am assuming I need to calculate based on its axes because mean and mean_den are based on it.
    |> subtract(mean)
    |> power(2)
    |> sum()
    |> divide(mean_den - ddof)
  end
```

josevalim:
Oh, I see. Let’s not worry about the axes version for now then!

tiagodavi:
@josevalim done. Fixed spaces, removed standard_scale function and added the ddof option to variance, standard_deviation without axes for the time being.

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,70 @@ defmodule Nx do
...
+      iex> Nx.variance(Nx.tensor([[1, 2], [3, 4]]), ddof: 1)
+      #Nx.Tensor<
+        f32
+        1.6666666269302368
+      >
+  """
+  @doc type: :aggregation
+  @spec variance(tensor :: Nx.Tensor.t(), opts :: Keyword.t()) :: Nx.Tensor.t()
+  def variance(tensor, opts \\ []) do
+    %T{shape: shape} = tensor = to_tensor(tensor)
```
polvalente, line 9194:
```suggestion
    %T{shape: shape} = tensor = to_tensor(tensor)
    
    opts = keyword!(opts, ddof: 0)
```

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,70 @@ defmodule Nx do
...
+
+    total = Tuple.product(shape)
+    ddof = Keyword.get(opts, :ddof, 0)
```
polvalente, line 9197:
```suggestion
    ddof = Keyword.fetch!(opts, :ddof)
```

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,70 @@ defmodule Nx do
...
+      #Nx.Tensor<
+        f32
+        1.29099440574646
+      >
+  """
+  @doc type: :aggregation
+  @spec standard_deviation(tensor :: Nx.Tensor.t(), ddof :: Keyword.t()) :: Nx.Tensor.t()
+  def standard_deviation(tensor, opts \\ []) do
+    tensor = to_tensor(tensor)
+
```
polvalente, line 9230:
```suggestion
```

In this specific case we can just delegate to variance

nx/lib/nx.ex
tiagodavi: done

nx/lib/nx.ex
tiagodavi: done

nx/lib/nx.ex
tiagodavi: done

polvalente (approved):

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,69 @@ defmodule Nx do
     end
   end
 
+  @doc """
+  Finds the variance of a tensor.
+
+  The variance is the average of the squared deviations from the mean.
+  The mean is typically calculated as sum(x) / N, where N = total of elements.
+  If, however, ddof is specified, the divisor N - ddof is used instead.
```
josevalim, line 9175:
```suggestion
  The mean is typically calculated as `sum(tensor) / n`, where `n` is the total
  of elements. If, however, `:ddof` (delta degrees of freedom) is specified, the
  divisor `n - ddof` is used instead.
```

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,69 @@ defmodule Nx do
...
+    |> subtract(mean)
+    |> power(2)
+    |> sum()
+    |> divide(total - ddof)
+  end
+
+  @doc """
+  Finds the standard deviation of a tensor.
+
+  if ddof is specified, the divisor N - ddof is used to calculate the variance.
```
josevalim, line 9211:
```suggestion
  The standard deviation is taken as the square root of the variance.
  If the `:ddof` (delta degrees of freedom) option is given, the divisor
  `n - ddof` is used to calculate the variance. See `variance/2`.
```

nx/lib/nx.ex
```diff
@@ -9167,6 +9167,69 @@ defmodule Nx do
...
+      >
+
+      iex> Nx.standard_deviation(Nx.tensor([[1, 2], [3, 4]]), ddof: 1)
+      #Nx.Tensor<
+        f32
+        1.29099440574646
+      >
+  """
+  @doc type: :aggregation
+  @spec standard_deviation(tensor :: Nx.Tensor.t(), ddof :: Keyword.t()) :: Nx.Tensor.t()
```
josevalim, line 9228:
```suggestion
  @spec standard_deviation(tensor :: Nx.Tensor.t(), opts :: Keyword.t()) :: Nx.Tensor.t()
```

josevalim:

PR #617, opened 2022-02-03
Pass 'mix test' on Windows platform

howard0su:
Mainly change the test case to handle \r\n on Windows v.s. \n on Linux.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -9347,8 +9347,7 @@ defmodule Nx do
   end
 
   defp binary_to_numbers(string) do
-    string = String.replace(string, "\r\n", "\n")
-    for row <- String.split(string, "\n", trim: true) do
+    for row <- String.split(string, ["\n", "\r\n"], trim: true) do
       row
       |> String.split(" ", trim: true)
       |> Enum.map(fn str ->
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -9347,6 +9347,7 @@ defmodule Nx do
   end
 
   defp binary_to_numbers(string) do
+    string = String.replace(string, "\r\n", "\n")
     for row <- String.split(string, "\n", trim: true) do
       row
       |> String.split(" ", trim: true)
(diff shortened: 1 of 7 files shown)


nx/lib/nx.ex
```diff
@@ -9347,6 +9347,7 @@ defmodule Nx do
   end
 
   defp binary_to_numbers(string) do
+    string = String.replace(string, "\r\n", "\n")
     for row <- String.split(string, "\n", trim: true) do
```
polvalente, line 9351:
```suggestion
    for row <- String.split(string, ["\r\n", "\n"], trim: true) do
```

nx/test/nx/defn/evaluator_test.exs
```diff
@@ -4,6 +4,8 @@ defmodule Nx.Defn.EvaluatorTest do
   alias Nx.Tensor, as: T
   import Nx.Defn
 
+  defp s(str), do: String.replace(str, "\r\n", "\n")
```
polvalente, line 7: I don't think this fix is easy to maintain. If we're going to support testing on Windows we'll also need a GitHub Actions workflow which runs them. Otherwise, each new test is probably not having this function being called anyway because it doesn't breaking on Mac or Linux.

@josevalim is there another fix around this? Perhaps an ExUnit.CaptureIO config or something?

josevalim: I think we need it for `capture_io` but it should not be necessary for inspect.

nx/test/nx/defn/expr_test.exs
```diff
@@ -198,7 +200,7 @@ defmodule Nx.Defn.ExprTest do
       b = Expr.parameter(nil, {:s, 64}, {2, 2}, 1)
 
       assert Nx.sum(Nx.add(Nx.add(Nx.dot(a, a), Nx.tanh(b)), 2))
-             |> inspect(safe: false) == """
+             |> inspect(safe: false) == s """
```
josevalim, line 203: Hi @howard0su! This change should not be necessary. `inspect` should not add `\r\n`. Can you please double check if changing the `inspect` comparison are necessary? Instead, \r\n will be added only when writing to the terminal on Windows.

josevalim (approved):

josevalim:

PR #619, opened 2022-02-04
feat: optional callbacks

polvalente:


nx/lib/nx/shared.ex
```diff
@@ -387,4 +387,35 @@ defmodule Nx.Shared do
             "#{inspect(struct1)} and #{inspect(struct2)}. You may need to call Nx.backend_transfer/1 " <>
             "(or Nx.backend_copy/1) on one or both of them to transfer them to a common implementation"
   end
+
+  @doc """
+  Used to define an Nx callback with an optional implementation.
+
+  The given body is used as the default implementation otherwise.
+  """
+  defmacro defoptional(function_header, body) do
```
polvalente, line 396: @josevalim I ended up trying to implement a macro for this, but I think that perhaps this isn't the way.

For one, I couldn't figure out how to call the `_implementation` defp from within the macro.
Secondly, I'm a bit confused on the expected arity for the optional callback. Normally when we define callbacks there's also the expected output holder.

Should we force the functions to define their expected outputs so it can be passed as  `[expected_output_holder | args]` to the `apply/3` call below?

Other than this, the code is currently working as expected, using the Torchx pseudo implementation when the given tensors are from Torchx.Backend and calling the default one for Nx.BinaryBackend

polvalente: I'm thinking about reverting to the original suggestion like below, but with the addition of the output holder:

```elixir
def optional(function_name, args, default_impl) do
  arity = length(args)
  backend = impl!(args)
  cond do
    backend == Nx.Defn.Expr -> raise "not implemented"
    function_exported?(backend, name, arity + 1) -> 
      apply(backend, name, args)
    :otherwise -> 
      apply(default_impl, args)
  end
end
```

Which can be used like so:

```elixir
def solve(a, b) do
  # keep the input validations here
  output_holder = b
  optional(:solve, [output_holder, a, b], fn out, a, b -> 
    # main linalg code here
  end)
end
```

josevalim: Yeah, I wouldn't start with the macro, it feels backward. Let's first start with the building blocks and then add macro functionality on top if required.

josevalim: I think the output holder be part of the args, yeah, because that's what we would call in the backend.

josevalim: Actually, I am not sure about the output. We have three options:

1. say the output is not included for optional callbacks
2. automatically compute the output by applying the anonymous functions with a tensor expression
3. require the output explicitly (which most likely requires doing 2 anyway)

nx/lib/nx/shared.ex
```diff
@@ -387,4 +393,27 @@ defmodule Nx.Shared do
...
+  """
+  def default_implementation(function_name, args, default_impl)
+      when is_atom(function_name) and is_list(args) and is_function(default_impl) do
+    arity = length(args)
+
+    backend = impl!(args)
+
+    cond do
+      backend == Nx.Defn.Expr ->
+        apply(default_impl, args)
```
polvalente, line 410: @josevalim I'm a little bit confused as to why, but passing the Expr tensors to the default implementation got the broken tests to pass (some of which were grad tests as well)

nx/test/nx/backend_test.exs
```diff
@@ -0,0 +1,103 @@
...
+    def from_binary(tensor, data, _opts) do
+      %{tensor | data: %__MODULE__{state: data}}
+    end
+
+    def backend_transfer(%{data: %{state: data}} = tensor, backend, _opts) do
+      %{tensor | data: struct(backend, state: data)}
+    end
+
+    def solve(_out, a, b) do
+      Logger.info("called custom implementation")
```
josevalim, line 19: Instead of use `Logger`, send a message to self and use `assert_received`.

polvalente: Duh. I tried registering the process and sending by name. Didn't occur to me that this is all within the same process!

nx/test/nx/backend_test.exs
```diff
@@ -0,0 +1,103 @@
+defmodule Nx.BackendTest do
```
josevalim, line 1: Suggestion to rename this to `nx/optional_test.exs`. Especially because we will need to add more tests once we add the `defn` implementation?

nx/lib/nx/shared.ex
josevalim: I don't know what to say except that this implementation looks correct to me. :D However, it is only part 1, because if you use `defn`, we won't currently dispatch to the optional implementation. :) But I think this is mergeable as is.

josevalim (commented):
LGTM, I have dropped some initial comments. :)

josevalim (approved):

PR #625, opened 2022-02-06
feat: add defn optional callbacks

polvalente:


nx/lib/nx/defn/expr.ex
```diff
@@ -816,6 +816,12 @@ defmodule Nx.Defn.Expr do
     end
   end
 
+  def optional(expr, name, [out, %{data: %{context: context}} | _] = args) do
+    out = %T{names: [], shape: {}, type: out.type}
```
josevalim, line 820: The names and shapes and types here are wrong. It should be the same as the output.

nx/lib/nx/shared.ex
```diff
@@ -407,7 +407,7 @@ defmodule Nx.Shared do
 
     cond do
       backend == Nx.Defn.Expr ->
-        apply(default_impl, args)
+        Nx.Defn.Expr.optional(apply(default_impl, args), function_name, args)
```
josevalim, line 511: I would like to check, either here or inside optional, that the result of `apply(default_impl, args)` matches the output shape in `hd(args)`. We should likely check it in the third clause too.

nx/lib/nx/defn/expr.ex
josevalim: I think this function should be something like this:

```elixir
# args is already without out
def optional(%{data: %{context: context}} = expr, name, args) do
  call = expr(expr, context, name, args)
  expr(expr, context, :default_implementation, [call, expr])
end
```

Then the grad will happen on the second argument. :)

You also want to change the inspect implementation to only consider the first argument. You don't want to show the `default_implementation` nor the second argument.

polvalente: I see! This makes sense

nx/lib/nx/defn/evaluator.ex
```diff
@@ -54,6 +54,25 @@ defmodule Nx.Defn.Evaluator do
...
+         state,
+         cache
+       ) do
+    {backend, _backend_options} = Nx.default_backend()
+
+    if function_exported?(backend, expr.data.op, length(expr.data.args)) do
+      eval_apply(expr.data.op, expr, state, cache)
+    else
+      eval_apply(default_impl_expr.data.op, default_impl_expr, state, cache)
+    end
```
josevalim, line 73:
```suggestion
    %{data: %{op: op, args: args}} = expr

    if function_exported?(Nx.Shared.list_impl!(args), op, length(args)) do
      eval(expr, state, cache)
    else
      eval(default_impl_expr, state, cache)
    end
```

josevalim: Btw, you may want to move this whole clause to eval_apply. The difference is that the results of `eval_apply` are cached and we likely want to cache this so we don't lookup the backend several times? 🤔

nx/lib/nx/defn/grad.ex
```diff
@@ -136,6 +136,10 @@ defmodule Nx.Defn.Grad do
   defp reduce_args(:while, %{data: %{args: [initial | _]}}, acc, fun),
     do: Composite.reduce(initial, acc, fun)
 
+  defp reduce_args(:optional, %{data: %{args: [_expr, default_impl_expr]}}, acc, fun) do
+    fun.(default_impl_expr, acc)
+  end
```
josevalim, line 141:
```suggestion
  defp reduce_args(:optional, %{data: %{args: [_expr, default_impl_expr]}}, acc, fun),
    do: fun.(default_impl_expr, acc)
```

torchx/test/torchx/defn_test.exs
```diff
@@ -15,9 +15,9 @@ defmodule Torchx.DefnTest do
   end
 
   describe "scalar" do
-    defn float_scalar(x), do: Nx.add(1.0, x)
-    defn integer_scalar(x), do: Nx.add(1, x)
-    defn broadcast_scalar(x), do: Nx.add(Nx.broadcast(1, {2, 2}), x)
+    defn(float_scalar(x), do: Nx.add(1.0, x))
+    defn(integer_scalar(x), do: Nx.add(1, x))
+    defn(broadcast_scalar(x), do: Nx.add(Nx.broadcast(1, {2, 2}), x))
```
josevalim, line 20: Please revert this. :)

nx/test/nx/optional_test.exs
```diff
@@ -72,23 +86,86 @@ defmodule Nx.OptionalTest do
...
+             \s\s
+               Nx.Defn.Expr
+               parameter a:0                            s64[3][3]
+               b = determinant a                        f32
+               c = sum b, axes: nil, keep_axes: false   f32
+             >
+             """
+    end
+
+    test "works with direct call" do
```
polvalente, line 154: @josevalim I'm not sure if we need to support this use case, but this call yields the unrolled default implementation instead of the expected expression. I'm still investigating why in the meantime

nx/test/nx/optional_test.exs
```diff
@@ -72,23 +106,93 @@ defmodule Nx.OptionalTest do
...
+      a = Nx.tensor([[1, 1], [2, 1]], backend: NonCustomImplTestBackend)
+
+      Nx.LinAlg.determinant(a)
+
+      assert_receive {:called_default_impl, :reshape}
+    end
+  end
+
+  describe "inspect" do
+    for backend <- [NonCustomImplTestBackend, CustomImplTestBackend] do
```
polvalente, line 159: The test for when the backend has the custom implementation works as expected, but it seems that the `defn` is yielding a Nx.Defn.Expr tensor instead when called from the default implementation.

This is a bug in the evaluator, right?

nx/lib/nx/shared.ex
```diff
@@ -501,14 +501,14 @@ defmodule Nx.Shared do
...
-  def default_implementation(function_name, args, default_impl)
+  def optional(function_name, args, default_impl)
       when is_atom(function_name) and is_list(args) and is_function(default_impl) do
     arity = length(args)
     backend = list_impl!(args)
 
     cond do
       backend == Nx.Defn.Expr ->
-        apply(default_impl, args)
+        Nx.Defn.Expr.optional(apply(default_impl, args), function_name, args)
```
josevalim, line 511: I would like to check, either here or inside optional, that the result of `apply(default_impl, args)` matches the output shape in `hd(args)`. We should likely check it in the third clause too.

nx/test/nx/optional_test.exs
```diff
@@ -1,7 +1,7 @@
 defmodule Nx.OptionalTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case, async: false
```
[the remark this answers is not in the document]
polvalente: It's due to the capture_io calls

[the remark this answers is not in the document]
josevalim: `capture_io` can be async: true in recent Elixir versions as long as you are doing matches with `=~` instead of `==`, as it may include the capture of other tests. I would say this is OK to do here, so I would go with `async: true` and `=~`.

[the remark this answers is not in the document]
polvalente: I know, but can't the output of another process get interlaced with the current one?

nx/test/nx/optional_test.exs
```diff
@@ -72,23 +161,88 @@ defmodule Nx.OptionalTest do
...
+               \s\s
+                 Nx.Defn.Expr
+                 parameter a:0       s64[3][3]
+                 b = determinant a   f32
+               >
+               """
+      end
+    end
+
+    test "works with direct call" do
```
josevalim, line 231: I would remove this test because it is not any different from the test above.

josevalim (approved):

PR #639, opened 2022-02-09
Use linalg API instead of the deprecated ones

howard0su:


polvalente (approved):

josevalim:
Should we update the supported versions in the README then? 🤔

howard0su:
> Should we update the supported versions in the README then? 🤔

The following API is changed from 1.9 to 1.10.
TORCH_API at::Tensor linalg_cholesky(const at::Tensor & self);

Let's only change qr for now, which is complaining deprecation.

josevalim:

PR #640, opened 2022-02-11
create new access

tiagodavi:


nx/test/nx/tensor_test.exs
```diff
@@ -127,4 +127,69 @@ defmodule Nx.TensorTest do
       assert_raise RuntimeError, "not supported", fn -> Nx.random_uniform({2, 2}) end
     end
   end
+
+  describe "access with :all" do
+    test "should access the entire tensor" do
+      t = Nx.tensor([[1, 2, 3], [4, 5, 6], [7, 8, 9]])
+
+      assert t[[:all]] == t
```
polvalente, line 135: @josevalim I'm not sure why this is behaving like this. Shouldn't this only bring the first row?

nx/test/nx/tensor_test.exs
```diff
@@ -127,4 +127,69 @@ defmodule Nx.TensorTest do
...
+      assert t[[0..1, :all]] == Nx.tensor([[1, 2, 3], [4, 5, 6]])
+    end
+
+    test "should take the last two rows" do
+      t = Nx.tensor([[1, 2, 3], [4, 5, 6], [7, 8, 9]])
+
+      assert t[[1..2, :all]] == Nx.tensor([[4, 5, 6], [7, 8, 9]])
+    end
+
+    test "should raise when out of bounds" do
```
polvalente, line 187: I think we should have a higher dimensional test as well, like a 3D tensor having :all along each o the axes

nx/test/nx/tensor_test.exs
jonatanklosko: Isn't this conceptually the same as `t[:all]`, therefore taking all rows?

polvalente: Oh right. So all implict dimensions are also ":all", right?

jonatanklosko: Yup!

josevalim:
Thank you @tiagodavi! Btw let's hold on a bit because we are likely going to add `..` to Elixir. So we can do:

    t[[.., 2]]

josevalim:
Pull request in Elixir is here: https://github.com/elixir-lang/elixir/pull/11623

josevalim:
Added to Elixir, thank you @tiagodavi for starting the discussion!

PR #644, opened 2022-02-16
Add logical_not as an optional callback in backend

howard0su:


nx/lib/nx.ex
```diff
@@ -4246,10 +4246,20 @@ defmodule Nx do
...
-    out = %T{shape: {}, type: type, names: []}
-    zero = Nx.BinaryBackend.from_binary(out, number_to_binary(0, type), [])
-    element_wise_pred_op(tensor, zero, :equal)
+    output = Nx.template(tensor.shape, {:u, 8}, names: tensor.names)
+
+    Nx.Shared.optional(:logical_not, [tensor], output, fn tensor ->
+      type = tensor.type
+
+      zero =
+        Nx.BinaryBackend.from_binary(
```
polvalente, line 4255: I don't think this should be using the binary backend. It should be a backend-agnostic implementation.

Perhaps some form of Nx.broadcast can help with this

nx/lib/nx.ex
```diff
@@ -4246,10 +4246,20 @@ defmodule Nx do
...
+          %T{shape: {}, type: type, names: []},
+          number_to_binary(0, type),
+          []
+        )
+
+      element_wise_pred_op(tensor, zero, :equal)
```
polvalente, line 4261: By the way, can't you just pass 0 instead of a scalar tensor?

nx/lib/nx.ex
josevalim: It is fine to use the BinaryBackend here, we are basically creating a constant. I understand it is confusing though!

nx/lib/nx.ex
josevalim: 0 will lead to a different type coercion logic, that's why we do the dance above!

josevalim (approved):

josevalim:

PR #646, opened 2022-02-17
Update XLA to TF 2.8.0

jonatanklosko:
See https://github.com/elixir-nx/xla/pull/15.

We still need to fix XLA compilation for cuda and publish to Hex, but I want to make sure CI passes (since the Linux/CPU binaries compiled successfully).


exla/c_src/exla/exla_client.cc
```diff
@@ -177,7 +177,8 @@ xla::StatusOr<ExlaBuffer*> ExlaClient::BufferFromBinary(const ErlNifBinary& bina
   xla::PjRtClient::HostBufferSemantics semantics = xla::PjRtClient::HostBufferSemantics::kImmutableUntilTransferCompletes;
 
   EXLA_ASSIGN_OR_RETURN(xla::PjRtDevice* device, client_->LookupDevice(device_id));
-  EXLA_ASSIGN_OR_RETURN(auto buffer, client_->BufferFromHostBuffer(binary.data, shape, semantics, nullptr, device));
+  EXLA_ASSIGN_OR_RETURN(auto buffer, client_->BufferFromHostBuffer(
+    binary.data, shape.element_type(), shape.dimensions(), absl::nullopt, semantics, nullptr, device));
```
jonatanklosko, line 181: The corresponding XLA change: https://github.com/tensorflow/tensorflow/commit/59077ac39cc64139f889863081bdb3d0f88d39b6

The `absl::nullopt` is for strides, which I don't think we need. The referenced commit message explains that NumPy keeps strides information on tensors, which allows transposition without any changes to the buffer. They now allow passing it to this function, so the buffer translation doesn't happen on the Python side.

jonatanklosko:
@josevalim the precompiled CUDA binaries are finally in place, I released xla 0.3.0 and this should be ready to go!

seanmor5 (approved):

josevalim:
@polvalente @seanmor5 can one of you run this branch on the GPU and let us know if all is good?

josevalim:

PR #647, opened 2022-02-18
Add conv for torchx

howard0su:


torchx/c_src/torchx.cpp
```diff
@@ -786,6 +786,32 @@ NIF(solve)
...
+  c10::optional<at::Tensor> bias_tensor;
+
+  std::vector<int64_t> output_padding;
+  // aten::_convolution(Tensor input, Tensor weight, Tensor? bias,
+  //      int[] stride, int[] padding, int[] dilation, bool transposed,
+  //      int[] output_padding, int groups, bool benchmark, bool deterministic, bool cudnn_enabled, bool allow_tf32) -> Tensor
+  TENSOR(at::_convolution(*tensorA, *tensorB, bias_tensor,
+    stride, padding, dilation, transposed, output_padding, groups,
+    true,   // benchmark
+    true,   // deterministic
```
seanmor5, line 809: Just a note that for performance this should probably be set to false, but we should later consider or see how we can add a deterministic API to Nx compilers/backends to guarantee consistent results in different algorithms

howard0su: ok

torchx/lib/torchx/backend.ex
```diff
@@ -848,6 +848,30 @@ defmodule Torchx.Backend do
...
+    # output_permutation = opts[:output_permutation]
+
+    if padding == :valid do
+      padding = List.duplicate(0, length(t.shape))
+    end
+
+    t = from_nx(t)
+    k = from_nx(k)
+
+    to_nx(Torchx.conv(t, k, strides, padding, input_dilation, false, feature_groups), out)
```
polvalente, line 872:
```suggestion
    t
    |> Torchx.conv(k, strides, padding, input_dilation, false, feature_groups)
    |> to_nx(out)
```

torchx/lib/torchx/backend.ex
```diff
@@ -870,6 +870,36 @@ defmodule Torchx.Backend do
...
+    # Not supported yet
+    # input_dilation = opts[:input_dilation]
+    # batch_groups = opts[:batch_group_size]
+    # input_permutation = opts[:input_permutation]
+    # kernel_permutation = opts[:kernel_permutation]
+    # output_permutation = opts[:output_permutation]
+
+    pad_config =
+      padding
+      |> Enum.map(fn {a, b} -> [a, b] end)
```
howard0su, line 889: shall I use Tuple.to_list here? which one is better?

josevalim: Both are fine. Given this case you always have two elements, I prefer the current version!

torchx/lib/torchx/backend.ex
```diff
@@ -870,6 +870,36 @@ defmodule Torchx.Backend do
...
+    strides = opts[:strides]
+    kernel_dilation = opts[:kernel_dilation]
+    feature_groups = opts[:feature_group_size]
+
+    # Not supported yet
+    # input_dilation = opts[:input_dilation]
+    # batch_groups = opts[:batch_group_size]
+    # input_permutation = opts[:input_permutation]
+    # kernel_permutation = opts[:kernel_permutation]
+    # output_permutation = opts[:output_permutation]
```
josevalim, line 885: Should we raise if any of those options are set?

nx/lib/nx.ex
```diff
@@ -7823,6 +7823,71 @@ defmodule Nx do
...
+              [25.0],
+              [20.0]
+            ],
+            [
+              [35.0],
+              [30.0]
+            ]
+          ]
+        ]
+      >
```
josevalim, line 7889: Let's please move this test to torchx/test/nx_test.exs. My concern about having it here is that it is too long and therefore not very useful as documentation. :(

torchx/lib/torchx/backend.ex
```diff
@@ -870,6 +870,57 @@ defmodule Torchx.Backend do
...
 
+  @impl true
+  def conv(%T{type: type} = out, t, k, opts) do
+    unsupported_option!(opts, :batch_group_size, 1)
+
+    input_dilation = opts[:input_dilation]
+
+    Enum.each(input_dilation, fn a ->
+      if a != 1 do
+        raise "input_dilation other than 1 is not supported"
```
josevalim, line 881:
```suggestion
        raise ArgumentError,
              "input_dilation other than 1 is not supported, got: #{inspect(input_dilation)}"
```

josevalim (commented):
This is looking great. Just two minor comments and we can ship it!!

nx/lib/nx.ex
howard0su: This should be a generic tests not specific to torchx.

josevalim (approved):

josevalim:

PR #650, opened 2022-02-19
Fix heisentests in regards to cache

josevalim:


exla/lib/exla.ex
```diff
@@ -294,12 +294,13 @@ defmodule EXLA do
 
   ## Examples
 
+      iex> fun = fn a, b -> Nx.add(a, b) end
```
polvalente, line 297: Shouldn't this be the same as &Nx.add/2?

polvalente (approved):

exla/lib/exla.ex
polvalente: Ok, so from what I gathered the BEAM will treat captures differently than `fn` functions, is that right?

josevalim: Yeah, an optimization in the future may break it, but they should be distinct.

josevalim:

PR #653, opened 2022-02-19
Add Nx.axis_index/2

jonatanklosko:
We currently use `Nx.Shape.normalize_axis/3` a lot internally, but a similar functionality could be useful outside. Specifically, when writing custom generic functions that accept `:axis`, in order to validate shapes we may need the axis index to operate on shape tuples.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -2764,6 +2764,27 @@ defmodule Nx do
   end
 
   @doc """
+  Returns the index of the given axis in the tensor.
+
+  ### Examples
+
+      iex> Nx.axis_index(Nx.iota({100, 10, 20}), 0)
+      0
+
+      iex> Nx.axis_index(Nx.iota({100, 10, 20}), -1)
+      2
+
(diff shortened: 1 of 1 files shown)


nx/lib/nx.ex
```diff
@@ -2763,6 +2763,27 @@ defmodule Nx do
...
+
+      iex> Nx.axis_index(Nx.iota({100, 10, 20}), -1)
+      2
+
+      iex> Nx.axis_index(Nx.iota({100, 10, 20}, names: [:batch, :x, :y]), :x)
+      1
+
+  """
+  @doc type: :shape
+  def axis_index(tensor, axis) do
```
jonatanklosko, line 2782: Technically the returned value is also a valid axis, so I'm not sure about the naming. Alternatives: `normalized_axis`, `plain_axis`, wdyt?

josevalim:
I think the current name is great! We just need an example of how this function raises for an invalid index.

jonatanklosko:
@josevalim done!

josevalim (approved):

josevalim:

PR #656, opened 2022-02-21
Try to fix memory problem

howard0su:


torchx/c_src/torchx.cpp
```diff
@@ -129,15 +137,22 @@ create_tensor_resource(ErlNifEnv *env, torch::Tensor tensor)
   ret = enif_make_resource(env, tensorPtr);
   enif_release_resource(tensorPtr);
 
+  printf("Tensor %p is created\n", tensorPtr);
```
polvalente, line 140: I think this should be using std::cout if this is going to be kept in the code

```suggestion
  std::cout << "Tensor "<< tensorPtr << " is created" << std::endl;
```

howard0su: this is just for quick debug. I will remove it after I sort out the problem

polvalente: Don't mind me then haha!

josevalim:
Hi @howard0su! Is this PR is still relevant? :)

josevalim:
Continuing in #758

PR #659, opened 2022-02-22
Add EXLA.set_as_nx_default/2

josevalim:
EXLA will have a fully functional backend in the future, so I thought we should rename the function to something more generic. We could even add a similar function to Torchx in the future.


seanmor5 (approved):

josevalim:

PR #664, opened 2022-02-23
feat(torchx): clamp indexed_add/gather indices

polvalente:


josevalim (approved):

polvalente:
@josevalim should we keep raising on binary backend or should we clamp there as well?

josevalim:
I would keep raising for now, we can always revisit later.

PR #667, opened 2022-03-03
feat(torchx): window_max

polvalente:


torchx/lib/torchx/backend.ex
```diff
@@ -979,6 +979,41 @@ defmodule Torchx.Backend do
...
+    pool_shape =
+      List.to_tuple(
+        List.duplicate(1, 4 - tuple_size(tensor.shape)) ++ Tuple.to_list(tensor.shape)
+      )
+
+    tensor
+    |> Nx.multiply(1.0)
+    |> Nx.reshape(pool_shape)
+    |> from_nx()
+    |> Torchx.max_pool_3d(window_dims, strides, List.duplicate(0, 3), window_dilations)
```
polvalente, line 1011: @josevalim From what I found, these `max_*_3d` functions are gonna help us with the implementation for our `window_*` functions.

The only problem is that it only accepts floating point values. Should I just raise for integer types, or should I promote them to floats and return them as floats still?

torchx/lib/torchx/backend.ex
```diff
@@ -979,6 +979,41 @@ defmodule Torchx.Backend do
     |> to_nx(out)
   end
 
+  @impl true
+  def window_max(out, tensor, window_dims_tuple, opts) do
+    if Enum.any?(opts[:padding], fn conf -> conf |> Tuple.to_list() |> Enum.any?(&(&1 != 0)) end) do
```
polvalente, line 984: @josevalim @seanmor5 I'm going to need a little help regarding the padding config. It's not really clear to me what's the equivalence there.

torchx/lib/torchx/backend.ex
polvalente: Other than this, the PR only fails on padded calls :)

josevalim (approved):
LGTM!

torchx/lib/torchx/backend.ex
```diff
@@ -979,6 +979,43 @@ defmodule Torchx.Backend do
...
+      List.to_tuple(
+        List.duplicate(1, 4 - tuple_size(tensor.shape)) ++ Tuple.to_list(tensor.shape)
+      )
+
+    padding = List.duplicate(0, 3)
+
+    tensor
+    |> from_nx()
+    |> Torchx.reshape(pool_shape)
+    |> Torchx.to_type(:double)
```
howard0su, line 1012: We don't need to convert to double if it is already float?

torchx/lib/torchx/backend.ex
```diff
@@ -979,6 +979,43 @@ defmodule Torchx.Backend do
...
+
+    window_dims =
+      List.duplicate(1, 3 - tuple_size(window_dims_tuple)) ++ Tuple.to_list(window_dims_tuple)
+
+    pool_shape =
+      List.to_tuple(
+        List.duplicate(1, 4 - tuple_size(tensor.shape)) ++ Tuple.to_list(tensor.shape)
+      )
+
+    padding = List.duplicate(0, 3)
```
howard0su, line 1007: why not create a static variable?

torchx/lib/torchx/backend.ex
```diff
@@ -979,6 +979,43 @@ defmodule Torchx.Backend do
...
+      raise ArgumentError, "padding unsupported"
+    end
+
+    strides = opts[:strides]
+
+    if tuple_size(window_dims_tuple) > 3 do
+      raise ArgumentError, "cannot have more than 3 dimensions for the window"
+    end
+
+    window_dilations = opts[:window_dilations]
```
howard0su, line 994: suggest to move fetch opts into the beginning of the function to make it easy to understand what is supported opts.

torchx/lib/torchx/backend.ex
polvalente: This should be removed soon, and it's not like it's a costly operation to justify inlining it

josevalim (approved):

PR #670, opened 2022-03-11
feat(torchx): overhaul window functions (+ sum and product)

polvalente:
This PR is a major refactor upon the Torchx implementation for windowed functions. It's a little bit of a step back in the sense that dilations aren't supported anymore, ~but padding will be easier to implement now through `Torchx.pad`~ but padding is now supported through `Torchx.pad` and we get the basic implementations for `window_sum` and `window_product` working as well

### DIFF
--- a/torchx/c_src/torchx.cpp
+++ b/torchx/c_src/torchx.cpp
@@ -368,6 +368,16 @@ NIF(flip)
   TENSOR(torch::flip(*input, dims));
 }
 
+NIF(unfold)
+{
+  TENSOR_PARAM(0, input);
+  PARAM(1, int64_t, dim);
+  PARAM(2, int64_t, size);
+  PARAM(3, int64_t, step);
+
+  TENSOR(at::native::unfold(*input, dim, size, step));
+}
+
 NIF(permute)
 {
   TENSOR_PARAM(0, t);
@@ -940,6 +950,7 @@ static ErlNifFunc nif_functions[] = {
     DF(indexed_add, 4),
     DF(argsort, 3),
     DF(flip, 2),
+    DF(unfold, 4),
 
     DF(add, 2),
     DF(subtract, 2),
--- a/torchx/lib/torchx/backend.ex
+++ b/torchx/lib/torchx/backend.ex
@@ -992,6 +992,7 @@ defmodule Torchx.Backend do
       tensor,
       window_dims_tuple,
       opts,
+      tensor.type |> Nx.Constants.min_finite() |> Nx.to_number(),
       &Torchx.amax(&1, &2, false)
     )
   end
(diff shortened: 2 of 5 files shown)


torchx/c_src/torchx.cpp
```diff
@@ -368,6 +368,16 @@ NIF(flip)
...
 }
 
+NIF(unfold)
+{
+  TENSOR_PARAM(0, input);
+  PARAM(1, int64_t, dim);
+  PARAM(2, int64_t, size);
+  PARAM(3, int64_t, step);
+
+  TENSOR(at::native::unfold(*input, dim, size, step));
```
polvalente, line 378: I still need to check if this function is supported on GPU. It wasn't documented and I found it by chance looking from the Python perspective of libtorch.

torchx/lib/torchx/backend.ex
```diff
@@ -981,66 +987,78 @@ defmodule Torchx.Backend do
...
-  end
-
-  @impl true
-  def window_min(out, tensor, window_dims_tuple, opts) do
-    # libtorch does not expose min_pool3d or similar
-    # Therefore, we use the fact that for any set of negative numbers,
-    # the maximum value is also the one with the smallest absolute value.
-
-    # For instance abs(min([1, 2, 3, 4])) == abs(max([-1, -2, -3, -4])).
+    # TO-DO: support padding
```
polvalente, line 1039: Padding is not supported in this PR because for some reason Nx.Constants.max_finite and min_finite aren't playing well with libtorch

torchx/test/torchx/nx_test.exs
```diff
@@ -634,9 +636,11 @@ defmodule Torchx.NxTest do
       end
     end
 
+    # window dilations temporarily broken
+    @tag :skip
     test "works with non-default options" do
       t = Nx.tensor([[[4, 2, 1, 3], [4, 2, 1, 7]], [[1, 2, 5, 7], [1, 8, 9, 2]]])
-      opts = [strides: [2, 1, 1], padding: :valid, window_dilations: [1, 2, 2]]
+      opts = [strides: [2, 1, 1], padding: :valid, window_dilations: [1, 2, 1]]
```
polvalente, line 643:
```suggestion
      opts = [strides: [2, 1, 1], padding: :valid, window_dilations: [1, 2, 2]]
```

torchx/c_src/torchx.cpp
polvalente: All tests pass with `LIBTORCH_TARGET=cu111`, so it works!

torchx/lib/torchx/backend.ex
polvalente: For some reason Torchx was behaving a lot differently if we passed an Nx.tensor instead of a number for the padding constant. This has been fixed.

howard0su (approved):
LGTM

josevalim (approved):

PR #672, opened 2022-03-13
implement backend_deallocate properly

howard0su:


torchx/c_src/torchx.cpp
```diff
@@ -3,6 +3,8 @@
 
 #include "nx_nif_utils.hpp"
 
+#define EMPTY_TAG 0xABADFACE
```
josevalim, line 6: Could it be the null byte instead?

torchx/c_src/torchx.cpp
```diff
@@ -136,10 +141,17 @@ NIF(delete_tensor)
...
+  {
+    t->~Tensor();
+    *(intptr_t*)t = EMPTY_TAG;
 
-  return nx::nif::ok(env);
+    return nx::nif::ok(env);
+  }
+  else
+  {
+    return nx::nif::error(env, "already deleted tensor");
```
josevalim, line 153: Do we ever get to this clause? 🤔

josevalim (commented):
This is great! I have added some comments.

torchx/c_src/torchx.cpp
howard0su: No. this is memory content not pointer itself.

torchx/c_src/torchx.cpp
howard0su: we should not as we already checked in TENSOR_PARAM macro. But I don't know what is proper way to express that here.

torchx/c_src/torchx.cpp
josevalim: Does it mean we can have a false positive? Could the memory content ever be `0xABADFACE`? I would add a comment then explaining it is memory content and why this is a "safe" value to use. :)

torchx/c_src/torchx.cpp
josevalim: One option would be to remove the conditional altogether and keep only the "if" branch. If you don't want to do that, then I would add a comment:

```suggestion
    // We should never reach this clause as TENSOR_PARAM would have raised
    return nx::nif::error(env, "already deleted tensor");
```

I have also just realized that this code can have race conditions. What happens if two Elixir processes try to delete the same tensor at the same time? Do we have a simple and cheap solution in C/C++ where we would have a mutex/lock per tensor? Otherwise we could solve this on the Elixir side though and have this operation happen behind a pool of Elixir processes.

torchx/c_src/torchx.cpp
howard0su: yes, it is possible in theory. But in reality, torch structure is a C++ object which has vtable as first element. the value here is not a valid vtable ptr.

torchx/c_src/torchx.cpp
howard0su: good point. didn't realize NIF need to deal with multi threads. Is this the only function worry about threading?

josevalim: I think this is the only function performing mutations... or am I wrong and none of libtorch/at API is thread-safe?

howard0su: I found NIF has mutex APIs we can use.

torchx/lib/torchx/backend.ex
```diff
@@ -165,7 +165,13 @@ defmodule Torchx.Backend do
   end
 
   @impl true
-  def backend_deallocate(%T{} = t), do: Torchx.delete_tensor(from_nx(t))
+  def backend_deallocate(%T{} = t) do
+    try do
+      Torchx.delete_tensor(from_nx(t))
+    rescue
+      ArgumentError -> :already_deallocated
+    end
```
polvalente, line 173:
```suggestion
    Torchx.delete_tensor(from_nx(t))
  rescue
    ArgumentError -> :already_deallocated
```

howard0su:
in case if one thread is doing operation while holding a tensor reference, and another thread calling deallocate tensor, i think it will crash. I checked exla, i think it doesn't protect as well. @josevalim

josevalim:
> in case if one thread is doing operation while holding a tensor reference, and another thread calling deallocate tensor, i think it will crash. I checked exla, i think it doesn't protect as well. @josevalim

You are correct. I need to think more about this then. We may need to have readwrite locks around all tensor operations.

howard0su:
> > in case if one thread is doing operation while holding a tensor reference, and another thread calling deallocate tensor, i think it will crash. I checked exla, i think it doesn't protect as well. @josevalim
> 
> You are correct. I need to think more about this then. We may need to have readwrite locks around all tensor operations.

There is rwlock in NIF. It is not hard to add rwlock around tensor. However the lock order problem become hard to solve. I suggest we should not go with this direction.

josevalim:
One suggestion I got was to use an atomic. Before each operation, we set the atomic to true and at the end we set it to false. If when setting it to true, it was already true, we raise an exception. I think this is a fair compromise that will give us safety but prevent anyone from deallocating/transferring they should not. WDYT?

For EXLA, I wonder if this is handled by their streams where all operations are scheduled.

torchx/c_src/torchx.cpp
```diff
@@ -21,8 +22,79 @@ inline const std::string* type2string(const torch::ScalarType type)
...
+    // setup
+    if (!enif_get_resource(env, arg, TENSOR_TYPE, (void **)&ptr))
+    {
+      // didn't give this parameter
+      return;
+    }
+
+    refcount = (std::atomic<int> *)(ptr + 1);
+
+    if (refcount->load() == 0)
```
polvalente, line 40:
```suggestion
    if (refcount->load() <= 0)
```

polvalente: Actually I think that the situations in which this could be negative would imply segfaults, right?

howard0su: if this is negative, then there is something terrible wrong.

torchx/c_src/torchx.cpp
```diff
@@ -769,7 +822,7 @@ NIF(cholesky)
 
 NIF(pad)
 {
-  TENSOR_PARAM(0, tensor)
+  TENSOR_PARAM(0, tensor);
```
josevalim, line 825:
```suggestion
  TENSOR_PARAM(0, tensor)
```

?

torchx/c_src/torchx.cpp
```diff
@@ -134,12 +208,10 @@ create_tensor_resource(ErlNifEnv *env, torch::Tensor tensor)
...
 NIF(delete_tensor)
 {
-  TENSOR_PARAM(0, t);
-
-  t->~Tensor();
-  enif_release_resource(t);
+  TensorP tensor(env, argv[0]);
 
-  return nx::nif::ok(env);
+  return tensor.deallocate()? nx::nif::ok(env):
```
josevalim, line 213:
```suggestion
  return tensor.deallocate() ? nx::nif::ok(env) :
```

torchx/c_src/torchx.cpp
```diff
@@ -134,12 +208,10 @@ create_tensor_resource(ErlNifEnv *env, torch::Tensor tensor)
...
+    enif_make_badarg(env);
```
josevalim, line 214: Maybe instead of raising we should return `:already_deallocated`?

torchx/c_src/torchx.cpp
josevalim, line 40: Are we sure that this can never be negative? Imagine you have three threads: one is holding the tensor on a computation. Now two other threads call deallocate at the same time. Because the tensor is still being used, the tensor is valid, and both deallocate counts will decrease the refcount which can then be negative.

Can this be a thing?

If this is possible, then there is another scenario where two threads are using the tensor, two threads deallocate it, and now the first thread using the tensor will deallocate it, before the other thread using it finishes.

torchx/c_src/torchx.cpp
```diff
@@ -750,7 +822,7 @@ NIF(cholesky)
 
 NIF(pad)
 {
-  TENSOR_PARAM(0, tensor)
+  TENSOR_PARAM(0, tensor);
   LIST_PARAM(1, std::vector<int64_t>, config)
   SCALAR_PARAM(2, constant)
```
josevalim, line 827:
```suggestion
  LIST_PARAM(1, std::vector<int64_t>, config);
  SCALAR_PARAM(2, constant);
```

torchx/c_src/torchx.cpp
howard0su: this will not protect calling deallocate twice. elixir side will have such issue?

howard0su: but it will not harm if you calling twice. we will still deallocate once on tensor. but both callers will get success message.

torchx/c_src/torchx.cpp
howard0su: i will fix this in a separate PR. and use do-while(0) to wrapper all macro to make sure it in the future.

howard0su: also do you think we should maintain a shared copy of nx_nif_util.h for both torch and exla?

torchx/c_src/torchx.cpp
howard0su: I prefer elixir code keep the API interface. I don't prefer C++ code has special atom here.

torchx/c_src/torchx.cpp
josevalim: Yes, you can call deallocate twice. I mean, it is definitely bad code to do so, but it could happen? The issue is that if we call deallocate twice, this condition may never be met because `refcount->load()` will now be negative?

Then, even if we change this to be `refcount->load() < 0`, then there is the issue in that: if you have two threads calling deallocate twice, and then two threads using and blocking a tensor, then when the first thread using and blocking a tensor terminates, it will deallocate before the other thread finishes.

I feel like we have two options:

1. Use a atomic to avoid unsafe writes. If you call deallocate and the refcount is not zero, we raise saying we cannot deallocate. We push the concern to the user, but I think we are fine.

2. Use two atomics: one to track reads and another to track deallocations.

I think we should go with 1 for simplity and we document that, if you try to deallocate while it is being used elsewhere, it will raise. It should not happen in practice anyway. WDYT?

josevalim: Actually, I think my suggestions above are wrong. We need one atomic and one flag to keep it safe:

1. On deallocate, we set the `to_deallocate` flag to false
2. When reading the tensor, we can only bump the refcount if the tensor is not be deallocated
3. When the refcounting is zero and the tensor was set to be deallocated, we effectively deallocate it

The `deallocate` function itself can use the tensor pointer, as all it does it to set a flag to true, and if the refcount is zero at the end, it will deallocate.

howard0su: I don't think this approach will get anything better. The current refcount is a standard implementation. As I mentioned, the only problem is that you will get two successful deallocate call. But deallocate will still happen once.

josevalim: The issue is that if you call deallocate twice, refcount->load() will now return `-1` and this conditional won't match. Setting this clause to `< 0` will also have bugs. I think you need to guarantee deallocation is either idempotent or atomic.

howard0su: This grantee atomic. let me try to explain in graph.

howard0su: image

josevalim: Great idea with the drawings. Here is what I am thinking:

```
threadA    threadB    threadC     ref_count
                                      1
 enter                                2
  use
            enter                     3
          deallocate                  2
                       enter          3
                     deallocate       2
            leave                     1
                       leave (!!!)    0
```

In the scenario above we will leave while it is still being used, no?

torchx/c_src/torchx.cpp
```diff
@@ -870,9 +953,10 @@ NIF(max_pool_3d)
 void free_tensor(ErlNifEnv *env, void *obj)
 {
   torch::Tensor* tensor = reinterpret_cast<torch::Tensor*>(obj);
-  if (tensor != nullptr) {
+
+  if (tensor != nullptr && IS_TENSOR_ALIVE(tensor))
```
josevalim, line 957: Can the tensor ever be a nullptr now?

torchx/c_src/torchx.cpp
```diff
@@ -21,8 +22,89 @@ inline const std::string* type2string(const torch::ScalarType type)
...
+      return;
+    }
+
+    refcount = (std::atomic<int> *)(ptr + 1);
+    deleted = (std::atomic_flag *)(refcount + 1);
+
+    if (refcount->load() == 0)
+    {
+      // already deallocated
+      ptr = nullptr;
```
josevalim, line 44: The PR looks good but I am bit confused about this code path that sets the pointer to `nullptr`. This path will only be executed if you try to access a deallocated tensor, so it feels weird that trying to access a deallocated tensor makes pointer changes.

We should either not set the pointer to null or do it on deallocate. I think we don't need to set it to nil because the atomic variables gives us everything we need to know about the tensor state.

howard0su: this ptr is class variable not tensor itself.

josevalim:

PR #673, opened 2022-03-13
Fix test cases for backend_transfer

howard0su:
fix the cases we still need the original variable. use _copy instead.


josevalim (approved):

josevalim:

PR #674, opened 2022-03-16
feat(torchx): scatter_window_max

polvalente:


torchx/lib/torchx/backend.ex
```diff
@@ -1021,41 +1021,85 @@ defmodule Torchx.Backend do
...
+      |> Nx.argmax(axis: -1)
+
+    indices_to_flatten =
+      tensor
+      |> Nx.axes()
+      |> Enum.map(fn axis ->
+        tensor
+        |> Nx.iota(axis: axis)
+        |> then(unfold_flat)
+        |> Nx.take_along_axis(Nx.new_axis(argmax, -1), axis: -1)
```
polvalente, line 1065: @josevalim This one is kinda tricky due to indexed_add and take_along_axis. Should I refactor it to use Torchx directly instead of going through Nx? The code might become far less readable going that route, though

polvalente: I thought using the torchs `tensor` as a template would allocate it through torchx. I'll look into this and force the backend if it's indeed on the BinaryBackend

torchx/lib/torchx/backend.ex
```diff
@@ -1069,6 +1113,29 @@ defmodule Torchx.Backend do
...
+  def unfold_windows(%T{} = tensor, padding, pad_constant, window_dims_tuple, strides) do
+    unfold_windows(from_nx(tensor), padding, pad_constant, window_dims_tuple, strides)
+  end
+
+  def unfold_windows(tensor, padding, pad_constant, window_dims_tuple, strides) do
+    padding =
+      padding
+      |> Enum.map(fn {a, b} -> [a, b] end)
+      |> Enum.reverse()
+      |> List.flatten()
```
josevalim, line 1125: You can probably do this much more efficiently with a single reduce:

```suggestion
      Enum.reduce(padding, [], fn {a, b}, acc -> [a, b | acc] end)
```

There may be other usages here. It may be worth extracting to a function too. Like flatten_padding. :D

torchx/lib/torchx/backend.ex
josevalim: I don't think it does but you are right. Perhaps it should!

polvalente: @josevalim I think it would be better if we leave allocating the iotas on the same device for another PR.

This ends up touching lots of other functionalities as well.
WDYT?

polvalente: (the code currently has the WIP for this, but I need to hunt a segfault)

josevalim: Fine by me, so please add a TODO and ship it. :)

josevalim (approved):

PR #678, opened 2022-03-21
Change Nx.shuffle/2 along axis to keep order along other axis

jonatanklosko:
Updates the behaviour to match [tf.random.shuffle](https://www.tensorflow.org/api_docs/python/tf/random/shuffle) and [np.random.shuffle](https://numpy.org/doc/stable/reference/random/generated/numpy.random.shuffle.html).

The behaviour with no `:axis` stays the same.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -931,23 +931,57 @@ defmodule Nx do
   @doc """
   Shuffles tensor elements.
 
+  By default shuffles elements within the whole tensor. When `:axis`
+  is given, shuffles the tensor along the specific axis instead.
+
   ## Options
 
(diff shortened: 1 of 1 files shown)


nx/lib/nx.ex
```diff
@@ -931,23 +931,57 @@ defmodule Nx do
   @doc """
   Shuffles tensor elements.
 
+  By default shuffles elements within the whole tensor. When `:axis`
```
polvalente, line 934:
```suggestion
  By default, shuffles elements within the whole tensor. When `:axis`
```

polvalente (approved):

PR #689, opened 2022-03-23
feat: minimal complex number implementation for Nx

polvalente:
First draft for complex numbers in Nx


nx/mix.lock
```diff
@@ -1,5 +1,7 @@
 %{
+  "complex": {:git, "https://github.com/polvalente/elixir-complex.git", "e37cb0629e445dd72a5a9a1a9935ec35872d0236", [branch: "main"]},
   "earmark_parser": {:hex, :earmark_parser, "1.4.19", "de0d033d5ff9fc396a24eadc2fcf2afa3d120841eb3f1004d138cbf9273210e8", [:mix], [], "hexpm", "527ab6630b5c75c3a3960b75844c314ec305c76d9899bb30f71cb85952a9dc45"},
+  "elixir_complex": {:git, "https://github.com/polvalente/elixir-complex.git", "0ba7011a167e5dd81b53e652218b9151f1c23981", []},
```
josevalim, line 4:
```suggestion
```

nx/lib/nx/binary_backend.ex
```diff
@@ -737,6 +746,23 @@ defmodule Nx.BinaryBackend do
   defp element_logical_xor(_, _, r) when r == 0, do: 1
   defp element_logical_xor(_, _, _), do: 0
 
+  defp validate_numbers(a, b, fun) do
```
josevalim, line 749: This validation should not be here. It should happen in the `Nx` module. Those functions should fail early if they get a complex number.

nx/lib/nx/binary_backend.ex
```diff
@@ -693,49 +693,48 @@ defmodule Nx.BinaryBackend do
     from_binary(out, data)
   end
 
-  defp element_add(_, a, b), do: a + b
-  defp element_subtract(_, a, b), do: a - b
-  defp element_multiply(_, a, b), do: a * b
-  defp element_divide(_, a, b), do: a / b
+  defp element_add(_, a, b), do: Complex.add(a, b)
```
jonatanklosko, line 696: Putting convenience aside, I think it's confusing that `Complex.add(1, 1)` returns `2` instead of `%Complex{re: 2, im: 0}`. It is similar to how `Nx.add(1, 1)` returns a tensor and not `2`.

If we want to keep the current API, one option would be to document a canonical complex form, where numbers with imaginary part `0` are always returned as regular numbers, as in `Complex.add(~c(1, 2), ~c(-1, -2)) #=> 0`. Then it would be expected that `Complex.add(1, 1) #=> 2`. Not sure if that's the best idea though.

Alternatively, instead of matching for numeric arguments in `Complex`, we may as well have these matches here. I guess the main issue is that then we would need to duplicate the implementation of `erf`, etc?

jonatanklosko: Third idea, we could extend the complex package, and define a `Number` module. Then `Number.add(a, b)` would support both built-in numbers and `Complex` numbers.

polvalente: I would rather have the Number module keeping this "no upcasts" behavior

polvalente: The problem with upcasting is that we cannot differ a complex with no imaginary part from a pure number after the result without checking the input types

jonatanklosko: That's where canonicalization would help, if the imaginary part is 0 we would always downcast, regardless the input types. Then matching on regular number arguments in Complex would just be an optimisation.

But yeah, `Number` module seems like the most transparent option to me.

jonatanklosko: Could also be `Number.Complex`, but this probably means more aliases (and I can imagine `Number` being taken by Elixir at some point). @josevalim will know best :)

nx/lib/nx.ex
```diff
@@ -1483,6 +1498,13 @@ defmodule Nx do
     backend.constant(out, number, options)
   end
 
+  def to_tensor(%Complex{} = number) do
+    {backend, options} = default_backend()
+    {_, size} = number.re |> Nx.Type.infer() |> Nx.Type.merge(Nx.Type.infer(number.im))
```
josevalim, line 1503:
```suggestion
  def to_tensor(%Complex{re: re, im: im} = number) do
    {backend, options} = default_backend()
    {_, size} = re |> Nx.Type.infer() |> Nx.Type.merge(Nx.Type.infer(im))
```

nx/lib/nx.ex
```diff
@@ -3083,8 +3105,30 @@ defmodule Nx do
...
+    :greater_equal,
+    :logical_and,
+    :logical_or,
+    :logical_xor
+  ]
+
   defp element_wise_bin_op(left, right, op, fun) do
     type = binary_type(left, right) |> fun.()
+
+    if op not in @allow_complex_type_bin_ops do
```
josevalim, line 3128: Instead of hard coding it here, I would have a `non_complex_element_wise_bin_op` that raises for complex and then calls this one.

nx/lib/nx/binary_backend.ex
```diff
@@ -783,7 +782,7 @@ defmodule Nx.BinaryBackend do
   def floor(out, tensor), do: element_wise_unary_op(out, tensor, &:erlang.floor/1)
```
josevalim, line 782: What about floor and round? :)

nx/lib/nx/binary_backend.ex
polvalente: https://github.com/polvalente/elixir-complex/pull/4

@jonatanklosko @josevalim this is what Complex.Math would look like. Should we commit to this?

nx/lib/nx/lin_alg.ex
```diff
@@ -1151,6 +1174,9 @@ defmodule Nx.LinAlg do
     Nx.Defn.Kernel.assert_shape_pattern(tensor, {n, n})
 
     transform(tensor, fn tensor ->
+      type = Nx.type(tensor)
+      Nx.Shared.raise_complex_not_supported(type, "LinAlg.determinant", 1)
```
josevalim, line 1178: They are indeed not supported? Or not for now?

nx/lib/nx/binary_backend.ex
polvalente: Missed these ones :)
I'll take a look into torch/numpy to see if they support this

nx/lib/nx/lin_alg.ex
polvalente: Not for now! The definitions for complex numbers are different (transposes become transpose+conjugate, for instance + correction factors here and there)

polvalente: Should I add "for now" in the message and function name?

nx/lib/nx.ex
```diff
@@ -5013,6 +5057,9 @@ defmodule Nx do
   end
 
   ## Unary ops
+  @allow_complex_type_unary_ops Nx.Shared.unary_math_funs()
+                                |> Keyword.keys()
+                                |> Kernel.--([:erf, :erfc, :erf_inv])
```
josevalim, line 5062: Wait, is `erf` not a complex operation? So why do we have Complex.erf? For completeness?

nx/lib/nx/lin_alg.ex
polvalente: or "not supported yet"

josevalim: Yeah, I would raise a different message saying "not yet".

nx/lib/nx.ex
polvalente: There is a complex definition for erf, as per: https://mathworld.wolfram.com/Erf.html
I left it unimplemented for now and chose to not support it in Nx until then. Perhaps we should remove it altogether (and bring the erf and friends implementations back to Nx.Shared)

josevalim: I guess the question is: is it supported by torch/xla? If so, I think we should have it working here and raise inside the `Complex` library.

nx/lib/nx/binary_backend.ex
polvalente: Torch doesn't support any of floor, round and ceil. Numpy supports round, but neither floor nor ceil. How should we go about this?

polvalente: Numpy's round behaves the same as Kernel.round/1

nx/lib/nx/binary_backend.ex
josevalim: Thanks for trying it out. Personally, I prefer `Complex`, without `Complex.Math`. If you always want a complex back, you can simply call `Complex.new()` before calling the other `Complex` functions.

polvalente: Theres always either ~z or ~i or ~j


z because it's normally what people use for complex variables, i and j due to the imaginary constant

jonatanklosko: `~z` is kinda like `~x` for real numbers, but that said it seems the best to me, because it points to the number as a whole rather that the imaginary constant.

nx/lib/nx.ex
```diff
@@ -5319,6 +5362,7 @@ defmodule Nx do
     def unquote(name)(tensor) do
       case to_tensor(tensor) do
         %T{type: {type, _}} = tensor when type in [:s, :u] -> tensor
+        %T{type: {:c, _}} -> Nx.Shared.raise_complex_not_supported(unquote(name), 1)
```
josevalim, line 5365: Is it not supported now or not supported permanently? If not supported now, I would keep it working here and let it raise elsewhere.

nx/lib/nx/binary_backend.ex
josevalim: Oh, I see. Let's not support it for now then.

nx/lib/nx.ex
josevalim: Ignore me. This was answered later.

nx/lib/nx.ex
```diff
@@ -5013,6 +5043,9 @@ defmodule Nx do
   end
 
   ## Unary ops
+  @allow_complex_type_unary_ops Nx.Shared.unary_math_funs()
+                                |> Keyword.keys()
+                                |> Kernel.--([:erf, :erfc, :erf_inv])
```
josevalim, line 5048:
```suggestion
  @disallow_complex_type_unary_ops [:erf, :erfc, :erf_inv]
```

josevalim (approved):
Just one tiny comment and ship it!

Issue #682, opened 2022-03-21
Add missing tests for skipped doctests


polvalente:
We need to add tests at `torchx/test/nx_test.exs` for the following functions:

- [ ] broadcast
- [x] dot
- [x] make_diagonal
- [x] mean
- [x] quotient
- [ ] sigil_M
- [ ] sigil_V

PR #697, opened 2022-03-25
Add make_diagonal tests for Torchx

Benjamin-Philip:
This commit adds the make_diagonal doctests and a make_diagonal
nx test to the Torchx backend.

See #682.

### DIFF
--- a/torchx/test/torchx/nx_test.exs
+++ b/torchx/test/torchx/nx_test.exs
@@ -148,6 +148,16 @@ defmodule Torchx.NxTest do
                Nx.tensor([[0, 1, 2], [3, 4, 5]], backend: Nx.BinaryBackend)
     end
 
+    test "make_diagonal" do
+      t =
+        [1, 2, 3]
+        |> Nx.tensor()
+        |> Nx.make_diagonal()
+
+      assert Nx.backend_transfer(t) ==
+               Nx.tensor([[1, 0, 0], [0, 2, 0], [0, 0, 3]], backend: Nx.BinaryBackend)
+    end
+
     test "random_uniform" do
       t = Nx.random_uniform({30, 50})
 
--- a/torchx/test/torchx/nx_test.exs
+++ b/torchx/test/torchx/nx_test.exs
@@ -154,8 +154,7 @@ defmodule Torchx.NxTest do
         |> Nx.tensor()
         |> Nx.make_diagonal()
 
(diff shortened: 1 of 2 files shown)


torchx/test/torchx/nx_test.exs
```diff
@@ -148,6 +148,16 @@ defmodule Torchx.NxTest do
...
     end
 
+    test "make_diagonal" do
+      t =
+        [1, 2, 3]
+        |> Nx.tensor()
+        |> Nx.make_diagonal()
+
+      assert Nx.backend_transfer(t) ==
+               Nx.tensor([[1, 0, 0], [0, 2, 0], [0, 0, 3]], backend: Nx.BinaryBackend)
```
polvalente, line 158: You cam use assert_all_close to avoid the need to backend_transfer

polvalente (approved):

polvalente:
Thank you!

torchx/test/torchx/nx_test.exs
Benjamin-Philip: > You cam use assert_all_close to avoid the need to backend_transfer

@polvalente quick question: Why aren't we using assert_all_close in the `Nx.iota` test?

https://github.com/elixir-nx/nx/blob/506ac004efdd1076e033b784db752d6996f012da/torchx/test/torchx/nx_test.exs#L144-L149

polvalente: It's probably one of the tests that came before the assertion? 

Although thinking about it we should actually assert exactly for when the tensor is expected to be exactly equal to the other...

I'll send a PR with a new "assert_*" function for this

Benjamin-Philip: > Although thinking about it we should actually assert exactly for when the tensor is expected to be exactly equal to the other...

In that case should `iota` and `make_diagonal` use the `assert_*` function?

polvalente: Yes. I'm thinking about calling it assert_equal. 

It would be something like assert_all_close but using left |> Nx.equal(right) |> Nx.all() |> Nx.to_number() == 1 as the equality check.

Do you want to send a PR on this or should I?

Benjamin-Philip: I think I'll send one.

polvalente: Great!

PR #698, opened 2022-03-25
Introduce assert_equal into Torchx.Case

Benjamin-Philip:
Introduce `assert_equal` to use when 2 tensors are expected to be exactly equal.

See https://github.com/elixir-nx/nx/pull/697#discussion_r8351958590.

### DIFF
--- a/torchx/test/torchx/nx_test.exs
+++ b/torchx/test/torchx/nx_test.exs
@@ -32,10 +32,10 @@ defmodule Torchx.NxTest do
     binary_a = Nx.backend_copy(a, Nx.BinaryBackend)
     binary_b = Nx.backend_transfer(b, Nx.BinaryBackend)
     binary_c = Kernel.apply(Nx, op, [binary_a, binary_b])
-    assert Nx.backend_transfer(c) == binary_c
+    assert_equal(c, binary_c)
 
     mixed_c = Kernel.apply(Nx, op, [a, binary_b])
-    assert Nx.backend_transfer(mixed_c) == binary_c
+    assert_equal(mixed_c, binary_c)
   end
 
   defp test_unary_op(op, data \\ [[1, 2], [3, 4]], type) do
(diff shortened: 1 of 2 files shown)


Benjamin-Philip:
@polvalente, should I include changes to be made to the test in `nx_test` in this PR or in a seperate PR?

polvalente:
> @polvalente, should I include changes to be made to the tests in `nx_test.exs` in this PR or in a seperate PR?

@Benjamin-Philip you can include them in this PR!

Benjamin-Philip:
I've updated all tests that had the following pattern:

```elixir
assert Nx.backend_transfer(tensor1) == tensor2
```

Let me know if there are changes to be added/removed.

torchx/test/torchx/nx_test.exs
```diff
@@ -121,11 +121,13 @@ defmodule Torchx.NxTest do
...
-                   type: {:bf, 16},
-                   backend: Nx.BinaryBackend
-                 )
+        assert_equal(
+          c,
+          Nx.tensor([[0.2001953125, 0.333984375], [0.427734375, 0.5]],
+            type: {:bf, 16},
+            backend: Nx.BinaryBackend
+          )
+        )
```
polvalente, line 130:
```suggestion
          Nx.tensor([[0.2001953125, 0.333984375], [0.427734375, 0.5]])
        )
```

torchx/test/torchx/nx_test.exs
```diff
@@ -249,42 +250,58 @@ defmodule Torchx.NxTest do
 
   describe "Nx.as_type" do
     test "basic conversions" do
```
polvalente, line 252: I think we should revert the ones under this describe because the equality will ensure the type is also equal, while assert_equal glosses over that

polvalente (approved):
Thanks again!

PR #700, opened 2022-03-27
Merge types in torchx select

seanmor5:


josevalim (approved):
We need either a doctest for select or a test on Torchx and we are good to go!!

PR #701, opened 2022-03-28
fix: support mixed backends in Nx.concatenate

polvalente:


polvalente:
@josevalim @seanmor5 we can also backend_transfer instead of failing. Perhaps we could provide an option for the target backend? If it's present, we backend_transfer on it, otherwise, raise pointing to said option

josevalim:
@polvalente I believe we have an implementation in Nx.Shared that receives a list of tensors and finds a common backends between them. The reason I mention this is because we should allow to mix a Torchx tensor with a BinaryBackend tensor, which is what we do in all other Nx operations. We should not support mixing other backends though, and I think the code in Nx.Shared already handles it.

josevalim (approved):

josevalim:
We need a test and we are good to go! I think we can have in Torchx that receives a binary backend, a Torchx backend, and concats them!

PR #706, opened 2022-04-02
add Kernel.power/2

vans163:


josevalim:
@vans163 can you please bump CI to v1.13? The operator is not available in previous versions.

nx/lib/nx/defn/kernel.ex
```diff
@@ -411,6 +411,21 @@ defmodule Nx.Defn.Kernel do
...
+  It delegates to `Nx.power/2` (supports broadcasting).
+
+  ## Examples
+
+      defn power(a, b) do
+        a ** b
+      end
+
+  """
+  def left ** right when Kernel.and(is_number(left), is_number(right)), do: Kernel.**(left, right)
```
polvalente, line 426:
```suggestion
  def unquote(:"**")(left, right) when Kernel.and(is_number(left), is_number(right)), do: Kernel.**(left, right)
```

Please change all occurences to something like this so it compiles on Elixir 1.12

josevalim: I think we can bump to 1.13, especially because it will allow us to use the `~M` and `~V` sigils!

polvalente:
> @vans163 can you please bump CI to v1.13? The operator is not available in previous versions.

@josevalim are we dropping Elixir 1.12?

nx/lib/nx/defn/kernel.ex
polvalente: Sure, makes sense!

vans163:
> @vans163 can you please bump CI to v1.13? The operator is not available in previous versions.

Where is the CI config?

polvalente:
> > @vans163 can you please bump CI to v1.13? The operator is not available in previous versions.
> 
> Where is the CI config?

It's in the .github directory. We should probably bump the requirement in all mix.exs files as well

PR #707, opened 2022-04-03
feat: support typecasting from complex

polvalente:
This mimics the behavior from Jax and Pytorch, in which casting a complex to float returns its real component

### DIFF
--- a/nx/lib/nx/shared.ex
+++ b/nx/lib/nx/shared.ex
@@ -121,7 +121,6 @@ defmodule Nx.Shared do
     if System.endianness() == :little do
       quote do
         case unquote(var) do
-          %Complex{re: x} -> binary_part(<<x::float-native-32>>, 2, 2)
           x when is_number(x) -> binary_part(<<x::float-native-32>>, 2, 2)
           x -> Nx.Shared.write_bf16(x)
         end :: binary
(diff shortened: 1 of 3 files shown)


nx/lib/nx/shared.ex
```diff
@@ -155,14 +157,14 @@ defmodule Nx.Shared do
...
+        %Complex{re: x} -> <<x::float-native-size(unquote(size))>>
         x when is_number(x) -> <<x::float-native-size(unquote(size))>>
         x -> Nx.Shared.write_non_finite(x, unquote(size))
       end :: binary
     end
   end
 
-  defp write_bin_modifier(var, type, size),
-    do: shared_bin_modifier(var, type, size)
+  defp write_bin_modifier(var, type, size), do: shared_bin_modifier(var, type, size)
```
polvalente, line 167: I couldn't find a way to get this match to work like the other ones above...
If we get this to work, perhaps we can avoid the `case` as it is in the BinaryBackends.

Any thoughts on this?

josevalim: Are we sure we want to change it here? I don't think we want to bring the behaviour of `as_type`, which may include loss of data, into our default conversion code. We can't even cast floats to integers at the moment outside of `as_type`. My suggestion is to solve all of this within `as_type`.

polvalente: Makes sense! I thought this was the better place at first, but I hadn't thought about loss of information

josevalim (approved):

Issue #693, opened 2022-03-24
Complex number utility functions


polvalente:
Add implementations for some complex number related functions:

- [x] Nx.conjugate: returns the complex conjugate of each entry
- [x] Nx.real: returns a float tensor with the real part of each entry
- [x] Nx.imag: returns a float tensor with the imaginary part of each entry
- [ ] Nx.phase: returns the phase for each entry

~Nx.conjugate can be implemented with Nx.Shared.optional + Nx.real + Nx.imag~

Benjamin-Philip:
@polvalente, Do you think having a Complex.imag and a Complex.real would be useful for Nx.imag and Nx.real?

That way we can make Complex manage cases with integers, floats and the general logic, while we just write an implementation that just maps the corresponding function.

PR #708, opened 2022-04-04
Add Nx.conjugate

Benjamin-Philip:
See #693.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -5199,12 +5199,17 @@ defmodule Nx do
 
   ## Examples
 
-       iex>Nx.conjugate(Complex.new(1, 2))
+       iex> Nx.conjugate(Complex.new(1, 2))
        #Nx.Tensor<
          c64
          1.0-2.0i
        >
 
+       iex> Nx.conjugate(1)
+       #Nx.Tensor<
+         c128
+         1.0+0.0i
+       >
 
        iex> Nx.conjugate(Nx.tensor([Complex.new(1, 2), Complex.new(2, -4)]))
        #Nx.Tensor<
@@ -5216,7 +5221,7 @@ defmodule Nx do
   def conjugate(tensor) do
     tensor = to_tensor(tensor)
 
-    impl!(tensor).conjugate(tensor, tensor)
+    impl!(tensor).conjugate(%{tensor | type: Nx.Type.to_complex(tensor.type)}, tensor)
   end
 
   @doc """
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -5193,6 +5193,33 @@ defmodule Nx do
   end
 
   @doc """
+  Calculates the complex conjugate of each element in the tensor.
+
+  If $z = a + bi$, $conjugate(z) = z^* = a - bi$
+
+  ## Examples
+
+       iex>Nx.conjugate(Complex.new(1, 2))
+       #Nx.Tensor<
+         c64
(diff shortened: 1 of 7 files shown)


Benjamin-Philip:
@polvalente,

So far I have only added an implementation for `BinaryBackend`. I also haven't written implementations for both EXLA and Torchx before. Are there any pointers/suggestions to keep in mind when writing one?

polvalente:
> 

We can hold off on Torchx and EXLA for now because they don't support Complex numbers yet. I have a WIP branch for Torchx

Benjamin-Philip:
> We can hold off on Torchx and EXLA for now because they don't support Complex numbers yet. I have a WIP branch for Torchx

In that case do we merge, or do we close and wait for complex support in Torchx and EXLA?

nx/lib/nx.ex
```diff
@@ -5192,6 +5192,33 @@ defmodule Nx do
...
+       #Nx.Tensor<
+         c64[2]
+         [1.0-2.0i, 2.0+4.0i]
+       >
+  """
+  @doc type: :element
+  def conjugate(tensor) do
+    tensor = to_tensor(tensor)
+
+    impl!(tensor).conjugate(tensor, tensor)
```
polvalente, line 5219: There's an issue with the output type here.

Try a test/example in which the input is s64 ou f32. I'd expect the output type to be inferred incorrectly (it should be the corresponding complex type).

I think we need a `to_complex` function in Nx.Types, which promotes everything below f32 to c64, f64 -> c128 and keeps complex types as the same.

Then you would have somthing like:
```suggestion
    impl!(tensor).conjugate(%{tensor | type: Nx.Types.to_complex(tensor.type)}, tensor)
```

polvalente: Regarding EXLA/Torchx, we can go ahead and merge this after this adjustment :)

polvalente: Oh, and regarding the failing torchx tests, there's a list to which you can add the tests in torchx/test/nx_doctest_test.exs

Benjamin-Philip: Ah, yes I forgot about it :)

nx/lib/nx.ex
```diff
@@ -5192,6 +5192,38 @@ defmodule Nx do
     end
   end
 
+  @doc """
+  Calculates the complex conjugate of each element in the tensor.
+
+  If $z = a + bi$, $conjugate(z) = z^* = a - bi$
```
polvalente, line 5198:
```suggestion
  If $z = a + bi = r e^\\theta$, $conjugate(z) = z^* = a - bi =  r e^{-\\theta}$
```

nx/lib/nx/type.ex
```diff
@@ -171,6 +171,30 @@ defmodule Nx.Type do
...
+      {:c, 64}
+      iex> Nx.Type.to_complex({:s, 64})
+      {:c, 128}
+      iex> Nx.Type.to_complex({:f, 64})
+      {:c, 128}
+
+  """
+  def to_complex({:c, size}), do: {:c, size}
+  def to_complex({_name, size}) when size <= 32, do: {:c, 64}
+  def to_complex(_type), do: {:c, 128}
```
polvalente, line 196: We only need to promote f64 to c128, everything else fits in c64 if it's not already complex

polvalente (approved):

josevalim:
Should we implement the gradient as we add these new functions?

PR #710, opened 2022-04-05
[Closed] Add Nx.phase

Benjamin-Philip:
See #693.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -5245,6 +5245,38 @@ defmodule Nx do
   end
 
   @doc """
+  Calculates the complex phase angle of each element in the tensor.
+
+  If $z = a + bi = r e^\\theta$, $phase(z) = {\rm atan2}(b, a)$
+
+  ## Examples
+
+       iex> Nx.phase(Complex.new(1, 2))
+       #Nx.Tensor<
+         c64
+         1.1071487665176392+0.0i
+       >
+
+       iex> Nx.phase(1)
+       #Nx.Tensor<
+         f32
+         0.0
+       >
+
+       iex> Nx.phase(Nx.tensor([Complex.new(1, 2), Complex.new(-2, 1)]))
+       #Nx.Tensor<
+         c64[2]
+         [1.1071487665176392+0.0i, 2.677945137023926+0.0i]
+       >
+  """
+  @doc type: :element
+  def phase(tensor) do
+    tensor = to_tensor(tensor)
+
+    impl!(tensor).phase(%{tensor | type: Nx.Type.to_floating(tensor.type)}, tensor)
+  end
+
+  @doc """
   Applies bitwise not to each element in the tensor.
 
   If you're using `Nx.Defn.defn/2`, you can use the `~~~` operator
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -5253,8 +5253,8 @@ defmodule Nx do
 
        iex> Nx.phase(Complex.new(1, 2))
        #Nx.Tensor<
-         c64
-         1.1071487665176392+0.0i
+         f32
+         1.1071487665176392
        >
 
        iex> Nx.phase(1)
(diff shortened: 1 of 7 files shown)


nx/lib/nx.ex
```diff
@@ -5244,6 +5244,38 @@ defmodule Nx do
...
+         f32
+         0.0
+       >
+
+       iex> Nx.phase(Nx.tensor([Complex.new(1, 2), Complex.new(-2, 1)]))
+       #Nx.Tensor<
+         c64[2]
+         [1.1071487665176392+0.0i, 2.677945137023926+0.0i]
+       >
+  """
```
Benjamin-Philip, line 5271: I feel like this should be a float and not a complex in all cases. Should I add a `Nx.Type.to_float` and set the output tensor type to float?

josevalim: What does EXLA/Numpy return here? It makes me think perhaps we should add support for complex numbers in EXLA before moving forward with adding more functions.

nx/lib/nx.ex
```diff
@@ -5244,6 +5244,38 @@ defmodule Nx do
...
+         1.1071487665176392+0.0i
+       >
+
+       iex> Nx.phase(1)
+       #Nx.Tensor<
+         f32
+         0.0
+       >
+
+       iex> Nx.phase(Nx.tensor([Complex.new(1, 2), Complex.new(-2, 1)]))
```
josevalim, line 5266: We can also start using `~V` and `~M`, since main depends on Elixir v1.13.

```suggestion
       iex> Nx.phase(~V[1+2i -2+1i])
```

WDYT @polvalente?

nx/lib/nx.ex
polvalente: I think we should return the corresponding float here c64 -> f32, c128 -> f64, otherwise f32

This would go for abs/real/imag as well. I imagine them as "taking the components apart", so it makes sense to decompose into the underlying datatype instead of keeping them as a complex with no imaginary part

Benjamin-Philip: > What does EXLA/Numpy return here? 

From what I understand from the docs, np.angle returns a float or an integer.

> It makes me think perhaps we should add support for complex numbers in EXLA before moving forward with adding more functions.

I think it doesn't hurt to add a few more functions. Maybe we could mention in the docs that these functions do not work with EXLA and Torchx?

nx/lib/nx.ex
```diff
@@ -5244,6 +5244,39 @@ defmodule Nx do
...
+       #Nx.Tensor<
+         f32[2]
+         [1.1071487665176392, 2.677945137023926]
+       >
+  """
+  @doc type: :element
+  def phase(tensor) do
+    tensor = to_tensor(tensor)
+
+    impl!(tensor).phase(%{tensor | type: Nx.Type.to_float(tensor.type)}, tensor)
```
seanmor5, line 5277: Looking at `Complex.phase`, could this not be implemented just as:

```suggestion
  def phase(tensor) do
    mask = less(tensor, 0)
    select(mask, Nx.Constants.pi(), atan2(imag(tensor), real(tensor))
  end
```

If so, it would save me from having to write something more verbose for EXLA :)

polvalente: Wouldn't this be atan2(imag, real) for the whole tensor?

x < 0 -> pi only makes sense for real numbers

seanmor5: Yeah actually I see what you mean, so then it could just be `atan2(imag(tensor), real(tensor))`

And as an optional callback instead of with a BinaryBackend impl?

polvalente: Make sense! Optional imolementation could be the way to go

Benjamin-Philip: > If so, it would save me from having to write something more verbose for EXLA :)

You would still need to use the same amount of verbosity for `Nx.real` and `Nx.imag`, as you need to iterate over the tensor. So, if you are doing for real and imag, why not for phase? Also, I think this would be more efficient than mapping with and anonymous function, especially if you're running natively.

However, I don't have any other issues with this approach and I am happy implement it this way.

nx/lib/nx.ex
```diff
@@ -5244,6 +5244,39 @@ defmodule Nx do
     impl!(tensor).conjugate(%{tensor | type: Nx.Type.to_complex(tensor.type)}, tensor)
   end
 
+  @doc """
+  Calculates the complex phase angle of each element in the tensor.
+
+  If $z = a + bi = r e^\\theta$, $phase(z) = {\rm atan2}(b, a)$
```
polvalente, line 5250:
```suggestion
  If $z = a + bi = r e^\\theta$, $phase(z) = {\\rm atan2}(b, a)$
```

What does `\rm` do?

nx/lib/nx.ex
polvalente: The advantage is that we'd only have to make mandatory callbacks (which force us to add grad impls) for real and imag.

You'll see, for instance, that Nx
LinAlg.determinant is a defn for the same reason this one should be, and invert has the perk of having a custom grad (because we didn't want to add the mandatory callback, but found a fast and easy impl for the grad)

polvalente: In short, it's not the worry of verbosity, but of the API surface that all backends need to implement

Benjamin-Philip:
Closing in favour of suggested implementation, and due to merge conflicts.

nx/lib/nx.ex
Benjamin-Philip: `\rm` sets a fixed font: upright, roman, medium weight. I used it because there's no built in \atan2 or \arctan2.
Now that I searched about it, I found that it is deprecated in LaTeX, so I am going to remove it.

PR #711, opened 2022-04-05
Adds support for complex numbers in EXLA

seanmor5:
Requires #709 first :)


josevalim:
@polvalente shall we close the other PR and merge the efforts here? :)

polvalente:
> @polvalente shall we close the other PR and merge the efforts here? :)

Sure! Lets work on this PR instead

seanmor5:
@polvalente I don't have anything more to add for EXLA, not sure what else needs to be done for Nx and Torchx

josevalim (commented):
I dropped one comment and I think we are good to go once it is addressed (given you agree with the proposed API).

nx/lib/nx.ex
```diff
@@ -5244,6 +5244,82 @@ defmodule Nx do
...
+    tensor = to_tensor(tensor)
+
+    type =
+      case tensor.type do
+        {:f, _} = type -> type
+        {:c, size} -> {:f, div(size, 2)}
+        _ -> {:f, 32}
+      end
+
+    impl!(tensor).real(%{tensor | type: type}, tensor)
```
josevalim, line 5282: I am thinking we should not call the implementation at all if the tensor is not complex.

josevalim, line 5282: I am thinking we should specialize this. If the tensor is floating, then we don't need to call the backend at all. If the tensor is complex, then we call the backend. Otherwise it is the equivalent of casting to float (we can use `as_type`).

nx/lib/nx.ex
```diff
@@ -5244,6 +5244,82 @@ defmodule Nx do
...
+        [1.0, 2.0]
+      >
+  """
+  @doc type: :element
+  def real(tensor) do
+    tensor = to_tensor(tensor)
+
+    type =
+      case tensor.type do
+        {:f, _} = type -> type
```
josevalim, line 5277: Use `Nx.Type.floating?` instead so it also handles brainfloats.

nx/lib/nx/binary_backend.ex
```diff
@@ -829,6 +829,26 @@ defmodule Nx.BinaryBackend do
...
   def conjugate(out, tensor), do: element_wise_unary_op(out, tensor, &Complex.conjugate/1)
 
+  @impl true
+  def real(out, tensor) do
+    real_fn = fn
+      %Complex{re: re} -> re
+      num -> num
+    end
+
+    element_wise_unary_op(out, tensor, real_fn)
```
josevalim, line 839: Per my comments on Nx, this should only be called if the tensor is a float and we can actually specialize the implementation: we traverse the binary getting only the real parts and write it down as float. We don't need to create the structs.

nx/lib/nx/binary_backend.ex
```diff
@@ -829,6 +829,26 @@ defmodule Nx.BinaryBackend do
...
+  end
+
+  @impl true
+  def imag(out, tensor) do
+    imag_fn = fn
+      %Complex{im: im} -> im
+      _ -> 0.0
+    end
+
+    element_wise_unary_op(out, tensor, imag_fn)
```
josevalim, line 849: We can specialize this too. For complex numbers, we get the second part of the binary. For all others, we can do a `:binary.duplicate(number_to_binary(0, out.type), size)`.

nx/lib/nx/defn/kernel.ex
```diff
@@ -946,6 +946,11 @@ defmodule Nx.Defn.Kernel do
             "got: #{Macro.to_string(other)}"
   end
 
+  def complex(real, imag \\ 0) do
+    imag_component = Nx.multiply(Nx.tensor(imag), Nx.tensor(Complex.new(0, 1)))
+    Nx.add(real, imag_component)
```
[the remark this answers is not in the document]
polvalente: Actually, perhaps we could just move this do Nx instead of keeping it here in the kernel.

It would be the inverse operation of `real` and `imag` such that: `t == complex(real(t), imag(t))`

WDYT?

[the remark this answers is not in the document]
josevalim: I am afraid this version would be an expensive operation because we need to get the real part, the imaginary part, multiply imaginary part by the constant `i`, and add them together. So I would go with a number only complex/2 Function, Which is also What python offers, and the `i()` constant in Nx.Constants (which people can import).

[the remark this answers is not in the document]
polvalente: It's not that we need to get the real and imaginary parts, but that we have them and want to construct a complex tensor from them. Pytorch allows for such cases:

```python
In [1]: import torch

In [2]: torch.complex(torch.tensor([1.0, 2.0, 3.0]), torch.tensor([4.0, 5.0, 6.0]))
Out[2]: tensor([1.+4.j, 2.+5.j, 3.+6.j])
```

We could either have that as an optional callback with this implementation as the default, or as a callback in itself.

nx/lib/nx/binary_backend.ex
polvalente: I'll rework `real` and `imag` here :)

nx/lib/nx/defn/kernel.ex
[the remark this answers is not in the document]
josevalim: Oh, I see. So what happens in this case:

    torch.complex(torch.tensor([1.0, 2.0, 3.0]), torch.tensor([1.+4.j, 2.+5.j, 3.+6.j]))

?

nx/lib/nx/binary_backend.ex
```diff
@@ -829,6 +829,30 @@ defmodule Nx.BinaryBackend do
   @impl true
   def conjugate(out, tensor), do: element_wise_unary_op(out, tensor, &Complex.conjugate/1)
 
+  @impl true
+  def real(out, %{type: {:c, s}} = tensor) do
+    data = to_binary(tensor)
+
+    result =
+      match_types [{:f, div(s, 2)}] do
+        for <<match!(real, 0), match!(_, 0) <- data>>, into: <<>>, do: real
```
polvalente, line 838: @josevalim I'm getting warnings on "this clause cannot match because of different types/sizes"

Any ideas?

josevalim, line 838: Can you try this?

```suggestion
    s = div(s, 2)

    result =
      for <<real::float-size(s), _::float-size(s) <- data>>, into: <<>>, do: real
```

The warning happens because the compiler knows that it can only be float.

nx/lib/nx/defn/kernel.ex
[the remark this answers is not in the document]
polvalente: torch only accepts floating tensors as torch.complex inputs:

```python
In [2]: import torch

In [3]: real = torch.tensor([1.0, 2.0])

In [4]: complex = torch.complex(real, torch.tensor([3.0, 4.0]))

In [5]: complex
Out[5]: tensor([1.+3.j, 2.+4.j])

In [6]: torch.complex(real, compleX)
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-6-191e657a427d> in <module>
----> 1 torch.complex(real, compleX)

NameError: name 'compleX' is not defined

In [7]: torch.complex(real, complex)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-7-949ebd635540> in <module>
----> 1 torch.complex(real, complex)

RuntimeError: Expected both inputs to be Float or Double tensors but got Float and ComplexFloat

In [8]: torch.complex(real, torch.tensor([3, 4]))
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-8-321f23cbb08d> in <module>
----> 1 torch.complex(real, torch.tensor([3, 4]))

RuntimeError: Expected both inputs to be Float or Double tensors but got Float and Long
```

nx/lib/nx.ex
```diff
@@ -5244,6 +5244,81 @@ defmodule Nx do
...
+        [1.0, 2.0]
+      >
+  """
+  @doc type: :element
+  def real(tensor) do
+    case to_tensor(tensor) do
+      %{type: {:c, size}} = tensor ->
+        impl!(tensor).real(%{tensor | type: {:f, div(size, 2)}}, tensor)
+
+      %{type: {:f, _}} = tensor ->
```
josevalim, line 5277: This will do a conversion for :bf. You need to use `Nx.Type.floating?` instead. Perhaps we need a test.

nx/lib/nx/binary_backend.ex
```diff
@@ -829,6 +829,32 @@ defmodule Nx.BinaryBackend do
...
+        into: <<>>,
+        do: real
+
+    from_binary(out, result)
+  end
+
+  @impl true
+  def imag(out, %{type: {:c, s}} = tensor) do
+    data = to_binary(tensor)
+    component_size = div(s, 2)
```
josevalim, line 848: The correct size is already in `out`. Perhaps we can use it.

nx/lib/nx/defn/compiler.ex
```diff
@@ -58,7 +58,7 @@ defmodule Nx.Defn.Compiler do
             when input: Nx.contained(), output: Nx.contained(), acc: Nx.contained()
 
   # Modules allowed in defn
-  @allowed_modules [Nx, Nx.Constants, Nx.Defn, Nx.Defn.Kernel, Nx.LinAlg]
+  @allowed_modules [Nx, Nx.Constants, Nx.Defn, Nx.Defn.Kernel, Nx.LinAlg, Nx.Defn.Complex]
```
josevalim, line 61: Revert

nx/lib/nx/defn/kernel.ex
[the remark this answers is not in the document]
josevalim: Ok, perfect. Let's go with Nx.complex/2 then. It can accept anything except complex. We still need `Nx.Constants.i()` though.

exla/mix.exs
```diff
@@ -47,6 +47,7 @@ defmodule EXLA.MixProject do
   defp deps do
     [
       {:nx, path: "../nx"},
+      {:complex, "~> 0.3.0", github: "elixir-nx/complex", branch: "main"},
```
josevalim, line 50:
```suggestion
```

josevalim (approved):
Beautiful! Just one comment about removing the complex dependency from EXLA, Which is up to you, and ship it!

exla/mix.exs
polvalente: I we remove this, it will re-lock upon mix deps.get, it seems.

josevalim (approved):

PR #714, opened 2022-04-10
feat(torchx): add complex support for torchx

polvalente:
Adds support for complex numbers in Torchx


torchx/lib/torchx/backend.ex
```diff
@@ -674,8 +703,20 @@ defmodule Torchx.Backend do
 
   ## Ops
 
+  @impl true
+  def atan2(%{type: {:c, _}}, _l, _r) do
+    raise ArithmeticError, "Torchx does not support complex values for atan2"
+  end
+
+  def atan2(out, left, right) do
```
josevalim, line 711: I think you can skip this clause? And still keep it being autogenerated… or do you get a warning?

josevalim (approved):

torchx/lib/torchx/backend.ex
polvalente: Yes, I could skip it! I hadn't even tried because I though it would warn, but it seems that the compiler is nicer when it comes to autogenerated functions.

torchx/lib/torchx/backend.ex
```diff
@@ -719,8 +753,32 @@ defmodule Torchx.Backend do
     end
   end
 
+  @impl true
+  def expm1(%{type: {:c, _}}, _t) do
+    raise ArithmeticError, "Torchx does not support complex values for expm1"
+  end
+
+  def expm1(out, tensor) do
```
josevalim, line 761: You should skip this one and log1p too. :)

PR #716, opened 2022-04-11
Add Nx.phase

Benjamin-Philip:
See #693.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -5269,10 +5269,13 @@ defmodule Nx do
   """
   def phase(tensor) do
     tensor = to_tensor(tensor)
+    output = %{tensor | type: Nx.Type.to_float(tensor.type)}
 
-    tensor
-    |> imag
-    |> atan2(real(tensor))
+    Nx.Shared.optional(:phase, [tensor], output, fn tensor ->
+      tensor
+      |> imag
+      |> atan2(real(tensor))
+    end)
   end
 
   @doc """
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -5272,7 +5272,7 @@ defmodule Nx do
   """
   def phase(tensor) do
     tensor = to_tensor(tensor)
-    output = %{tensor | type: Nx.Type.to_float(tensor.type)}
+    output = %{tensor | type: Nx.Type.to_real(tensor.type)}
 
     Nx.Shared.optional(:phase, [tensor], output, fn tensor ->
       tensor
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -5246,6 +5246,36 @@ defmodule Nx do
   end
 
   @doc """
+  Calculates the complex phase angle of each element in the tensor.
+  If $z = a + bi = r e^\\theta$, $phase(z) = atan2(b, a)$
+
+  ## Examples
+       iex> Nx.phase(Complex.new(1, 2))
+       #Nx.Tensor<
+         f32
+         1.1071487665176392
+       >
+       iex> Nx.phase(1)
+       #Nx.Tensor<
+         f32
+         0.0
+       >
+       iex> import Nx, only: [sigil_V: 2]
+       iex> Nx.phase(~V[1+2i -2+1i])
+       #Nx.Tensor<
+         f32[2]
+         [1.1071487665176392, 2.677945137023926]
+       >
+  """
+  def phase(tensor) do
+    tensor = to_tensor(tensor)
+
+    tensor
+    |> imag
+    |> atan2(real(tensor))
+  end
+
+  @doc """
   Returns the real component of each entry in a complex tensor
   as a floating point tensor.
 
@@ -8657,6 +8687,7 @@ defmodule Nx do
           [4, 3, 4]
         ]
       >
+
 
       iex> t = Nx.tensor([[1, 2], [3, 4]], names: [:x, :y])
       iex> Nx.take(t, Nx.tensor([1, 0, 1]), axis: :y)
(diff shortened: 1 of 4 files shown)


Benjamin-Philip:
I have not added an Optional Callback because I am not sure how you add optional callbacks in Elixir.

nx/lib/nx.ex
```diff
@@ -5245,6 +5245,38 @@ defmodule Nx do
     impl!(tensor).conjugate(%{tensor | type: Nx.Type.to_complex(tensor.type)}, tensor)
   end
 
+  @doc """
+  Calculates the complex phase angle of each element in the tensor.
+  If $z = a + bi = r e^\\theta$, $phase(z) = atan2(b, a)$
```
polvalente, line 5250:
```suggestion
  $phase(z) = atan2(b, a), z = a + bi \\in \\Complex$
```

nx/lib/nx.ex
```diff
@@ -5245,6 +5245,38 @@ defmodule Nx do
...
+
+  ## Examples
```
polvalente, line 5252:
```suggestion
  ### Examples

```

nx/lib/nx.ex
```diff
@@ -5245,6 +5245,38 @@ defmodule Nx do
...
+       iex> Nx.phase(Complex.new(1, 2))
+       #Nx.Tensor<
+         f32
+         1.1071487665176392
+       >
+       iex> Nx.phase(1)
+       #Nx.Tensor<
+         f32
+         0.0
+       >
```
polvalente, line 5262:
```suggestion
       >

```

nx/lib/nx.ex
```diff
@@ -5245,6 +5245,38 @@ defmodule Nx do
...
+       iex> import Nx, only: [sigil_V: 2]
+       iex> Nx.phase(~V[1+2i -2+1i])
+       #Nx.Tensor<
+         f32[2]
+         [1.1071487665176392, 2.677945137023926]
+       >
+  """
+  def phase(tensor) do
```
polvalente, line 5270:
```suggestion
  defn phase(tensor) do
```

You'll need to import this

torchx/test/torchx/nx_doctest_test.exs
```diff
@@ -16,8 +16,9 @@ defmodule Torchx.NxDoctestTest do
   @temporarily_broken_doctests [
     # broadcast - shape mismatch in one test
     broadcast: 3,
-    # conjugate, real, imag - Torchx does not support complex numbers
+    # conjugate, phase, real, imag - Torchx does not support complex numbers
     conjugate: 1,
+    phase: 1,
```
polvalente, line 21: This shouldn't be skipped here

Benjamin-Philip: Where should it be skipped then? `phase` depends on `real` and `imag` and thus won't work without them..

nx/lib/nx/type.ex
```diff
@@ -195,6 +195,29 @@ defmodule Nx.Type do
   def to_complex({:f, 64}), do: {:c, 128}
   def to_complex(_type), do: {:c, 64}
 
+  @doc """
+  Converts the given type to a float with the minimum size
+  necessary.
```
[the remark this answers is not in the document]
Benjamin-Philip: I didn't do that as `to_floating` and `to_complex` both have only one doctest.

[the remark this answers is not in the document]
polvalente: I would split the in this PR, but I'd wait on José to respond. Perhaps there's a reason I'm not seeing for them to be a single example/test

nx/lib/nx/type.ex
```diff
@@ -195,6 +195,29 @@ defmodule Nx.Type do
...
+  ## Examples
```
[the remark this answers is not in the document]
Benjamin-Philip: I also noticed that we are using level 2 headers for `to_{complex, floating}`. So I undid these changes.

nx/lib/nx.ex
Benjamin-Philip: I undid this too as the other functions under Nx use level 2 headers

nx/lib/nx.ex
Benjamin-Philip: I don't understand why we are making an exception for `phase` and using defn?

nx/lib/nx/type.ex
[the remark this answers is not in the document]
josevalim: It is generally level 2 headers indeed.

nx/lib/nx/type.ex
```diff
@@ -195,6 +195,31 @@ defmodule Nx.Type do
...
+      iex> Nx.Type.to_float({:c, 128})
+      {:f, 64}
+      iex> Nx.Type.to_float({:f, 32})
+      {:f, 32}
+      iex> Nx.Type.to_float({:f, 64})
+      {:f, 64}
+  """
+  def to_float({:f, size}), do: {:f, size}
+  def to_float({:c, s}), do: {:f, div(s, 2)}
+  def to_float(_type), do: {:f, 32}
```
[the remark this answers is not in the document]
josevalim: My suggestion is to call this function `to_real`. Because it is all about going from complex to something in the real domain.

nx/lib/nx/type.ex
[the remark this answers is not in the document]
Benjamin-Philip: I think we can resolve this in that case.

nx/lib/nx/type.ex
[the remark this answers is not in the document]
Benjamin-Philip: > This is going to convert bf16 to f32, and it should not.
> My suggestion is to call this function to_real. Because it is all about going from complex to something in the real domain.

My inital plan was to convert this to float regardless of the previous type, but since what we are trying to do is achieve an acceptable representation for real numbers, I think we can rename this to `to_real` like José suggested.

josevalim (approved):

Issue #692, opened 2022-03-24
Support complex numbers in Nx.LinAlg


polvalente:
This is an umbrella issue to cover the implementation of complex numbers for the following functions (and grads where applicable):

Grads are comprised in #695

- [x] cholesky
- [x] norm
- [x] triangular_solve
- [x] solve
- [x] invert
- [x] qr
- [ ] eigh
- [ ] svd
- [x] lu
- [x] matrix_power
- [x] determinant

polvalente:
Closing in favor of #792 and #793

PR #719, opened 2022-04-14
feat(complex): add QR decomposition support

polvalente:
Related to #692

### DIFF
--- a/nx/test/nx/lin_alg_test.exs
+++ b/nx/test/nx/lin_alg_test.exs
@@ -159,10 +159,36 @@ defmodule Nx.LinAlgTest do
       assert_all_close(Nx.dot(q, r), t, atol: 1.0e-10)
     end
 
+    test "works with complex matrix" do
+      import Nx, only: :sigils
+
+      t = ~M[
+        1 0 1i
+        0 2 -1i
+        1 1 1
+      ]
+
+      {q, r} = Nx.LinAlg.qr(t)
+
+      assert_all_close(q, ~M[
+        -0.7070 0.2357  0.6666
+         0      -0.9428   0.3333
+        -0.7070 -0.2357  -0.6666
+      ])
+
+      assert_all_close(r, ~M[
+        -1.4142 -0.7071 -0.7071-0.7071i
+        0      -2.1213  -0.2357+1.1785i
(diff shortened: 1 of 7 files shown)


nx/test/nx/lin_alg_test.exs
```diff
@@ -159,10 +159,36 @@ defmodule Nx.LinAlgTest do
       assert_all_close(Nx.dot(q, r), t, atol: 1.0e-10)
     end
 
+    test "works with complex matrix" do
+      import Nx, only: :sigils
```
polvalente, line 163: is it ok if I import this at the top of the module?

nx/test/nx/lin_alg_test.exs
```diff
@@ -159,10 +159,36 @@ defmodule Nx.LinAlgTest do
...
+        0 2 -1i
+        1 1 1
+      ]
+
+      {q, r} = Nx.LinAlg.qr(t)
+
+      assert_all_close(q, ~M[
+        -0.7070 0.2357  0.6666
+         0      -0.9428   0.3333
+        -0.7070 -0.2357  -0.6666
```
polvalente, line 176: During tests, I needed to copy and paste complex tensors from the terminal but couldn't just paste them into the sigil due to the brackets for each row.

Perhaps we can allow for that through a regex? Something like:

String.replace(string, ~r/\[([^\]+)\]/, "\\1")

nx/test/nx/lin_alg_test.exs
josevalim: Totally!

nx/test/nx/lin_alg_test.exs
josevalim: Another option is to print the tensor differently. Using a similar format as ~M. Higher dimensional tensors are printed as "lists" of matrices.

josevalim (approved):

nx/test/nx/lin_alg_test.exs
polvalente: So we would print all tensors like this every time?

josevalim: I don't know, somethjing to discuss :)

PR #732, opened 2022-04-17
feat(grad): support QR for complex tensors

polvalente:
#695


nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -203,8 +203,11 @@ defmodule Nx.BinaryBackend.Matrix do
 
   defp qr_decomposition(matrix, m, n, eps) when m >= n do
     # QR decomposition is performed by using Householder transform
+
+    max_i = if m == n, do: n - 2, else: n - 1
```
polvalente, line 207: This makes QR yield the same results as Jax for square matrices :)
The results before were mathematically sound, but when comparing more complex grads, things would become more difficult

josevalim (approved):

PR #735, opened 2022-04-19
Add complex tests for aggregate functions

Benjamin-Philip:
Add complex tests for aggregate functions. See #695.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -6310,10 +6310,10 @@ defmodule Nx do
         1
       >
 
-     iex> Nx.argmax(Nx.tensor([Complex.new(1, 2), Complex.new(3,2)]))
+     iex> Nx.argmax(Nx.tensor([Complex.new(1, 2), Complex.new(3, 0)]))
      #Nx.Tensor<
        s64
-       1
+       0
      >
 
 
@@ -6429,10 +6429,10 @@ defmodule Nx do
         0
       >
 
-      iex> Nx.argmin(Nx.tensor([Complex.new(1, 2), Complex.new(3,2)]))
+      iex> Nx.argmin(Nx.tensor([Complex.new(1, 2), Complex.new(3, 0)]))
       #Nx.Tensor<
         s64
-        0
+        1
       >
 
   ### Aggregating over an axis
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -6302,13 +6302,20 @@ defmodule Nx do
         10
       >
 
-  If a tensor of floats is given, it still returns integers:
+  If a tensor of floats or complex numbers is given, it still returns integers:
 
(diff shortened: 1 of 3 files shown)


nx/lib/nx.ex
```diff
@@ -6302,14 +6302,21 @@ defmodule Nx do
...
-  If a tensor of floats is given, it still returns integers:
+  If a tensor of floats or complex numbers is given, it still returns integers:
 
       iex> Nx.argmax(Nx.tensor([2.0, 4.0]))
       #Nx.Tensor<
         s64
         1
       >
 
+     iex> Nx.argmax(Nx.tensor([Complex.new(1, 2), Complex.new(3,2)]))
```
polvalente, line 6313:
```suggestion
     iex> Nx.argmax(Nx.tensor([Complex.new(1, 2), Complex.new(3,0)]))
```

I'm curious on what this would return. Also, if this is considered a tie

nx/lib/nx.ex
```diff
@@ -6414,14 +6421,20 @@ defmodule Nx do
...
-  If a tensor of floats is given, it still returns integers:
+  If a tensor of floats or complex numbers is given, it still returns integers:
 
       iex> Nx.argmin(Nx.tensor([2.0, 4.0]))
       #Nx.Tensor<
         s64
         0
       >
 
+      iex> Nx.argmin(Nx.tensor([Complex.new(1, 2), Complex.new(3,2)]))
```
polvalente, line 6432: Using the same example as I suggested in argmax here would be nice

nx/lib/nx.ex
Benjamin-Philip: On second thoughts, this should actually error out, as we can't really compare 2 complex numbers.

polvalente: Jax and Pytorch have a defined behavior for this. Perhaps ensure that we have the same behavior

nx/lib/nx.ex
```diff
@@ -6302,14 +6302,21 @@ defmodule Nx do
...
       iex> Nx.argmax(Nx.tensor([2.0, 4.0]))
       #Nx.Tensor<
         s64
         1
       >
 
+     iex> Nx.argmax(Nx.tensor([Complex.new(1, 2), Complex.new(3, 0)]))
+     #Nx.Tensor<
+       s64
+       0
```
polvalente, line 6316: This is not exactly what I was expecting, but at least it's consistent with what Nx.greater returns!

If both components are bigger, the comomex is bigger, otherwise it is not

polvalente (approved):

Benjamin-Philip:
@polvalente, Should I disable the argmax and argmin tests for torchx?

polvalente:
> @polvalente, Should I disable the argmax and argmin tests for torchx?

Yeah, add them to the "cannot support" list

polvalente (approved):

PR #427, opened 2021-06-19
Allow EXLA outside of defn

seanmor5:

PR #739, opened 2022-04-21
Add EXLA backend

jonatanklosko:
Builds on top of #427.


nx/lib/nx/binary_backend.ex
```diff
@@ -718,6 +718,11 @@ defmodule Nx.BinaryBackend do
   defp element_remainder(_, a, b), do: :math.fmod(a, b)
 
   defp element_atan2(_, a, b), do: Complex.atan2(a, b)
+  # TODO: we need to handle :infinity, :neg_infinity and :nan in all other operations
+  defp element_max(_, :infinity, _b), do: :infinity
+  defp element_max(_, _a, :infinity), do: :infinity
+  defp element_max(_, :neg_infinity, b), do: b
+  defp element_max(_, a, :neg_infinity), do: a
```
jonatanklosko, line 725: This came up in a window_max test, however it's a more general problem we need to handle for all the other operations in the binary backend.

exla/lib/exla/backend.ex
```diff
@@ -0,0 +1,192 @@
...
+  unary_ops =
+    [:exp, :expm1, :log, :log1p, :logistic, :cos, :sin, :tan] ++
+      [:cosh, :sinh, :tanh, :acos, :asin, :atan, :acosh, :asinh, :atanh] ++
+      [:sqrt, :rsqrt, :cbrt, :erf, :erfc, :erf_inv] ++
+      [:abs, :bitwise_not, :ceil, :conjugate, :floor, :negate, :round, :sign] ++
+      [:count_leading_zeros, :population_count, :real, :imag]
+
+  callbacks =
+    [
+      {:as_type, [:out, :tensor], [:tensor]},
```
jonatanklosko, line 130: We could reduce this to `{:as_type, [1]}}`, however I think that would be much more error prone.

exla/test/exla/backend_test.exs
```diff
@@ -0,0 +1,10 @@
+defmodule EXLA.BackendTest do
+  use ExUnit.Case, async: true
+
+  setup do
+    Nx.default_backend(EXLA.Backend)
+    :ok
+  end
+
+  doctest Nx, except: [compatible?: 2, default_backend: 1]
+end
```
josevalim, line 10:
```suggestion
end

```

nx/lib/nx/binary_backend.ex
josevalim: Yup, let's implement this for min and max for now. We will handle it for sure in other places.

exla/lib/exla/backend.ex
```diff
@@ -0,0 +1,192 @@
...
+    from_binary(out, data)
+  end
+
+  defp number_to_binary(number, type),
+    do: match_types([type], do: <<write!(number, 0)>>)
+
+  @impl true
+  def from_binary(t, binary, _backend_options), do: from_binary(t, binary)
+
+  defp from_binary(t, binary) when is_binary(binary), do: %{t | data: %B{state: binary}}
```
josevalim, line 45: The state for this should never a binary. It should always be a reference to something on the EXLA device. Please check the previous implementation :)

exla/lib/exla/backend.ex
```diff
@@ -0,0 +1,268 @@
...
+  @impl true
+  def backend_deallocate(%T{data: %B{buffer: buffer}}) do
+    EXLA.DeviceBuffer.deallocate(buffer)
+  end
+
+  @impl true
+  def to_batched_list(out, tensor, opts) do
+    binary_tensor = backend_copy(tensor, Nx.BinaryBackend, [])
+    Nx.BinaryBackend.to_batched_list(out, binary_tensor, opts)
+  end
```
jonatanklosko, line 78: Is there any issue with this approach?

jonatanklosko:
@josevalim the remaining tests should be fixed by defaulting `:keep_on_device` to `true`, because some doctests use `broadcast` etc to prepare arguments and it returns binary backend.

On a related note, if `:keep_on_device` if `false`, maybe we should transfer to the default backend, rather than BinaryBackend?

exla/lib/exla/backend.ex
```diff
@@ -0,0 +1,268 @@
...
+  """
+
+  @behaviour Nx.Backend
+  @enforce_keys [:buffer]
+  defstruct [:buffer]
+
+  alias Nx.Tensor, as: T
+  alias EXLA.Backend, as: B
+
+  import Nx.Shared
```
josevalim, line 27: This is private Nx API and we should not rely on it. Maybe you can try calling `Nx.BinaryBackend.constant() |> Nx.to_binary()` or something like that?

josevalim: Or create on binary backend and transfer.

exla/lib/exla/backend.ex
josevalim: I think so. I don't think we want to convert to a list of binarybackend tensors. If possible we would still keep them on the device... but I am not sure how easy it would be to implement it. So my suggestion is to raise on this function for now? 🤔

exla/lib/exla/backend.ex
```diff
@@ -0,0 +1,268 @@
...
+    Nx.BinaryBackend.to_batched_list(out, binary_tensor, opts)
+  end
+
+  @impl true
+  def to_binary(%T{data: %B{buffer: buffer}, type: {_, size}}, limit) do
+    EXLA.DeviceBuffer.read(buffer, limit * div(size, 8))
+  end
+
+  @impl true
+  def inspect(%T{} = tensor, inspect_opts) do
```
josevalim, line 86: Please update to the inspect implementation of this PR: https://github.com/elixir-nx/nx/pull/740 :D

exla/lib/exla/backend.ex
```diff
@@ -0,0 +1,268 @@
...
+        Nx.Backend.inspect(tensor, binary, inspect_opts)
+      else
+        "Tensors on the GPU cannot be inspected. Explicitly transfer the tensor by calling Nx.backend_transfer/1"
+      end
+
+    maybe_add_signature(result, tensor)
+  end
+
+  # TODO: Elixir v1.13 has a default_inspect_fun which
+  # we can use to customize this behaviour for tests.
```
josevalim, line 99:
```suggestion
```

exla/lib/exla/backend.ex
```diff
@@ -0,0 +1,268 @@
...
+  defp same_client_device?(buffer, opts) do
+    {client, device_id} = client_and_device_id(opts)
+    buffer.client_name == client.name and buffer.device_id == device_id
+  end
+
+  defp platform(%T{data: %B{buffer: buffer}}) do
+    client = EXLA.Client.fetch!(buffer.client_name)
+    client.platform
+  end
+
```
josevalim, line 140:
```suggestion

  ## JIT callbacks

```

exla/lib/exla/backend.ex
```diff
@@ -0,0 +1,268 @@
...
+  @impl true
+  def eye(%{shape: shape, type: type, names: names}, _) do
+    fun = fn ->
+      Nx.eye(shape, type: type, names: names, backend: Nx.Defn.Expr)
+    end
+
+    EXLA.jit(fun, [])
+  end
```
josevalim, line 148: This variation wouldn't work?

```suggestion
  def eye(out, tensor) do
    fun = fn -> Nx.Defn.Expr.eye(out, tensor) end
    EXLA.jit(fun, [])
  end
```

The same below...

exla/lib/exla/backend.ex
```diff
@@ -0,0 +1,268 @@
...
+      Nx.random_normal(shape, mu, sigma, type: type, names: names, backend: Nx.Defn.Expr)
+    end
+
+    EXLA.jit(fun, [])
+  end
+
+  @impl true
+  def concatenate(out, tensors, axis) do
+    expr_fn = fn tensors ->
+      apply(Nx.Defn.Expr, :concatenate, [out, Tuple.to_list(tensors), axis])
```
josevalim, line 180:
```suggestion
      Nx.Defn.Expr.concatenate(out, Tuple.to_list(tensors), axis)
```

exla/lib/exla/backend.ex
jonatanklosko: Ohhh, I mirrored the local version of that, thanks! :see_no_evil:

exla/lib/exla/backend.ex
jonatanklosko: I don't see a reason why it wouldn't, and if so, we can just generate it as all the others. Also for `random_uniform` and `random_normal` I think we want to pass the `min`, `max`, `sigma`, `mu` through JIT arguments. @seanmor5 please let us know if we miss something :D

exla/lib/exla/backend.ex
jonatanklosko: @josevalim in `Nx.Backend.inspect/2` the docs say

https://github.com/elixir-nx/nx/blob/bd4c499f95e23ae7d76bee08b5f678c719a6644b/nx/lib/nx/backend.ex#L148-L150

Should we `+ 1` or is it no longer relevant?

exla/lib/exla/backend.ex
jonatanklosko: Did `Nx.BinaryBackend.constant` + `Nx.BinaryBackend.backend_transfer`. Not sure if it's better to call `backend_transfer` on Nx or directly on the backend, but the latter seems more consistent given we call `constant` directly.

exla/lib/exla/backend.ex
jonatanklosko: My initial thought was that we could generate an expr that returns a tuple of slices, but with this approach the computation graph would grow with data size or batch granularity, right?

exla/lib/exla/backend.ex
```diff
@@ -0,0 +1,224 @@
...
+
+  defp same_client_device?(buffer, opts) do
+    {client, device_id} = client_and_device_id(opts)
+    buffer.client_name == client.name and buffer.device_id == device_id
+  end
+
+  ## JIT callbacks
+
+  # TODO: remove keep_on_device entirely
+  @jit_opts [run_options: [keep_on_device: true]]
```
jonatanklosko, line 127: Better to drop it in a separate PR because it involves a lot of changes.

exla/lib/exla/backend.ex
josevalim: Good find. We need to +1.

josevalim (approved):

josevalim:

PR #740, opened 2022-04-21
Inspect torchx on all environments

josevalim:


torchx/lib/torchx/backend.ex
```diff
@@ -1293,21 +1293,20 @@ defmodule Torchx.Backend do
     Enum.reduce(padding, [], fn {a, b}, acc -> [a, b | acc] end)
   end
 
+
```
josevalim, line 1296:
```suggestion
```

josevalim:
@polvalente can you pls check if this is good on the GPU and if so please merge?

torchx/lib/torchx/backend.ex
```diff
@@ -1295,19 +1294,12 @@ defmodule Torchx.Backend do
...
-      if device?(tensor, :cpu) do
-        binary = Torchx.to_blob(from_nx(tensor))
-        Nx.Backend.inspect(tensor, binary, inspect_opts)
-      else
-        "Tensors on the GPU cannot be inspected. Explicitly transfer the tensor by calling Nx.backend_transfer/1"
-      end
-
-    maybe_add_signature(result, tensor)
+    tensor
+    |> to_binary(min(inspect_opts.limit, Nx.size(tensor)))
```
josevalim, line 1298:
```suggestion
    limit = if inspect_opts.limit == :infinity, do: :infinity, else: inspect_opts.limit + 1

    tensor
    |> to_binary(min(limit, Nx.size(tensor)))
```

torchx/lib/torchx/backend.ex
```diff
@@ -1295,19 +1294,14 @@ defmodule Torchx.Backend do
...
-        binary = Torchx.to_blob(from_nx(tensor))
-        Nx.Backend.inspect(tensor, binary, inspect_opts)
-      else
-        "Tensors on the GPU cannot be inspected. Explicitly transfer the tensor by calling Nx.backend_transfer/1"
-      end
+    limit = if inspect_opts.limit == :infinity, do: :infinity, else: inspect_opts.limit + 1
 
-    maybe_add_signature(result, tensor)
+    tensor
+    |> to_binary(min(limit, Nx.size(tensor)))
```
polvalente, line 1300: @josevalim I'm getting a segfault whenever the tensor is transferred to cuda:

```elixir
iex(1)> Torchx.device_available? :cuda
true
iex(2)> t = Nx.iota {3}
#Nx.Tensor<
  s64[3]
  Torchx.Backend(cpu)
  [0, 1, 2]
>
iex(3)> t |> Torchx.from_nx
from_nx/1    
iex(3)> t |> Torchx.from_nx()
{:cpu, #Reference<0.793727343.3884580880.186395>}
iex(4)> t |> Torchx.from_nx() |> Torchx.to_device(:cuda)
{:cuda, #Reference<0.793727343.3884580886.186862>}
iex(5)> t |> Torchx.from_nx() |> Torchx.to_device(:cuda) |> Torchx.to_nx
[1]    22011 segmentation fault (core dumped)  iex -S mix
```

and

```elixir
iex(1)> t = Nx.iota({3}, backend: {Torchx.Backend, device: :cuda}); nil
nil
iex(2)> t = Nx.iota({3}, backend: {Torchx.Backend, device: :cuda})     
[1]    23015 segmentation fault (core dumped)  iex -S mix
```

polvalente: This also happens on main, though, on `to_binary`:

```elixir
iex(1)> t = Nx.iota({3}, backend: {Torchx.Backend, device: :cuda}); nil
nil
iex(2)> t
#Nx.Tensor<
  s64[3]
  Torchx.Backend(cuda)
  Tensors on the GPU cannot be inspected. Explicitly transfer the tensor by calling Nx.backend_transfer/1
>       
iex(3)> t |> Nx.to_binary
[1]    24818 segmentation fault (core dumped)  iex -S mix
```

josevalim: Hrm... is to_blob invalid for GPU devices then? Maybe we need to transfer them to the CPU first? Although I thought the C implementation would take care of that.

torchx/c_src/torchx.cpp
```diff
@@ -277,12 +277,18 @@ NIF(to_blob)
...
+  else if (device.has_value() && device.value().type() == torch::kCPU)
   {
     void *result_data = (void *)enif_make_new_binary(env, byte_size, &result);
     memcpy(result_data, data_ptr, byte_size);
     return nx::nif::ok(env, result);
   }
+  else
+  {
+    void *result_data = (void *)enif_make_new_binary(env, byte_size, &result);
+    memcpy(result_data, reshaped.to(torch::kCPU).data_ptr(), byte_size);
```
josevalim, line 292: This will transfer everything from the GPU, even if we are going to read only part of it. What if we trim down the tensor and make sure it has at most `limit` elements? We should also probably raise if `limit` is more than the number of elements?

polvalente: We could take a slice of the tensor given the limit. I wouldn't raise if limit is more than the number of elements, though, because it is just an upper bound (for instance, `limit: :infinity` wouldn't make sense if we only accepted a limit within the tensor bounds)

polvalente (approved):

PR #741, opened 2022-04-22
Add absolute value test for complex numbers

Benjamin-Philip:
See #695

### DIFF
--- a/nx/test/nx/complex_test.exs
+++ b/nx/test/nx/complex_test.exs
@@ -11,6 +11,10 @@ defmodule Nx.ComplexTest do
   @arg2 Complex.new(-2, 7)
 
   describe "unary operations" do
+    test "abs" do
+      assert_all_close(Nx.abs(@arg), 3.605551275)
+    end
+
     test "exp" do
       assert_all_close(Nx.exp(@arg), Complex.new(-7.315, 1.042))
     end


nx/test/nx/complex_test.exs
```diff
@@ -11,6 +11,10 @@ defmodule Nx.ComplexTest do
   @arg2 Complex.new(-2, 7)
 
   describe "unary operations" do
+    test "abs" do
+      assert_all_close(Nx.abs(@arg), 3.605551275)
```
polvalente, line 15: #695 is about ensuring the current grad implementations work (or providing fixes to them). Do you want to take a stab at that for abs?

Benjamin-Philip: Sure! I'll see what I can do.

polvalente: It might help to use Jax in Python for that. Something along the lines of:

```python
import jax
import jax.numpy as jno

jax.grad(lambda t: jnp.sum(jnp.abs(t)), holomorphic=True)(jnp.array(...))
```

This will give you a reference value for the test :)

polvalente: @Benjamin-Philip I ended up implementing this grad because I forgot that you had this PR open :(
I'm sorry about this!

If you want to take a look, it's on #755.
`put_slice`, `window_sum`, `reduce`, `window_reduce` and `pad` should all only need tests. Do you want to take those instead?

polvalente: If you want to tackle a grad implementation in itself, there's cholesky waiting there. Conv and bitcast haven't been properly implemented for complex yet.

Benjamin-Philip: > I ended up implementing this grad because I forgot that you had this PR open :(
I'm sorry about this!

Not a problem! I was actually quite busy this weekend, so I couldn't work on this PR. I'll have a look at your suggestion.

PR #743, opened 2022-04-22
Always keep data on the EXLA device after computation

jonatanklosko:
Removes the `:keep_on_device` option and effectively makes it the default behaviour.

Context: the `EXLA.Backend` (#739) allows for using regular Nx operations on a device tensor, so implicit transfer to the CPU doesn't really sense.

### DIFF
--- a/exla/test/support/exla_case.ex
+++ b/exla/test/support/exla_case.ex
@@ -1,0 +1,46 @@
+defmodule EXLA.Case do
+  @moduledoc """
+  Test case for tensor assertions
+  """
+
+  use ExUnit.CaseTemplate
+
+  using do
+    quote do
+      import EXLA.Case
+    end
+  end
+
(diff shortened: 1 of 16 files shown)


exla/test/support/exla_case.ex
```diff
@@ -0,0 +1,46 @@
...
+      import EXLA.Case
+    end
+  end
+
+  defmacro assert_equal(left, right) do
+    # Assert against binary backend tensors to show diff on failure
+    quote do
+      assert unquote(left) |> to_binary_backend() == unquote(right) |> to_binary_backend()
+    end
+  end
```
jonatanklosko, line 19: This makes the `code:` (in the error message) a bit obfuscated, but still more detailed than a simple exception. The important part is that we show the diff.

josevalim (approved):

josevalim:

PR #749, opened 2022-04-25
Raise a better error message for undefined functions in defn

### LINKED ISSUE #730: Raise better error message for undefined functions in defn
For local functions in the AST (the first element of the tuple is an atom and the third argument is a list) which are invalid, we should say "undefined function NAME/ARITY" in the error message.

jonatanklosko:
Closes #730.

### DIFF
--- a/nx/test/nx/defn_test.exs
+++ b/nx/test/nx/defn_test.exs
@@ -1319,6 +1319,18 @@ defmodule Nx.DefnTest do
       end
     end
 
+    test "undefined local function" do
+      assert_raise CompileError, ~r"#{location(+5)}: undefined function do_add/2 \(there is no such import\)", fn ->
+        defmodule Sample do
+          import Nx.Defn
+
+          defn add(a, b) do
+            do_add(a, b)
+          end
+        end
+      end
+    end
+
     test "non-variables used as arguments" do
       assert_raise CompileError,
                    ~r"#{location(+4)}: only variables, tuples, maps, and structs are allowed as patterns in defn",
--- a/nx/test/nx/defn_test.exs
+++ b/nx/test/nx/defn_test.exs
@@ -1321,7 +1321,7 @@ defmodule Nx.DefnTest do
 
     test "undefined local function" do
       assert_raise CompileError,
-                   ~r"#{location(+5)}: undefined function do_add/2 \(there is no such import\)",
+                   ~r"#{location(+6)}: undefined function do_add/2 \(there is no such import\)",
                    fn ->
                      defmodule Sample do
                        import Nx.Defn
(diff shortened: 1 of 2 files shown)


jonatanklosko:
The test

```elixir
    test "invalid numerical expression" do
      assert_raise CompileError, ~r"#{location(+5)}: invalid numerical expression", fn ->
        defmodule Sample do
          import Nx.Defn

          defn add(_a, _b) do
            receive do
              :ok -> :ok
            end
          end
        end
      end
    end
```

now fails with the new error:

```
test/nx/defn_test.exs:1314: undefined function receive/1 (there is no such import)
```

Is there any other "invalid numerical expression" that doesn't fall into that?

nx/test/nx/defn_test.exs
```diff
@@ -1319,6 +1319,18 @@ defmodule Nx.DefnTest do
...
     end
 
+    test "undefined local function" do
+      assert_raise CompileError, ~r"#{location(+5)}: undefined function do_add/2 \(there is no such import\)", fn ->
+        defmodule Sample do
+          import Nx.Defn
+
+          defn add(a, b) do
+            do_add(a, b)
+          end
```
polvalente, line 1329: What happens if you call a function which is defined but isnnot a defn?

jonatanklosko: Good call, it will be treated as undefined, I will add another error message for it.

polvalente (approved):

josevalim:
@jonatanklosko it seems tests are failing on CI? 🤔

jonatanklosko:
@josevalim yeah see my comment above, we either need to change the test case to something else or remove it. I just can't think of any other syntax that falls under this error :)

nx/test/nx/defn_test.exs
```diff
@@ -1319,6 +1319,36 @@ defmodule Nx.DefnTest do
...
+                   fn ->
+                     defmodule Sample do
+                       import Nx.Defn
+
+                       defn add(a, b) do
+                         do_add(a, b)
+                       end
+
+                       defp do_add(a, b), do: a + b
+                     end
```
polvalente, line 1348: @jonatanklosko perhaps this needs to be a string called with Code.eval_string or something like that for tests to pass?

jonatanklosko: Ohh, I see the other two tests fail now as I run the formatter!

jonatanklosko: We assert line relatively and formatting just moved it around. I've just fixed it. The remaining error is what I mentioned above.

josevalim:

PR #750, opened 2022-04-25
Add support for atom type shorthand

### LINKED ISSUE #742: Support atoms in :type
We should support :u8, :s16, and so on. We already have `Nx.Type.normalize!`, which is an entrypoint built exactly to support conveniences like these, and it should normalize them into `{:u, size}` and so on. And of course, we should update the docs.

jonatanklosko:
Closes #742.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7353,7 +7353,7 @@ defmodule Nx do
   def map(tensor, opts \\ [], fun) do
     %T{type: type} = tensor = to_tensor(tensor)
     opts = keyword!(opts, type: type)
-    output_type = opts[:type]
+    output_type = Nx.Type.normalize!(opts[:type])
     out = %{tensor | type: output_type}
     impl!(tensor).map(out, tensor, opts, fun)
   end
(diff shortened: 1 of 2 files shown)


nx/lib/nx.ex
```diff
@@ -7353,7 +7353,7 @@ defmodule Nx do
   def map(tensor, opts \\ [], fun) do
     %T{type: type} = tensor = to_tensor(tensor)
     opts = keyword!(opts, type: type)
-    output_type = opts[:type]
```
polvalente, line 7356: Did you take a look at things like `case tensor.type do` as well?

jonatanklosko: `tensor.type` should always be in the canonical form, because we always normalize in the creation functions, no?

polvalente: I've done a pass through this file and it seems everything is ok, because all tensor.type occurrences happen after a `to_tensor` call :)

polvalente (approved):

nx/lib/nx.ex
jonatanklosko: To be clear, my point was that we should never encounter `%Nx.Tensor{type: :f32}` (which is reflected in the typespec), if we do it's a bug elsewhere. As far as I understand we only want the short syntax to abbreviate options :)

josevalim:

nx/lib/nx.ex
josevalim: Correct!

PR #751, opened 2022-04-25
Raise a more specific error for tensors in a shape tuple

### LINKED ISSUE #668: `Nx.runtime_normal/4` throws an error when used inside defn and the shape is set via tuple
Hey all 👋 
first of all, many many thanks for the amazing work you all are doing with this library and the related ones.
I'm newbie in ML and I'm having fun learning it by playing with Nx.

## The issue

`Nx.random_normal/4` throws an error when used inside a numerical function `defn` and the shape is set via tuple (eg. `Nx.random_normal({10, 5})`). I'm using EXLA as backend.

## My system

I'm tested it on macOS X 11.6.3.

Erlang and Elixir versions below 👇 

```
❯ iex
Erlang/OTP 24 [erts-12.1.5] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Interactive Elixir (1.13.0) - press Ctrl+C to exit (type h() ENTER for help)
```

## How to reproduce

```elixir
Mix.install([
  {:exla, "~> 0.1.0-dev", github: "elixir-nx/nx", sparse: "exla"},
  {:nx, "~> 0.1.0-dev", github: "elixir-nx/nx", sparse: "nx", override: true}
])

Nx.Defn.global_default_options(compiler: EXLA)

defmodule RandomNormalTest do
  import Nx.Defn

  defn create_with_defn(shape) do
    Nx.random_normal(shape, 0.0, 1.0)
  end
end
```

```elixir
RandomNormalTest.create_with_defn({10, 4})
```

Throws this error:

```
** (ArgumentError) invalid dimension in axis 1 in shape. Each dimension must be a positive integer, got #Nx.Tensor<
  s64
  
  Nx.Defn.Expr
  parameter a:1   s64
> in shape {#Nx.Tensor<
   s64
   
   Nx.Defn.Expr
   parameter a:0   s64
 >, #Nx.Tensor<
   s64
   
   Nx.Defn.Expr
   parameter a:1   s64
 >}
    (nx 0.1.0) lib/nx/shape.ex:35: Nx.Shape.validate!/3
    (nx 0.1.0) lib/nx.ex:905: Nx.random_normal/4
    (nx 0.1.0) lib/nx/defn/compiler.ex:104: Nx.Defn.Compiler.runtime_fun/4
    (exla 0.1.0-dev) lib/exla/defn.ex:299: anonymous fn/2 in EXLA.Defn.compile/7
    (exla 0.1.0-dev) lib/exla/defn/locked_cache.ex:40: EXLA.Defn.LockedCache.run/2
    (exla 0.1.0-dev) lib/exla/defn.ex:298: EXLA.Defn.compile/7
    (exla 0.1.0-dev) lib/exla/defn.ex:225: EXLA.Defn.__jit__/4
```

It works fine if, instead of using a tuple to set the shape, a tensor is passed to infer it. Or when the `Nx.runtime_uniform/4` is used within a "standard" function `def …`.

I created a gist with a livebook to reproduce it https://gist.github.com/nickgnd/16b441cf818f20180b2d2afafd7c4db0

Let me know if there is anything else I can do :)
Thanks again 🙇‍♂️

jonatanklosko:
Closes #668.

### DIFF
--- a/nx/test/nx/defn_test.exs
+++ b/nx/test/nx/defn_test.exs
@@ -346,7 +346,7 @@ defmodule Nx.DefnTest do
 
     test "raise an error given a shape tuple with tensor values" do
       assert_raise ArgumentError,
-                   ~r"invalid dimension in axis 0 in shape[\s\S]*If you are trying to pass shape as an argument to a defn function",
+                   ~r"invalid dimension in axis 0 in shape.*If you are trying to pass shape as an argument to a defn function"s,
                    fn -> iota({10}) end
     end
   end
--- a/nx/test/nx/defn_test.exs
+++ b/nx/test/nx/defn_test.exs
@@ -343,6 +343,10 @@ defmodule Nx.DefnTest do
       assert %T{shape: {3}, data: %Expr{op: :random_normal, args: [%T{shape: {}}, %T{shape: {}}]}} =
                random_normal(Nx.tensor([1, 2, 3]))
     end
+
+    test "raise an error given a shape tuple with tensor values" do
+      assert_raise ArgumentError, ~r"invalid dimension in axis 0 in shape", fn -> iota({10}) end
+    end
   end
 
   describe "tensor ops" do
--- a/nx/test/nx/defn_test.exs
+++ b/nx/test/nx/defn_test.exs
@@ -345,7 +345,9 @@ defmodule Nx.DefnTest do
     end
 
     test "raise an error given a shape tuple with tensor values" do
-      assert_raise ArgumentError, ~r"invalid dimension in axis 0 in shape", fn -> iota({10}) end
+      assert_raise ArgumentError,
+                   ~r"invalid dimension in axis 0 in shape[\s\S]*If you are trying to pass shape as an argument to a defn function",
+                   fn -> iota({10}) end
     end
   end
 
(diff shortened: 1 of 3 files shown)


polvalente (approved):
It would be nice to have a defn test for this

nx/test/nx/defn_test.exs
```diff
@@ -343,6 +343,12 @@ defmodule Nx.DefnTest do
       assert %T{shape: {3}, data: %Expr{op: :random_normal, args: [%T{shape: {}}, %T{shape: {}}]}} =
                random_normal(Nx.tensor([1, 2, 3]))
     end
+
+    test "raise an error given a shape tuple with tensor values" do
+      assert_raise ArgumentError,
+                   ~r"invalid dimension in axis 0 in shape[\s\S]*If you are trying to pass shape as an argument to a defn function",
```
polvalente, line 349:
```suggestion
                   ~r"invalid dimension in axis 0 in shape.*If you are trying to pass shape as an argument to a defn function",
```

Isn't `[\s\S]` the same as `.` given that one is the negation of the other?

jonatanklosko:
```suggestion
                   ~r"invalid dimension in axis 0 in shape.*If you are trying to pass shape as an argument to a defn function"s,
```

jonatanklosko: By default dot doesn't match newlines, but there is an option!

polvalente (approved):

nx/test/nx/defn_test.exs
polvalente: Oh, I see!

PR #760, opened 2022-04-26
Add grad tests for Nx.window_sum, Nx.pad, and Nx.put_slice

Benjamin-Philip:
See #695.

### DIFF
--- a/nx/test/nx/defn/grad_test.exs
+++ b/nx/test/nx/defn/grad_test.exs
@@ -1276,6 +1276,20 @@ defmodule Nx.Defn.GradTest do
 
       assert_all_close(lhs, rhs)
     end
+
+    test "works with complex" do
+      x = Nx.iota({2, 1, 3, 3}, type: {:c, 64})
+
+      lhs = grad_sum_window_sum(x)
+
+      rhs =
+        Nx.tensor([
+          [[[1.0, 1.0, 1.0], [2.0, 2.0, 2.0], [1.0, 1.0, 1.0]]],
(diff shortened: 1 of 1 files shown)


Benjamin-Philip:
I've not added an imaginary part in the left hand side as they don't seem to make a difference to the rhs (at least for pad and put_slice).

polvalente:
> I've not added an imaginary part in the left hand side as they don't seem to make a difference to the rhs (at least for pad and put_slice).

@Benjamin-Philip could you multiply the tensors by Nx.Constants.i()? Just so we ensure they are well behaved for imaginary tensors.

Benjamin-Philip:
> > I've not added an imaginary part in the left hand side as they don't seem to make a difference to the rhs (at least for pad and put_slice).
> 
> @Benjamin-Philip could you multiply the tensors by Nx.Constants.i()? Just so we ensure they are well behaved for imaginary tensors.

I've multiplied all the tensors by `i`, but now the real part is zero. I hope that's okay? Or should I rewrite the tests to have a non-zero real and imaginary part?

polvalente:
> > > I've not added an imaginary part in the left hand side as they don't seem to make a difference to the rhs (at least for pad and put_slice).
> > 
> > @Benjamin-Philip could you multiply the tensors by Nx.Constants.i()? Just so we ensure they are well behaved for imaginary tensors.
> 
> I've multiplied all the tensors by `i`, but now the real part is zero. I hope that's okay? Or should I rewrite the tests to have a non-zero real and imaginary part?

I think it's ok that they are purely imaginary :)

polvalente (approved):

PR #761, opened 2022-04-26
Add cumulative functions

### LINKED ISSUE #485: Add cumulative reduction functions
This discussion came up in the Elixir slack. I think we should add cumulative reductions to the Nx API, namely:

- [ ] cumsum
- [ ] cumprod
- [ ] cummin
- [ ] cummax

Most backends will have a more specific, optimized implementation than what we can write with the current API. We can implement them in XLA using `reduce_window` - this will not work in Torchx obviously.

jonatanklosko:
Closes #485.

I went with a longer naming (`cumulative_sum` rather than `cumsum`) for consistency with the window functions. If there is a preference for the shorter one let me know :)

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7048,6 +7048,182 @@ defmodule Nx do
     do: aggregate_window_op(tensor, window_dimensions, opts, :window_product)
 
   @doc """
+  Returns the cumulative sum of elements along an axis.
+
+  ## Options
+
+    * `:axis` - the axis to sum elements along. Defaults to `0`
+
+  ## Examples
+
+      iex> Nx.cumulative_sum(Nx.tensor([1, 2, 3, 4]))
+      #Nx.Tensor<
+        s64[4]
+        [1, 3, 6, 10]
+      >
+
+      iex> Nx.cumulative_sum(Nx.iota({3, 3}), axis: 0)
+      #Nx.Tensor<
+        s64[3][3]
+        [
+          [0, 1, 2],
+          [3, 5, 7],
+          [9, 12, 15]
+        ]
+      >
+
+      iex> Nx.cumulative_sum(Nx.iota({3, 3}), axis: 1)
+      #Nx.Tensor<
+        s64[3][3]
+        [
+          [0, 1, 3],
+          [3, 7, 12],
+          [6, 13, 21]
+        ]
+      >
+  """
+  @doc type: :cumulative
+  def cumulative_sum(tensor, opts \\ []),
+    do: cumulative_window_op(tensor, opts, :window_sum)
+
+  @doc """
+  Returns the cumulative product of elements along an axis.
+
+  ## Options
+
+    * `:axis` - the axis to multiply elements along. Defaults to `0`
+
+  ## Examples
+
+      iex> Nx.cumulative_product(Nx.tensor([1, 2, 3, 4]))
+      #Nx.Tensor<
+        s64[4]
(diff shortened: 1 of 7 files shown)


josevalim (approved):

jonatanklosko:
@polvalente any idea why torch on windows returns `-1` for all functions? Interestingly it's only on windows.

polvalente:
> @polvalente any idea why torch returns `-1` for `window_min` calls? Interestingly it's only on windows.

@jonatanklosko it's really odd! It doesn't even just return a `-1` tensor, there are some `1` positions as well...

jonatanklosko:
Should we skip the doctest for now?

josevalim:
@jonatanklosko instead let's define the optional callbacks for those and implement them for Torchx. Apparently the window operations are multiple operations for torchx, so the optional callback is worth it.

polvalente (approved):

torchx/lib/torchx/backend.ex
```diff
@@ -701,6 +701,42 @@ defmodule Torchx.Backend do
     end
   end
 
+  @impl true
+  def cumulative_sum(%T{type: out_type} = out, %T{} = t, axis) do
+    check_type!(out_type)
```
jonatanklosko, line 706: @polvalente this matches `sum` and `product`, let me know if it's not what we want.

nx/lib/nx.ex
```diff
@@ -7047,6 +7047,185 @@ defmodule Nx do
...
+  @doc type: :cumulative
+  def cumulative_max(tensor, opts \\ []),
+    do: cumulative_op(tensor, opts, :cumulative_max, :window_max)
+
+  defp cumulative_op(tensor, opts, op, window_op) do
+    opts = keyword!(opts, axis: 0)
+    tensor = to_tensor(tensor)
+    axis = Nx.Shape.normalize_axis(tensor.shape, opts[:axis], tensor.names)
+
+    Nx.Shared.optional(op, [tensor, axis], tensor, fn tensor, axis ->
```
jonatanklosko, line 7211: @josevalim I defined the callbacks as

```elixir
@callback cumulative_sum(out :: tensor, t :: tensor, axis) :: tensor
```

however, we can't pass axis this way, because `EXLA.Backend` does JIT and converts `axis` to tensor. Thoughts on what approach to take?

jonatanklosko:
```suggestion
    Nx.Shared.optional(op, [tensor, axis], tensor, fn tensor, _axis ->
```

This would work, because we would capture and use the axis directly, it's just a little bit weird.

josevalim: @jonatanklosko we should change EXLA.Backend to JIT only on the first argument. We are being too eager on the JITing there.

jonatanklosko: Ahhh, we have the callback name, so even if we add more tensors, we can just differentiate based on the name, got it!

josevalim: Yeah, but even then I think all optional callbacks only need to JIT on the first arguyment!

jonatanklosko: Hmm, what do you mean?

```elixir
@callback solve(out :: tensor, a :: tensor, b :: tensor) :: tensor
```

We want to JIT on both `a` and `b`, no?

jonatanklosko: Updated!

josevalim:

PR #764, opened 2022-04-28
docs: update Torchx docs

### LINKED ISSUE #684: Add moduledoc for Torchx module
We need to add a moduledoc that documents the types and devices for Torchx.
It should make it clear they are Torchx-specific and that Torchx.Backend provides the mapping between Nx to the underlying Torchx types.

polvalente:
closes #684

### DIFF
--- a/torchx/lib/torchx/backend.ex
+++ b/torchx/lib/torchx/backend.ex
@@ -22,10 +22,10 @@ defmodule Torchx.Backend do
        and nan becomes zero. `Torchx` behaviour is type dependent with no clear
        rule across types.
 
-  ### Options
+  ## Options
 
-    * `:device` - an atom representing the device for the allocation of a given tensor.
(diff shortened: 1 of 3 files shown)


torchx/lib/torchx/backend.ex
```diff
@@ -22,6 +22,10 @@ defmodule Torchx.Backend do
        and nan becomes zero. `Torchx` behaviour is type dependent with no clear
        rule across types.
 
+  ## Options
+
+    * `:device` - Defaults to `:cpu`. An atom representing the device for the allocation of a given tensor.
+    Valid values can be seen at the main [`Torchx`](Torchx.html#module-devices) docs.
```
polvalente, line 28: Is there a better way to point to anchors like this?

josevalim: I think this is the way to go, yeah.

josevalim (approved):

PR #782, opened 2022-05-18
feat(complex): add support in Nx.Defn.Composite

polvalente:


nx/lib/nx/defn/composite.ex
```diff
@@ -138,7 +135,7 @@ defmodule Nx.Defn.Composite do
   defp flatten_each(%T{} = tensor, acc, _fun),
     do: [tensor | acc]
 
-  defp flatten_each(number, acc, fun) when is_number(number),
+  defp flatten_each(number, acc, fun) when is_number(number) or is_struct(number, Complex),
     do: [fun.(number) | acc]
 
   defp flatten_each(container, acc, fun),
```
polvalente, line 141: @josevalim I couldn't remember exactly what the issue was since I wasn't able to reproduce it through `if` in the `defn_test.exs` file. However, I've explicitly tested and fixed these functions directly.

The `@doc false`  functions below were kept as is because I didn't see any `is_number` nor `%T{}` matching

nx/lib/nx.ex
```diff
@@ -360,6 +360,8 @@ defmodule Nx do
 
   @non_finite [:neg_infinity, :infinity, :nan]
 
+  defguard is_tensor(t) when is_number(t) or is_struct(t, T) or is_struct(t, Complex)
```
josevalim, line 363: Let's doc it too! And we need to add a `@doc type: ...` thing. But no idea what to add here. Maybe its own called `:guard`?

nx/lib/nx/defn/composite.ex
josevalim: Perfect!

josevalim (approved):
Just one comment and ship it!

nx/lib/nx.ex
polvalente: Screenshot from 2022-05-18 13-33-32
Screenshot from 2022-05-18 13-33-37

I added it like this. Please take another look and merge if it makes sense!
I've added it as `type: :guards`, should it be `:guard` instead?

josevalim (approved):

PR #784, opened 2022-05-27
feat(non-finite): support non-finites in BinaryBackend

polvalente:
blocked by https://github.com/elixir-nx/complex/pull/20

### DIFF
--- a/nx/mix.exs
+++ b/nx/mix.exs
@@ -38,7 +38,7 @@ defmodule Nx.MixProject do
 
   defp deps do
     [
-      {:complex, "~> 0.4.0"},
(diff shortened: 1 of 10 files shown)


nx/mix.exs
```diff
@@ -38,7 +38,7 @@ defmodule Nx.MixProject do
 
   defp deps do
     [
-      {:complex, "~> 0.4.0"},
+      {:complex, "~> 0.4.0", path: "../../complex"},
```
polvalente, line 41: This needs to change with a ~> 0.4.1 release for complex due to some needed fixes in https://github.com/elixir-nx/complex/pull/20

PR #785, opened 2022-05-27
feat: add FFT

polvalente:


exla/mix.exs
```diff
@@ -57,8 +57,8 @@ defmodule EXLA.MixProject do
 
   defp deps do
     [
-      # {:nx, path: "../nx"},
-      {:nx, "~> 0.2.0"},
+      {:nx, path: "../nx"},
```
polvalente, line 60: This is needed for FFT to be implemented in both EXLA and Torchx (in their respective mix.exs files)

nx/lib/nx/binary_backend.ex
```diff
@@ -2153,6 +2153,72 @@ defmodule Nx.BinaryBackend do
...
+          read!(x, 0)
+        end
+      end
+
+    result =
+      if rem(n, 2) == 0 do
+        fft_list(data)
+      else
+        dft_list(data)
+      end
```
polvalente, line 2170: I think we can just use `dft_list` for all tensors. I'll try to compare both using benchee, but I don't expect `fft_list` to actually be faster. I'll bring back the results here

nx/lib/nx/binary_backend.ex
```diff
@@ -2153,6 +2153,72 @@ defmodule Nx.BinaryBackend do
...
+
+    left =
+      even
+      |> Enum.zip(t)
+      |> Enum.map(fn {x, y} -> x + y end)
+
+    right =
+      even
+      |> Enum.zip(t)
+      |> Enum.map(fn {x, y} -> x - y end)
```
josevalim, line 2217:
```suggestion
    left = Enum.zip_with(even, t, fn {x, y} -> x + y end)
    right = Enum.zip_with(even, t, fn {x, y} -> x - y end)
```

nx/lib/nx/binary_backend.ex
josevalim: > We can also add a pad option which pads the input tensor with zeros to the next power of two, which is a common DSP technique for best use of the FFT. WDYT?

Sounds good to me!

exla/mix.exs
josevalim: Note I have already added a path depednency on main. :)

nx/lib/nx/binary_backend.ex
jonatanklosko: Both [torch.fft.fft](https://pytorch.org/docs/stable/generated/torch.fft.fft.html#torch.fft.fft) and [numpy.fft.fft](https://numpy.org/doc/stable/reference/generated/numpy.fft.fft.html) have an `n` option that either trims or pads to the given size.

We could try implementing a more complex algorithm to optimise non-powers-of-2, but for performance people should be using a different backend anyway, so it's probably an overkill. Also, [a random reference](https://math.stackexchange.com/a/77152).

nx/lib/nx/binary_backend.ex
```diff
@@ -2153,6 +2153,72 @@ defmodule Nx.BinaryBackend do
...
+        Enum.reduce(data, {0, 0}, fn x, {acc, n} ->
+          {acc + x * Complex.exp(Complex.new(0, -2 * :math.pi() * k * n / bigN)), n + 1}
+        end)
+
+      xk
+    end
+  end
+
+  defp fft_list([]), do: []
+  defp fft_list([data]), do: [data]
```
jonatanklosko, line 2196: We should have this clause for `dft_list`, for fft we don't expect an odd number :)

polvalente: This is the base case for recursion

nx/lib/nx/binary_backend.ex
polvalente: It's easy to get the next power of 2 of a number through log and ceil, so maybe we can add a `pad: integer() | :power_of_two` option?

nx/lib/nx/binary_backend.ex
jonatanklosko: Oh, I think we should be calling the top-most fft recursively, rather than `fft_list` directly. For example if the length is 6, it splits into 3 and 3, and for those we need to use `dft_list`.

We could actually make it a single function with guards.

polvalente: I chose to split because I think it's a little messy to use different datatypes and signatures for the same function name ina recursion like this. So the top level formats the data and then calls this.

But I agree we can use a guard to merge dft_list into fft_list!

jonatanklosko: The datatype and signature for `dft_list` and `fft_list` is the same though, it's just that `fft_list` is a special case as an optimisation, no? My point main point is that currently it doesn't work for 6 (I think) :)

nx/lib/nx/binary_backend.ex
polvalente: Turns out Enum.zip was bugging out and the algorithm really only does work for powers of two. I've refactored to only apply the recursive version in said cases, and use the naive DFT otherwise.

I'll just add the padding option and this is ready for review again :)

exla/lib/exla/defn.ex
```diff
@@ -760,6 +760,28 @@ defmodule EXLA.Defn do
...
+        {^n} ->
+          tensor
+
+        {m} when m > n ->
+          EXLA.Op.slice(tensor, [0], [n], [1])
+
+        {m} when m < n ->
+          EXLA.Op.pad(tensor, EXLA.Op.constant_r0(state.builder, Complex.new(0), output_type), [
+            {0, n - m, 0}
+          ])
```
josevalim, line 779:
```suggestion
          constant = EXLA.Op.constant_r0(state.builder, Complex.new(0)
          EXLA.Op.pad(tensor, constant, output_type), [{0, n - m, 0}])
```

nx/lib/nx.ex
```diff
@@ -10295,6 +10295,86 @@ defmodule Nx do
...
   end
 
+  @doc """
+  Calculates the DFT of the given 1D tensor.
+
+  ## Options
+
+    * `:eps` - Threshold which backends can use for cleaning-up results. Defaults to `1.0e-10`.
+    * `:length` - Either a positive integer or `:power_of_two`. Will pad or slice the tensor
+    accordingly. `:power_of_two` will automatically pad to the next power of two.
```
josevalim, line 10305:
```suggestion
      accordingly. `:power_of_two` will automatically pad to the next power of two.
```

josevalim (approved):
Two nitpicks and SHIP IT! :D

nx/lib/nx/binary_backend.ex
jonatanklosko: Oh interesting, why does it make a difference if we compute the recursive call with explicit DFT or FFT?

polvalente: The FFT algorithm removes a lot of the complex exponential calculations by some clever recombining of the parts which were recursively calculated. https://en.wikipedia.org/wiki/Cooley%E2%80%93Tukey_FFT_algorithm

Since we have the Radix-2 implementation, it only works for powers of two, but there are versions which work in general length, but depend on prime factorizations and/or fast matrix multiplication, so I think it's probably overkill as well.

PR #788, opened 2022-05-28
Nx.logistic to Nx.sigmoid

### LINKED ISSUE #787: Rename logistic to sigmoid
According to the [Nx docs](https://hexdocs.pm/nx/Nx.html#logistic/1), the logistic function is defined as follows:

$$
f(x) = \frac{L}{1 + e^{-x}}
$$

However, after some googling, I found that the standard logistic function is defined as follows:

$$
f(x) = \frac{L}{1 + e^{-k(x - x_{0})}}
$$

And that we are using the sigmoid function. The sigmoid function [just happens to be a special case of the logistic function](https://stats.stackexchange.com/questions/204484/what-are-the-differences-between-logistic-function-and-sigmoid-function).

Therefore I propose that we rename `Nx.logistic` to `Nx.sigmoid`.

PS: It feels so good to finally be able to write in $\LaTeX$! To bad that they are using MathJax and not KaTeX.

josevalim:
Closes #787

### DIFF
--- a/exla/c_src/exla/exla.cc
+++ b/exla/c_src/exla/exla.cc
@@ -818,7 +818,7 @@ ERL_NIF_TERM log1p(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[]) {
   return xla_unary_op(env, argc, argv, xla::Log1p);
 }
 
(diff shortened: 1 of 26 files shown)


josevalim:

PR #789, opened 2022-05-31
feat: improve fft implementation for composite even numbers

polvalente:
As @jonatanklosko pointed out, we can take advantage of all factors of 2 in the recursive case


jonatanklosko (approved):
Perfect!

josevalim (approved):

PR #800, opened 2022-06-05
feat: add starter implementation for as_windowed

polvalente:
Tentative implementation for Nx.as_windowed

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -9875,7 +9875,7 @@ defmodule Nx do
 
   `iodata` is a list of binaries that can be written to any io device,
   such as a file or a socket. You can ensure the result is a binary by
-  calling `IO.iodata_to_binary/1`.  
+  calling `IO.iodata_to_binary/1`.
 
   ## Examples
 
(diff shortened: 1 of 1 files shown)


nx/lib/nx.ex
```diff
@@ -10375,6 +10375,198 @@ defmodule Nx do
...
+          ],
+          [
+            [-2, -3],
+            [-5, -6],
+            [-8, -9]
+          ]
+        ]
+      >
+  """
+  def as_windowed(tensor, window_dimensions, opts \\ []) do
```
polvalente, line 10460: @josevalim Are we keeping this here even though it can be implemented through pure Nx?
I'm almost certain the `for` can be turned into a `while` for a better `defn` version which could be used
as the `optional` default implementation.

For binary backend and torchx we can do better than this through binary parsing and through Torchx.unfold.
However, I'm also happy with keeping this in NxSignal instead.

nx/lib/nx.ex
```diff
@@ -10375,6 +10375,198 @@ defmodule Nx do
...
+    window_dimensions_list = Tuple.to_list(window_dimensions)
+
+    sliced_indices =
+      for {stride, axis} <- Enum.with_index(strides), reduce: scattered_indices do
+        idx ->
+          len = elem(window_min_source.shape, axis)
+          lengths = idx.shape |> Tuple.to_list() |> List.replace_at(axis, len)
+
+          if stride > 1 do
+            sorted = Nx.sort(idx, axis: axis, direction: :asc)
```
polvalente, line 10527: I can't convert this to a defn basically because I couldn't find a way to get the `axis` to increment in a `while`, for instance, and still keep it as a number.

Current defn implementation here: https://github.com/polvalente/nx-signal/blob/08c059d10d5bd90fcb433b2a5286e2a2e9441e99/lib/nx_signal.ex#L166

PR #801, opened 2022-06-05
feat: add Nx.ifft/2

polvalente:


nx/lib/nx.ex
```diff
@@ -10393,6 +10393,98 @@ defmodule Nx do
...
+      case opts[:length] do
+        :power_of_two ->
+          2 ** Kernel.ceil(:math.log2(n))
+
+        n when is_integer(n) and n > 0 ->
+          n
+
+        length ->
+          raise "expected an integer or :power_of_two as length, got: #{inspect(length)}"
+      end
```
josevalim, line 10475: We can share a lot here with fft, right?

nx/lib/nx/binary_backend.ex
```diff
@@ -2197,7 +2202,16 @@ defmodule Nx.BinaryBackend do
...
-        fft_list(row, n)
+        case kind do
+          :fft ->
+            fft_list(row, n)
+
+          :ifft ->
+            row
+            |> Enum.map(&Complex.conjugate/1)
+            |> fft_list(n)
+            |> Enum.map(&(Complex.conjugate(&1) / n))
```
josevalim, line 2213: Does it mean it could be implemented using optional callbacks? Or not actually advisable?

polvalente: It could, yeah, but we would have implementations for both Torchx and EXLA, so I don't see much advantage in doing this.

I'd also expect everywhere that provides DFT to provide inverse DFT as well.

josevalim: Ship it!

PR #804, opened 2022-06-06
fix(complex): Enum.reduce instead of Enum.sum

polvalente:
Some edge cases failed when mixing real and complex tensors


josevalim:
Could we add tests to avoid regressions? :)

polvalente:
The one in `conv` I'll add on a separate PR because `conv` doesn't support complex numbers yet. So I'm rolling that back in this one :)

josevalim (approved):

PR #805, opened 2022-06-07
feat(complex): add support in Nx.conv

### LINKED ISSUE #791: Support complex numbers in Nx.conv
We need to add support for complex numbers for the function and for it's grad implementation

- [x] BinaryBackend
- [x] EXLA
- [x] Torchx
- [x] grad implementation

polvalente:
closes #791

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -8484,6 +8484,19 @@ defmodule Nx do
         ]
       >
 
+  Complex tensors are also supported
+
+      iex> left = Nx.tensor([[[Complex.new(1, 1), 2, Complex.new(3, -3)]]])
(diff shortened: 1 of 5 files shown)


polvalente:
EXLA seems to support it out of the box, given that exla/test/exla/backend_test.exs runs the doctests and there's a new doctest for conv

nx/lib/nx.ex
```diff
@@ -8484,6 +8484,19 @@ defmodule Nx do
         ]
       >
 
+  Complex tensors are also supported
```
josevalim, line 8487:
```suggestion
  Complex tensors are also supported:
```

josevalim (approved):

Issue #795, opened 2022-06-04
Add ||, && and raise to defn


josevalim:

jeantux:
can i work on it?

PR #806, opened 2022-06-08
Added: Logical operator to && and ||

jeantux:
#795 Add && and || as logical operator

### DIFF
--- a/nx/lib/nx/defn/kernel.ex
+++ b/nx/lib/nx/defn/kernel.ex
@@ -1,4 +1,6 @@
 defmodule Nx.Defn.Kernel do
+  import Kernel, except: [raise: 1]
+
   @moduledoc """
   All imported functionality available inside `defn` blocks.
   """
@@ -540,7 +542,7 @@ defmodule Nx.Defn.Kernel do
       end
 
   """
-  def left && right, do: boolean_and(left, right)
+  def left && right, do: Kernel.&&(left, right)
 
   @doc """
   Element-wise logical OR operation.
@@ -587,7 +589,7 @@ defmodule Nx.Defn.Kernel do
       end
 
   """
-  def left || right, do: boolean_or(left, right)
+  def left || right, do: Kernel.||(left, right)
 
   @doc """
   Element-wise logical NOT operation.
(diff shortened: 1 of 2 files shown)


nx/lib/nx/defn/kernel.ex
```diff
@@ -522,6 +522,34 @@ defmodule Nx.Defn.Kernel do
...
+          "boolean value passed to Nx.Defn.Kernel.and/2, " <>
+            "values passed to Nx.Defn.Kernel.and/2 must be " <>
+            "tensors or numbers, consider using 1 for true " <>
+            "and 0 for false as an alternative"
+  end
+
+  def left && right when Kernel.and(is_number(left), is_number(right)),
+    do: logical_and(left, right)
+
+  def left && right, do: Nx.logical_and(left, right)
```
polvalente, line 551: I'm not sure if this is actually the implementation we want. This was discussed because with `&&` and `||` we can mostly avoid going into `transform` just to extract opts.

For example: `eps = opts[:eps] || 1.0e-5`, So I think it's more in line with adding something closer to `if` or `cond`

jeantux: ok it makes sense, in this simple implementation i just reflected the behavior of the `and` and `or` operator.

I can adjust.

jeantux:
I am delegating the operation to `Kernel.&&` and `Kernel.||`.

I don't know if it will be necessary to write this operation in another way, such as how it is done in the `lib/nx/binary_tensor.ex` module .. 🤔

josevalim:
Sorry, I should have detailed what those functions do in the original issue. But now I realize that their introduction will cause confusion, so we need to change something.

In particular, note that the current implementation of `||` and `&&` won't work because they can no longer be chained. For example, I can't do `foo || maybe_nil || maybe_another`. We also always evaluate both sides, which means we can't write `opts[:foo] || raise ...`. Perhaps the safest best is to assume that they return a tensor (or a number), but that would make their implementation quite more complicated.

josevalim:
For example, what should `0 || 20` return? For `or`, it would be 20. But right now it is 0 and that's inconsistent.

Also, I don't think we need to worry about this case:

    eps = opts[:eps] || 1.0e-5

because `keyword!` can handle default values. I guess what we really need is "get this option and, if it is false, raise this error message". But the meaning of "is false" is different in `defn` and I don't think we want to raise for `opts[:eps] || raise "expected eps"` even if eps is zero.

At the end of the day, all we want is the nil check, which perhaps could be written like this:

    opts[:eps] |> if_nil(raise "expected eps")

But I don't like departing from Elixir built-in idioms.

josevalim:
In any case, thank you for the PR @jeantux! I wouldn't be able to see those issues without you taking a stab at the problem. The main issues are caused by the problem definition, not the implementation. :)

PR #808, opened 2022-06-08
changed sigma to variance in :rand.normal()

### LINKED ISSUE #809: Error in sampling from Normal distribution
There is an error in how Nx calls Erlang when creating random samples from a Normal distribution-- Nx is sending the standard deviation but Erlang expects the variance.  I created a pull request that fixes it.

wadehenning:
There is an error in how Nx is sampling random variables from a Normal distribution.  This request fixes it.
closes #809


polvalente (approved):

polvalente:
@wadehenning I forgot to actually send the comment before :sweat_smile:
Thanks for catching and fixing the bug!

I've sent a separate PR adding a property test to ensure everything works

PR #813, opened 2022-06-20
Fix Nx.Defn.Expr.as_type to properly cast constants

jonatanklosko:
See the corresponding test.

### DIFF
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -584,6 +584,13 @@ defmodule Nx.Defn.Expr do
     tensor = to_expr(tensor)
 
     if c = maybe_constant(tensor) do
+      c =
+        if is_float(c) and Nx.Type.integer?(out.type) do
+          floor(c)
+        else
+          c
+        end
+
       constant(out, c)
     else
       expr(out, tensor.data.context, :as_type, [tensor])
@@ -988,8 +995,8 @@ defmodule Nx.Defn.Expr do
         is_integer(number) and Nx.Type.float?(type) ->
           Complex.multiply(1.0, number)
 
-        is_float(number) and Nx.Type.integer?(type) ->
-          floor(number)
+        not is_integer(number) and Nx.Type.integer?(type) ->
+          raise "value #{inspect(number)} is not valid for constant of type #{inspect(type)}"
 
         number ->
           number
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -586,7 +586,7 @@ defmodule Nx.Defn.Expr do
     if c = maybe_constant(tensor) do
       c =
         if is_float(c) and Nx.Type.integer?(out.type) do
-          floor(c)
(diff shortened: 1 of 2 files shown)


nx/lib/nx/defn/expr.ex
```diff
@@ -984,9 +984,16 @@ defmodule Nx.Defn.Expr do
...
     number =
-      if is_integer(number) and Nx.Type.float?(type),
-        do: Complex.multiply(1.0, number),
-        else: number
+      cond do
+        is_integer(number) and Nx.Type.float?(type) ->
+          Complex.multiply(1.0, number)
+
+        is_float(number) and Nx.Type.integer?(type) ->
+          floor(number)
```
josevalim, line 992: I don't think we should do this here, it most often is a bug. I think here we should raise and we do explicit conversions on as_type.

jonatanklosko: I think it's a valid case, see

https://github.com/elixir-nx/nx/blob/1f4ddfea59849da1291083cac51177915511d0f5/nx/lib/nx/defn/expr.ex#L586-L587

This is an optimisation when casting constants, so we need to cast the actual value accordingly.

josevalim: Yes, the bug is in `as_type`. We can cast from integer -> float, but never from float to integer transparently, unless using `as_type`. So the logic belongs there! :)

nx/lib/nx/defn/expr.ex
```diff
@@ -984,9 +991,16 @@ defmodule Nx.Defn.Expr do
...
     number =
-      if is_integer(number) and Nx.Type.float?(type),
-        do: Complex.multiply(1.0, number),
-        else: number
+      cond do
+        is_integer(number) and Nx.Type.float?(type) ->
+          Complex.multiply(1.0, number)
+
+        not is_integer(number) and Nx.Type.integer?(type) ->
+          raise "value #{inspect(number)} is not valid for constant of type #{inspect(type)}"
```
jonatanklosko, line 999: Let me know if having this assertion is worth it.

nx/lib/nx/defn/expr.ex
```diff
@@ -584,6 +584,13 @@ defmodule Nx.Defn.Expr do
     tensor = to_expr(tensor)
 
     if c = maybe_constant(tensor) do
+      c =
+        if is_float(c) and Nx.Type.integer?(out.type) do
+          floor(c)
```
josevalim, line 589: what the binary backend uses is:

```suggestion
          trunc(c)
```

nx/lib/nx/defn/expr.ex
josevalim:
```suggestion
        not is_integer(number) and Nx.Type.integer?(type) ->
          raise ArgumentError, "value #{inspect(number)} is not valid for constant of type #{inspect(type)}"
```

josevalim (approved):

josevalim:

PR #821, opened 2022-06-26
Make `Nx.flatten` an optional callback

Benjamin-Philip:
- Make `Nx.flatten` an optional callback
- Implement flatten for `Torchx`

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -2160,7 +2160,17 @@ defmodule Nx do
   """
   @doc type: :shape
   def flatten(tensor) do
-    reshape(tensor, {size(tensor)})
+    tensor = to_tensor(tensor)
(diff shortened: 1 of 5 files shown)


josevalim:
I don't think there is a benefit in making this an optional callback because, in all cases, it is still just a single Nx operation. Optional callbacks should be used when we are replacing either multiple Nx operations or multiple operations in the backend. :)

In any case, thank you!

Benjamin-Philip:
In that case would making it a proper callback be better?

I feel using the native `flatten` equivalent in backends as opposed to the more generalized `reshape` may be more efficient and performant in the case of larger tensors.

josevalim:
I don't think that is the reality. Flatten is meant to be a convenient reshape operation. Unless we have concrete benchmarks showing otherwise.

PR #822, opened 2022-06-26
Fix warnings with implementing `Nx.phase` as an optional callback

Benjamin-Philip:
The phase optional callback was not added to the `Nx.Backend` module.
Thus, when implementing the callback for a backend, warnings are
passed during compilation.

This commit fixes those warnings, by adding the callback as optional
in the `Nx.Backend` module


josevalim:

PR #826, opened 2022-06-27
feat: add :reverse to cumulative functions

### LINKED ISSUE #771: Support :reverse option in cumulative functions
To match Explorer: https://github.com/elixir-nx/explorer/pull/200

polvalente:
closes #771

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7396,8 +7396,8 @@ defmodule Nx do
     axis = Nx.Shape.normalize_axis(tensor.shape, opts[:axis], tensor.names)
 
     Nx.Shared.optional(op, [tensor, [axis: axis, reverse: reverse]], tensor, fn tensor,
-                                                                              axis: axis,
-                                                                              reverse: reverse ->
+                                                                                axis: axis,
+                                                                                reverse: reverse ->
       shape = shape(tensor)
       axis_size = elem(shape, axis)
       rank = rank(shape)
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7155,6 +7155,7 @@ defmodule Nx do
   ## Options
 
     * `:axis` - the axis to sum elements along. Defaults to `0`
+    * `:reverse` - if `true`, reverses the direction on the axis of accumulation. Defaults to `false`
 
   ## Examples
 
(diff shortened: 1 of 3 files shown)


josevalim (approved):
Mix format and ship it!!!

nx/lib/nx.ex
```diff
@@ -7183,6 +7184,26 @@ defmodule Nx do
...
+          [9, 11, 13],
+          [9, 12, 15]
+        ]
+      >
+
+      iex> Nx.cumulative_sum(Nx.iota({3, 3}), axis: 1, reverse: true)
+      #Nx.Tensor<
+        s64[3][3]
+        [
+          [2, 3, 3],
```
jonatanklosko, line 7202:
```elixir
iex> [0, 1, 2] |> Explorer.Series.from_list() |> Explorer.Series.cumulative_sum(reverse: true)
#Explorer.Series<
  integer[3]
  [3, 3, 2]
>
```

Reverse only determines the direction we go. The result is equivalent to `reverse + cumulative_* + reverse`.

The default windowed implementation could be optimised by padding right rather than left, but for torch I think we will need the double reverse.

jonatanklosko: Neither numpy nor pytorch support this option, but [tf.math.cumsum](https://www.tensorflow.org/api_docs/python/tf/math/cumsum) does.

torchx/lib/torchx/backend.ex
```diff
@@ -705,39 +705,107 @@ defmodule Torchx.Backend do
...
+    result = Torchx.cumulative_sum(t_tx, axis)
+
+    if reverse do
+      result
+      |> Torchx.flip([axis])
+      |> to_nx(out)
+    else
+      to_nx(result, out)
+    end
   end
```
jonatanklosko, line 732: We can refactor it into `cumulative_op(out, t, opts, &Torchx.cumulative_sum/2)` :)

josevalim (approved):

nx/lib/nx.ex
```diff
@@ -7194,6 +7215,7 @@ defmodule Nx do
   ## Options
 
     * `:axis` - the axis to multiply elements along. Defaults to `0`
+    * `:reverse` - if `true`, reverses the direction on the axis of accumulation. Defaults to `false`
```
jonatanklosko, line 7218:
```suggestion

    * `:reverse` - whether to perform accumulation in the opposite direction. Defaults to `false`
```

polvalente: I don't like this because it doesn't lead the person to pass a boolean

jonatanklosko: I'd say `` Defaults to `false` `` makes it clear, but your call!

nx/lib/nx.ex
```diff
@@ -7324,7 +7411,20 @@ defmodule Nx do
...
+          tensor
+        end
+
+      result = aggregate_window_op(t, window_shape, [padding: padding], window_op)
+
+      if reverse do
+        Nx.reverse(result, axes: [axis])
+      else
+        result
+      end
```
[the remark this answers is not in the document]
polvalente: This works beautifully!
It eliminates both reverses :)

nx/lib/nx.ex
polvalente: How about:

`* `:reverse` - a boolean which determines whether to perform accumulation in the opposite direction. Defaults to `false``

jonatanklosko: Works for me :) Sidenote, we currently have `` `:keep_axis` - whether or not ...``

PR #827, opened 2022-06-27
Add transform_if/2 and raise/2

### LINKED ISSUE #795: Add ||, && and raise to defn

josevalim:
Closes #795, closes #806.


josevalim:
This approach is still flawed and confusing. We will go with assert_keys instead.

PR #828, opened 2022-06-28
feat: add deftransform and deftransformp

polvalente:
This aims to deprecate `Nx.Defn.Kernel.transform` in favor of a more stable approach for handling mixing Nx-land with Elixir-land.
Using separate definitions avoid leaking the Nx.Defn.Kernel overrides into the former `transform` `fn`s, for instance


nx/test/nx/defn_test.exs
```diff
@@ -1597,4 +1597,48 @@ defmodule Nx.DefnTest do
...
+      result = deftransform_test(Nx.tensor(1), Nx.tensor(2), b: 3, c: 4)
+
+      assert String.trim("""
+             #Nx.Tensor<
+               s64
+             \s\s
+               Nx.Defn.Expr
+               parameter a:0       s64
+               parameter c:1       s64
+               b = add 3, a        s64
```
polvalente, line 1634: I found it really odd that the second parameter ended up with the letter `c` instead of `b`. Is this a possible bug related to `deftransform`?

nx/lib/nx/defn/compiler.ex
```diff
@@ -268,11 +268,14 @@ defmodule Nx.Defn.Compiler do
...
   end
 
-  defp get_and_normalize_definition(def, state) do
+  defp get_and_normalize_definition(def, function_type, state) do
     {:v1, kind, meta, clauses} = Module.get_definition(state.module, def)
     state = %{state | function: def, line: meta[:line] || state.line, rewrite_underscore?: true}
 
     case clauses do
+      [{meta, args, [], ast}] when function_type == :transform ->
+        {{kind, meta, args, ast}, state}
```
polvalente, line 277: I think we can relax the conditions for when `function_type == :transform` here, right? Like, allow multiple clauses, guards and so on?

nx/lib/nx/defn.ex
```diff
@@ -722,7 +791,13 @@ defmodule Nx.Defn do
         %{}
       end
 
-    exports = Map.put(exports, {name, arity}, %{kind: kind, defaults: defaults})
+    exports =
+      Map.put(exports, {name, arity}, %{
+        function_type: function_type,
```
josevalim, line 796:
```suggestion
        type: function_type,
```

We know it is a function. :)

nx/lib/nx/defn/compiler.ex
josevalim: I would not allow it for now, I would keep it as restricted as `defn`. You may need to execute those lines on `args` and `ast` though:

https://github.com/elixir-nx/nx/blob/863ea5a6040dc6d620aabc80fd62b07df1ae5b31/nx/lib/nx/defn/compiler.ex#L427-L430

nx/lib/nx/defn.ex
```diff
@@ -654,12 +654,73 @@ defmodule Nx.Defn do
   all local function calls within `defn`.
   """
   defmacro defnp(call, do: block) do
-    define(:defp, call, block, __CALLER__)
+    define(:defp, call, block, __CALLER__, :numerical)
+  end
+
+  @doc """
+  Defines a transform that executes the given `fun` with `arg`
+  when building `defn` expressions.
```
josevalim, line 662:
```suggestion
  Defines a function that transforms a numerical definition expression.
```

nx/lib/nx/defn.ex
```diff
@@ -654,12 +654,73 @@ defmodule Nx.Defn do
...
+      deftransform my_inspect(value), do: IO.inspect(value)
+
+  Or:
+
+      defn tanh_power(a, b) do
+        res = Nx.tanh(a) + Nx.power(b, 2)
+        my_inspect(res)
+        res
+      end
+
```
josevalim, line 688:
```suggestion

```

nx/lib/nx/defn.ex
```diff
@@ -654,12 +654,73 @@ defmodule Nx.Defn do
...
+  Let's see a trivial example, which is to use `IO.inspect/1` to
+  print a tensor expression at definition time:
+
+      defn tanh_power(a, b) do
+        Nx.tanh(a) + Nx.power(b, 2) |> my_inspect()
+      end
+
+      deftransform my_inspect(value), do: IO.inspect(value)
```
josevalim, line 679:
```suggestion
      deftransformp my_inspect(expr), do: IO.inspect(expr)
```

nx/lib/nx/defn/compiler.ex
polvalente: Is there a test I could add where things would break without this `prewalk`?

nx/lib/nx/defn/compiler.ex
```diff
@@ -421,6 +474,18 @@ defmodule Nx.Defn.Compiler do
     {{call, meta, [ast, fun]}, state}
   end
 
+  defp normalize({{:., _, [Nx.Defn.Assertions, _function]} = call, meta, [ast, fun]}, state) do
```
polvalente, line 477: This should probably not be here, just added to get the tests to pass/code to compile without warnings.

I think this is related to what we talked about over remote calls. Basically, I don't want to normalize args passed to a transform in this case. This came up when replacing the `transform` in `assert_shape_pattern`

polvalente: (this is basically a copy-paste from the :transform clause above, by the way)

nx/lib/nx/lin_alg.ex
```diff
@@ -38,10 +38,17 @@ defmodule Nx.LinAlg do
...
+
+  deftransformp __adjoint__(t, opts) do
+    case t do
+      %{type: {:c, _}} ->
+        t |> Nx.transpose(opts) |> Nx.conjugate()
+
+      _ ->
+        Nx.transpose(t, opts)
+    end
   end
```
polvalente, line 52: I wonder what would be the consequences of exposing `deftransform adjoint` instead of `defn adjoint`

polvalente: Same with `norm`, below

nx/lib/nx/defn.ex
```diff
@@ -686,6 +740,33 @@ defmodule Nx.Defn do
     end
   end
 
+  defp define_transform(kind, call, block, env) do
+    # Note name here is not necessarily an atom due to unquote(name) support
+    {name, args} = decompose_call!(kind, call, env)
+    arity = length(args)
+
+    defaults =
```
polvalente, line 748: we still need to handle defaults, guards and multiclause for deftransform. Are these in scope for this PR?

josevalim: Guards and multiclause yes, those should be easy. I would skip on defaults for now!

josevalim: Actually, let's do it all. It seems the logic for defaults is already ready in the compiler. :)

nx/lib/nx/defn/compiler.ex
```diff
@@ -189,8 +189,16 @@ defmodule Nx.Defn.Compiler do
...
     defns =
-      for {{name, arity}, %{defaults: defaults}} <- exports,
+      for {{name, arity}, %{defaults: defaults}} <- defn_exports,
+          arity <- (arity - map_size(defaults))..arity,
+          do: {name, arity}
+
+    transforms =
+      for {{name, arity}, %{defaults: defaults}} <- transform_exports,
           arity <- (arity - map_size(defaults))..arity,
           do: {name, arity}
```
josevalim, line 203: Extract this to a common function maybe?

nx/lib/nx/lin_alg.ex
```diff
@@ -38,10 +38,17 @@ defmodule Nx.LinAlg do
...
   """
   defn adjoint(t, opts \\ []) do
-    transform(t, fn
-      %{type: {:c, _}} = t -> t |> Nx.transpose(opts) |> Nx.conjugate()
-      t -> Nx.transpose(t, opts)
-    end)
+    __adjoint__(t, opts)
+  end
+
+  deftransformp __adjoint__(t, opts) do
```
josevalim, line 44: Let's call it `do_adjoint` or something?

nx/lib/nx/lin_alg.ex
josevalim: In this case, none, we can probably do it today. In general, the main differences are:

1. Inputs are not converted into expressions
2. and therefore they are not jitted

But given it is a small function, it is fine. We should probably add those notes to `deftransformer` docs.

nx/lib/nx/defn/compiler.ex
josevalim: Yeah... the only way I can think to address this is to introduce some sort of `match_shape` macro. Then we can also introduce `match_type` (which will help clean up adjoint/transform). `match_shape` is easy... so perhaps we can start with it. But `match_type` also has complications because often we just want to ask if something is a float and have it work for both f/bf. Similar for integers.

Anyway, let's add `match_shape`?

nx/lib/nx/defn.ex
```diff
@@ -654,12 +654,67 @@ defmodule Nx.Defn do
...
+    block = body[:do]
+    define_transform(:def, call, block, __CALLER__)
+  end
+
+  @doc """
+  Private function version for `deftransform`
+  """
+  defmacro deftransformp(call, body \\ []) do
+    block = body[:do]
+    define_transform(:defp, call, block, __CALLER__)
```
polvalente, line 712: style question: the current or the following?

```suggestion
    define_transform(:defp, call, body[:do], __CALLER__)
```

polvalente: Or even a separate /1 clause with it's own docs (what would even go there, actually?)

josevalim: I think you should define two separate clauses. Your code as-is  would accept this:

    deftransform foo(a, b), doe: a + b

So make the arity 1 their own functions and explicitly match on `deftransform (call, do: block)` for arity 2. :)

polvalente: What's the issue with accepting the definition like that?

polvalente: Oooh I see, the typo was intentional :)

nx/lib/nx/defn/compiler.ex
josevalim: You can remove this now. :)

nx/test/nx/defn_test.exs
```diff
@@ -198,7 +198,7 @@ defmodule Nx.DefnTest do
   end
 
   describe "arguments" do
-    defn identity(t), do: t
+    defn(identity(t), do: t)
```
josevalim, line 201: Oops, you want to revert those. Something got messed up in the formatter. :)

josevalim (commented):
Looks great! I have added only some minor comments and we can ship this!!!

josevalim (approved):
There is one comment left and we need to also update the module docs for Nx.Defn. Then we can ship it!!!

Issue #817, opened 2022-06-24
Functions set_diagonal(tensor, opts \\ []) , and indexed_set(target, indices, updates)


RD314:
It would be great if `set_diagonal(tensor, opts \\ []) ` and ` indexed_set(target, indices, updates) ` can be added. 

It would complement the operations allowed already. Slicing and access syntax allow setting the values to columns or rows, or individual "cells". 

Setting diagonal or other set of "cells" can be now done only by some cycle accessing and setting each diagonal value. 
There are functions that can be used as a head start: 
1) Accesign diagonal: `make_diagonal(tensor, opts \\ [])`, `take_diagonal(tensor, opts \\ [])`.
2) Adding value to the enumerated list of indexes: ` indexed_add(target, indices, updates)`.

For setting the diagonal the workaround is to take the diagonal values and subtract them from matrix, e.g. 
```
#make diagonal zero and add new value
t = Nx.random_uniform({5,5},1,0.1)
new_diagonal = Nx.random_uniform({5},1, 0.1) 
s = t |> Nx.take_diagonal() |> Nx.make_diagonal()
Nx.subtract(t, s) |> Nx.add(Nx.make_diagonal(new_diagonal))
```
The issue I see with this is that subtracting leads to rounding errors, and efficiency of the workaround.

The function ` indexed_set(target, indices, updates) ` has the similar logic and would be great complement to `indexed_add(target, indices, updates)`.

josevalim:
We would love a PR that adds `Nx.put_diagonal/3`. `Nx.indexed_put/3` would be welcome as well.

polvalente:
@RD314 do you plan on sending a PR on this? Just asking so we avoid having multiple people working on the same thing :)

RD314:
@polvalente the short answer Is no. It would be great if you can address this need.

polvalente:
> @polvalente the short answer Is no. It would be great if you can address this need.

@RD314 don't worry! I just didn't want to steal this from you :)
Hopefully we'll have a PR on this later this week

Benjamin-Philip:
> Hopefully we'll have a PR on this later this week

I am working on this right now.

PR #829, opened 2022-06-28
Add `indexed_put`

Benjamin-Philip:
See #817.

Adds the `indexed_put` function.

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -1064,6 +1064,54 @@ defmodule EXLA.Defn do
     )
   end
 
+  defp to_operator(
+         :indexed_put,
+         [target, indices, updates],
+         %{type: type},
+         state
+       ) do
+    target = to_type(target, type)
+    updates = to_type(updates, type)
+
+    args = [%{type: type, shape: {}}, %{type: type, shape: {}}]
+    scatter_fn = op_computation(:add, args, state)
+
+    rank = target |> op_shape() |> tuple_size()
+    # indices_rank is guaranteed to be 2 by Nx.Shape
+    indices_rank = 2
+    rank_diff = rank - indices_rank + 1
+
+    indices_shape = op_shape(indices)
+
+    indices_shape =
+      [List.duplicate(1, rank_diff) | Tuple.to_list(indices_shape)]
+      |> List.flatten()
+      |> List.to_tuple()
+
+    indices = EXLA.Op.reshape(indices, indices_shape)
+
+    # If indices has shape {x, y}, updates is guaranteed by Nx.Shape to
+    # have shape {x}, so if we reshaped indices to {..., x, y}, we need to
+    # reshape updates to {..., x}
+
+    updates_shape = Tuple.delete_at(indices_shape, tuple_size(indices_shape) - 1)
+
+    updates = EXLA.Op.reshape(updates, updates_shape)
+
+    axes = axes_for_rank(rank)
+
+    EXLA.Op.scatter(
+      target,
+      indices,
+      updates,
+      scatter_fn,
+      rank,
+      [],
+      axes,
+      axes
+    )
+  end
+
   defp to_operator(:map, [arg, _opts, fun], %{shape: shape, type: type}, _state) do
     arg = to_type(arg, type)
     EXLA.Op.map(arg, fun, Nx.axes(shape))
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -1018,105 +1018,28 @@ defmodule EXLA.Defn do
 
   defp to_operator(
          :indexed_add,
-         [target, indices, updates],
-         %{type: type},
+         tensors,
+         %{type: type} = out,
          state
        ) do
-    target = to_type(target, type)
-    updates = to_type(updates, type)
-
     args = [%{type: type, shape: {}}, %{type: type, shape: {}}]
     scatter_fn = op_computation(:add, args, state)
 
-    rank = target |> op_shape() |> tuple_size()
-    # indices_rank is guaranteed to be 2 by Nx.Shape
-    indices_rank = 2
-    rank_diff = rank - indices_rank + 1
-
-    indices_shape = op_shape(indices)
-
-    indices_shape =
-      [List.duplicate(1, rank_diff) | Tuple.to_list(indices_shape)]
-      |> List.flatten()
-      |> List.to_tuple()
-
-    indices = EXLA.Op.reshape(indices, indices_shape)
-
-    # If indices has shape {x, y}, updates is guaranteed by Nx.Shape to
-    # have shape {x}, so if we reshaped indices to {..., x, y}, we need to
-    # reshape updates to {..., x}
-
-    updates_shape = Tuple.delete_at(indices_shape, tuple_size(indices_shape) - 1)
-
-    updates = EXLA.Op.reshape(updates, updates_shape)
-
-    axes = axes_for_rank(rank)
-
-    EXLA.Op.scatter(
-      target,
-      indices,
-      updates,
-      scatter_fn,
-      rank,
-      [],
-      axes,
-      axes
-    )
+    scatter(scatter_fn, tensors, out)
   end
 
-  defp to_operator(
-         :indexed_put,
-         [target, indices, updates],
-         %{type: type},
-         state
-       ) do
-    target = to_type(target, type)
-    updates = to_type(updates, type)
-
+  defp to_operator(:indexed_put, tensors, out, state) do
     # Build update computation
 
     subbuilder = subbuilder(state.builder, "scatter_reduction")
 
-    param_shape = EXLA.Shape.make_shape(type, {})
+    param_shape = EXLA.Shape.make_shape(out.type, {})
     _left = EXLA.Op.parameter(subbuilder, 0, param_shape, "left")
     right = EXLA.Op.parameter(subbuilder, 1, param_shape, "right")
 
     scatter_fn = EXLA.Builder.build(right)
 
-    rank = target |> op_shape() |> tuple_size()
-    # indices_rank is guaranteed to be 2 by Nx.Shape
-    indices_rank = 2
-    rank_diff = rank - indices_rank + 1
-
-    indices_shape = op_shape(indices)
-
-    indices_shape =
-      [List.duplicate(1, rank_diff) | Tuple.to_list(indices_shape)]
-      |> List.flatten()
-      |> List.to_tuple()
-
-    indices = EXLA.Op.reshape(indices, indices_shape)
-
-    # If indices has shape {x, y}, updates is guaranteed by Nx.Shape to
-    # have shape {x}, so if we reshaped indices to {..., x, y}, we need to
-    # reshape updates to {..., x}
-
-    updates_shape = Tuple.delete_at(indices_shape, tuple_size(indices_shape) - 1)
-
-    updates = EXLA.Op.reshape(updates, updates_shape)
-
-    axes = axes_for_rank(rank)
-
-    EXLA.Op.scatter(
-      target,
-      indices,
-      updates,
-      scatter_fn,
-      rank,
-      [],
-      axes,
-      axes
-    )
+    scatter(scatter_fn, tensors, out)
   end
 
   defp to_operator(:map, [arg, _opts, fun], %{shape: shape, type: type}, _state) do
@@ -1338,6 +1261,46 @@ defmodule EXLA.Defn do
       end
 
     apply(exla_op, [tensor, n])
+  end
+
+  defp scatter(scatter_fn, [target, indices, updates], %{type: type}) do
+    target = to_type(target, type)
+    updates = to_type(updates, type)
+
+    rank = target |> op_shape() |> tuple_size()
+    # indices_rank is guaranteed to be 2 by Nx.Shape
+    indices_rank = 2
+    rank_diff = rank - indices_rank + 1
+
+    indices_shape = op_shape(indices)
+
+    indices_shape =
+      [List.duplicate(1, rank_diff) | Tuple.to_list(indices_shape)]
+      |> List.flatten()
+      |> List.to_tuple()
+
+    indices = EXLA.Op.reshape(indices, indices_shape)
+
+    # If indices has shape {x, y}, updates is guaranteed by Nx.Shape to
+    # have shape {x}, so if we reshaped indices to {..., x, y}, we need to
+    # reshape updates to {..., x}
+
+    updates_shape = Tuple.delete_at(indices_shape, tuple_size(indices_shape) - 1)
+
+    updates = EXLA.Op.reshape(updates, updates_shape)
+
+    axes = axes_for_rank(rank)
+
+    EXLA.Op.scatter(
+      target,
+      indices,
+      updates,
+      scatter_fn,
+      rank,
+      [],
+      axes,
+      axes
+    )
   end
 
   ## Cache and hook helpers helpers
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -1073,8 +1073,15 @@ defmodule EXLA.Defn do
     target = to_type(target, type)
     updates = to_type(updates, type)
 
-    args = [%{type: type, shape: {}}, %{type: type, shape: {}}]
-    scatter_fn = op_computation(:add, args, state)
+    # Build update computation
+
+    subbuilder = subbuilder(state.builder, "scatter_reduction")
+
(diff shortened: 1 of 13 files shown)


nx/lib/nx.ex
```diff
@@ -5176,15 +5176,93 @@ defmodule Nx do
...
+      #Nx.Tensor<
+        s64[1][2][3]
+        [
+          [
+            [0, 1, 2],
+            [3, 4, 5]
+          ]
+        ]
+      >
+      iex> indices = Nx.tensor([[0, 0, 0], [0, 1, 1], [0, 0, 0], [0, 0, 2], [0, 1, 2]])
```
Benjamin-Philip, line 5203: Will we allow repeats of a index?

polvalente: I don't see a reason to prohibit them. indexed_add already accepts repeated indices, as well as take/gather and friends.

There would be a performance hit in checking for repeated indices as well

polvalente: We should document the behaviour for this explictly, though

nx/lib/nx.ex
```diff
@@ -5176,15 +5176,93 @@ defmodule Nx do
...
+  @doc """
+  Performs an indexed `set` operation on the `target` tensor,
+  setting the `updates` into the corresponding `indices` positions.
+
+  `indices` must be a fully qualified tensor of shape `{n, Nx.rank(target)}`, with `n`
+  being an arbitrary number of indices, while `updates` must have a compatible `{n}` shape.
+
+  ### Examples
+
+      iex> t = Nx.iota({1, 2, 3})
```
polvalente, line 5193: I think the first example should happen o a zero-tensor instead, for clarity. 
It should also leave one position unset

nx/lib/nx.ex
polvalente, line 5203: I think we should put repeated indices as a separate example to highlight the behavior

nx/lib/nx.ex
```diff
@@ -5176,15 +5176,93 @@ defmodule Nx do
...
+
+      iex> Nx.indexed_set(Nx.tensor([[1], [2]]), Nx.tensor([[1, 2, 3]]), Nx.tensor([0]))
+      ** (ArgumentError) expected indices to have shape {*, 2}, got: {1, 3}
+
+      iex> Nx.indexed_set(Nx.tensor([[1], [2]]), Nx.tensor([[1, 2]]), Nx.tensor([0, 1]))
+      ** (ArgumentError) expected updates tensor to match the first axis of indices tensor with shape {1, 2}, got {2}
+  """
+  @doc type: :indexed
+  def indexed_set(target, indices, updates) do
+    {target, indices, updates, type} = indexed_op(target, indices, updates)
```
polvalente, line 5251: You could pass : indexed_add/:indexed_set as an argument and call apply/3 inside of indexed_op/4

nx/lib/nx.ex
```diff
@@ -5176,15 +5176,93 @@ defmodule Nx do
...
+      ** (ArgumentError) updates must be a rank 1 tensor, got: 2
+
+      iex> Nx.indexed_set(Nx.tensor([[1], [2]]), Nx.tensor([[1, 2, 3]]), Nx.tensor([0]))
+      ** (ArgumentError) expected indices to have shape {*, 2}, got: {1, 3}
+
+      iex> Nx.indexed_set(Nx.tensor([[1], [2]]), Nx.tensor([[1, 2]]), Nx.tensor([0, 1]))
+      ** (ArgumentError) expected updates tensor to match the first axis of indices tensor with shape {1, 2}, got {2}
+  """
+  @doc type: :indexed
+  def indexed_set(target, indices, updates) do
```
jonatanklosko, line 5250: My vote goes to `indexed_put` (the same wording as most built-in data structures), or alternatively `indexed_replace` if we prefer to match `List.replace_at`.

polvalente: `indexed_put` goes well with `put_slice` :)

polvalente: I think José suggested `indexed_put` as the name for this function, actually

Benjamin-Philip: In that case I'll rename the PR and the function.

Benjamin-Philip:
@polvalente, I have implemented `indexed_put` for the binary backend, however the patch is not compiling due to some errors (and the `--warnings-as-errors` flag in the ci). So, in order to run the tests, you will have to checkout locally and test until I fix the warnings.

Benjamin-Philip:
I am going to work on a Torchx implementation now.

nx/lib/nx/shape.ex
```diff
@@ -831,8 +831,8 @@ defmodule Nx.Shape do
 
   defp validate_strides!(_, _), do: :ok
 
-  @doc "Validates the input shapes for `Nx.indexed_add/3`"
-  def indexed_add(
+  @doc "Validates the input shapes for `Nx.indexed_*/3`"
+  def indexed(
```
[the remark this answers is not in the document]
jonatanklosko: This is a private module, so we are good :)

nx/lib/nx.ex
```diff
@@ -5176,15 +5176,114 @@ defmodule Nx do
...
+    impl = fn target, indices, updates, out ->
+      impl!(target, indices, updates).indexed_add(out, target, indices, updates)
+    end
+
+    indexed_op(target, indices, updates, impl)
+  end
+
+  @doc """
+  Performs an indexed `set` operation on the `target` tensor,
+  setting the `updates` into the corresponding `indices` positions.
```
jonatanklosko, line 5188:
```suggestion
  Puts individual values from `updates` into the given tensor at the corresponding `indices`.
```

nx/lib/nx.ex
```diff
@@ -5176,15 +5176,114 @@ defmodule Nx do
...
+      ** (ArgumentError) updates must be a rank 1 tensor, got: 2
+
+      iex> Nx.indexed_put(Nx.tensor([[1], [2]]), Nx.tensor([[1, 2, 3]]), Nx.tensor([0]))
+      ** (ArgumentError) expected indices to have shape {*, 2}, got: {1, 3}
+
+      iex> Nx.indexed_put(Nx.tensor([[1], [2]]), Nx.tensor([[1, 2]]), Nx.tensor([0, 1]))
+      ** (ArgumentError) expected updates tensor to match the first axis of indices tensor with shape {1, 2}, got {2}
+  """
+  @doc type: :indexed
+  def indexed_put(target, indices, updates) do
```
jonatanklosko, line 5267: For consistency with other functions :)

```suggestion
  def indexed_put(tensor, indices, updates) do
```

nx/lib/nx/binary_backend.ex
```diff
@@ -1671,12 +1671,40 @@ defmodule Nx.BinaryBackend do
...
+
+  # # Parameters
+
+  # - `out` - output tensor
+  # - `target` - target tensor
+  # - `indices` - tensor with indices to update
+  # - `updates` - tensor with updates to make
+  # - `resolve_updates` - anonymous function to resolve repeated updates
+  # - `update_element` - anonymous function that updates an element, given
+  #    the current element and the update
```
jonatanklosko, line 1699: I think it's fine to skip this, since it's a private function and from context it's pretty straightforward what the arguments are :)

nx/test/nx_test.exs
```diff
@@ -818,6 +818,88 @@ defmodule NxTest do
...
+      n = 5
+
+      indices =
+        Nx.tensor(
+          for row <- 0..(n - 1), col <- 0..(n - 1) do
+            [row, col]
+          end
+        )
+
+      updates = 0..(n * n - 1) |> Enum.map(fn _ -> 1 end) |> Nx.tensor()
```
jonatanklosko, line 832:
```suggestion
      updates = List.duplicate(1, n * n) |> Nx.tensor()
```

nx/test/nx_test.exs
```diff
@@ -818,6 +818,88 @@ defmodule NxTest do
...
+
+    test "single-index regression" do
+      n = 5
+
+      # 1D case
+      zeros = List.duplicate(0, n)
+      upd = Nx.tensor([1])
+
+      for i <- 0..(n - 1) do
+        result = Nx.tensor(List.update_at(zeros, i, fn _ -> 1 end))
```
jonatanklosko, line 856:
```suggestion
        result = Nx.tensor(List.replace_at(zeros, i, 1))
```

nx/lib/nx.ex
```diff
@@ -5176,15 +5176,114 @@ defmodule Nx do
...
+
+  @doc """
+  Performs an indexed `set` operation on the `target` tensor,
+  setting the `updates` into the corresponding `indices` positions.
+
+  `indices` must be a fully qualified tensor of shape `{n, Nx.rank(target)}`, with `n`
+  being an arbitrary number of indices, while `updates` must have a compatible `{n}` shape.
+
+  In case of repeating indices, the last occurence of index and its corresponding update
+  value wins.
```
polvalente, line 5194:
```suggestion
  value takes precedence.
```

nx/lib/nx.ex
```diff
@@ -5176,15 +5176,114 @@ defmodule Nx do
...
+
```
polvalente, line 5195:
```suggestion

  See also: `indexed_add/3`, `gather/2`, `take/3`, `take_along_axis/3`
```

nx/lib/nx.ex
```diff
@@ -5176,15 +5176,114 @@ defmodule Nx do
...
+        s64[1][2][3]
+        [
+          [
+            [2, 1, -2],
+            [3, 3, 5]
+          ]
+        ]
+      >
+
+  Type promotions should happen automatically.
```
polvalente, line 5233:
```suggestion
  Type promotions should happen automatically, with the resulting type being the combination of the `target` type and the `updates` type.
```

nx/lib/nx.ex
polvalente: I would keep it as `target` for consistency with `indexed_add` and because we're operating on 3 different tensors here

nx/lib/nx.ex
```diff
@@ -5176,15 +5176,114 @@ defmodule Nx do
...
 
     type = binary_type(target, updates)
 
-    Nx.Shape.indexed_add(target, indices, updates)
+    Nx.Shape.indexed(target, indices, updates)
+
+    out = %{target | type: type}
 
-    impl!(target, indices, updates).indexed_add(%{target | type: type}, target, indices, updates)
+    impl.(target, indices, updates, out)
```
polvalente, line 5286:
```suggestion
    apply(impl!(target, indices, updates), function, [out, target, indices, updates])
```

nx/lib/nx.ex
```diff
@@ -5176,15 +5176,114 @@ defmodule Nx do
...
+  @doc type: :indexed
+  def indexed_put(target, indices, updates) do
+    impl = fn target, indices, updates, out ->
+      impl!(target, indices, updates).indexed_put(out, target, indices, updates)
+    end
+
+    indexed_op(target, indices, updates, impl)
+  end
+
+  defp indexed_op(target, indices, updates, impl) do
```
polvalente, line 5275:
```suggestion
    indexed_op(target, indices, updates, :indexed_put)
  end

  defp indexed_op(target, indices, updates, function) do
```

polvalente: This and the suggestion below seems better to avoid floating anonymous functions around. Also makes things clearer in the sense that the implicit contract for the anonymous function had `out` as the last argument instead of the usual first one

nx/lib/nx/backend.ex
```diff
@@ -92,6 +92,7 @@ defmodule Nx.Backend do
   @callback window_scatter_max(out :: tensor, tensor, tensor, tensor, shape, keyword) :: tensor
   @callback window_scatter_min(out :: tensor, tensor, tensor, tensor, shape, keyword) :: tensor
   @callback indexed_add(out :: tensor, tensor, tensor, tensor) :: tensor
+  @callback indexed_put(out :: tensor, tensor, tensor, tensor) :: tensor
```
polvalente, line 95:
```suggestion
  @callback indexed_add(out :: tensor, target :: tensor, indices :: tensor, updates :: tensor) :: tensor
  @callback indexed_put(out :: tensor, target :: tensor, indices :: tensor, updates :: tensor) :: tensor
```

nx/lib/nx/binary_backend.ex
```diff
@@ -1671,12 +1671,40 @@ defmodule Nx.BinaryBackend do
...
   @impl true
-  def indexed_add(
-        %T{} = out,
-        %T{shape: shape, type: {_, target_size}} = target,
-        %T{shape: {indices_rows, _indices_cols} = indices_shape} = indices,
-        %T{shape: {indices_rows}} = updates
-      ) do
+  def indexed_add(out, target, indices, updates) do
+    resolve_updates = fn upds -> Enum.reduce(upds, 0, &+/2) end
+    update_element = fn cur, upd -> cur + upd end
```
polvalente, line 1676:
```suggestion
    update_element = &+/2
```

nx/lib/nx.ex
jonatanklosko: Yeah, what I meant was to align `indexed_add` as well. If we want `target` then we should align `put_slice` as well, though I think `tensor` works for the leading argument, especially if it matches the wording in the docs. Sidenote: `take`/`take_along_axis`/`gather` use `tensor` instead of `source` :D

polvalente: I'd rather align other functions with `target`/`source` terminology, but I'm not set too hard on this

jonatanklosko: Me neither, @josevalim feel free to opine on this :D

Benjamin-Philip: I think that indexed_add, and indexed_put should retain the target terminology as it resolves any ambiguity in the difference between the 3 tensors.

On the other hand, for other functions, I don't think it makes any difference except the 3 tensors.

Benjamin-Philip:
Should I make the changes that apply to indexed_add as well in this PR, or in a seperate PR?

polvalente:
> Should I make the changes that apply to indexed_add as well in this PR, or in a seperate PR?

@Benjamin-Philip you can use this same PR!

exla/lib/exla/defn.ex
```diff
@@ -1064,6 +1064,54 @@ defmodule EXLA.Defn do
...
+         :indexed_put,
+         [target, indices, updates],
+         %{type: type},
+         state
+       ) do
+    target = to_type(target, type)
+    updates = to_type(updates, type)
+
+    args = [%{type: type, shape: {}}, %{type: type, shape: {}}]
+    scatter_fn = op_computation(:add, args, state)
```
Benjamin-Philip, line 1077: I'm having some trouble to find an update computation that will return the rhs, so that we can scatter in place instead of adding.

jonatanklosko: You can build the computation from scratch, something like this:

```elixir
defp scatter_reduction_computation(state) do
  subbuilder = subbuilder(state.builder, :scatter_reduction)

  param_shape = EXLA.Shape.make_shape(type, {})
  left = EXLA.Op.parameter(subbuilder, 0, param_shape, "left")
  right = EXLA.Op.parameter(subbuilder, 1, param_shape, "right")
  
  tuple = EXLA.Op.tuple(subbuilder, [left, right])
  result = EXLA.Op.get_tuple_element(subbuilder, 1)
  
  EXLA.Builder.build(result)
end
```

Benjamin-Philip:
Done!

exla/lib/exla/defn.ex
Benjamin-Philip: Thanks!

exla/lib/exla/defn.ex
```diff
@@ -1064,6 +1064,61 @@ defmodule EXLA.Defn do
...
+    EXLA.Op.scatter(
+      target,
+      indices,
+      updates,
+      scatter_fn,
+      rank,
+      [],
+      axes,
+      axes
+    )
```
jonatanklosko, line 1119: The only difference from `:indexed_add` is the `scatter_fn` right? We can move all of that to private `scatter` function and pass the built computation as an additional argument.

Benjamin-Philip: Yes. That's what I meant by "DRYing".

jonatanklosko (approved):
Fantastic :)

Benjamin-Philip:
Sure thing! I also want to implement put_diagonal.

nx/lib/nx.ex
josevalim: I would go with `tensor` for consistency. indices and updates is enough to make it clear what it is.

PR #831, opened 2022-07-02
feat(defn): support booleans

polvalente:
This PR adds support for booleans as tensor inputs (true -> 1, false -> 0),
which means that:

- We now support literal true and false being used as explicit tensor constants (i.e. `Nx.tensor(true)`)
~- Booleans are now valid inputs for defn and for Nx functions~
- logical and comparison expressions also work accordingly with booleans

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -375,7 +375,7 @@ defmodule Nx do
   See also: `t:t/0`
   """
   @doc type: :guards
-  defguard is_tensor(t) when is_number(t) or is_struct(t, T) or is_struct(t, Complex)
+  defguard is_tensor(t) when is_number(t) or is_struct(t, T) or is_struct(t, Complex) or is_boolean(t)
 
   ## Creation API
 
@@ -698,7 +698,7 @@ defmodule Nx do
 
   defp tensor_or_number_to_binary(true, type), do: tensor_or_number_to_binary(1, type)
   defp tensor_or_number_to_binary(false, type), do: tensor_or_number_to_binary(0, type)
-  
+
   defp tensor_or_number_to_binary(%Complex{re: re, im: im}, {:c, size}) do
     number_to_binary(re, {:f, div(size, 2)}) <> number_to_binary(im, {:f, div(size, 2)})
   end
@@ -1638,6 +1638,9 @@ defmodule Nx do
     out = %T{shape: {}, type: {:c, size * 2}, names: []}
     backend.constant(out, number, options)
   end
+
+  def to_tensor(true), do: to_tensor(1)
+  def to_tensor(false), do: to_tensor(0)
 
   def to_tensor(t) do
     raise ArgumentError, "expected a %Nx.Tensor{} or a number, got: #{inspect(t)}"
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -610,10 +610,9 @@ defmodule Nx do
     Enum.reduce(tail, infer_type(head), &Nx.Type.merge(infer_type(&1), &2))
   end
 
-  defp infer_type(value) when is_boolean(value), do: {:u, 8}
-
   defp infer_type(number)
-       when is_number(number) or is_struct(number, Complex) or number in @non_finite do
+       when is_number(number) or is_struct(number, Complex) or number in @non_finite or
+              is_boolean(number) do
     Nx.Type.infer(number)
   end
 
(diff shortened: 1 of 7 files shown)


polvalente:
@josevalim I'm not 100% sure if this is the approach you wanted to take. This PR ended up supporting `true` and `false` basically anywhere where `1` and `0` would be upcast to tensor expressions.

There's a test that shows that `true + 1 == 2`, which smells a little to me, though.

nx/lib/nx.ex
```diff
@@ -375,7 +375,8 @@ defmodule Nx do
   See also: `t:t/0`
   """
   @doc type: :guards
-  defguard is_tensor(t) when is_number(t) or is_struct(t, T) or is_struct(t, Complex)
+  defguard is_tensor(t)
+           when is_number(t) or is_struct(t, T) or is_struct(t, Complex) or is_boolean(t)
```
josevalim, line 379: I don't think we should do this change (for the same reason we don't include `is_list(t)` here).

nx/lib/nx.ex
```diff
@@ -1609,6 +1639,14 @@ defmodule Nx do
     backend.constant(out, number, options)
   end
 
+  def to_tensor(bool) when is_boolean(bool) do
```
josevalim, line 1642: We don't want this change either. :) The goal is to only allow it in `Nx.tensor`.

nx/lib/nx/defn/composite.ex
```diff
@@ -135,8 +135,9 @@ defmodule Nx.Defn.Composite do
   defp flatten_each(%T{} = tensor, acc, _fun),
     do: [tensor | acc]
 
-  defp flatten_each(number, acc, fun) when is_number(number) or is_struct(number, Complex),
-    do: [fun.(number) | acc]
+  defp flatten_each(number, acc, fun)
+       when is_number(number) or is_struct(number, Complex) or is_boolean(number),
```
josevalim, line 139: We can also revert this. :)

nx/lib/nx/shared.ex
```diff
@@ -377,16 +377,19 @@ defmodule Nx.Shared do
...
   Builds the type of an element-wise binary operation.
   """
-  def binary_type(a, b) when is_number(a) and is_number(b), do: Nx.Type.infer(a + b)
-  def binary_type(a, b) when is_number(a), do: Nx.Type.merge_number(type(b), a)
-  def binary_type(a, b) when is_number(b), do: Nx.Type.merge_number(type(a), b)
+  def binary_type(a, b) when (is_number(a) or is_boolean(a)) and (is_number(b) or is_boolean(b)),
+    do: Nx.Type.infer(a + b)
+
+  def binary_type(a, b) when is_number(a) or is_boolean(a), do: Nx.Type.merge_number(type(b), a)
+  def binary_type(a, b) when is_number(b) or is_boolean(b), do: Nx.Type.merge_number(type(a), b)
```
josevalim, line 384: We should revert these changes too, as the code here is meant to work with tensors and true/false are not tensor equivalents.

nx/lib/nx/shared.ex
```diff
@@ -377,16 +377,19 @@ defmodule Nx.Shared do
...
   def binary_type(a, b), do: Nx.Type.merge(type(a), type(b))
 
   # For unknown types, return {:f, 32} as the caller
   # should validate the input in a later pass.
   defp type(%T{type: type}), do: type
   defp type({_, _} = type), do: type
   defp type(%Complex{}), do: {:c, 64}
+  defp type(bool) when is_boolean(bool), do: {:u, 8}
```
josevalim, line 392: Same here maybe?

nx/lib/nx/type.ex
```diff
@@ -495,6 +496,10 @@ defmodule Nx.Type do
     {:f, 32}
   end
 
+  def merge_number(type, bool) when is_boolean(bool) do
```
josevalim, line 499: I think this one we should revert.

nx/lib/nx/type.ex
```diff
@@ -130,6 +130,7 @@ defmodule Nx.Type do
   """
   def infer(value) when is_integer(value), do: {:s, 64}
   def infer(value) when is_float(value), do: {:f, 32}
+  def infer(value) when is_boolean(value), do: {:u, 8}
```
josevalim, line 133: :+1:

josevalim:
> @josevalim I'm not 100% sure if this is the approach you wanted to take. This PR ended up supporting true and false basically anywhere where 1 and 0 would be upcast to tensor expressions.

Yeah, we want to do less. We will support booleans only on `Nx.tensor` and on the operators in Kernel that already support booleans, that's it. :)

nx/lib/nx/defn/composite.ex
```diff
@@ -192,8 +193,12 @@ defmodule Nx.Defn.Composite do
   def to_inputs(args) do
     {args, _} =
       Enum.map_reduce(args, 0, fn arg, i ->
-        traverse(arg, i, fn arg, i ->
-          {Expr.parameter(Nx.to_tensor(arg), :root, i), i + 1}
+        traverse(arg, i, fn
+          arg, _i when is_boolean(arg) ->
+            raise ArgumentError, "booleans are not supported as defn inputs"
```
polvalente, line 198: I've added this for better error messages when booleans are passed as inputs. Since defns are now aware of booleans, I thought it was better than letting a ProtocolUndefinedError raise for Nx.Container

polvalente: This is also a point where we could decide to support booleans as defn inputs through calling Nx.tensor on them as well, though

josevalim (approved):

josevalim:

josevalim:
I went ahead and merged it cause I will work on top of this branch. :)

PR #832, opened 2022-07-04
Add raise/1-2 and inspect/1-2 to Nx.Defn.Kernel

josevalim:
* inspect_expr and inspect_value have been deprecated
  in favor of print_expr and print_value

* assert_shape and assert_shape_pattern have been deprecated
  in favor of case+raise

### DIFF
--- a/nx/lib/nx/defn/kernel.ex
+++ b/nx/lib/nx/defn/kernel.ex
@@ -859,8 +859,7 @@ defmodule Nx.Defn.Kernel do
           {:badkey, key} ->
             Kernel.raise(
               ArgumentError,
-              "unknown key #{Kernel.inspect(key)} in #{Kernel.inspect(keyword)}, " <>
-                "expected one of #{Kernel.inspect(keys)}"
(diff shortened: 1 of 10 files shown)


nx/lib/nx/defn/kernel.ex
```diff
@@ -1488,111 +1421,143 @@ defmodule Nx.Defn.Kernel do
...
 
   @doc """
-  Asserts the `tensor` has a certain `shape`.
+  Raises a runtime exception with the given `message`.
 
-  If it succeeds, it returns the given tensor. Raises
-  an error otherwise.
+  See `raise/2` for more information on exceptions inside `defn`.
+  """
+  defmacro raise(message) do
```
josevalim, line 1438:
```suggestion
  # It needs to be a macro so we don't add stacktrace entries.
  # Since there is no defdelegate for macros, we do it manually.
  defmacro raise(message) do
```

nx/lib/nx/lin_alg.ex
```diff
@@ -458,7 +458,7 @@ defmodule Nx.LinAlg do
   ### Error cases
```
polvalente, line 458: It seems that `norm` doesn't need transforms anymore

PR #833, opened 2022-07-04
fix: support unary plus and minus for case

polvalente:


nx/lib/nx/shared.ex
```diff
@@ -290,8 +290,8 @@ defmodule Nx.Shared do
       defmacro unquote(name)(unquote_splicing(args)) do
         {module, name} =
           case __CALLER__.context do
-            :guard -> {Kernel, unquote(name)}
-            _ -> {__MODULE__, unquote(fallback)}
+            ctx when ctx not in [:guard, :match] -> {__MODULE__, unquote(fallback)}
```
polvalente, line 293: As I'm not sure which values are valid, I chose to explicitly allow :guard and :match only

josevalim (approved):

nx/test/nx/defn_test.exs
```diff
@@ -243,6 +243,26 @@ defmodule Nx.DefnTest do
   describe "unary ops" do
     defn exp(t), do: Nx.exp(t)
 
+    defn unary_plus_minus_guards(t, opts \\ []) do
+      case opts[:value] do
+        value when value == -2 ->
+          Nx.add(t, value)
```
josevalim, line 249: Make this return -1 or 1 in order to simplify the tests. See the tests I have added for `case` itself.

josevalim (changes requested):
Let's simplify the tests a bit!

nx/lib/nx/shared.ex
josevalim: it is only guard/match/nil.

josevalim (approved):

PR #839, opened 2022-07-18
feat: add grad for indexed_put and indexed_add

polvalente:
Adds grad for `Nx.indexed_put/3` and `Nx.indexed_add/3`. This also enabled grad for `Nx.put_diagonal/3` and `Nx.make_diagonal/2`, which use `indexed_put` in their implementations

### DIFF
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -418,6 +418,26 @@ defmodule Nx.Defn.Grad do
     [{x, operand_t}, {update, update_t}]
   end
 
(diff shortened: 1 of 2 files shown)


polvalente:
@Benjamin-Philip I went ahead and implemented these grads because I noticed that `indexed_add` also didn't have them implemented

josevalim (approved):

Benjamin-Philip:
> @Benjamin-Philip I went ahead and implemented these grads because I noticed that `indexed_add` also didn't have them implemented

Cool. No worries.

PR #840, opened 2022-07-18
feat: add Nx.is_nan/1

### LINKED ISSUE #814: Identify NaN in tensors
I think it would be useful to have inside Nx.defn functions like is_number or is_nan. 
It would be useful especially when working with grad that could return NaN and gradient is used in subsequent formulas.

polvalente:
closes #814


exla/c_src/exla/exla.cc
```diff
@@ -2294,152 +2299,152 @@ ERL_NIF_TERM start_log_sink(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+    {"tanh", 1, tanh},
+    {"acosh", 1, acosh},
+    {"asinh", 1, asinh},
+    {"atanh", 1, atanh},
+    {"real", 1, real},
+    {"imag", 1, imag},
+    {"sqrt", 1, sqrt},
+    {"rsqrt", 1, rsqrt},
+    {"cbrt", 1, cbrt},
+    {"is_nan", 1, is_nan},
```
polvalente, line 2382: I'm reverting this formatting

nx/lib/nx/binary_backend.ex
```diff
@@ -841,6 +841,24 @@ defmodule Nx.BinaryBackend do
   @impl true
   def bitwise_not(out, tensor), do: element_wise_unary_op(out, tensor, &:erlang.bnot/1)
 
+  @impl true
+  def is_nan(out, %{type: {t, n_bits}}) when t in [:u, :s] do
+    # integers cannot represent nans, so we can just create
+    # a zero boolean tensor
+
+    size = Tuple.product(out.shape) * div(n_bits, 8)
```
josevalim, line 849:
```suggestion
    size = Nx.size(out.shape) * div(n_bits, 8)
```

That seems to be what we use in this module.

josevalim (approved):

PR #843, opened 2022-07-19
feat: add is_infinity

polvalente:


exla/lib/exla/op.ex
```diff
@@ -212,6 +212,29 @@ defmodule EXLA.Op do
...
+  end
+
+  def is_infinity(op, type) do
+    %{ref: ref} =
+      case type do
+        {t, _} when t in [:f, :bf] ->
+          op
+
+        _ ->
+          convert_element_type(op, Nx.Type.to_floating(type))
```
josevalim, line 231: Sorry, I should have caught this in the nan review. :) If the tensor is not f or bf, shouldn't we simply multiply it by zero instead?

nx/lib/nx/defn/expr.ex
```diff
@@ -366,7 +366,20 @@ defmodule Nx.Defn.Expr do
...
+        :sqrt,
+        :rsqrt,
+        :cbrt,
+        :is_nan,
+        :is_infinity,
+        :negate,
+        :sign,
+        :abs,
+        :bitwise_not
+      ] ++
```
josevalim, line 382: Pls move `is_nan` and `is_infinity` to a separate line so this doesn't break apart. :)

exla/lib/exla/op.ex
polvalente: Yeah, that makes sense :)

exla/lib/exla/op.ex
```diff
@@ -198,20 +198,45 @@ defmodule EXLA.Op do
...
   end
 
+  def is_infinity(op, {t, _}, _shape, _axes, _state) when t in [:f, :bf] do
+    %{ref: ref} = op
+    result_ref = EXLA.NIF.is_infinity(ref) |> unwrap!()
+    %{op | ref: result_ref}
+  end
+
+  def is_infinity(_op, _type, shape, axes, %{builder: builder}) do
+    # For non-floating types, we can just return
```
polvalente, line 232: @josevalim is this close to what you wanted?
I went down this route because multiplication would also need to be followed by conversion to u8 afterwards

polvalente: Also: should we merge is_nan and is_infinity into a is_non_finite function here that receives the operation as an argument? One is the copy-paste of the other aside from the NIF function itself

seanmor5:
Just a nitpick, what is the reason for negating is_finite and offering this API over just is_finite?

polvalente:
> Just a nitpick, what is the reason for negating is_finite and offering this API over just is_finite? 

Libtorch provides "isinf" so either Torchx or EXLA would need to be negated in that sense.

I chose the name before the implementation in either one, actually, mostly because we assume everything to be finite numbers by default. It also pairs up nicely with is_nan, in the sense that both of them search for special entry values

exla/lib/exla/op.ex
josevalim: Looks good! Should we encapsulate it in a function to reuse it?

josevalim (approved):

josevalim:
We may want to add is_finite anyway because many times we want to lookup for the non-finite values.

exla/lib/exla/op.ex
polvalente: Yeah, I think I'll reuse the same code of is_nan and is_infinity

polvalente:
Turns out there was Xla::IsInf available (and using IsFinite was a bug due to nan evaluating as non-finite as well). Merging this as soon as CI passes :)

Issue #333, opened 2021-03-13
Backend-agnostic stateless PRNGs


polvalente:
As discussed in #331, we need an approach for having portable PRNGs that can be applied for any backend.

> We definitely need the ability to seed RNGs, and I'm confident this approach will work for Torchx and other backends, but I'm not 100% sure how it will extend to EXLA and other Tensor compilers that can't depend on a stateful RNG. As an example, we can set the RNG seed as an executable run option in EXLA, but that would apply to the entire execution and not individual calls to `random_uniform` within `defn`. It would also need to be passed as a compile option rather than directly to calls to `random_x`.
> 
> I think it probably is best to move forward instead with Jax-style stateless PRNGs because we can implement them with our current API and have a solution that extends to every compiler and backend. The JAX PRNGs also tout themselves as perfect for distribution/parallel computing, and I think that aligns with some of our future goals. In order to rework this from the EXLA perspective, it would probably involve using `RngBitGenerator` and reimplementing our current random functions in terms of that and other primitives.

_Originally posted by @seanmor5 in https://github.com/elixir-nx/nx/issues/331#issuecomment-798775291_

vans163:
Leaving some notes here

JAX PRNG https://github.com/google/jax/blob/537e35b0fa2c2126cdd22a2e346b65ce11395f80/jax/_src/prng.py  
uses ThreeFry https://bashtage.github.io/randomgen/bit_generators/threefry.html  which allows parallel application use cases via its `key system`  

If we set the RngSeed in XLA, https://github.com/pytorch/xla/blob/master/torch_xla/csrc/tensor.cpp#L346, would this apply to all future kernels as well (as it seems to be set on the DeviceCtx)? Would future kernels reset seed to 0 if the context does not restart? (Just thinking of a temporary quick stopgap that can be depreciated later once proper PRNG support rolls out)

PR #846, opened 2022-07-19
Stateless PRNG with splittable keys proposal #333

john-sonz:
- Implementation of stateless PRNG, based on splittable keys and the Threefry algorithm 
- Based on the JAX PRNG implementation and yield the same results
- At the moment includes only `randint` and `uniform` generators

#333


polvalente (commented):
This is looking great overall!

I had to stop the review session, but I'm submitting the comments so far :)

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+    y2 =
+      rotate_left(xs[1], rot)
+      |> Nx.bitwise_xor(y1)
+
+    Nx.stack([y1, y2])
+    |> Nx.as_type({:u, 32})
+  end
+
+  defnp rolled_loop_step(i, {xs, ks, rs}) do
+    {k1, k2, k3} = {ks[0], ks[1], ks[2]}
```
polvalente, line 249: I'm not sure I like this. It's just to use single-line assignments?

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+    {r1, r2} = {rs[0], rs[1]}
+
+    {_, xs, _} =
+      while {x = 0, xs, rs}, Nx.less(x, 4) do
```
polvalente, line 253:
```suggestion
      while {x = 0, xs, rs}, x < 4 do
```

This works exactly the same :)

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+        {x + 1, apply_round(xs, rs[0][x]), rs}
+      end
+
+    xs1 =
+      Nx.broadcast(k1, xs[0])
+      |> Nx.add(xs[0])
```
polvalente, line 259:
```suggestion
    xs1 = k1 + xs[0]
```

doesn't this work?

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+
+    xs2 =
+      Nx.broadcast(k2 + i + 1, xs[1])
+      |> Nx.add(xs[1])
```
polvalente, line 263:
```suggestion
    xs2 = k2 + i + 1 + xs[1]
```

if this does indeed work, I'm not 100% sure on the shapes here

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+
+    new_ks = Nx.stack([k2, k3, k1])
+    new_rs = Nx.stack([r2, r1])
+
+    {new_xs, new_ks, new_rs}
+  end
+
+  defnp rotate_left(x, rot) do
+    nbits = 32
+    x <<< rot ||| x >>> (nbits - rot)
```
polvalente, line 277: nbits should be a module attribute IMO

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+
+  @doc """
+  Sample uniform random integer values in [minval, maxval).
+  
+  ## Options
+  
+    * `:type` - an int type for the returned tensor
+  
+    * `:shape` - shape of the returned tensor
+
```
polvalente, line 297:
```suggestion
    * `:type` - the integer type for the returned tensor
    * `:shape` - shape of the returned tensor
  

```

Does it need to be integer, though? I'd guess f32 would work just fine, for instance...

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+  Sample uniform random integer values in [minval, maxval).
+  
+  ## Options
+  
+    * `:type` - an int type for the returned tensor
+  
+    * `:shape` - shape of the returned tensor
+  
+  """
+  defn randint(key, minval, maxval, opts \\ []) do
```
polvalente, line 299:
```suggestion
  defn random_integer(key, min_val, max_val, opts \\ []) do

```

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+    opts = keyword!(opts, shape: {1}, type: {:s, 32})
+    shape = opts[:shape]
+    type = {_, nbits} = opts[:type]
+
+    transform(key, &randint_transform(&1, type))
```
polvalente, line 304: use deftransformp for randint_transform instead

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+
+    keys = transform(key, &split/1)
```
polvalente, line 306: deftransformp as well

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+
+    minval = Nx.broadcast(minval, shape)
+    maxval = Nx.broadcast(maxval, shape)
+
+    higher_bits = transform(keys[0], &random_bits(&1, shape))
+    lower_bits = transform(keys[1], &random_bits(&1, shape))
```
polvalente, line 312: deftransformp for random_bits

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+    span = maxval - minval
+
+    multiplier =
+      Nx.power(2, Nx.quotient(nbits, 2))
+      |> Nx.remainder(span)
+      |> Nx.power(2)
+      |> Nx.remainder(span)
```
polvalente, line 319: I think a comment on what this multiplier is would be nice

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+
+    offset =
+      higher_bits
```
polvalente, line 322: same with the offset, it would be nice to add a comment on where this calculation comes from (or a general reference for both at the top of the function)

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+  
+  ## Options
+  
+    * `:type` - an int type for the returned tensor
+  
+    * `:shape` - shape of the returned tensor
+  
+    * `:minval` - minimum value, default is 0.0
+  
+    * `:maxval` - maximum value, default is 1.0
```
polvalente, line 351:
```suggestion
    * `:type` - an int type for the returned tensor  
    * `:shape` - shape of the returned tensor  
    * `:minval` - minimum value, default is 0.0  
    * `:maxval` - maximum value, default is 1.0

```

polvalente: the type description doesn't match the usage

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+  
+  """
+  defn uniform(key, opts \\ []) do
+    opts = keyword!(opts, shape: {1}, type: {:f, 32}, minval: 0.0, maxval: 1.0)
+
+    transform(key, &uniform_transform(&1, opts))
+
+    info = transform(opts[:type], &finfo(&1))
```
polvalente, line 359: deftransformp

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+
+    shape = opts[:shape]
+    type = {_dtype, nbits} = opts[:type]
+    minval = opts[:minval]
+    maxval = opts[:maxval]
+
+    u_one = Nx.tensor(1.0, type: type) |> Nx.bitcast({:u, nbits})
```
polvalente, line 366: Is there any way to avoid bitcast here?
We can't have bitcast in Torchx.

Since it's the beggining of the function, I would rather do the bitcast in a transform in pure elixir if it is indeed necessary.

nx/lib/nx/random.ex
polvalente: finfo and mant should be replaced with more explicit names

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+
+    transform(key, &random_bits(&1, shape))
+    |> Nx.as_type({:u, nbits})
+    |> Nx.right_shift(Nx.tensor(nbits - info[:mant], type: {:u, nbits}))
+    |> Nx.bitwise_or(u_one)
+    |> Nx.bitcast(type)
+    |> Nx.subtract(Nx.tensor(1.0, type: type))
```
polvalente, line 373:
```suggestion
    |> Nx.subtract(1)

```

nx/lib/nx/random.ex
polvalente: u_one isn't actually one, right? What is the idea here?

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+    |> Nx.multiply(maxval - minval)
+    |> Nx.add(minval)
+    |> Nx.reshape(shape)
+    |> Nx.max(minval)
```
polvalente, line 377: is this `max` necessary? What is the range of numbers output at `|> Nx.bitcast(type)`?

nx/lib/nx/random.ex
VenomSolo: > u_one isn't actually one, right? What is the idea here?

It let's us preserve the exponent of 1 (we are only randomizing fraction/mantissa)

nx/lib/nx/random.ex
VenomSolo: I believe we are randomizing integers and floats in distinct enough ways to justify existence of both randint and uniform. I may, however, look into merging those.

nx/lib/nx/random.ex
VenomSolo: It is. Would it be better to just make separate assignments?

polvalente: Yes :)

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+
+  defnp apply_round(xs, rot) do
+    y1 = Nx.add(xs[0], xs[1])
+
+    y2 =
+      rotate_left(xs[1], rot)
+      |> Nx.bitwise_xor(y1)
+
+    Nx.stack([y1, y2])
+    |> Nx.as_type({:u, 32})
```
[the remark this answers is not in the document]
VenomSolo: I'll add comments temporarily since the implementations works mostly for 32bit numbers right now.

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+    even? = rem(count_size, 2) == 0
+
+    if even? do
+      count
+    else
+      count
+      |> Nx.flatten()
+      |> Nx.slice_along_axis(0, count_size, axis: 0)
+    end
+  end
```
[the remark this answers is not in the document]
VenomSolo: That's pretty neat suggestion, I'll try it out since it was my first time wrapping my head around defn which needs that kind of branching.

nx/lib/nx/random.ex
polvalente: They can remain separate. Randint will only return integers and uniform returns floats in the range (a discrete vs continuous kind of domain). My comment was just related to the output typing.

[1,2,3] can be represented by an f32 tensor without any problems

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+
+  def fold_in(key, data) when is_integer(data) do
+    assert_key(key)
+
+    impl(key, key(data))
+  end
+
+  def random_bits(key, shape \\ {1}) do
+    assert_key(key)
```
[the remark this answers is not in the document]
VenomSolo: We left assert_key as it is to raise an appropriate error in case of wrong key type because there are mentioned functions that validate shape but I didn't find any that could validate type in defn.

Also, it's the first time I've heard of deftransform. It could use it's entry in documentation.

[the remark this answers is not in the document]
polvalente: It will be once 0.3 is released! It's a recent addition in an effort to make transforms easier to work with :)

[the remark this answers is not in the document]
VenomSolo: That's cool. I'll leave some memo to come back and change this once 0.3 hits the road.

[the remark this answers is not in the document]
polvalente: Oh, but since you're coding in Nx itself it already is available!

You can either generate the docs locally or search for the definition (in Nx.Defn or Nx.Defn.Kernel, can't recall which one :sweat-smile:)

VenomSolo:
I've implemented most of the suggestions, some of them are still a little clunky since we don't really have support for 64bits. Probably missed some things, will look for them later.

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,374 @@
+defmodule Nx.Random do
+  @moduledoc """
+
+  ## PRNG
+  (Pseudo-Random Number Generator)
```
polvalente, line 5:
```suggestion
  ## PRNG (Pseudo-Random Number Generator)

```

The intent is so that we don't introduce an acronym without it's meaning directly associated to it

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,374 @@
...
+        u32[2]
+        [232, 3567587327]
+      >
+  """
+  defn key(seed) do
+    k1 = Nx.right_shift(seed, 32)
+    k2 = Nx.bitwise_and(seed, 0xFFFFFFFF)
+
+    Nx.stack([k1, k2])
+    |> Nx.as_type({:u, 32})
```
polvalente, line 101:
```suggestion
  defn key(seed) do
    seed = Nx.as_type(seed, {:u, 32})
    
    k1 = Nx.right_shift(seed, 32)
    k2 = Nx.bitwise_and(seed, 0xFFFFFFFF)

    Nx.stack([k1, k2])
```

nx/lib/nx/random.ex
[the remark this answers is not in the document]
polvalente: We can leave the scope of this PR as supporting only 32 bits, and then we can see if and how we extend this to more bits

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,374 @@
...
+  defnp rolled_loop_step(i, {xs, ks, rs}) do
+    k1 = ks[0]
+    k2 = ks[1]
+    k3 = ks[2]
+    r1 = rs[0]
+    r2 = rs[1]
+
+    {_, xs, _} =
+      while {x = 0, xs, rs}, x < 4 do
+        {x + 1, apply_round(xs, rs[0][x]), rs}
```
polvalente, line 225:
```suggestion
      while {x = 0, xs, r1}, x < 4 do
        {x + 1, apply_round(xs, r1[x]), rs}

```

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,374 @@
...
+    Nx.stack([y1, y2])
+    |> Nx.as_type({:u, 32})
+  end
+
+  defnp rolled_loop_step(i, {xs, ks, rs}) do
+    k1 = ks[0]
+    k2 = ks[1]
+    k3 = ks[2]
+    r1 = rs[0]
+    r2 = rs[1]
```
polvalente, line 221: I think we should receive these values separately as a 5-element tuple or as 2 separate tuples. You unwrap the tensor where, just to wrap it back up at the end of the function

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,374 @@
...
+    xs = Nx.add(ks, xs)
+
+    ks =
+      Nx.stack([
+        key2,
+        Nx.bitwise_xor(key1, key2)
+        |> Nx.bitwise_xor(0x1BD11BDA),
+        key1
+      ])
+      |> Nx.as_type({:u, 32})
```
polvalente, line 191: This wouldn't need to be stacked with my suggestion of passing the values as a tuple of tensors throughout the rolled loop below.

xs1 and xs2 can also be defined as `xs1 = xs + ks[0]` and `xs2 = xs + ks[1]`, such that we can also pass them without having to keep stacking and slicing tensors

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,374 @@
...
+    #we lose precision on purpose
+
+    state = {xs, ks, rotations}
+
+    {_, {nxs, _, _}} =
+      while {x = 0, state}, Nx.less(x, 5) do
+        {x + 1, rolled_loop_step(x, state)}
+      end
+
+    nxs
```
polvalente, line 201: We can leave stack of the nxs tuple to this last line

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,374 @@
...
+    |> Nx.bitwise_or(u_one)
+    |> bitcast(type)
+    |> Nx.subtract(Nx.tensor(1.0, type: type))
+    |> Nx.multiply(max_val - min_val)
+    |> Nx.add(min_val)
+    |> Nx.reshape(shape)
+    |> Nx.max(min_val)
+  end
+
+  deftransformp bitcast(tensor, type), do: Nx.bitcast(tensor, type)
```
polvalente, line 352: why does this need to be deftransform?

VenomSolo: > We can't have bitcast in Torchx.
>
> Since it's the beggining of the function, I would rather do the bitcast in a transform in pure elixir if it is indeed necessary.

I followed that suggestion and made it deftransform because transform() is getting deprecated

jonatanklosko: Yeah, the point is that you can use `Nx.bitcast` directly in `defn`, doesn't need to be a transform :)

VenomSolo: > Yeah, the point is that you can use `Nx.bitcast` directly in `defn`, doesn't need to be a transform :)

Sorry, but now I'm confused on relationship between defn and torchx. Should we use another solution if we can't have bitcast in Torchx or what exactly is the problem? I know we can use Nx.bitcast in defn and it was like that at the beginning but I was refering to @polvalente suggestion.

jonatanklosko: Ah, I see the other comment.

Wrapping it in a `deftransform` doesn't change anything. `deftransform` allows you to freely call arbitrary functions `Enum.*` or use constructs like `for`, however calling functions from `Nx.*` works the same way, it builds up the numerical expression.

So the point was that ideally we would avoid `Nx.bitcast` altogether. So given

```elixir
u_one = Nx.tensor(1.0, type: type) |> Nx.bitcast({:u, nbits})
```

we know all the operands when building the expression (`1.0`, `type`, `nbits`), so we could compute `u_one` in pure elixir, without `Nx.bitcast`.

But I see we still use `bitcast` in the other place and I'm not sure if we can easily eliminate it, because it operates on the generated tensor.

If this is not clear or there's anything about `defn`/transform that confuses you let me know!

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,374 @@
...
+
+    info = transform(opts[:type], &float_info(&1))
+
+    shape = opts[:shape]
+    type = {_dtype, nbits} = opts[:type]
+    min_val = opts[:min_val]
+    max_val = opts[:max_val]
+
+    u_one = Nx.tensor(1065353216, type: {:u, nbits})
+    #Equivalent of constant - Nx.tensor(1.0, type: type) |> Nx.bitcast({:u, nbits})
```
jonatanklosko, line 338: The value `1065353216` will be different depending on the `type`, no? We can compute `u_one` at compile time, but it is worth the effort only if we can eliminate the other `bitcast` too.

VenomSolo: Yes, it is a 32bit placeholder until we figure out what to do with bitcast.

polvalente: I'll look into supporting bitcast from integer to float in Torchx, so don't worry about this :)

jonatanklosko: I did an initial exploration of an alternative approach that doesn't rely on bit transformations, results in [this notebook](https://gist.github.com/jonatanklosko/462e0b2f0dbe06a42ae2c84224a59ba4). It seems that it could work, but could use someone with more expertise.

@polvalente regardless, I've just realised is that we use unsigned integers all across the implementation, so this alone kills torchx, right?

Sidenote: in `jax.random` [they say](https://github.com/google/jax/blob/07fcf793240645053987f918681da2c60c8c7103/jax/_src/random.py#L258-L261):

> The strategy here is to randomize only the mantissa bits with an exponent of
> 1 (after applying the bias), then shift and scale to the desired range. The
> bit-level transformation we use relies on Numpy and XLA having bit-for-bit
> equivalent float representations, which might not be true on all platforms.

polvalente: You're right! Let's not support torchx for now, then.

We can perhaps extend the code in a way that it operates the positive parte of the s64 domain in the future

VenomSolo: @jonatanklosko @polvalente Do you guys have any resources on integration with xla/torchx and possible caveats that we should be aware of in the future? To be honest as new to Nx as I am, I can't really follow your discussion without more knowledge.

jonatanklosko: @VenomSolo no worries! Things should generally work fine with exla. As for torchx, the main differences is that u16/u32/u64 types are not supported, and there are a couple functions missing:

https://github.com/elixir-nx/nx/blob/8eadd775cf08b871c364794b231924170ff06e3d/torchx/lib/torchx/backend.ex#L1557-L1559

To verify you can configure either as a backend and try out some examples. In a script/livebook you can do

```elixir
Mix.install([
  {:nx, path: "~/path/to/nx/nx"},
  # We could also point to released versions, but since we have it locally
  # it's convenient to point there during development:
  {:exla, path: "~/path/to/nx/exla"},
  {:torchx, path: "~/path/to/nx/torchx"}
])

# Set the default backend for computing Nx operations
# Nx.default_backend(Torchx.Backend)

key = Nx.Random.key(10)
Nx.Random.uniform(key)
```

Note that by default all operations are computed using `Nx.BinaryBackend`, which is a pure Elixir implementation.

nx/test/nx/random_test.exs
```diff
@@ -0,0 +1,140 @@
...
+    end
+  end
+
+  defp to_hex(tensor) do
+    tensor
+    |> Nx.to_flat_list()
+    |> Enum.map(&Integer.to_string(&1, 16))
+  end
+
+  describe "threefry2x32/2" do
```
polvalente, line 85: We could add property tests for the uniform and random_integer distributions that generate a sizeable number of samples so we can assert on the values for the first and second moments of the uniform distribution.

For uniform, we can approximate to the continuous case: https://en.wikipedia.org/wiki/Continuous_uniform_distribution#Moments

For random_integer, we can treat it as a discrete uniform distribution: https://mathworld.wolfram.com/DiscreteUniformDistribution.html

(first moment = mean, second moment = variance)

A similar test can be found here (where I needed to test a normal distribution that's generated through a given uniform distribution generator)

nx/test/nx/random_test.exs
```diff
@@ -0,0 +1,140 @@
...
+    test "matches known results from reference implementation" do
+      # values from https://github.com/DEShawResearch/Random123-Boost/blob/65e3d874b67aa7b3e02d5ad8306462f52d2079c0/libs/random/test/test_threefry.cpp#L30-L32
+
+      expected_results = [
+        ["6B200159", "99BA4EFE"],
+        ["1CB996FC", "BB002BE7"],
+        ["C4923A9C", "483DF7A0"]
```
polvalente, line 92:
```suggestion
        Nx.tensor([0x6B200159, 0x99BA4EFE], type: :u32),
        Nx.tensor([0x1CB996FC, 0xBB002BE7], type: :u32),
        Nx.tensor([0xC4923A9C, 0x483DF7A0], type: :u32)
```

this way you can skip the `to_hex` function and use assert_all_close or even assert left == right to ensure the values are the same

josevalim:
Hi @john-sonz! I am checking on the status of this PR. Anything pending? :) Can we help? TY! :heart:

VenomSolo:
> Hi @john-sonz! I am checking on the status of this PR. Anything pending? :) Can we help? TY! ❤️

This PR should be good to go after the code formatting ( not counting test errors from linalg :') ) and we'll hopefully start working on vmap. @john-sonz mostly covered batching in linalg functions.

seanmor5:
Thanks for this work! I have opened a corollary to this PR in Axon here: https://github.com/elixir-nx/axon/pull/350

I noticed that right now Nx.Random.uniform does not correctly preserve the requested type, you may have to cast at the end or cast some of the intermediate values

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,419 @@
...
+        ]
+      >
+  """
+  defn split(key, num \\ 2) do
+    assert_key(key)
+
+    impl(key, Nx.iota({num, 2}))
+  end
+
+  defn fold_in(key, data) do
```
josevalim, line 136: Should this be documented? Or should it be private?

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,419 @@
...
+      Nx.reshape(count, {:auto})
+      |> Nx.as_type({:u, 32})
+
+    threefry2x32(reshaped_key, reshaped_count)
+    |> Nx.reshape(shape)
+    |> Nx.as_type({:u, 32})
+  end
+
+  # Check count
+  defn threefry2x32(key, count) do
```
josevalim, line 193: Should this be private or documented?

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,419 @@
...
+  
+  ## Options
+  
+    * `:type` - the integer type for the returned tensor
+    * `:shape` - shape of the returned tensor
+  
+  
+  """
+  defn randint(key, min_val, max_val, opts \\ []) do
+    opts = keyword!(opts, shape: {1}, type: {:s, 32})
```
josevalim, line 295: We should default to s64 which is the default for integers, WDYT?

```suggestion
    opts = keyword!(opts, shape: {1}, type: {:s, 64})
```

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,419 @@
...
+      _ ->
+        raise(
+          ArgumentError,
+          "expected key to have shape {2}, got tensor with shape #{inspect(shape)}"
+        )
+    end
+
+    case type do
+      {:u, _} ->
+        :ok
```
josevalim, line 409: Should we assert the key is explicitly u32, since that's how it is defined? For example, would the algorithms still work if the keys are s64 or u16?

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,419 @@
...
+      shape:
+        {rem(Nx.size(shape) * bit_width, 32)
+         |> Nx.any()
+         |> Nx.add(Nx.size(shape) * bit_width / 32)}
+    ]
+  end
+
+  defn random_bits(key, opts \\ []) do
+    assert_key(key)
+    opts = keyword!(opts, shape: {1}, bit_width: 32)
```
josevalim, line 153:
```suggestion
    opts = keyword!(opts, shape: {}, bit_width: 32)
```

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,419 @@
...
+  
+    * `:shape` - shape of the returned tensor
+  
+    * `:min_val` - minimum value, default is 0.0
+  
+    * `:max_val` - maximum value, default is 1.0
+  
+  """
+  defn uniform(key, opts \\ []) do
+    opts = keyword!(opts, shape: {1}, type: {:f, 32}, min_val: 0.0, max_val: 1.0)
```
josevalim, line 353:
```suggestion
    opts = keyword!(opts, shape: {}, type: {:f, 32}, min_val: 0.0, max_val: 1.0)
```

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,419 @@
...
+
+    ks =
+      Nx.stack([
+        key2,
+        Nx.bitwise_xor(key1, key2)
+        |> Nx.bitwise_xor(0x1BD11BDA),
+        key1
+      ])
+      # we lose precision on purpose
+      |> Nx.as_type({:u, 32})
```
josevalim, line 221: Both xs and ks are u8, so I think the only reason we are getting a s64 is because of the constant. So this should address it:

```suggestion
        |> Nx.bitwise_xor(Nx.tensor(0x1BD11BDA, type: {:u, 8})),
        key1
      ])
```

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,419 @@
...
+
+  @doc """
+  Sample uniform random integer values in [min_val, max_val).
+  
+  ## Options
+  
+    * `:type` - the integer type for the returned tensor
+    * `:shape` - shape of the returned tensor
+  
+
```
josevalim, line 292: Please include some examples here!

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,419 @@
...
+  ## Options
+  
+    * `:type` - a float type for the returned tensor
+  
+    * `:shape` - shape of the returned tensor
+  
+    * `:min_val` - minimum value, default is 0.0
+  
+    * `:max_val` - maximum value, default is 1.0
+
```
josevalim, line 350: Please include some examples here!

josevalim:
Excellent @john-sonz and @VenomSolo! I have done my review and it is mostly minor details to clean up the code and docs. I have also started working on the basic foundation for vmap and I hope to have something to share soon. :)

nx/lib/nx/random.ex
VenomSolo: Included

nx/lib/nx/random.ex
VenomSolo: Included

nx/lib/nx/random.ex
VenomSolo: ks should never be u8 because key is by definition u32

nx/lib/nx/random.ex
VenomSolo: Adjusted with a fix to random_bits for support of constants

nx/lib/nx/random.ex
VenomSolo: Adjusted with a fix to random_bits for support of constants

nx/lib/nx/random.ex
VenomSolo: Yes, key has to be explicitly u32

nx/lib/nx/random.ex
VenomSolo: Sure, adjusted.

nx/lib/nx/random.ex
VenomSolo: Should be private but I will leave tests commented for now.

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,419 @@
...
+  deftransformp max_count_size(shape, bit_width) do
+    [
+      shape:
+        {rem(Nx.size(shape) * bit_width, 32)
+         |> Nx.any()
+         |> Nx.add(Nx.size(shape) * bit_width / 32)}
+    ]
+  end
+
+  defn random_bits(key, opts \\ []) do
```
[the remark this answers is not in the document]
VenomSolo: Should be private but I will leave tests commented for now.

nx/lib/nx/random.ex
VenomSolo: Should be private but I will leave tests commented for now.

VenomSolo:
Made an adjustment to reduce slicing suggested earlier by @polvalente but somehow glossed over.

nx/test/nx/random_test.exs
```diff
@@ -0,0 +1,202 @@
...
+
+  describe "split/2" do
+    test "splits key into multiple keys" do
+      key = Nx.Random.key(33)
+
+      two_keys = Nx.Random.split(key)
+      multiple_keys = Nx.Random.split(key, 12)
+
+      assert Nx.shape(two_keys) == {2, 2}
+      assert Nx.shape(multiple_keys) == {12, 2}
```
polvalente, line 25: I would also assert on the values here, not only shapes

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+  ## PRNG
+  (Pseudo-Random Number Generator)
+
+  Unlike the stateful pseudo-random number
+  generators (PRNGs) that users of most programming
+  languages and numerical libraries
+  may be accustomed to, Nx random functions all
+  require an explicit PRNG state to be passed as
+  a first argument. The random state is described by
+  an `Nx.Tensor` composed by 2 unsigned 32-bit integers
```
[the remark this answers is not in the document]
VenomSolo: Revised

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+    assert_key(key)
+    opts = keyword!(opts, shape: {1}, bit_width: 32)
+    shape = opts[:shape]
+    bit_width = opts[:bit_width]
+
+    bits = case bit_width do
+      64 -> bits =
+            impl(key, Nx.concatenate([Nx.iota(shape), Nx.iota(shape)+Nx.size(shape)]))
+            |> Nx.reshape({2, :auto})
+            |> Nx.as_type({:u, bit_width})
```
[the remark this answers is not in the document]
VenomSolo: Adjusted

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+            (bits[0] <<< 32) ||| bits[1]
+      32 -> impl(key, Nx.iota(shape))
+      _ ->  impl(key, Nx.iota(shape))
```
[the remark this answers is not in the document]
VenomSolo: Yes, it does but defaults to s64, added explicit type.

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,390 @@
...
+            |> Nx.as_type({:u, bit_width})
+    end
+
+    Nx.reshape(bits, shape)
```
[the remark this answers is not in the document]
VenomSolo: Adjusted

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,475 @@
...
+        [232, 3567587327]
+      >
+  """
+  defn key(seed) do
+    k1 = Nx.right_shift(seed, 32)
+    k2 = Nx.bitwise_and(seed, 0xFFFFFFFF)
+
+    Nx.stack([k1, k2])
+    |> Nx.as_type({:u, 32})
+  end
```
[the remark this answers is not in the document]
VenomSolo: I don't quite get the question? Clause in random_bits has nothing to do with key length, it corresponds with the requested bits number.

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,475 @@
...
+    shape = opts[:shape]
+    bit_width = opts[:bit_width]
+
+    bits =
+      case bit_width do
+        64 ->
+          bits =
+            impl(key,
+            Nx.stack([Nx.iota(shape), Nx.iota(shape) + Nx.size(shape)])
+            |> Nx.reshape({:auto}))
```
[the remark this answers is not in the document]
VenomSolo: It sure works as an equivalent, there seems to be no need for deftransformp. I'll take that any day over that little monster I created because of Nx.concatenate limitations.

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,475 @@
...
+    Nx.reshape(bits, shape)
+  end
+
+  defnp impl(key, count) do
+    shape = Nx.shape(count)
+
+    reshaped_key = Nx.reshape(key, {2, 1})
+
+    reshaped_count =
+      Nx.reshape(count, {:auto})
```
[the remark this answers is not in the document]
VenomSolo: Out of those only flattening count is still necessary, moved it to threefry2x32 implementation

nx/test/nx/random_test.exs
```diff
@@ -0,0 +1,202 @@
...
+    end
+  end
+
+  # describe "fold_in/2" do
```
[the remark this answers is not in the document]
VenomSolo: These functions were public earlier (and so they were tested) but they are no longer and so I was reluctant to completely delete all those tests. Should we regardless?

nx/test/nx/random_test.exs
```diff
@@ -0,0 +1,139 @@
...
+  import Nx.Helpers
+
+  doctest Nx.Random
+
+  describe "key/1" do
+    test "transforms given integer into PRNG key" do
+      key = Nx.Random.key(44)
+
+      assert key |> Nx.type() |> Nx.Type.integer?()
+      assert Nx.shape(key) == {2}
```
jonatanklosko, line 13: Now that we assert on the value, both of these are covered :)

```suggestion
```

nx/lib/nx/random.ex
```diff
@@ -0,0 +1,468 @@
...
+  end
+
+  deftransformp max_count_size(shape, bit_width) do
+    [
+      shape:
+        {rem(Nx.size(shape) * bit_width, 32)
+         |> Nx.any()
+         |> Nx.add(Nx.size(shape) * bit_width / 32)}
+    ]
+  end
```
jonatanklosko, line 149: Looks like we don't use it anywhere.

```suggestion
```

nx/lib/nx/random.ex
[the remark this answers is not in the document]
polvalente: I meant that while the rest of the API accepts type parameters, this implementation only seems to work by receiving a 64bit value and outputting u32. Perhaps we should explicitly upcast u8, u16 and u32 to u64 and s8, s16 and s32 to s64 before passing into `right_shift` and `bitwise_and`

Otherwise, from your comment I understand that the key width doesn't change regardless of the output type and width for the generator. Is this right?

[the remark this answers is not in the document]
VenomSolo: That is right, key width always stay the same.

Is the issue that implementation SEEMS to work only for 64bit seed or that it WORKS only for 64bit (which isn't true)? We can, of course, explicitly cast them if that improves readability.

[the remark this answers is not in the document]
polvalente: I think we should raise if the input seed has less than 64 bits to ensure we have the expected entropy in the keyset generated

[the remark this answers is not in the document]
VenomSolo: What keyset? We get entropy from the threefry function, key is literally 64bit seed divided in half or zero and seed if it's 32bit or smaller. Now I'm even more confused?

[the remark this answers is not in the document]
polvalente: What I mean is that the `key` function expects to output 2 useful numbers k1 and k2, ultimately.
However, if the input has 32 bits k1 will always be 0 and k2 will always be the seed. There's less information being output than what would otherwise be expected. For 16 and 8 bits this is a bigger problem.

Just to make sure we're on the same page, we have to keep in mind that `seed` is a scalar Nx.Tensor, not a plain number, even if you pass a plain number as the input.

Given all of this, I'm not sure if having less useful information in the seed will affect the algorithm's output that much. This is why I think it's probably best that we enforce the `seed` to be given as a 64bit input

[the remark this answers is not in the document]
VenomSolo: The thing is, it's exactly what is expected. Even 63 consecutive 0s are as valuable for us as any other number. There's never "less" information in that context. And it won't affect the algorithm at all. Even if we enforce 64bit input there is no rule that invalidates input like 1 or 2 which are essentialy 2bit numbers. We simply don't care if we get 63 0sand 1 or 7 0s and 1 because we either way get the same amount of information. In both cases we'll get [0,1]. I think we are misunderstanding something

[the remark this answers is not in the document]
polvalente: I get what you mean. If you say the algorithm won't be affected by having less information (which there is because we have half the useful bits and therefore 2ˆ32 less keys to choose from), we can move forward ignoring this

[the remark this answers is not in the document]
VenomSolo: We can safely ignore this as this is something entirely up to user's decision. Doesn't matter what seed he provides, we can split up to 2^64 keys regardless.

polvalente (approved):

PR #857, opened 2022-07-31
Add tests for dot on torchx

Ian-GL:
Add tests for dot in `torchx/test/nx_test.exs` given that the doctests are disabled due batching not being supported.

Issue #682

### DIFF
--- a/torchx/test/torchx/nx_test.exs
+++ b/torchx/test/torchx/nx_test.exs
@@ -403,20 +403,33 @@ defmodule Torchx.NxTest do
 
       assert {2, 1, 1, 3} == out.shape
 
-      assert_equal(out,
-               Nx.tensor([
-                 [
-                   [
-                     [3_731_448, 3_745_476, 3_759_504]
-                   ]
-                 ],
-                 [
-                   [
-                     [10_801_560, 10_843_812, 10_886_064]
-                   ]
-                 ]
-               ])
+      assert_equal(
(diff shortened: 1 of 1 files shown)


torchx/test/torchx/nx_test.exs
```diff
@@ -368,6 +368,57 @@ defmodule Torchx.NxTest do
...
+    test "dot with vector and scalar" do
+      t = Nx.tensor([[1, 2, 3]])
+      assert Nx.shape(t) == {1, 3}
+      out = Nx.dot(t, 3)
+
+      assert Nx.shape(out) == {1, 3}
+      assert_equal(out, Nx.tensor([[3, 6, 9]]))
+    end
+
+    test "dot does not re-sort the contracting axes" do
```
polvalente, line 398: Could you add a test that shows that batching is not supported? 

This way when/if we support it, a new test will break and point towards adding more tests

polvalente (approved):

torchx/test/torchx/nx_test.exs
Ian-GL: For sure! Just added it here https://github.com/elixir-nx/nx/pull/857/commits/3653d7a652767c633f61edc4af06ee374189fcf6

polvalente (approved):

polvalente:
Thanks!!

PR #859, opened 2022-08-01
Add mean tests for Torchx

Ian-GL:
Adding `Nx.mean/2` tests for Torchx. These tests are added as part of the tests since the doctests are skipped given that Torchx does not support unsigned integers.

Issue #682


polvalente (approved):

polvalente:
Thanks!!

PR #860, opened 2022-08-02
Add quotient test for Torchx. Remove argmax and argmin from exclusion list.

Ian-GL:
Added tests for `Nx.quotient/2` since the doctests are ignored due to unsigned integers not being supported on Torchx.

`argmax` and `argmin` doctests no longer use complex numbers, which allows them to be restored.

Related to issue #682


polvalente (approved):

polvalente:
Thanks again! Keep'em coming :sweat_smile:

PR #861, opened 2022-08-02
feat(torchx): add batching support for dot

polvalente:
helps with #682

### DIFF
--- a/torchx/lib/torchx/backend.ex
+++ b/torchx/lib/torchx/backend.ex
@@ -921,6 +921,29 @@ defmodule Torchx.Backend do
     |> to_nx(out)
   end
 
+  def dot(
+        %T{type: out_type} = out,
+        %T{type: left_type, data: %TB{ref: left_ref}},
+        left_axes,
+        left_batched_axes,
+        %T{type: right_type, data: %TB{ref: right_ref}},
+        right_axes,
+        right_batched_axes
+      ) do
+    left_axes = Enum.map(left_axes, fn x -> x - Enum.count(left_batched_axes, &(&1 < x)) end)
+    right_axes = Enum.map(right_axes, fn x -> x - Enum.count(right_batched_axes, &(&1 < x)) end)
+
+    Torchx.batched_tensordot(
+      to_typed_ref(left_ref, left_type, out_type),
+      to_typed_ref(right_ref, right_type, out_type),
+      left_axes,
+      left_batched_axes,
+      right_axes,
+      right_batched_axes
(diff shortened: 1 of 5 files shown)


torchx/lib/torchx/backend.ex
```diff
@@ -921,6 +921,29 @@ defmodule Torchx.Backend do
...
+  def dot(
+        %T{type: out_type} = out,
+        %T{type: left_type, data: %TB{ref: left_ref}},
+        left_axes,
+        left_batched_axes,
+        %T{type: right_type, data: %TB{ref: right_ref}},
+        right_axes,
+        right_batched_axes
+      ) do
+    left_axes = Enum.map(left_axes, fn x -> x - Enum.count(left_batched_axes, &(&1 < x)) end)
```
polvalente, line 933: @josevalim @jonatanklosko any idea on how to make this more efficient? Basically, I need to "remove" the batching axes from the axis numbering for libtorch

torchx/lib/torchx/backend.ex
```diff
@@ -921,6 +921,29 @@ defmodule Torchx.Backend do
...
+    right_axes = Enum.map(right_axes, fn x -> x - Enum.count(right_batched_axes, &(&1 < x)) end)
+
+    Torchx.batched_tensordot(
```
polvalente, line 936: Should I try to make tensordot optionally aware of batching instead of exposing a separate function?

josevalim: I think it is fine as is, I would just remove the `batched_` prefix.

torchx/lib/torchx/backend.ex
josevalim: You could do manual recursion to avoid multiple traversals but it probably doesn't matter.

josevalim (approved):

torchx/test/torchx/nx_test.exs
```diff
@@ -420,18 +420,6 @@ defmodule Torchx.NxTest do
       )
     end
 
-    test "dot currently raises when using batching" do
```
Ian-GL, line 423: The rest of the dot related tests here could also be removed as they are testing the same thing as the doctests, so now that those are restored they are not needed.

torchx/lib/torchx/backend.ex
polvalente: I ended up merging all into a single tensordot/6 clause for the backend and a single function at the cpp level

Issue #229, opened 2021-02-18
Support updates with put_in and update_in


josevalim:
This is so we can have a more complete implementation of the Access protocol for Tensors

polvalente:
@josevalim It seems that #817 would enable this issue through `indexed_set`. WDYT?

josevalim:
get_in uses slices, which I think the put_slice is already enough to implement this. :)

josevalim:
Reopening this because the way of updating is quite painful. We need to roll your own version of Access for now.

josevalim:
Closing as we will relax indexed_add and indexed_put instead.

PR #864, opened 2022-08-04
Implement the get_and_update Access callback on tensors

Ian-GL:
This is an initial implementation of the `get_and_update` callback so that tensors can support `put_in` and `update_in`. It uses slices by creating a slice of the section before the update and another one afterwards, and at the end concatenating them with the new value in the middle.

I also added this to the documentation and doctests.

Issue #229

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -323,6 +323,79 @@ defmodule Nx do
   For a more complex slicing rules, including strides, you
   can always fallback to `Nx.slice/4`.
 
+  Access also supports updating values with put_in/3 and update_in/3,
+  where the path to the value is a list of integers indicating the sub-dimensions.
+
+  The value that will be used to substitute should be a tensor of the same
+  shape as the one being updated.
+
+      iex> t = Nx.tensor([1, 2, 3])
+      iex> put_in(t, [2], Nx.tensor(2))
+      #Nx.Tensor<
+        s64[3]
+        [1, 2, 2]
+      >
+
+  In a multi-dimensional array, the path will determine at which level
+  the update will be made.
+
+  A whole tensor can be updated:
+
+      iex> t = Nx.tensor([[1, 2], [3, 4]])
+      iex> put_in(t, [1], Nx.tensor([4, 5]))
+      #Nx.Tensor<
+        s64[2][2]
+        [
+          [1, 2],
+          [4, 5]
+        ]
+      >
+
+  Or a scalar in a deeper level:
+
+      iex> t = Nx.tensor([[[1, 2], [3, 4]], [[5, 6], [7, 8]]])
+      iex> put_in(t, [0, 1, 1], Nx.tensor(99))
+      #Nx.Tensor<
+        s64[2][2][2]
+        [
+          [
+            [1, 2],
+            [3, 99]
+          ],
+          [
+            [5, 6],
+            [7, 8]
+          ]
+        ]
+      >
+
+  As long as the update value is the same shape as the target one
+  to be substituted it will replace it, if not it will raise:
+
+      iex> t = Nx.tensor([[1, 2, 3], [4, 5, 6]])
+      iex> put_in(t, [0], Nx.tensor([1, 2]))
+      ** (ArgumentError) the shape {2} of the new value does not match the current value shape {3}
+
+  Out of bound attempts to update will also raise:
+      iex> t = Nx.tensor([[1, 2], [3, 4]])
+      iex> put_in(t, [2], Nx.tensor([5, 6]))
+      ** (ArgumentError) index 2 is out of bounds for axis 0 in shape {2, 2}
+
+  Unlike the access get syntax, it does not accept negative indices.
+
+  As for the update_in function, it accepts any function that can take
+  the tensor as an argument and will apply it to the target:
+
+      iex> t = Nx.tensor([[1, 2, 3], [4, 5, 6]])
+      iex> update_in(t, [0], fn t -> Nx.multiply(t, 2) end)
+      #Nx.Tensor<
+        s64[2][3]
+        [
+          [2, 4, 6],
+          [4, 5, 6]
+        ]
+      >
+
   ## Backends
 
   The `Nx` library has built-in support for multiple backends.
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -336,10 +336,24 @@ defmodule Nx do
         [1, 2, 2]
       >
 
+  The access update syntax also supports negative values, which will access the
+  element from the back:
+
+      iex> t = Nx.tensor([1, 2, 3])
+      iex> put_in(t, [-1], Nx.tensor(4))
+      #Nx.Tensor<
+        s64[3]
+        [1, 2, 4]
+      >
+
+  The length of the path list should always be equal or less than the rank
+  of the tensor.
+
   In a multi-dimensional array, the path will determine at which level
   the update will be made.
 
-  A whole tensor can be updated:
+  The update value can also be a tensor with multiple entries as long
+  as the shape matches the target to be updated:
 
       iex> t = Nx.tensor([[1, 2], [3, 4]])
       iex> put_in(t, [1], Nx.tensor([4, 5]))
@@ -351,7 +365,7 @@ defmodule Nx do
         ]
       >
 
-  Or a scalar in a deeper level:
+  We can also update a single tensor entry at a deeper level:
 
       iex> t = Nx.tensor([[[1, 2], [3, 4]], [[5, 6], [7, 8]]])
       iex> put_in(t, [0, 1, 1], Nx.tensor(99))
@@ -369,8 +383,8 @@ defmodule Nx do
         ]
       >
 
-  As long as the update value is the same shape as the target one
-  to be substituted it will replace it, if not it will raise:
+  If the update value is not the same shape as the target one
+  it will raise:
 
       iex> t = Nx.tensor([[1, 2, 3], [4, 5, 6]])
       iex> put_in(t, [0], Nx.tensor([1, 2]))
@@ -380,8 +394,6 @@ defmodule Nx do
       iex> t = Nx.tensor([[1, 2], [3, 4]])
       iex> put_in(t, [2], Nx.tensor([5, 6]))
       ** (ArgumentError) index 2 is out of bounds for axis 0 in shape {2, 2}
-
-  Unlike the access get syntax, it does not accept negative indices.
 
   As for the update_in function, it accepts any function that can take
   the tensor as an argument and will apply it to the target:
--- a/nx/lib/nx/tensor.ex
+++ b/nx/lib/nx/tensor.ex
@@ -193,6 +193,7 @@ defmodule Nx.Tensor do
   end
 
   defp get_slice(_tensor, _indices, [0 | _]), do: nil
+
   defp get_slice(tensor, indices, lengths) do
     impl = Nx.Shared.impl!(tensor)
 
--- a/nx/lib/nx/tensor.ex
+++ b/nx/lib/nx/tensor.ex
@@ -143,8 +143,61 @@ defmodule Nx.Tensor do
   end
 
   @impl true
-  def get_and_update(_tensor, _index, _update) do
-    raise "Access.get_and_update/3 is not yet supported by Nx.Tensor"
+  def get_and_update(%Nx.Tensor{shape: {}} = tensor, _index, _update) do
+    raise ArgumentError,
+          "cannot use the tensor[index] syntax on scalar tensor #{inspect(tensor)}"
+  end
+
+  def get_and_update(tensor, index, update) do
+    if index < 0 or index >= elem(tensor.shape, 0) do
+      raise ArgumentError,
+            "index #{index} is out of bounds for axis 0 in shape #{inspect(tensor.shape)}"
+    end
+
+    {:ok, current_value} = fetch(tensor, index)
+    {_, new_value} = update.(current_value)
+
+    if current_value.shape != new_value.shape do
+      raise ArgumentError,
+            "the shape #{inspect(new_value.shape)} of the new value does not match the current value shape #{inspect(current_value.shape)}"
+    end
+
+    {before_indices, after_indices} = get_indices(index, tensor.shape)
+    {before_lengths, after_lengths} = get_lengths(index, tensor.shape)
+
+    slice_before = get_slice(tensor, before_indices, before_lengths)
+    slice_after = get_slice(tensor, after_indices, after_lengths)
+
+    wrapper_shape =
+      tensor.shape
+      |> Tuple.delete_at(0)
+      |> Tuple.insert_at(0, 1)
+
+    new_value_wrapped = Nx.broadcast(new_value, wrapper_shape)
+
+    slices = Enum.reject([slice_before, new_value_wrapped, slice_after], &is_nil/1)
+
+    {current_value, Nx.concatenate(slices)}
+  end
+
+  defp get_indices(index, shape) do
+    before_indices = List.duplicate(0, Nx.rank(shape))
+    after_indices = List.replace_at(before_indices, 0, index + 1)
+
+    {before_indices, after_indices}
+  end
+
+  defp get_lengths(index, shape) do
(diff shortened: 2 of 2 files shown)


nx/lib/nx.ex
```diff
@@ -323,6 +323,79 @@ defmodule Nx do
...
+      iex> t = Nx.tensor([[1, 2, 3], [4, 5, 6]])
+      iex> put_in(t, [0], Nx.tensor([1, 2]))
+      ** (ArgumentError) the shape {2} of the new value does not match the current value shape {3}
+
+  Out of bound attempts to update will also raise:
+      iex> t = Nx.tensor([[1, 2], [3, 4]])
+      iex> put_in(t, [2], Nx.tensor([5, 6]))
+      ** (ArgumentError) index 2 is out of bounds for axis 0 in shape {2, 2}
+
+  Unlike the access get syntax, it does not accept negative indices.
```
Ian-GL, line 384: On this I am not sure. It seems a bit weird for me to support negative indices to update, but to be consistent with the access syntax used to get values it should be supported. I'll take a look to add it.

nx/lib/nx.ex
```diff
@@ -323,6 +323,79 @@ defmodule Nx do
...
+
+      iex> t = Nx.tensor([[1, 2], [3, 4]])
+      iex> put_in(t, [1], Nx.tensor([4, 5]))
+      #Nx.Tensor<
+        s64[2][2]
+        [
+          [1, 2],
+          [4, 5]
+        ]
+      >
```
polvalente, line 352: "A whole tensor can be updated" doesn't seem to be compatible with the example (in which only part of the tensor is updated)

nx/lib/nx.ex
```diff
@@ -323,6 +323,79 @@ defmodule Nx do
...
+
+  Or a scalar in a deeper level:
```
polvalente, line 354:
```suggestion
  We can also update a single tensor entry at a deeper level:
```

nx/lib/nx.ex
polvalente:
```suggestion
  One can use a tensor with various entries as an update value:

      iex> t = Nx.tensor([[1, 2], [3, 4]])
      iex> put_in(t, [1], Nx.tensor([4, 5]))
      #Nx.Tensor<
        s64[2][2]
        [
          [1, 2],
          [4, 5]
        ]
      >
```

nx/lib/nx.ex
```diff
@@ -323,6 +323,79 @@ defmodule Nx do
...
+          ],
+          [
+            [5, 6],
+            [7, 8]
+          ]
+        ]
+      >
+
+  As long as the update value is the same shape as the target one
+  to be substituted it will replace it, if not it will raise:
```
polvalente, line 373:
```suggestion
  As long as the update value is the same shape as the one
  to be substituted, the replacement will happen. Otherwise, it will raise:
```

nx/lib/nx.ex
polvalente: I would accept negative values as well. I'm not totally sure on the index format, though. I would expect the index path to be more similar to Nx.gather/indexed_add/indexed_put

nx/lib/nx.ex
```diff
@@ -323,6 +323,79 @@ defmodule Nx do
...
+
+      iex> t = Nx.tensor([[1, 2, 3], [4, 5, 6]])
+      iex> update_in(t, [0], fn t -> Nx.multiply(t, 2) end)
+      #Nx.Tensor<
+        s64[2][3]
+        [
+          [2, 4, 6],
+          [4, 5, 6]
+        ]
+      >
```
polvalente, line 397: I'm not sure if we should support `update_in`. @josevalim thoughts?

nx/lib/nx.ex
polvalente: For supporting negative indices, you can do `dimension_size + negative_index` to turn it into the corresponding non-negative index

nx/lib/nx/tensor.ex
```diff
@@ -143,8 +143,62 @@ defmodule Nx.Tensor do
...
+    if current_value.shape != new_value.shape do
+      raise ArgumentError,
+            "the shape #{inspect(new_value.shape)} of the new value does not match the current value shape #{inspect(current_value.shape)}"
+    end
+
+    {before_indices, after_indices} = get_indices(index, tensor.shape)
+    {before_lengths, after_lengths} = get_lengths(index, tensor.shape)
+
+    slice_before = get_slice(tensor, before_indices, before_lengths)
+    slice_after = get_slice(tensor, after_indices, after_lengths)
```
polvalente, line 169: do we really need to get slices?

There is Nx.indexed_put and Nx.put_slice, one of which I would expect to be used here instead of getting slices + concatenation

Ian-GL: Yes! That is a very good catch! I just re-read the issue and see it was already mentioned there, I misread before and thought it said slices. I just tried it with `put_slices` and it simplifies the code a lot.

nx/lib/nx.ex
Ian-GL: I will add the support for negative indexes then. As for the index format, `Kernel.put_in` does expect a list, I actually tried it so it could be used with both tensors and lists and it doesn't work. I also think this is intuitive enough as the list of keys is what is used with `put_in` and `update_in`.

nx/lib/nx.ex
Ian-GL: Hmm, what I'm trying to say is that the inner element which is also a tensor can be updated as a whole. I agree that what I wrote is confusing but I'm thinking if there could be another way to state it in the sense that an N-dimensional tensor can be used as the update value as long as the shape matches with the element at the level it tries to replace it.

nx/lib/nx.ex
polvalente: So the idea is to support a list of 1 to n entries, where n is the rank of the tensor being Access-ed, right?

nx/lib/nx.ex
Ian-GL: This basically comes out of the box by implementing the `get_and_update` Access behavior. Though if it seems problematic a note could be added to advise against it. I think it is consistent with the `update_in` of other structs, so don't see a reason not to support it although it might not be as useful.

nx/lib/nx.ex
Ian-GL: Yes. That's a very clear way to put it, I will add it to the docs.

Ian-GL: Well, n is <= rank, you can also access `[0]` and substitute an N-dimensional array instead of a single entry.

polvalente: What would one expect when updating a nested list of tensors? I'm trying to wrap my head around how things would work

For instance: `[1, 2, [Nx.tensor([[3, 4], [5, 6]]), 7]]`

Valid keys would be like below? I'm representing `key -> value accessed`

```elixir
[0] -> 1
[1] -> 2
[2] -> [Nx.tensor(...), 7]
[2, 0] -> Nx.tensor([[3, 4], [5, 6]])
[2, 0, [0]] -> Nx.tensor([3, 4])
[2, 0, [1]] -> Nx.tensor([5, 6])
[2, 0, [0, 0]] -> Nx.tensor(3)
[2, 0, [1, 1]] -> Nx.tensor(6)
[2, 1] -> 7
```

nx/lib/nx/tensor.ex
```diff
@@ -143,8 +143,47 @@ defmodule Nx.Tensor do
...
+
+    new_value_wrapped = Nx.broadcast(new_value, wrapper_shape)
+
+    indices = get_slice_indices(index, tensor.shape)
+
+    impl = Nx.Shared.impl!(tensor)
+
+    updated_tensor =
+      tensor
+      |> impl.put_slice(tensor, indices, new_value_wrapped)
```
polvalente, line 178:
```suggestion
    updated_tensor = Nx.put_slice(tensor, indices, new_value_wrapped)
```

I don't think you need to go below the public API for this

nx/lib/nx.ex
polvalente: The main issue here is when using with data which is not allocated on the CPU. You would bring the data from the other device to the CPU just for this update

nx/lib/nx.ex
Ian-GL: Hmm, well, lists do not implement the Access behaviour, but if it all were tensors then yes it would be similar to that, for example:

```elixir
Nx.tensor([[[1, 2]],  [[3, 4]]])

[0] -> [[1, 2]]
[0, 0] -> [1, 2]
[0, 0, 1] -> 2
[1, 0, 1] -> 4
[0, 1] -> Out of boundaries
```

So for a shape {a, b, c} you can make a list of up to [x, y, z] where x <= a, etc. but you can also do [x, y] and that will give you a tensor of shape {b, c}, meaning for each element of the list to access the rank of the tensor to update will diminish by 1.

polvalente: I forgot to add the `Access.at` calls:

```elixir
iex(3)> l = [1, 2, [3, [4, 5]]]
[1, 2, [3, [4, 5]]]
iex(4)> get_in(l, [Access.at(0)])
1
iex(5)> get_in(l, [Access.at(2), Access.at(2)])
nil
iex(6)> get_in(l, [Access.at(2), Access.at(1)])
[4, 5]
```

Same idea as what I specified there, though.

polvalente: My main concern is on the index format and how you would access a tensor in a nested list, though

Ian-GL: So yeah, it would be a combination of `Access.at(i)` and integers. I just tried this and it does work:

```elixir
list = [1, 2, Nx.tensor([1, 2, 3])]
put_in(list, [Access.at(2), 2], Nx.tensor(99))

[
  1,
  2,
  #Nx.Tensor<
    s64[3]
    [1, 2, 99]
  >
]
```

polvalente: Yeah, this is starting to make sense to me now :)

nx/lib/nx.ex
Ian-GL: Changed how this was structured here https://github.com/elixir-nx/nx/pull/864/commits/f279b5490efec83586be4876655f87cafb8c6776

nx/lib/nx.ex
Ian-GL: Changed here https://github.com/elixir-nx/nx/pull/864/commits/f279b5490efec83586be4876655f87cafb8c6776

nx/lib/nx.ex
Ian-GL: WDYT about how it is written now, @polvalente ? https://github.com/elixir-nx/nx/pull/864/commits/f279b5490efec83586be4876655f87cafb8c6776

nx/lib/nx/tensor.ex
Ian-GL: For this I was taking as an example how the slices are generated in `fetch`, which uses the implementation slice directly. What are the main tradeoffs between one and other?

josevalim: put_slice will attempt to normalize the data but we have already normalize the data here. That's why we skipped. I think skipping here is fine!

nx/lib/nx.ex
josevalim: @polvalente Inside defn, update_in will work as a read + write operation that will be compiled to the GPU. It won't bring the data to the CPU but it will do a read which can most likely be skipped with for some operations such as `indexed_add`. So I think this approach is fine, all things considered it, but perhaps we don't want to make it first class (i.e. we can import `put_in` in `Nx.Defn.Kernel` but probably best to skip `update_in`). Does that make sense?

nx/lib/nx.ex
```diff
@@ -323,6 +323,91 @@ defmodule Nx do
   For a more complex slicing rules, including strides, you
   can always fallback to `Nx.slice/4`.
 
+  Access also supports updating values with put_in/3 and update_in/3,
+  where the path to the value is a list of integers indicating the sub-dimensions.
```
josevalim, line 327:
```suggestion
  Access also supports updating values with `put_in/3`,
  where the path to the value is a list of integers indicating the sub-dimensions.
```

nx/lib/nx.ex
```diff
@@ -323,6 +323,91 @@ defmodule Nx do
...
+      >
+
+  If the update value is not the same shape as the target one
+  it will raise:
+
+      iex> t = Nx.tensor([[1, 2, 3], [4, 5, 6]])
+      iex> put_in(t, [0], Nx.tensor([1, 2]))
+      ** (ArgumentError) the shape {2} of the new value does not match the current value shape {3}
+
+  Out of bound attempts to update will also raise:
```
josevalim, line 393:
```suggestion
  Out of bound attempts to update will also raise:

```

nx/lib/nx/tensor.ex
```diff
@@ -143,8 +143,44 @@ defmodule Nx.Tensor do
...
+  def get_and_update(%Nx.Tensor{shape: {}} = tensor, _index, _update) do
+    raise ArgumentError,
+          "cannot use the tensor[index] syntax on scalar tensor #{inspect(tensor)}"
+  end
+
+  def get_and_update(tensor, index, update) do
+    index = normalize_index(index, 0, tensor.shape)
+
+    {:ok, current_value} = fetch(tensor, index)
+    {_, new_value} = update.(current_value)
```
josevalim, line 155:
```suggestion
    new_value =
      case update.(current_value) do
        {_, new_value} -> new_value
        :pop -> raise "cannot pop a tensor"
      end
```

nx/lib/nx/tensor.ex
```diff
@@ -143,8 +143,44 @@ defmodule Nx.Tensor do
...
+
+    if current_value.shape != new_value.shape do
```
josevalim, line 157: I am thinking we should probably allow broadcasts? For example:

```elixir
iex> tensor = Nx.iota({2, 2})
iex> put_in tensor[0], 13
#Nx.Tensor<
  [
    [13, 13],
    [2, 3]
  ]
>
```

WDYT?

josevalim:
Actually, I think this approach may not work. :;I will write more about it soon. the code and docs are fantastic though, thanks @Ian-GL !

josevalim:
Ok, so there is an issue with how `Access` works in Elixir. `Access` works dimension by dimension, which means that, if you do this:

    put_in tensor[1][4][3], new_tensor

It will get index 1 at axis 0 and emit it. Then get index 4 at axis 1 and emit it. Then get index 3 at axis 2 and emit it. This is fine. However, it also means the updates happen in reverse: we will replace the last tensor by the new one, then we will store the result of that in its upper tensor, then store the result of that in its upper tensor. This means we are doing additional updates for each additional dimension.

We would actually be able to optimize this in `defn` but we can't optimize this outside `defn`. It is not about the implementation, the implementation is solid within the constraints of the `Access` protocol in Elixir. The issue is really due to how access works.

How can we solve this? One option would be to allow lists in access:

    put_in tensor[1, 4, 3], new_tensor

But this would require changes to Elixir and honestly the difference between approaches would be very brittle.

We could also allow a list to be given:

    put_in tensor[[1, 4, 3]], new_tensor

This does not require changes to Elixir but still very brittle.

So unfortunately I think the best option for now is to tell folks to use `Nx.put_slice/3`. :(

The work was great @Ian-GL, I am sorry we won't be able to merge it though due to higher order problems. Thanks and have a great weekend!

Ian-GL:
@josevalim , thanks for the thorough explanation! Yes, that does make a lot of sense. I actually didn't know that was how `put_in` worked under the hood and was surprised when I was working on this and saw that on the index I always got just a single number instead of the list (and then I saw the source code on how it works using recursion).

I agree that allowing lists in access would be very different to what there is today and I do see that changing it could cause all sort of different troubles.

All in all I agree with your conclusion. Thanks for taking the time to explain! Hope you have a great weekend too!

PS. Would it be worth it to add in the docs a summarized explanation of why tensors don't implement the `get_and_update` callback of Access and a nudge to use `Nx.put_slice/3` directly?

josevalim:
@Ian-GL oh! Yeah, I think we should change the exception we raise in get_and_update and point people to `put_slice` instead. :)

PR #870, opened 2022-08-11
Support batched tensors in `Nx.LinAlg.qr`

john-sonz:
Makes `Nx.LinAlg.qr` work with batched tensors. 

When given batched tensor `Nx.LinAlg.qr` will perform QR decomposition across the matrices in inner-most 2 dimensions, keeping the rest of the shape unaffected.

### DIFF
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -672,15 +672,15 @@ defmodule Nx.LinAlg do
   end
 
   @doc """
-  Calculates the QR decomposition of a 2-D tensor with shape `{M, N}`.
+  Calculates the QR decomposition of a tensor with shape `{..., M, N}`.
 
   ## Options
 
     * `:mode` - Can be one of `:reduced`, `:complete`. Defaults to `:reduced`
       For the following, `K = min(M, N)`
 
-      * `:reduced` - returns `q` and `r` with shapes `{M, K}` and `{K, N}`
-      * `:complete` - returns `q` and `r` with shapes `{M, M}` and `{M, N}`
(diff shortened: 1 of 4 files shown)


nx/lib/nx/lin_alg.ex
```diff
@@ -727,6 +727,40 @@ defmodule Nx.LinAlg do
...
+        [
+          [
+            [1.0, 0.0, 0.0],
+            [0.0, 1.0, 0.0],
+            [0.0, 0.0, 1.0]
+          ],
+          [
+            [1.0, 0.0, 0.0],
+            [0.0, 1.0, 0.0],
+            [0.0, 0.0, 1.0]
```
polvalente, line 743: Minor nitpick, but I'll change this later myself: I think it would be nice to show a matrix where one of the Q matrices is different from the other

polvalente (approved):

nx/test/nx/lin_alg_test.exs
```diff
@@ -260,16 +260,60 @@ defmodule Nx.LinAlgTest do
...
+            [0.0, 0.0, 20.0]
+          ]
+        ])
+
+      assert_all_close(r, expected_r, atol: 1.0e-10)
+
+      assert_all_close(Nx.dot(q, [2], [0], r, [1], [0]), t, atol: 1.0e-10)
+    end
+
     test "property" do
```
polvalente, line 307: Reminder for myself to check if there's a property test in torchx and exla as well to change

polvalente: I think we'll also have to do some work to get the grad for the batched version to work because it uses Nx.LinAlg.invert

PR #872, opened 2022-08-12
fix: plain non-finites in bin_reduce

### LINKED ISSUE #871: Handle infinities in reduction operations in BinaryBackend
Currently this fails:

```elixir
iex> Nx.Constants.infinity() |> Nx.broadcast({10}) |> Nx.sum()
** (ArgumentError) expected a number or a scalar tensor of type {:f, 32}, got: :infinity
```

Same with other operations using `bin_reduce`.

polvalente:
closes #871


jonatanklosko (approved):
Awesome, thanks!

PR #878, opened 2022-08-14
feat(grad): Nx.product

polvalente:
more work towards #161

### DIFF
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -446,6 +446,72 @@ defmodule Nx.Defn.Grad do
     [{x, reduce_g(x, opts, g)}]
   end
 
+  defp grad(:product, [x, opts], ans, g) do
+    # From https://people.maths.ox.ac.uk/gilesm/files/NA-08-01.pdf
+    # 2.2.4 we take that the reverse-mode gradient of a determinant
+    # is A_bar = C_bar . C . inv(A^T)
+
+    # We also know that the determinant of a diagonal matrix is the
+    # product of the diagonal items. This means that we can imagine
+    # our condensed product axes as being the diagonal of a matrix
+    # and apply the determinant grad on it.
+
+    # Also, if A is diagonal, A^T = A and inv(A) is as trivial
+    # as inverting the items. Also, C_bar and C are scalars.
+    # This means that a_bar = c_bar |> Nx.multiply(c) |> Nx.divide(a).
+
+    # First, let's condition our input matrix by substituting 0s with 1s
+    # Then, we move the reducing axes into a condensed innermost axis.
+
+    axes =
+      case opts[:axes] do
+        [] -> Nx.axes(x.shape)
+        nil -> Nx.axes(x.shape)
+        axes -> axes
+      end
+
+    outermost_axes = Nx.axes(x.shape) -- axes
+    permutation = outermost_axes ++ axes
+
+    inverse_permutation =
+      permutation
+      |> Enum.with_index()
+      |> Enum.sort_by(fn {x, _} -> x end)
+      |> Enum.map(fn {_, i} -> i end)
+
+    zero_selector = Nx.equal(x, 0)
+
+    reduce_axes_size = axes |> Enum.reduce(1, fn axis, acc -> acc * elem(x.shape, axis) end)
+
+    reduced_shape =
+      outermost_axes
+      |> Enum.map(&elem(x.shape, &1))
+      |> List.insert_at(-1, reduce_axes_size)
+      |> List.to_tuple()
+
+    formatted_x =
+      zero_selector
+      |> Nx.select(1, x)
+      |> Nx.transpose(axes: permutation)
+      |> Nx.reshape(reduced_shape)
+
+    g_ans_new_axes = Tuple.to_list(g.shape) ++ List.duplicate(1, Nx.rank(x) - Nx.rank(g))
+
+    g_ans =
+      g
+      |> Nx.multiply(ans)
+      |> Nx.reshape(List.to_tuple(g_ans_new_axes))
+      |> Nx.tile(List.duplicate(1, tuple_size(g.shape)) ++ [reduce_axes_size])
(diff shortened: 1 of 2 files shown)


nx/lib/nx/defn/grad.ex
```diff
@@ -446,6 +446,76 @@ defmodule Nx.Defn.Grad do
...
+    # This means that a_bar = c_bar |> Nx.multiply(c) |> Nx.divide(a).
+
+    # First, let's condition our input matrix by substituting 0s with 1s
+    # Then, we move the reducing axes into a condensed innermost axis.
+
+    # After we have our matrix well conditioned, we can operate on the
+    # innermost axis as described above, and then we transpose the items
+    # back into place
+
+    axes =
```
polvalente, line 470: @jonatanklosko This one could really use your insight on some of the tensor shape  manipulations below

polvalente: perhaps the code can be simplified somewhere using `if opts[:keep_axes]`

nx/lib/nx/defn/grad.ex
```diff
@@ -446,6 +446,76 @@ defmodule Nx.Defn.Grad do
...
+    reduce_axes_size = axes |> Enum.reduce(1, fn axis, acc -> acc * elem(x.shape, axis) end)
+
+    reduced_shape =
+      outermost_axes
+      |> Enum.map(&elem(x.shape, &1))
+      |> List.insert_at(-1, reduce_axes_size)
+      |> List.to_tuple()
+
+    formatted_x =
+      zero_selector
```
polvalente, line 497: The way that Jax defines the result for 0-entries is closer to the forward-mode calculation:

A = product([a b c ... z])
dA = da.b.c....z + a.db.c..... + a.b...dz
and we can notice that each component of the addition on the right side is the same as:

A*da/a, A*db/b and so on (which is closer to the definition I reached as well).
However, this last definition is ill-defined when a given entry is 0.

@jonatanklosko any insights given this?

polvalente: For now the code doesn't break, but also doesn't reach the same result as Jax for 0 entries

jonatanklosko: Does jax special case or do they use the more explicit implementation?

polvalente: They use a different implementation in which ijstead of taking the product and dividing by each item, they take the product with each item removed.

I didn't mimic the implementation at first because they define only the forward-mode derivative. But looking at the current state, I'm thinking about using while to calculate this in a more iterative way.

WDYT?

jonatanklosko: FTR we discussed offline and have some ideas :)

nx/lib/nx/defn/grad.ex
```diff
@@ -446,6 +446,47 @@ defmodule Nx.Defn.Grad do
...
+    ans_removed_zero =
+      zero?
+      |> Nx.select(1, x)
+      |> Nx.product(axes: axes, keep_axes: true)
+
+    zeros_in_product = Nx.sum(zero?, axes: axes, keep_axes: true)
+    one_zero? = Nx.equal(zeros_in_product, 1)
+    many_zeros? = Nx.greater(zeros_in_product, 1)
+
+    dx = Nx.multiply(g, Nx.divide(ans, x))
```
polvalente, line 483: Isn't the division by zero here a problem?

nx/lib/nx/defn/grad.ex
```diff
@@ -446,6 +446,47 @@ defmodule Nx.Defn.Grad do
...
   end
 
+  defp grad(:product, [x, opts], ans, g) do
+    axes =
+      case opts[:axes] do
+        [] -> Nx.axes(x.shape)
+        nil -> Nx.axes(x.shape)
+        axes -> axes
+      end
+
```
jonatanklosko, line 456:
```suggestion
    axes = opts[:axes] || Nx.axes(x)
```

@polvalente in the other place we do this:

https://github.com/elixir-nx/nx/blob/84407cf85244faa83ef22995bbee3c35d5ec2f2e/nx/lib/nx/defn/grad.ex#L574-L576

Interestingly, binary backend raises on empty list, EXLA returns the tensor unchanged and Torchx multiples over all axes. We should unify this edge case, I think it would make sense to validate the list is not empty in the Nx functions upfront. wdyt?

nx/lib/nx/defn/grad.ex
jonatanklosko: Wanted to ask that, I would expect inf/nan that doesn't matter given we replace all the invalid values. But perhaps it's better not to rely on this?

nx/lib/nx/defn/grad.ex
polvalente: Yes, this makes sense!

nx/lib/nx/defn/grad.ex
jonatanklosko: Well, it raises for binary backend

```
iex(1)> Nx.divide(Nx.tensor(1), Nx.tensor(0))
** (ArithmeticError) bad argument in arithmetic expression
    (complex 0.4.2) lib/complex.ex:579: Complex.divide/2
    (nx 0.3.0) lib/nx/binary_backend.ex:2371: Nx.BinaryBackend."-binary_to_binary/4-lbc$^1/2-18-"/5
    (nx 0.3.0) lib/nx/binary_backend.ex:662: Nx.BinaryBackend.element_wise_bin_op/4
    iex:1: (file)
```

Though the grad tests pass, so that's a little surprising.

jonatanklosko: Doesn't hurt to use the zeroless tensor since we need it anyway, I will change!

jonatanklosko:
I rewrote it without transpose, I think it came out pretty simple in the end :D

polvalente (commented):
I can't approve my own PR for _some_ reason. The new impl looks a lot cleaner!

josevalim (approved):

PR #879, opened 2022-08-16
add unique/1 to nx to remove duplicate entries from 1D tensor

tiagodavi:
The idea is to be able to do something like that: https://scikit-learn.org/stable/modules/generated/sklearn.utils.multiclass.unique_labels.html

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -10907,6 +10907,43 @@ defmodule Nx do
   end
 
   @doc """
+  Remove all duplicate elements from 1D tensor.
+  ## Examples
+
+      iex> Nx.unique(Nx.tensor([1,1,2,3]))
+      #Nx.Tensor<
+        s64[3]
+        [1, 2, 3]
+      >
+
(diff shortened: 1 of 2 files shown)


nx/lib/nx.ex
```diff
@@ -10906,6 +10906,43 @@ defmodule Nx do
...
+  @doc type: :creation
+  @spec unique(tensor :: T.t()) :: T.t()
+  def unique(tensor) do
+    t = to_tensor(tensor)
+    t_rank = rank(tensor)
+
+    if t_rank == 1 do
+      t_items =
+        t
+        |> to_flat_list()
```
polvalente, line 10937: Thanks for the proposal.

However, this function cannot currently be supported in Nx because the defn compiler must know the output shapes at compile-time, which is impossible for this function.

For future contributions, also keeo in mind that we try to stay on "Nx land" as much as possible because when dealing with GPUs and other devices, there's overhead in transferring data to the CPU, so this implementation would need to be changed as well.

polvalente (changes requested):

PR #880, opened 2022-08-18
Support batched tensors in more of `Nx.LinAlg` functions

john-sonz:
Batched tensors support for:
- `Nx.LinAlg.adjoint`
- `Nx.LinAlg.cholesky`
- `Nx.LinAlg.lu`
- `Nx.LinAlg.svd`


nx/lib/nx/binary_backend.ex
```diff
@@ -1233,12 +1233,18 @@ defmodule Nx.BinaryBackend do
...
       ) do
     data = to_binary(tensor)
+    rank = tuple_size(output_shape)
+    n = elem(output_shape, rank - 1)
 
-    l = B.Matrix.cholesky(data, input_type, {m, n}, output_type)
+    l =
+      bin_batch_reduce(data, n * n, input_type, <<>>, fn matrix, acc ->
+        l = B.Matrix.cholesky(matrix, input_type, {n, n}, output_type)
+        <<acc::bitstring, l::bitstring>>
```
polvalente, line 1246:
```suggestion
        acc <> l
```

The same applies to all similar occurences below

nx/lib/nx/binary_backend.ex
```diff
@@ -1286,17 +1286,31 @@ defmodule Nx.BinaryBackend do
...
   @impl true
   def svd(
-        {%{shape: {m, _}} = u_holder, s_holder, %{shape: {_, n}} = v_holder} = outputs,
+        {u_holder, %{type: output_type} = s_holder, v_holder},
         %{type: input_type, shape: input_shape} = tensor,
         opts
       ) do
+    bin = to_binary(tensor)
+    rank = tuple_size(input_shape)
+    {m, n} = {elem(input_shape, rank - 2), elem(input_shape, rank - 1)}
```
polvalente, line 1295: Avoid multi-assigment like this in favor of one match per line please

nx/lib/nx/binary_backend.ex
```diff
@@ -1286,17 +1286,31 @@ defmodule Nx.BinaryBackend do
...
+
     if m < n do
       raise ArgumentError,
-            "SVD not implemented for wide matrices (tensors with shape {m, n} where m < n)"
+            "SVD not implemented for batches of wide matrices (matrices with shape {m, n} where m < n)"
```
polvalente, line 1299: A rank-2 tensor is not a batch, is it?

nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -546,11 +546,7 @@ defmodule Nx.BinaryBackend.Matrix do
       |> Enum.map(fn {row, idx} -> Enum.at(row, idx) end)
       |> Enum.reject(&is_nil/1)
 
-    {s, [vt_row | _] = vt} = apply_singular_value_corrections(s, vt)
-
-    if length(vt) != vt_rows or length(vt_row) != vt_cols do
```
polvalente, line 551: Seems like we lost this check during the refactor

nx/lib/nx/lin_alg.ex
```diff
@@ -38,31 +38,51 @@ defmodule Nx.LinAlg do
...
+        Nx.transpose(tensor, opts)
+    end
+  end
+
+  deftransformp adjoint_opts(shape, opts) do
+    rank = tuple_size(shape)
+
+    if rank > 2 do
+      axes = Enum.concat(0..(rank - 3), [rank - 1, rank - 2])
+      Keyword.put(opts, :axes, axes)
```
polvalente, line 58: This might make things confusing if the axes option was passed by the user

polvalente: Although it seems no options are actually documented, so perhaps we can remove the ability to pass any options as input to adjoint

nx/lib/nx/lin_alg.ex
```diff
@@ -38,31 +38,51 @@ defmodule Nx.LinAlg do
...
-  The matrix must be positive-definite and either Hermitian
+  The matrices must be positive-definite and either Hermitian
   if complex or symmetric if real. An error is raised by the
   default backend if those conditions are not met. Other
   backends may emit undefined behaviour.
 
   ### Examples
 
-      iex> Nx.LinAlg.cholesky(Nx.tensor([[20.0, 17.6], [17.6, 16.0]]))
+      iex> Nx.LinAlg.cholesky(Nx.tensor([[[20.0, 17.6], [17.6, 16.0]], [[1.0, 0.0], [0.0, 1.0]]]))
```
polvalente, line 74: We should keep the original example and add another one for the batched version

nx/lib/nx/lin_alg.ex
```diff
@@ -1009,113 +1029,104 @@ defmodule Nx.LinAlg do
...
+  Calculates the A = PLU decomposition of batched square 2-D matrices A.
 
   ## Options
 
     * `:eps` - Rounding error threshold that can be applied during the factorization
 
   ## Examples
 
-      iex> {p, l, u} = Nx.LinAlg.lu(Nx.tensor([[1, 2, 3], [4, 5, 6], [7, 8, 9]]))
+      iex> {p, l, u} = Nx.LinAlg.lu(Nx.tensor([[[1, 2, 3], [4, 5, 6], [7, 8, 9]], [[1, 0, 1], [-1, 0, -1], [1, 1, 1]]]))
```
polvalente, line 1052: Keep the original example please

nx/lib/nx/shape.ex
```diff
@@ -1721,24 +1721,36 @@ defmodule Nx.Shape do
...
   """
-  def cholesky({n, n}, names), do: {{n, n}, names}
+  def cholesky(shape, names) when tuple_size(shape) > 1 do
+    rank = tuple_size(shape)
+    matrix_shape = {elem(shape, rank - 2), elem(shape, rank - 1)}
+
+    unless match?({n, n}, matrix_shape) do
+      raise(
+        ArgumentError,
+        "tensor must be a batch of square matrices, got shape: #{inspect(shape)}"
```
polvalente, line 1742: I'm not sure if we should mention batches in the case where the input is a plain matrix, for all batch-related error messages

polvalente (commented):
Nice work!

nx/lib/nx/binary_backend/matrix.ex
john-sonz: Whoops 🙈

nx/lib/nx/lin_alg.ex
john-sonz: I did remove it, not sure why it was there in the first place

nx/lib/nx/shape.ex
john-sonz: I changed it to `must be a square matrix or a batch of square matrices`, but if you come up with a better message I'll be happy to change it

polvalente: This is a bit verbose, but it's clear on what's expect, so I think it's fine!

nx/lib/nx/lin_alg.ex
```diff
@@ -1065,7 +1093,6 @@ defmodule Nx.LinAlg do
           [7.0, 8.0, 9.0]
         ]
       >
-
       iex> {p, l, u} = Nx.LinAlg.lu(Nx.tensor([[1, 0, 1], [-1, 0, -1], [1, 1, 1]]))
```
josevalim, line 1096:
```suggestion

      iex> {p, l, u} = Nx.LinAlg.lu(Nx.tensor([[1, 0, 1], [-1, 0, -1], [1, 1, 1]]))
```

nx/lib/nx/lin_alg.ex
```diff
@@ -1009,25 +1037,25 @@ defmodule Nx.LinAlg do
...
-  Calculates the A = PLU decomposition of a 2-D tensor A with shape `{N, N}`.
+  Calculates the A = PLU decomposition of batched square 2-D matrices A.
 
   ## Options
 
     * `:eps` - Rounding error threshold that can be applied during the factorization
 
   ## Examples
-
       iex> {p, l, u} = Nx.LinAlg.lu(Nx.tensor([[1, 2, 3], [4, 5, 6], [7, 8, 9]]))
```
josevalim, line 1059:
```suggestion

      iex> {p, l, u} = Nx.LinAlg.lu(Nx.tensor([[1, 2, 3], [4, 5, 6], [7, 8, 9]]))
```

josevalim (approved):

polvalente (approved):

PR #888, opened 2022-08-25
Fix typo in intro-to-nx

ftes:
I think what that sentence is meant to say is

> `defn` is a subset of the language allowed in the `def` block.

### DIFF
--- a/nx/guides/intro-to-nx.livemd
+++ b/nx/guides/intro-to-nx.livemd
@@ -522,10 +522,9 @@ import Nx.Defn
 ```
 
 Just as the Elixir language supports `def`, `defmacro`, and `defp`,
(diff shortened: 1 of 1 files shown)


nx/guides/intro-to-nx.livemd
```diff
@@ -522,10 +522,9 @@ import Nx.Defn
 ```
 
 Just as the Elixir language supports `def`, `defmacro`, and `defp`,
-Nx supports `defn`. There are a few restrictions. It allows only
+Nx supports `defn`. `There are a few restrictions. It allows only
```
josevalim, line 525:
```suggestion
Nx supports `defn`. There are a few restrictions. It allows only
```

josevalim (approved):

josevalim:

PR #891, opened 2022-08-30
Enforce a noop when an empty :axes option is passed

jonatanklosko:
Semantically, in most cases empty `:axes` should be a noop, but currently binary backend raises an error, EXLA returns the tensor unchanged and Torchx multiples over all axes. This adds a check ensuring the axes are never empty.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -2606,12 +2606,15 @@ defmodule Nx do
 
       iex> Nx.squeeze(Nx.tensor([[[[[1]]]]]), axes: [0, 0])
       ** (ArgumentError) axes [0, 0] must be unique integers between 0 and 4
+
+      iex> Nx.squeeze(Nx.tensor([[1]]), axes: [])
+      ** (ArgumentError) :axes must be a non-empty list, got: []
   """
   @doc type: :shape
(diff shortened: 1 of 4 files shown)


josevalim (approved):

josevalim (commented):
it feels like noop is the correct behavior on all of them? Summing no axes should return the original tensor. Squeezing no axes as well… and so on.

nx/lib/nx.ex
```diff
@@ -8872,7 +8881,7 @@ defmodule Nx do
   def reverse(tensor, opts \\ []) do
     opts = keyword!(opts, [:axes])
     %{shape: shape, names: names} = tensor = to_tensor(tensor)
-    axes = opts[:axes] || axes(shape)
+    axes = assert_axes!(opts[:axes]) || axes(shape)
 
     case Nx.Shape.normalize_axes(shape, axes, names) do
       [] -> tensor
```
josevalim, line 8887: This clause is already specifically handling the empty list.

josevalim:
Ok, so I changed my mind. I think we should just be a no-op and not invoke the backend at all. Thoughts?

josevalim (approved):

PR #893, opened 2022-09-01
Implement cumulative operations using associative scan

### LINKED ISSUE #887: Cumulative sum runs slowly with EXLA backend
This takes around 200s for me:

```elixir
Mix.install([
  {:nx, github: "elixir-nx/nx", sparse: "nx", override: true},
  {:exla, github: "elixir-nx/nx", sparse: "exla", override: true}
])

x = Nx.iota({200000})
Nx.cumulative_sum(x)
```

Meanwhile, the corresponding jax script finishes immediately:

```python
import jax.numpy as jnp

x = jnp.arange(200000)
jnp.cumsum(x)
```

jonatanklosko:
Closes #887.

See https://github.com/elixir-nx/nx/issues/887#issuecomment-1232755018.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -371,6 +371,7 @@ defmodule Nx do
 
   import Nx.Shared
   import Nx.Defn.Kernel, only: [keyword!: 2]
+  import Nx.Defn
 
   alias Nx.Tensor, as: T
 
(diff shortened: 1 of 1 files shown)


nx/lib/nx.ex
```diff
@@ -7835,40 +7836,168 @@ defmodule Nx do
...
+  #           [0, 4]
+  #         ],
+  #         [
+  #           [8, 0],
+  #           [0, 8]
+  #         ]
+  #       ]
+  #     >
+  #
+  defnp associative_scan(tensor, fun, opts \\ []) do
```
josevalim, line 7903: We don't want this to be `defnp` because everything in the `Nx` module is considered a `defn` already. 

Otherwise I am worried we will run into weird scenarios like jitting this function while compiling another JIT and so on.

So please remove `import Nx.Defn`, and rewrite those to `defp` and we should be good!

PR #896, opened 2022-09-06
Fix EXLA compiler error with subsequent `if`s returning tuples

jonatanklosko:
Just a failing test for now.

### DIFF
--- a/exla/test/exla/defn/expr_test.exs
+++ b/exla/test/exla/defn/expr_test.exs
@@ -1196,6 +1196,16 @@ defmodule EXLA.Defn.ExprTest do
       assert_equal(if_tuple_match(Nx.tensor(1), Nx.tensor(10), Nx.tensor(20)), Nx.tensor(-10))
     end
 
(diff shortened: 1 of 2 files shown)


exla/test/exla/defn/expr_test.exs
```diff
@@ -1196,6 +1196,16 @@ defmodule EXLA.Defn.ExprTest do
       assert_equal(if_tuple_match(Nx.tensor(1), Nx.tensor(10), Nx.tensor(20)), Nx.tensor(-10))
     end
 
+    defn if_tuple_match_twice(a, b) do
+      {a, b} = if(a, do: {a, b}, else: {a, b})
+      {a, b} = if(a, do: {a, b}, else: {a, b})
```
jonatanklosko, line 1201: Note that the following passes

```suggestion
      {a, b} = if(a == b, do: {a, b}, else: {a, b})
      {a, b} = if(a == b, do: {a, b}, else: {a, b})
```

josevalim (approved):

PR #898, opened 2022-09-08
ToLiteral() is as synchronous operation that uses the host shape

josevalim:


josevalim:

PR #900, opened 2022-09-13
fix: compatibilize shape in some operations

polvalente:
Part of the bugs caught by Axon

1. conv output_permutations were in disagreement with Nx.Shape
2. binary ops didn't broadcast args properly

### DIFF
--- a/torchx/test/torchx/nx_test.exs
+++ b/torchx/test/torchx/nx_test.exs
@@ -938,6 +938,32 @@ defmodule Torchx.NxTest do
         ])
       )
     end
+
+    test "output_permutation" do
+      assert_all_close(
+        Nx.conv(Nx.iota({1, 3, 3, 6}), Nx.broadcast(1, {2, 6, 1, 1}),
+          input_permutation: [0, 3, 1, 2],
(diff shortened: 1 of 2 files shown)


torchx/test/torchx/nx_test.exs
```diff
@@ -62,6 +62,25 @@ defmodule Torchx.NxTest do
 
         test_binary_op(op, type_a, type_b)
       end
+
+      test "#{op}(#{Nx.Type.to_string(type_a)}, #{Nx.Type.to_string(type_b)}) broadcast" do
```
polvalente, line 66: not sure if we should add all of these, though, because this adds 1k tests to the file (praise the power of permutations)

josevalim: Probably too many. Perhaps we should get a few as reference only?

josevalim (approved):

torchx/test/torchx/nx_test.exs
polvalente: I've changed the generator to only test for same-type broadcasts instead (1k -> 100 tests instead)

PR #905, opened 2022-09-14
refactor(torchx): fake u64 with s64

polvalente:
Torchx currently doesn't play nice with Nx.sum and other similar 
aggregate functions for unsigned integers. By faking u64 with s64,
we can support a broader range of inputs.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -4016,10 +4016,10 @@ defmodule Nx do
       >
 
       iex> left = Nx.tensor([[10, 20]], type: {:u, 8}, names: [:x, :y])
-      iex> right = Nx.tensor([[1], [2]], type: {:u, 32})
+      iex> right = Nx.tensor([[1], [2]], type: {:u, 64})
       iex> Nx.quotient(left, right)
       #Nx.Tensor<
-        u32[x: 2][y: 2]
+        u64[x: 2][y: 2]
         [
           [10, 20],
           [5, 10]
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -4016,10 +4016,10 @@ defmodule Nx do
       >
 
       iex> left = Nx.tensor([[10, 20]], type: {:u, 8}, names: [:x, :y])
-      iex> right = Nx.tensor([[1], [2]], type: {:u, 64})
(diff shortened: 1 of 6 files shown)


nx/lib/nx.ex
```diff
@@ -4016,10 +4016,10 @@ defmodule Nx do
       >
 
       iex> left = Nx.tensor([[10, 20]], type: {:u, 8}, names: [:x, :y])
-      iex> right = Nx.tensor([[1], [2]], type: {:u, 32})
+      iex> right = Nx.tensor([[1], [2]], type: {:u, 64})
       iex> Nx.quotient(left, right)
       #Nx.Tensor<
-        u32[x: 2][y: 2]
+        u64[x: 2][y: 2]
```
polvalente, line 4022: I changed this test/example just to avoid having to implement u32 while getting rid of the last "temporarily broken" doctest. Let me know if I should revert this

torchx/lib/torchx/backend.ex
```diff
@@ -226,15 +231,23 @@ defmodule Torchx.Backend do
...
+    |> maybe_pad_binary(type)
+    |> Torchx.from_blob(
       shape,
       to_torch_type(type),
       device_option(backend_options)
     )
     |> to_nx(out)
   end
 
+  defp maybe_pad_binary(bin, {_, size} = type) when type in [{:u, 16}, {:u, 32}] do
```
josevalim, line 244:
```suggestion
  defp maybe_pad_binary(bin, {:u, size}) when size in [16, 32] do
```

josevalim (approved):

nx/lib/nx.ex
```diff
@@ -4016,10 +4016,10 @@ defmodule Nx do
       >
 
       iex> left = Nx.tensor([[10, 20]], type: {:u, 8}, names: [:x, :y])
-      iex> right = Nx.tensor([[1], [2]], type: {:u, 32})
+      iex> right = Nx.tensor([[1], [2]], type: {:u, 64})
```
polvalente, line 4019:
```suggestion
      iex> right = Nx.tensor([[1], [2]], type: {:u, 32})
```

nx/lib/nx.ex
polvalente:
```suggestion
        u32[x: 2][y: 2]
```

PR #910, opened 2022-09-16
feat(torchx): support window dilations

polvalente:


torchx/lib/torchx/backend.ex
```diff
@@ -1545,26 +1547,16 @@ defmodule Torchx.Backend do
...
-    # if Enum.any?(opts[:padding], fn conf -> conf |> Tuple.to_list() |> Enum.any?(&(&1 != 0)) end) do
-    #   raise ArgumentError, "padding unsupported"
-    # end
-
-    intermediate_type =
-      tensor.type
-      |> Nx.Type.to_floating()
-      |> to_torch_type()
-
     t_tx =
```
polvalente, line 1550: I'm not sure what was the deal with the intermediate_type casting... The floating point conversion was actually breaking s64 window_min

josevalim (approved):

PR #916, opened 2022-09-19
Preserve output type for uniform

seanmor5:
This fixes failing tests in Axon's RNG


polvalente (approved):

nx/lib/nx/random.ex
```diff
@@ -401,6 +401,9 @@ defmodule Nx.Random do
     shape = Nx.shape(opts[:shape])
     u_one = Nx.tensor(1.0, type: type) |> Nx.bitcast({:u, nbits})
 
+    min_value = Nx.as_type(min_value, type)
+    max_value = Nx.as_type(max_value, type)
```
polvalente, line 405: Should we add a test for this?

PR #918, opened 2022-09-20
Fix Torchx dot product with multiple batch axes

jonatanklosko:
Currently it fails with

```
terminate called after throwing an instance of 'c10::Error'
  what():  bdim.level() > prev_levelINTERNAL ASSERT FAILED at "../aten/src/ATen/BatchedTensorImpl.cpp":73, please report a bug to PyTorch.
```

Each batch dimension is effectively a vmap and we need to specify an increasing vmap level.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -8714,6 +8714,17 @@ defmodule Nx do
         ]
       >
 
+      iex> u = Nx.tensor([[[1, 1]], [[2, 2]]])
+      iex> v = Nx.tensor([[[1, 2]], [[1, 2]]])
+      iex> Nx.dot(u, [2], [0, 1], v, [2], [0, 1])
+      #Nx.Tensor<
+        s64[2][1]
+        [
+          [3],
+          [6]
+        ]
+      >
+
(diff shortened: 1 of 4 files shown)


nx/lib/nx.ex
```diff
@@ -8714,6 +8714,17 @@ defmodule Nx do
...
+      iex> u = Nx.tensor([[[1, 1]], [[2, 2]]])
+      iex> v = Nx.tensor([[[1, 2]], [[1, 2]]])
+      iex> Nx.dot(u, [2], [0, 1], v, [2], [0, 1])
+      #Nx.Tensor<
+        s64[2][1]
+        [
+          [3],
+          [6]
+        ]
+      >
```
josevalim, line 8726: Please add this test to Torchx suite directly instead? The doctests should be about docs mainly. :)

jonatanklosko: I was thinking this way we validate multiple axes work for every backend that runs the doctests, but we can move it to torchx specifically yeah.

josevalim (approved):

PR #923, opened 2022-09-21
Fix operations for u16/u32/u64 on Torchx

josevalim:


josevalim:
Thanks to @polvalente on the pairing!

nx/lib/nx/random.ex
```diff
@@ -129,33 +129,6 @@ defmodule Nx.Random do
...
-
-      32 ->
-        threefry2x32(key, Nx.iota(shape, type: {:s, 64}))
-        |> Nx.as_type({:u, 32})
-
-      _ ->
-        threefry2x32(key, Nx.iota(shape, type: {:s, 64}))
-        |> Nx.as_type({:u, bit_width})
-    end
-  end
```
josevalim, line 157: No changes here, I just moved it down.

josevalim: The rest of the changes in the file is to require fewer `as_type`. No changes to the algo.

torchx/lib/torchx/backend.ex
```diff
@@ -808,17 +814,46 @@ defmodule Torchx.Backend do
     raise ArithmeticError, "Torchx does not support complex values for atan2"
   end
 
+  defp bitmask(tensor, {:u, 16}),
+    do: Torchx.bitwise_and(tensor, Torchx.scalar_tensor(0xFFFF, :int, :cpu))
+
+  defp bitmask(tensor, {:u, 32}),
+    do: Torchx.bitwise_and(tensor, Torchx.scalar_tensor(0xFFFF_FFFF, :long, :cpu))
+
+  defp bitmask(tensor, _),
```
polvalente, line 823: What happens if we mask with 0xffff_ffff here instead?

polvalente: Or is the idea to emulate as best as possible u64?

josevalim:

PR #926, opened 2022-09-22
fix: set negative image for signed values

polvalente:


torchx/nx.code-workspace
```diff
@@ -0,0 +1,7 @@
+{
+  "folders": [
+    {
+      "path": ".."
+    }
+  ]
+}
```
jonatanklosko, line 7: :cat2:

jonatanklosko (approved):
Unless anyone has objections to enforcing a specific behaviour, this looks good to me! @josevalim @seanmor5

josevalim (approved):

PR #929, opened 2022-09-23
Support generators and unrolling in while

josevalim:


nx/lib/nx/defn/kernel.ex
```diff
@@ -1046,42 +1046,79 @@ defmodule Nx.Defn.Kernel do
...
+
+      while acc = 0, i <- tensor, unroll: true do
+        acc + i
+      end
+
+  Or unroll it in batches:
+
+      while acc = 0, i <- tensor, unroll: 4 do
+        acc + i
+      end
```
jonatanklosko, line 1080:
```suggestion
      end
  
  Unrolling means that the the `while` body is automatically duplicated
  a certain amount of times, as if you wrote all iterations by hand. This
  makes the final expression larger, which causes a longer compilation
  time, however it enables additional compile-time optimizations (such as
  fusion), improving the runtime efficiency.
```

nx/lib/nx/defn/kernel.ex
```diff
@@ -1046,42 +1046,79 @@ defmodule Nx.Defn.Kernel do
...
```
jonatanklosko, line 1081: I think we should also add a warning note that `while` should generally be avoided unless there is no way to express the algorithm with regular tensor operations.

nx/lib/nx/defn/expr.ex
```diff
@@ -305,18 +305,99 @@ defmodule Nx.Defn.Expr do
...
-        {parameter(expr, context, counter), {counter + 1, merge_context!(expr, acc)}}
-      end)
+    case generator do
+      :none ->
+        if opts != [] do
+          raise CompileError,
+            line: line,
+            file: file,
+            description:
+              "options are only supported on while with conditions, only with generators, got: #{inspect(opts)}"
```
jonatanklosko, line 318: ?

```suggestion
              "options are only supported on while with generators, got: #{inspect(opts)}"
```

josevalim:
```suggestion
              "options are not supported on while with conditions, only with generators, got: #{inspect(opts)}"
```

nx/lib/nx/defn/kernel.ex
josevalim: Done!

PR #931, opened 2022-09-23
Get deps folder and mix.lock file with project name on CI

### LINKED ISSUE #867: Mix deps cache always misses on CI
The key which is based on mix.lock fails to point to the correct file (probably due to the working_directory change).

Ian-GL:
This is an attempt to solve the issue of deps cache on CI. If I understand this correctly, the key field does need to get the info for the `mix.lock` corresponding to each of the projects and also the deps field will need the path that includes the project name.

Closes #867


polvalente:
closing in favor of #941

PR #933, opened 2022-09-25
Adapt `eigh` function to Hermitian

kenken-neko:
# Change
The input to function `eigh` is extended from a symmetric matrix to a Hermitian so that eigenvalues can be obtained.

# Supplements
- Reference Algorithm:  https://people.inf.ethz.ch/arbenz/ewp/Lnotes/chapter4.pdf
- I have verified that the results of [Doctest](https://github.com/elixir-nx/nx/blob/a573745447180bd4f1d4817740e590f4040d7aab/nx/lib/nx/lin_alg.ex#L1025-L1040) match the output of `numpy.linalg.eigh` with the same input Hermitian matrix.
- The `lin_alg_test.exs` was extended to Hermitian, but the number of test cases generated was [reduced from 10 to 3 ](https://github.com/elixir-nx/nx/blob/a573745447180bd4f1d4817740e590f4040d7aab/nx/test/nx/lin_alg_test.exs#L435)because convergence takes time.


kenken-neko:
An error has occurred in the automated test of `torchx`. I am not familiar with the implementation of `torchx`. Please wait a moment as I need some time to investigate.

kenken-neko:
I have finished addressing the CI check errors. Next time I will remember to run the `torchx` test as well as `nx` before opening PR.

kenken-neko:
I have made some additional commits in response to CI checks since PR opened. This PR is currently available for review.

nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -679,7 +686,7 @@ defmodule Nx.BinaryBackend.Matrix do
     # receives a_reverse as a list of numbers and returns the reflector as a
     # k x k matrix
 
-    norm_a_squared = Enum.reduce(a, 0, fn x, acc -> x * x + acc end)
+    norm_a_squared = Enum.reduce(a, 0, fn x, acc -> Complex.abs(x) * Complex.abs(x) + acc end)
```
polvalente, line 689:
```suggestion
    norm_a_squared = Enum.reduce(a, 0, fn x, acc -> x * Complex.conjugate(x) + acc end)
```

nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -849,7 +856,7 @@ defmodule Nx.BinaryBackend.Matrix do
       |> Enum.map(fn col ->
         row
         |> Enum.zip(col)
-        |> Enum.reduce(0, fn {x, y}, acc -> acc + x * y end)
+        |> Enum.reduce(0, fn {x, y}, acc -> acc + Complex.multiply(x, y) end)
```
polvalente, line 859:
```suggestion
        |> Enum.reduce(0, fn {x, y}, acc -> acc + x * Complex.conjugate(y) end)
```

nx/lib/nx/lin_alg.ex
```diff
@@ -1022,20 +1022,35 @@ defmodule Nx.LinAlg do
         ]
       >
 
+      iex> a = Nx.tensor([[1, Complex.new(0, 2), 2], [Complex.new(0, -2), -3, Complex.new(0, 2)], [2, Complex.new(0, -2), 1]])
```
polvalente, line 1025: we should probably keep this as a test instead of a doctest due to readability on the eigenvec matrix

polvalente (approved):
Thanks!

kenken-neko:
@polvalente 
I'm currently in the process of correcting two things: changing doctest to test and one commit suggestion that seems to be wrong. Sorry, please wait a moment.

nx/lib/nx/binary_backend/matrix.ex
kenken-neko: The process here is correct to multiply complex elements (x * y), but the process is different from the (x*y^*) suggested. I misunderstood and committed the suggestion once myself, but I reverted to the original code. [56e6662](https://github.com/elixir-nx/nx/pull/933/commits/56e66629ad7cdcf22c07230a6793947713e858c5)

nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -860,7 +867,7 @@ defmodule Nx.BinaryBackend.Matrix do
...
```
polvalente, line 870: This is wrong. It must be one times the complex conjugate of the other.

By the way, this module imports Complex.Kernel, so the * operator is the same as Complex.multiply

polvalente: I saw that fixing this had broken a test, but the definition must be so, and one way to think about that is that a dot(v, v) must yield norm(v) squared, and that will only happen if you conjugate one of the operands

kenken-neko: > This is wrong. It must be one times the complex conjugate of the other.

In `dot_matrix`, I just want to do a simple matrix product (A*B), not A*B^*. Why do I have to take the complex conjugate here?  By the way, if I take the complex conjugate here, I get an error in the test.

> By the way, this module imports Complex.Kernel, so the * operator is the same as Complex.multiply

Thank you for letting us know!

polvalente: The dot product for complex vectors is defined as A^*.B instead of A^t.B. The thing here is that it's the adjoint vector that matters, not the transpose in itself, and the adjoint for a complex vector is the transpose conjugate.

Rule of thumb is that every time you see a transpose in real matrices you swap it with a transpose conjugate for complex extension. 

One of the reasons is what I mentioned about dot(v, v) === norm(v)^2

https://en.m.wikipedia.org/wiki/Dot_product#Complex_vectors

kenken-neko: Thank you, I will fix the `dot_matrix` to that of suggestion and correct the incorrect use of `dot_matrix`.

kenken-neko:
This commit [2c58346](https://github.com/elixir-nx/nx/pull/933/commits/2c583462d97bd98ec7fda35f3802be614653cd7c) mistake. Please wait a moment while we reissue the correct commit.

nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -214,15 +214,18 @@ defmodule Nx.BinaryBackend.Matrix do
...
+          h_cc = conjugate_matrix(h)
+          r_cc = conjugate_matrix(r)
+
           # If we haven't allocated Q yet, let Q = H1
           q =
             if is_nil(q) do
               h
             else
-              dot_matrix(q, h)
+              dot_matrix(q, h_cc)
```
polvalente, line 225: Why are you taking the conjugate here?

kenken-neko:
Sorry. I have company business to attend to so I will proceed with the revision in about 8 hours.

polvalente:
> Sorry. I have company business to attend to so I will proceed with the revision in about 8 hours.

No rush!

kenken-neko:
The function `dot_matrix` was fixed to handle A*B^* processing as suggested. On the other hand, a new function `multiply_matrix` was defined and used it to handle A*B processing.

nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -854,6 +861,18 @@ defmodule Nx.BinaryBackend.Matrix do
...
+      |> transpose_matrix()
+      |> Enum.map(fn col ->
+        row
+        |> Enum.zip(col)
+        |> Enum.reduce(0, fn {x, y}, acc -> acc + x * Complex.conjugate(y) end)
+      end)
+    end)
+  end
+
+  defp multiply_matrix(m1, m2) do
```
polvalente, line 875:
```suggestion
  defp dot_matrix_real(m1, m2) do
```

nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -377,16 +381,19 @@ defmodule Nx.BinaryBackend.Matrix do
             if is_nil(q) do
               h
             else
-              dot_matrix(q, h)
+              multiply_matrix(q, h)
```
polvalente, line 384: Is the `h` matrix complex? wouldn't taking the real part of that suffice?
using `multiply_matrix` with 2 complex matrices can lead to problems

nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -377,16 +381,19 @@ defmodule Nx.BinaryBackend.Matrix do
...
             end
 
           # Hessenberg matrix H updating
-          h_t = transpose_matrix(h)
+          h_cc =
+            h
+            |> transpose_matrix()
+            |> Enum.map(fn a_row -> Enum.map(a_row, &Complex.conjugate(&1)) end)
```
polvalente, line 391: I would write an `ajdoint_matrix` function that transposes and conjugates in a single pass

nx/test/nx/lin_alg_test.exs
```diff
@@ -1,5 +1,6 @@
 defmodule Nx.LinAlgTest do
   use ExUnit.Case, async: true
+  use Complex.Kernel
```
polvalente, line 3: I don't thing we should use Complex.Kernel here

nx/test/nx/lin_alg_test.exs
```diff
@@ -426,14 +427,50 @@ defmodule Nx.LinAlgTest do
...
+
+      assert {eigenvals, eigenvecs} = Nx.LinAlg.eigh(t)
+
+      # Eigenvalues
+      assert eigenvals ==
+               Nx.tensor([Complex.new(-5, 0), Complex.new(3, 0), Complex.new(1, 0)])
+
+      # Eigenvectors
+      assert complex_round(eigenvecs, 3) ==
+               Nx.tensor([
```
polvalente, line 447: shouldn't these results also be rounded to 3 places?

nx/test/nx/lin_alg_test.exs
```diff
@@ -426,14 +427,50 @@ defmodule Nx.LinAlgTest do
...
+      # Generate real Hermitian matrices with different eigenvalues
+      # from random matrices based on the relation A = Q.Λ.Q^*
+      # where Λ is the diagonal matrix of eigenvalues and Q is unitary matrix.
 
-      for _ <- 1..10 do
-        # Orthogonal matrix from a random matrix
-        {q, _} = Nx.random_uniform({3, 3, 3}) |> Nx.LinAlg.qr()
+      for _ <- 1..3 do
+        # Unitary matrix from a random matrix
+        {q, _} = Nx.random_uniform({3, 3, 3}, type: {:c, 64}) |> Nx.LinAlg.qr()
```
polvalente, line 473: we should ensure it still works with floats, so I would iterate on the type as well
```suggestion
      for _ <- 1..3, type <- [f: 32, c: 64] do
        # Unitary matrix from a random matrix
        {q, _} = Nx.random_uniform({3, 3, 3}, type: type) |> Nx.LinAlg.qr()
```

nx/test/nx/lin_alg_test.exs
```diff
@@ -450,33 +487,34 @@ defmodule Nx.LinAlgTest do
...
-      # from random matrices based on the relation A = Q.Λ.Q^t
+      # Generate real Hermitian matrices with close eigenvalues
+      # from random matrices based on the relation A = Q.Λ.Q^*
 
-      for _ <- 1..10 do
-        # Orthogonal matrix from a random matrix
-        {q, _} = Nx.random_uniform({3, 3, 3}) |> Nx.LinAlg.qr()
+      for _ <- 1..3 do
+        # Unitary matrix from a random matrix
+        {q, _} = Nx.random_uniform({3, 3, 3}, type: {:c, 64}) |> Nx.LinAlg.qr()
```
polvalente, line 517:
```suggestion
      for _ <- 1..3, type <- [f: 32, c: 64] do
        # Unitary matrix from a random matrix
        {q, _} = Nx.random_uniform({3, 3, 3}, type: type) |> Nx.LinAlg.qr()
```

nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -952,7 +971,7 @@ defmodule Nx.BinaryBackend.Matrix do
   end
 
   defp approximate_zeros(matrix, tol) do
-    do_round = fn x -> if abs(x) < tol, do: 0, else: x end
+    do_round = fn x -> if Complex.abs(x) < tol, do: 0, else: x end
```
polvalente, line 974:
```suggestion
    do_round = fn x -> if Complex.abs(x) < tol, do: 0 * x, else: x end
```

to ensure that if x is complex, it stays complex

nx/lib/nx/binary_backend/matrix.ex
kenken-neko: `q`, `h` may also have complex numbers as elements. What problems may arise?

nx/lib/nx/binary_backend/matrix.ex
kenken-neko: I did not know what to do. Do you mean that I should define a private function `adjoint_matrix` ?

nx/test/nx/lin_alg_test.exs
kenken-neko: This description is unnecessary and should be deleted.

nx/test/nx/lin_alg_test.exs
kenken-neko: Does it mean that the value should not be rounded ?

nx/lib/nx/binary_backend/matrix.ex
polvalente: yes

nx/lib/nx/binary_backend/matrix.ex
polvalente: The QR definition is wrong, for instance.

The algorithm is defined such that each iteration builds Q and R by composing H_i matrices over a pair of matrices, such that premultiplying with H_i zeroes out the column i under the diagonal.

So we have {I, A} -> {H1, H1*.A} -> {H1.H2, H2*.H1*.A} and so on until {Q, R}. But the definition is only correct if Q = H1.H2...Hn and R = Hn*....H2*.H1.A, because when doing Q.R you get the H_i.H_i* = I pairs to cancel out. That is:

Q.R = H1.H2...(Hn.Hn*)....H2*.H1.A = H1.(H2.H2*).H1*.A = H1.H1*.A = A

By using "multiply_matrix" instead of "dot_matrix", you're removing that property.
If something went wrong, it's probably on line 382 because there's an implicit dot_matrix(identity, h) there, that is, you must return adjoint(h) there

polvalente: same implicit dot_matrix at line 220

nx/test/nx/lin_alg_test.exs
polvalente: I don't get what you mean

nx/test/nx/lin_alg_test.exs
polvalente: I meant that it doesn't make sense to write out that many decimal places if you're comparing to a rounded-to-3-places vector on the left

nx/test/nx/lin_alg_test.exs
kenken-neko: I'll delete `use Complex.Kernel`.

nx/lib/nx/binary_backend/matrix.ex
kenken-neko: I defined private function `adjoint_matrix`.

nx/test/nx/lin_alg_test.exs
kenken-neko: I deleted `use Complex.Kernel`.

nx/lib/nx/binary_backend/matrix.ex
kenken-neko: I changed the function in lines 220 and 384 to `dot_matrix` as [the formula shown](https://github.com/elixir-nx/nx/pull/933#discussion_r982887087). However, an error occurred in tests such as `eigh` using the function of QR decomposition. On the other hand, if the functions in lines 220 and 384 are left as `dot_matrix_real`, no error occurs.

polvalente: I checked your branch out locally, and there's for sure some inconsistency with the householder reflector that turns that dot_matrix_real necessary, but we can keep it for now.

kenken-neko: Thanks for the advice. Yes, I agree.
I added the TODO comments and discussion URL so we don't forget.  [5e670a7](https://github.com/elixir-nx/nx/pull/933/commits/5e670a7d91485623ebcf9407f01d43d4b7ef2ce7)

kenken-neko:
Fixed a point where [the test could not pass due to a difference in sign of zero](https://github.com/elixir-nx/nx/actions/runs/3147752019/jobs/5117554146), perhaps due to a slight change in numerical error.  d3d7875

polvalente (approved):

PR #941, opened 2022-10-01
Fix dependency caching in CI

### LINKED ISSUE #867: Mix deps cache always misses on CI

Benjamin-Philip:
This commit fixes dependency caching by doing the following:

- Upgrading from cache@v1 to cache@v3
- Using a more verbose cache path
- Changing the key format to be a bit more clearer

Caching attempts were failing as result of an invalid path, which did not account for the change in working directory.
By using a more verbose path, this issue has been eliminated.

Closes #867

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,8 @@ jobs:
         id: mix-cache # id to use in retrieve action
         with:
           path: deps
-          key: v1-${{ runner.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-mix-${{ hashFiles('**/mix.lock') }}
+          key: v1-${{ runner.os }}-${{ matrix.elixir }}-${{ matrix.otp }}-mix-${{ hashFiles('**/mix.lock') }}
+          restore-keys: v1-${{ runner.os }}-${{ matrix.elixir }}-${{ matrix.otp }}-mix-
       - name: Retrieve libtorch cache
         if: ${{ matrix.working_directory == 'torchx' }}
         uses: actions/cache@v1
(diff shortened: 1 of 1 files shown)


Benjamin-Philip:
I hate debugging CIs since this. Inability to test something on my own machine, and instead pushing to github and waiting for a runner is infuriating. I wish we had an alternative to GH Actions based on git hooks and bash scripts instead!

.github/workflows/ci.yml
```diff
@@ -28,11 +28,13 @@ jobs:
...
+        uses: actions/cache@v3
         id: mix-cache # id to use in retrieve action
         with:
-          path: deps
-          key: v1-${{ matrix.os }}-${{ matrix.otp }}-${{ matrix.elixir }}-mix-${{ hashFiles(format('{0}{1}', github.workspace, '/mix.lock')) }}
+          path: ${{ github.workspace }}/${{ matrix.working_directory }}/deps
+          key: ${{ runner.os }}-Elixir-v${{ matrix.elixir }}-OTP-${{ matrix.otp }}-${{ hashFiles(format('{0}/{1}/mix.lock', github.workspace, matrix.working_directory)) }}
       - name: Retrieve libtorch cache
         if: ${{ matrix.working_directory == 'torchx' }}
         uses: actions/cache@v1
```
polvalente, line 40: There's at least this other usage of cache that needs to be bumped. I've updated the allowed actions list to allo cache v3

Benjamin-Philip: Ah yes. I had forgotten about the libtorch caches. I will update them soon.

Benjamin-Philip: Fixed.

Benjamin-Philip:
You will have to rerun the job for the deps to cache.

polvalente (approved):

polvalente:
Thanks! It seems that both libtorch and mix deps cache are working now!

PR #942, opened 2022-10-02
Remove elements from cache if they are no longer going to be used

### LINKED ISSUE #920: Perform cache lifetime analysis in Nx.Defn.Evaluator
Today we keep the cache alive forever, but we need to purge entries once they are all used.

josevalim:
Closes #920.

### DIFF
--- a/nx/lib/nx/defn/evaluator.ex
+++ b/nx/lib/nx/defn/evaluator.ex
@@ -156,7 +156,7 @@ defmodule Nx.Defn.Evaluator do
         {Map.new(clause_cache), seen_ids_acc}
       end)
 
-    {Map.put(meta, id, {clauses_cache, last_cache, all_ids}), cache}
+    {Map.put(meta, id, {clauses_cache, last_cache, Map.keys(all_ids)}), cache}
   end
 
   defp compute_cache(:token, %{data: %Expr{args: [token], id: id}}, state, meta, cache) do
@@ -215,6 +215,9 @@ defmodule Nx.Defn.Evaluator do
     {other, caches}
   end
 
+  defp decrement_cache(cache, id, 1, _res), do: Map.delete(cache, id)
+  defp decrement_cache(cache, id, counter, res), do: %{cache | id => {counter - 1, res}}
+
   defp eval_parent([cache | caches], id, op, ans, state, acc) do
     case cache do
       %{^id => {_count, res}} ->
@@ -235,8 +238,13 @@ defmodule Nx.Defn.Evaluator do
             "Please report this bug with the relevant code that triggers it: https://github.com/elixir-nx/nx"
   end
 
-  defp decrement_cache(cache, id, 1, _res), do: Map.delete(cache, id)
-  defp decrement_cache(cache, id, counter, res), do: %{cache | id => {counter - 1, res}}
+  defp decrement_parents([cache | caches], id) do
+    case cache do
+      %{^id => {count, value}} -> [decrement_cache(cache, id, count, value) | caches]
+      %{^id => count} -> [%{cache | id => count - 1} | caches]
+      %{} -> [cache | decrement_parents(caches, id)]
+    end
+  end
 
   defp eval_apply(:parameter, %{data: %Expr{args: [i]}}, state, caches) do
     case Enum.fetch!(state.params, i).() do
@@ -283,7 +291,7 @@ defmodule Nx.Defn.Evaluator do
   end
 
   defp eval_apply(:cond, %{data: %Expr{args: [clauses, last], id: id}}, state, caches) do
-    {clauses_cache, last_cache, add_ids} = Map.fetch!(state.meta, id)
+    {clauses_cache, last_cache, parent_ids} = Map.fetch!(state.meta, id)
 
     {chosen, chosen_cache} =
       clauses
@@ -291,30 +299,7 @@ defmodule Nx.Defn.Evaluator do
       |> cond_clause(last, last_cache, state, caches)
 
     {res, [_ | caches]} = composite_eval(chosen, state, chosen_cache)
-
-    # caches =
-    #   Enum.reduce(all_ids, caches, fn id, caches ->
-    #     case caches do
-    #       # Value was cached, decrement it after cond
-    #       %{^id => {count, value}} ->
-    #         decrement_cache(cache, id, count, value)
-
-    #       # Value was not cached but it was only used in cond
-    #       %{^id => 1} ->
-    #         Map.delete(cache, id)
-
-    #       # Value was not cached but it is meant to be used elsewhere...
-    #       %{^id => count} when is_integer(count) ->
-    #         case res_cache do
-    #           # and we computed it in cond, so we cache it!
-    #           # TODO: Do not expire cond caches if value is used upstream
-    #           %{^id => {_, value}} -> decrement_cache(cache, id, count, value)
-    #           # and we did not compute it in cond, so we only decrement it!
-    #           %{} -> %{cache | id => count - 1}
-    #         end
-    #     end
-    #   end)
-
+    caches = Enum.reduce(parent_ids, caches, &decrement_parents(&2, &1))
     {res, caches}
   end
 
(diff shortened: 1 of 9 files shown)


nx/lib/nx/defn/evaluator.ex
```diff
@@ -42,135 +42,304 @@ defmodule Nx.Defn.Evaluator do
...
-  defp eval(%Nx.Tensor{data: %Expr{op: :tensor, args: [t]}}, _state, cache) do
-    {t, cache}
+  defp precompile(fun, vars, hooks) do
+    expr = fun.(vars)
+    state = %{hooks: hooks, parent_ids: nil, current_ids: nil}
+    cache = init_compute_cache(expr, state, %{})
+    {expr, cache}
+  end
+
+  defp init_compute_cache(expr, state, parent_ids) do
```
jonatanklosko, line 63: fwiw currently we always call it with `parent_ids = %{}`.

nx/lib/nx/defn/evaluator.ex
```diff
@@ -42,135 +42,304 @@ defmodule Nx.Defn.Evaluator do
...
+    {elem(tuple, i), caches}
+  end
+
+  defp eval_apply(:attach_token, %Nx.Tensor{data: %Expr{args: [token, expr]}}, state, caches) do
+    {_, caches} = eval(token, state, caches)
+    eval(expr, state, caches)
+  end
+
+  defp eval_apply(:fun, %{data: %Expr{args: [args, expr, _mfa], id: id}}, state, caches) do
+    fun_cache = Map.fetch!(hd(caches), [:fun | id])
```
jonatanklosko, line 260: Curious: what's the benefit over `{:fun, id}`?

nx/lib/nx/defn/evaluator.ex
```diff
@@ -42,135 +42,304 @@ defmodule Nx.Defn.Evaluator do
...
 
-  defp eval(%Nx.Tensor{data: %Expr{op: op, id: id}} = ans, state, cache) do
+  defp eval(%Nx.Tensor{data: %Expr{op: op, id: id}} = ans, state, [cache | caches]) do
     case cache do
-      %{^id => res} ->
-        {res, cache}
+      %{^id => count} when is_integer(count) ->
+        {res, [cache | caches]} = eval_apply(op, ans, state, [cache | caches])
+        state.gc && :erlang.garbage_collect(self())
+        {res, [decrement_cache(cache, id, count, res) | caches]}
```
jonatanklosko, line 189: Why don't we garbage collect after decrementing (possibly pruning) the cache?

Also, why not garbage collect in the branch below?

nx/lib/nx/defn/evaluator.ex
josevalim: It is a (bad?) habit I picked: `[:fun | id]` uses one word fewer compared to tuple, but honestly unlikely to matter here. :)

nx/lib/nx/defn/evaluator.ex
josevalim: The value that is being pruned from the cache is `res`, which we are returning from this function, which means it is still alive. In another word, we purge from the catch the last time a value will be used, but it still has to be used.

jonatanklosko: Ah, makes sense!

nx/lib/nx/defn/evaluator.ex
jonatanklosko: Good to know, thanks :) For a moment it made me wonder if `id` isn't a list, it probably sacrifices some readability.

nx/test/nx/defn/evaluator_test.exs
```diff
@@ -478,4 +478,144 @@ defmodule Nx.Defn.EvaluatorTest do
...
+
+    test "with nested map" do
+      assert cond_cache_map(%{iteration: 0}) ==
+               %{iteration: Nx.tensor(1)}
+
+      assert cond_cache_map(%{iteration: 0, abc: 1}) ==
+               %{iteration: Nx.tensor(1), abc: Nx.tensor(1)}
+
+      assert cond_cache_map(%{iteration: 0, xyz: 1}) ==
+               %{iteration: Nx.tensor(1), xyz: Nx.tensor(1)}
```
jonatanklosko, line 584: Aren't these the same?

```suggestion
      assert cond_cache_map(%{iteration: 0, abc: 1}) ==
               %{iteration: Nx.tensor(1), abc: Nx.tensor(1)}
```

nx/test/nx/defn/evaluator_test.exs
```diff
@@ -478,4 +478,144 @@ defmodule Nx.Defn.EvaluatorTest do
...
+
+      assert cond_cache_map(%{iteration: 4}) ==
+               %{iteration: Nx.tensor(5)}
+
+      assert cond_cache_map(%{iteration: 4, abc: 1}) ==
+               %{iteration: Nx.tensor(5), abc: Nx.tensor(1)}
+
+      assert cond_cache_map(%{iteration: 4, xyz: 1}) ==
+               %{iteration: Nx.tensor(5), xyz: Nx.tensor(1)}
```
jonatanklosko, line 593:
```suggestion
      assert cond_cache_map(%{iteration: 4, abc: 1}) ==
               %{iteration: Nx.tensor(5), abc: Nx.tensor(1)}
```

nx/test/nx/defn/evaluator_test.exs
```diff
@@ -478,4 +478,144 @@ defmodule Nx.Defn.EvaluatorTest do
...
+      {state, prev} =
+        if state.iteration < 12 do
+          prev_timestep = if(state.iteration < 2, do: 1, else: -1)
+
+          factor =
+            if prev_timestep >= 0 do
+              2
+            else
+              1
+            end
```
jonatanklosko, line 606:
```suggestion
          sign = if(state.iteration < 2, do: 1, else: -1)

          factor =
            if sign >= 0 do
              2
            else
              1
            end
```

nx/test/nx/defn/evaluator_test.exs
josevalim: No! they changed the map ordering which lead to different errors :D I will drop a comment.

nx/test/nx/defn/evaluator_test.exs
```diff
@@ -478,4 +478,144 @@ defmodule Nx.Defn.EvaluatorTest do
...
+        end
+
+      %{state | iteration: state.iteration + 1}
+    end
+
+    test "with nested map" do
+      assert cond_cache_map(%{iteration: 0}) ==
+               %{iteration: Nx.tensor(1)}
+
+      assert cond_cache_map(%{iteration: 0, abc: 1}) ==
```
josevalim, line 580:
```suggestion
      # Use :abc/:xyz so we try different key orderings.
      assert cond_cache_map(%{iteration: 0, abc: 1}) ==
```

nx/test/nx/defn/evaluator_test.exs
```diff
@@ -478,4 +478,144 @@ defmodule Nx.Defn.EvaluatorTest do
...
+               %{iteration: Nx.tensor(1), abc: Nx.tensor(1)}
+
+      assert cond_cache_map(%{iteration: 0, xyz: 1}) ==
+               %{iteration: Nx.tensor(1), xyz: Nx.tensor(1)}
+
+      assert cond_cache_map(%{iteration: 4}) ==
+               %{iteration: Nx.tensor(5)}
+
+      assert cond_cache_map(%{iteration: 4, abc: 1}) ==
```
josevalim, line 589:
```suggestion
      # Use :abc/:xyz so we try different key orderings.
      assert cond_cache_map(%{iteration: 4, abc: 1}) ==
```

nx/test/nx/defn/evaluator_test.exs
jonatanklosko: Oh wow :smile:

jonatanklosko (approved):

jonatanklosko:
Beautiful!! :shipit:

josevalim:

PR #947, opened 2022-10-04
Add weighted_mean

### LINKED ISSUE #937: Add weighted_mean function
I think it will be helpful if we add an option `:sample_weights`. It will default to `nil` and then it will calculate the arithmetic mean. Otherwise, it will calculate the weighted mean. We can add the analogical functionality to `Nx.mean`.

msluszniak:
Closes #937

I implemented a version which accepts only one axis to aggregate. That's because it is unclear what this function would do for more than one dimension.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -6616,12 +6616,11 @@ defmodule Nx do
   Returns the weighted mean for the tensor and the weights.
 
   If the `:axis` option is given, it aggregates over
-  that dimension, effectively removing it. `axes: [0]`
+  that dimension, effectively removing it. `axis: 0`
   implies aggregating over the highest order dimension
-  and so forth. If the axis is negative, then counts
-  the axis from the back. For example, `axis: [-1]` will
-  always aggregate all rows. You can only aggregate over
-  one axis.
+  and so forth. If the axis is negative, then the axis will
+  be counted from the back. For example, `axis: -1` will
+  always aggregate over the last dimension.
 
   You may optionally set `:keep_axis` to true, which will
   retain the rank of the input tensor by setting the averaged
@@ -6711,24 +6710,24 @@ defmodule Nx do
       if shape != weights_shape do
         cond do
           axis == nil ->
-            raise ArgumentError, "Axis must be specified when shapes of input and weights differ."
+            raise ArgumentError, "axis must be specified when shapes of input and weights differ."
 
           tuple_size(weights_shape) != 1 ->
-            raise ArgumentError, "1D weights expected when shapes of a and weights differ."
+            raise ArgumentError, "rank-1 weights tensor expected when input shapes differ."
 
           elem(weights_shape, 0) != elem(shape, axis) ->
-            raise ArgumentError, "Length of weights not compatible with specified axis."
+            raise ArgumentError, "length of weights not compatible with specified axis."
 
           true ->
             nil
         end
 
         dims_to_broadcast =
-          for(_ <- 0..(tuple_size(shape) - 2), do: 1) ++ Tuple.to_list(weights_shape)
+          List.duplicate(1, tuple_size(shape) - 1) ++ Tuple.to_list(weights_shape)
 
         dims_to_broadcast = List.to_tuple(dims_to_broadcast)
         weights = broadcast(weights, dims_to_broadcast)
-        dims_to_swap = for i <- 0..(tuple_size(shape) + tuple_size(weights_shape) - 2), do: i
+        dims_to_swap = Enum.to_list(0..(tuple_size(shape) + tuple_size(weights_shape) - 2))
 
         dims_to_swap = swap(dims_to_swap, axis, tuple_size(shape) + tuple_size(weights_shape) - 2)
 
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -6718,13 +6718,13 @@ defmodule Nx do
       if shape != weights_shape do
         cond do
           axes == nil ->
-            raise ArgumentError, "axis must be specified when shapes of input and weights differ."
+            raise ArgumentError, "axis must be specified when shapes of input and weights differ"
 
           tuple_size(weights_shape) != 1 ->
-            raise ArgumentError, "rank-1 weights tensor expected when input shapes differ."
+            raise ArgumentError, "rank-1 weights tensor expected when input shapes differ"
 
           elem(weights_shape, 0) != elem(shape, List.first(axes)) ->
-            raise ArgumentError, "length of weights not compatible with specified axis."
+            raise ArgumentError, "length of weights not compatible with specified axis"
 
           true ->
             nil
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -6655,42 +6655,50 @@ defmodule Nx do
       >
 
       iex> t = Nx.tensor([[[1, 2, 3], [4, 5, 6]], [[7, 8, 9], [10, 11, 12]]], names: [:x, :y, :z])
-      iex> weights = Nx.tensor([[[8, 2, 4], [0, 3, 5]], [[4, 2, 6], [3, 1, 5]]])
+      iex> weights = Nx.tensor([[[0, 1, 2], [1, 1, 0]], [[-1, 1, -1], [1, 1, -1]]])
       iex> Nx.weighted_mean(t, weights, axis: :x)
       #Nx.Tensor<
         f32[y: 2][z: 3]
         [
-          [3.0, 5.0, 6.599999904632568],
-          [10.0, 6.5, 9.0]
+          [7.0, 5.0, -3.0],
+          [7.0, 8.0, 12.0]
         ]
       >
 
       iex> t = Nx.tensor([[[1, 2, 3], [4, 5, 6]], [[7, 8, 9], [10, 11, 12]]], names: [:x, :y, :z])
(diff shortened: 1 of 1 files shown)


polvalente:
https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.average.html
Jax (as well as NumPy) seem to only accept a single axis, so no worries about that!

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
   defp mean_den(shape, [axis | axes]) when axis >= 0,
     do: elem(shape, axis) * mean_den(shape, axes)
 
+  @doc """
+  Returns the weighted mean for the tensor and the weights.
+
+  If the `:axis` option is given, it aggregates over
+  that dimension, effectively removing it. `axes: [0]`
```
polvalente, line 6619:
```suggestion
  that dimension, effectively removing it. `axis: 0`
```

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+  implies aggregating over the highest order dimension
+  and so forth. If the axis is negative, then counts
+  the axis from the back. For example, `axis: [-1]` will
```
polvalente, line 6622:
```suggestion
  and so forth. If the axis is negative, then the axis will
  be counted from the back. For example, `axis: -1` will
```

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+  always aggregate all rows. You can only aggregate over
+  one axis.
```
polvalente, line 6624:
```suggestion
  always aggregate over the last dimension.
```

We shouldn't use rows for N-dim tensors. Also, "axis" being singular is our standard for "single-axis aggregation"

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+
+  You may optionally set `:keep_axis` to true, which will
```
polvalente, line 6626: I'm not sure if we should use keep_axis or keep_axes here. Both would make sense in that keep_axis -> "keep the aggregated axis" and keep_axes -> "keep all axes present in the original tensor"

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+        ]
+      >
+
+      iex> t = Nx.tensor([[[1, 2, 3], [4, 5, 6]], [[7, 8, 9], [10, 11, 12]]], names: [:x, :y, :z])
+      iex> weights = Nx.tensor([[[8, 2, 4], [0, 3, 5]], [[4, 2, 6], [3, 1, 5]]])
+      iex> Nx.weighted_mean(t, weights, axis: -1)
+      #Nx.Tensor<
+        f32[x: 2][y: 2]
+        [
+          [1.7142857313156128, 5.625],
```
polvalente, line 6675: I think we should have cleaner numbers in the example.
I suggest having a weights tensor with lots of 1s, -1s and 0s so that it's easier to calculate everything

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+        else: opts[:axis]
+
+    weights =
+      if shape != weights_shape do
+        cond do
+          axis == nil ->
+            raise ArgumentError, "Axis must be specified when shapes of input and weights differ."
+
+          tuple_size(weights_shape) != 1 ->
+            raise ArgumentError, "1D weights expected when shapes of a and weights differ."
```
polvalente, line 6717:
```suggestion
            raise ArgumentError, "1D weights expected when input shapes differ."
```

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+    axis =
+      if opts[:axis] != nil,
+        do: Nx.Shape.normalize_axes(shape, [opts[:axis]], names),
+        else: opts[:axis]
+
+    weights =
+      if shape != weights_shape do
+        cond do
+          axis == nil ->
+            raise ArgumentError, "Axis must be specified when shapes of input and weights differ."
```
polvalente, line 6714:
```suggestion
            raise ArgumentError, "axis must be specified when shapes of input and weights differ."
```

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+
+          tuple_size(weights_shape) != 1 ->
+            raise ArgumentError, "1D weights expected when shapes of a and weights differ."
+
+          elem(weights_shape, 0) != elem(shape, axis) ->
+            raise ArgumentError, "Length of weights not compatible with specified axis."
```
polvalente, line 6720:
```suggestion
            raise ArgumentError, "length of weights not compatible with specified axis."
```

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+
+          true ->
+            nil
+        end
+
+        dims_to_broadcast =
+          for(_ <- 0..(tuple_size(shape) - 2), do: 1) ++ Tuple.to_list(weights_shape)
```
polvalente, line 6727:
```suggestion
          List.duplicate(1, tuple_size(shape) - 1) ++ Tuple.to_list(weights_shape)
```

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+
+        dims_to_broadcast = List.to_tuple(dims_to_broadcast)
+        weights = broadcast(weights, dims_to_broadcast)
+        dims_to_swap = for i <- 0..(tuple_size(shape) + tuple_size(weights_shape) - 2), do: i
```
polvalente, line 6731:
```suggestion
        dims_to_swap = Enum.to_list(0..(tuple_size(shape) + tuple_size(weights_shape) - 2))
```

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+
+        dims_to_swap = swap(dims_to_swap, axis, tuple_size(shape) + tuple_size(weights_shape) - 2)
+
+        transpose(weights, axes: dims_to_swap)
```
polvalente, line 6735: why is this transpose needed?

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+      else
+        weights
+      end
+
+    weights_sum = sum(weights, axes: axis, keep_axes: opts[:keep_axis])
+
+    if to_number(any(equal(weights_sum, 0.0))) == 1 do
+      raise ArgumentError, "Weights sum to zero, can't be normalized"
+    end
```
polvalente, line 6744: I would just let it crash in the operation. This check will take the operation from the device instead of running everything in-device

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+
+    divide(
+      sum(multiply(tensor, weights), axes: axis, keep_axes: opts[:keep_axis]),
+      weights_sum
+    )
```
polvalente, line 6749: Please turn this into a pipeline

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,152 @@ defmodule Nx do
...
+  @doc type: :aggregation, from_backend: false
+  def weighted_mean(tensor, weights, opts \\ []) do
+    opts = keyword!(opts, axis: nil, keep_axis: false)
+    %T{shape: shape, names: names} = tensor = to_tensor(tensor)
+    %T{shape: weights_shape} = weights = to_tensor(weights)
+
+    axis =
+      if opts[:axis] != nil,
+        do: Nx.Shape.normalize_axes(shape, [opts[:axis]], names),
+        else: opts[:axis]
```
polvalente, line 6708: either call the variable `axes` or extract it into a number | nil again. I found it confusing that the singular axis refers to a list here

nx/lib/nx.ex
msluszniak: I agree. What do you think @josevalim?

josevalim: We have both `keep_axis` and `keep_axes` in Nx, so I think it should be `keep_axis` when the other option is `axis` (and `keep_axes` for `axes`).

nx/lib/nx.ex
```diff
@@ -6612,6 +6612,141 @@ defmodule Nx do
...
+      if shape != weights_shape do
+        cond do
+          axes == nil ->
+            raise ArgumentError, "axis must be specified when shapes of input and weights differ."
+
+          tuple_size(weights_shape) != 1 ->
+            raise ArgumentError, "rank-1 weights tensor expected when input shapes differ."
+
+          elem(weights_shape, 0) != elem(shape, List.first(axes)) ->
+            raise ArgumentError, "length of weights not compatible with specified axis."
```
josevalim, line 6727:
```suggestion
            raise ArgumentError, "axis must be specified when shapes of input and weights differ"

          tuple_size(weights_shape) != 1 ->
            raise ArgumentError, "rank-1 weights tensor expected when input shapes differ"

          elem(weights_shape, 0) != elem(shape, List.first(axes)) ->
            raise ArgumentError, "length of weights not compatible with specified axis"
```

josevalim:

PR #954, opened 2022-10-12
Add Median

### LINKED ISSUE #953: Add median function
I'll implement this functionality.

msluszniak:
Closes #953


nx/lib/nx.ex
```diff
@@ -6747,6 +6747,131 @@ defmodule Nx do
...
+        s64[2][2]
+        [
+          [2, 3],
+          [4, 7]
+        ]
+      >
+
+  ### Keeping axis
+
+      iex> t = Nx.tensor([[[1, 2, 2], [3, 4, 2]], [[4, 5, 2], [7, 9, 2]]])
```
polvalente, line 6821: I would be nice to add a test (not a doctesr, though) where the input tensor is not sorted

nx/lib/nx.ex
```diff
@@ -6747,6 +6747,131 @@ defmodule Nx do
...
+    axis =
+      if opts[:axis] != nil,
+        do: Nx.Shape.normalize_axis(shape, opts[:axis], names),
+        else: opts[:axis]
+
+    if axis != nil do
+      axis_size = axis_size(tensor, axis)
+      tensor = sort(tensor, axis: axis)
+
+      if rem(axis_size, 2) == 1 do
```
polvalente, line 6851: Instead of nested ifs, perhaps cond would be more readable?

polvalente (approved):

nx/lib/nx.ex
msluszniak: I think that nested ifs are good because if we use cond, then we need to sort a tensor in every branch.

ghostkid130: Would concur with @polvalente lots of nesting, perhaps we can leverage pattern matching. I took a stab at refactoring _(untested)_ let me know what you think when you get the chance. I hope is useful 🙏 

```
  def median(tensor, opts \\ []) do
    opts = keyword!(opts, axis: nil, keep_axis: false)
    tensor = to_tensor(tensor)
    process_median_axis(opts[:axis], tensor, opts)
  end

  defp process_median_axis(axis, %T{names: names, shape: shape}, opts) when not is_nil(axis) do
    axis = Nx.Shape.normalize_axis(shape, opts[:axis], names)
    axis_size = axis_size(tensor, axis)
    tensor = sort(tensor, axis: axis)
    half_idx = div(axis_size, 2)

    process_remainer = fn
      1 ->
        res = slice_along_axis(tensor, half_idx, 1, axis: axis)
        if opts[:keep_axis], do: res, else: squeeze(res, axes: [axis])
      _ ->
        two_elems = slice_along_axis(tensor, half_idx - 1, 2, axis: axis)
        mean(two_elems, axes: [axis], keep_axes: opts[:keep_axis])
      end

    rem(axis_size, 2)
    |> (&process_remainer.(&1)).()
  end


  defp process_median_axis(nil, %T{shape: shape}, _opts) do
    tensor =
      flatten(tensor)
      |> sort()

    {tensor_size} = shape(tensor)
    half_idx = div(tensor_size, 2)

    process_tensor_remainder = fn
      1 ->
        tensor[[half_idx]]
      _ ->
        tensor[[half_idx - 1]]
        |> add(tensor[[half_idx]])
        |> divide(2)
    end

    rem(tensor_size, 2)
    |> (&process_tensor_remainder.(&1)).()
  end

```

polvalente: I think cond or case can do the job fine here without having to split into multiple clauses. I'll merge as-is and refactor in a separate commit :)

polvalente: https://github.com/elixir-nx/nx/commit/fe3210479810c6847eea5e109a1298b5cd22ff85

PR #955, opened 2022-10-13
feat(exla): convert :optional to xla Call

polvalente:
This is part of our effort to implement compilation boundaries to allow for exchanging of memory and CPU time


seanmor5:
Nice! Do we have any benchmarks that can compare between using `xla_call` vs. not?

exla/lib/exla/defn.ex
```diff
@@ -1449,6 +1473,31 @@ defmodule EXLA.Defn do
     {EXLA.Builder.build(res), update_outfeed(cache, comp_cache)}
   end
 
+  defp optional_computation(name, arg, expr, state, cache) do
```
josevalim, line 1476: This seems to be the same as `while_computation`. Perhaps we can call it "maybe_token_computation" and reuse it on both?

exla/lib/exla/defn.ex
```diff
@@ -541,15 +541,39 @@ defmodule EXLA.Defn do
...
+        List.to_tuple(computation_args),
+        default_body,
+        state,
+        cache
+      )
+
+    builder =
+      case call_args do
+        [%{builder: builder} | _] ->
+          builder
```
josevalim, line 568: Why do we need this? It should be always state.builder, no?

exla/lib/exla/defn.ex
```diff
@@ -541,15 +541,39 @@ defmodule EXLA.Defn do
...
+
+        _ ->
+          state.builder
+      end
+
+    call = EXLA.Op.call(builder, token, call_args, call_body)
```
josevalim, line 574: We should pass the token as a regular argument (iirc the first argument) in `call_args`, no?

josevalim: I.e. EXLA.Op.call(builder, computation, args)

polvalente:
> Nice! Do we have any benchmarks that can compare between using `xla_call` vs. not?

@seanmor5 not yet! Any suggestions on how to check that?

josevalim:
@seanmor5 my understanding is that this does not affect performance. It is mostly a way to encapsulate logic to avoid multiple repetitions (the cache is coming later today!).

josevalim (approved):

josevalim:

PR #957, opened 2022-10-14
Add @reuse true annotation

josevalim:
* [ ] TODO: test grad
* [ ] TODO: test tuple returns in boundaries
* [ ] TODO: implement inspect
* [ ] TODO: change optional to have similar signature
* [ ] TODO: fix calling a boundary from within `defn` with a number
* [ ] TODO: test boundary with multiple arguments
* [ ] TODO: rename @reuse to @boundary

### DIFF
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -253,7 +253,8 @@ defmodule Nx.Defn.Expr do
 
   @doc false
   def boundary(type, key, fun, args) do
-    {params, context} = Enum.map_reduce(args, nil, &to_param_expr(&1, :root, &2))
+    param_context = new_context(:boundary)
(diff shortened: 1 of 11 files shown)


nx/lib/nx/defn/expr.ex
```diff
@@ -251,6 +251,15 @@ defmodule Nx.Defn.Expr do
     attach_token(token, expr)
   end
 
+  @doc false
+  def boundary(type, key, fun, args) do
+    {params, context} = Enum.map_reduce(args, nil, &to_param_expr(&1, :root, &2))
```
polvalente, line 256: I was gonna ask if you should be using `:root` here, but I think that since we're always at the start of a given defn here, it's always `:root`, right?

polvalente (commented):
This is looking great so far!

PR #958, opened 2022-10-16
Reduce Nx.Random.split tree

josevalim:
We know the counts are even,
so we can skip padding and reshaping.

### DIFF
--- a/nx/lib/nx/random.ex
+++ b/nx/lib/nx/random.ex
@@ -104,7 +104,7 @@ defmodule Nx.Random do
 
     Nx.iota({2, num}, type: :u32)
     |> threefry2x32_20(key)
-    |> Nx.reshape({:auto, 2})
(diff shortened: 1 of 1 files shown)


nx/lib/nx/random.ex
```diff
@@ -102,24 +102,28 @@ defmodule Nx.Random do
   defn split(key, num \\ 2) do
     assert_key!(key)
 
-    key
-    |> threefry2x32(Nx.iota({num, 2}))
-    |> Nx.as_type({:u, 32})
+    Nx.iota({2, num}, type: :u32)
+    |> threefry2x32_20(key)
+    |> Nx.reshape({:auto, 2})
```
josevalim, line 107:
```suggestion
    |> Nx.reshape({num, 2})
```

josevalim:

PR #959, opened 2022-10-16
Add fold-in

seanmor5:


nx/lib/nx/random.ex
```diff
@@ -105,6 +105,25 @@ defmodule Nx.Random do
     threefry2x32(key, {num, 2})
   end
 
+  @doc """
+  Folds in new data to a PRNG key.
+  """
+  defn fold_in(key, data) do
```
polvalente, line 111: One test or two would be nice. Perhaps a doctest?

polvalente (approved):

PR #961, opened 2022-10-18
Add Mode

msluszniak:


josevalim (approved):

nx/lib/nx.ex
```diff
@@ -6881,6 +6881,216 @@ defmodule Nx do
     end
   end
 
+  @doc """
+  Returns the mode for the tensor and the weights.
```
josevalim, line 6885:
```suggestion
  Returns the mode (the values that appear most often) of a tensor.
```

nx/lib/nx.ex
```diff
@@ -6881,6 +6881,216 @@ defmodule Nx do
...
+          |> reshape({num_elements, 1})
+      end)
+      |> concatenate(axis: 1)
+
+    {n, _} = shape(gen_indices)
+
+    largest_group_indices =
+      broadcast(0, sorted)
+      |> indexed_add(gen_indices, broadcast(1, {n}))
+      |> argmax(axis: opts[:axis], keep_axis: true)
```
msluszniak, line 7062:
```suggestion
      |> argmax(axis: axis, keep_axis: true)
```

nx/lib/nx.ex
```diff
@@ -6881,6 +6881,216 @@ defmodule Nx do
...
+          |> iota(axis: axis)
+          |> reshape({num_elements, 1})
+      end)
+      |> concatenate(axis: 1)
+
+    {n, _} = shape(gen_indices)
+
+    largest_group_indices =
+      broadcast(0, sorted)
+      |> indexed_add(gen_indices, broadcast(1, {n}))
```
jonatanklosko, line 7061:
```suggestion
    largest_group_indices =
      broadcast(0, sorted)
      |> indexed_add(gen_indices, broadcast(1, {num_elements}))
```

nx/lib/nx.ex
```diff
@@ -6881,6 +6881,216 @@ defmodule Nx do
...
+            slice_along_axis(sorted, 1, axis_size(sorted, axis) - 1, axis: axis)
+          )
+        ],
+        axis: axis
+      )
+      |> cumulative_sum(axis: axis)
+
+    num_elements = tensor_shape |> Tuple.product()
+
+    gen_indices =
```
jonatanklosko, line 7044: Let's rename to `counting_indices` or something, otherwise it'c easy to confuse for `group_intices`.

nx/lib/nx.ex
```diff
@@ -6881,6 +6881,216 @@ defmodule Nx do
...
+    tensor_size = size(tensor)
+
+    cond do
+      tensor_rank == 0 ->
+        if opts[:keep_axis], do: new_axis(tensor, -1), else: tensor
+
+      tensor_size == 1 and axis == nil ->
+        tensor = flatten(tensor)
+        res = take(tensor, tensor([0]))
+        if opts[:keep_axis], do: res, else: squeeze(res)
```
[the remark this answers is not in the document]
msluszniak: This will be not consistent with sklearn mode. for input `Nx.tensor([[[[[[1]]]]]])` and `:keep_axis` set to `true` in sklearn it returns `Nx.tensor([1])`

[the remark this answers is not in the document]
jonatanklosko: We should be consistent with other Nx functions, not sklearn, and that's the behaviour we currently have, for example:

```elixir
Nx.iota({2, 2, 2, 2}) |> Nx.sum(keep_axes: true)
#Nx.Tensor<
  s64[1][1][1][1]
  [
    [
      [
        [120]
      ]
    ]
  ]
>
```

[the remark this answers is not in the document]
jonatanklosko: Actually `Nx.sum` is different because we implicitly it's `axes: all_axes`. For a single axis, I'm not sure if `keep_axis` makes much sense in the first place. Another example:

```elixir
Nx.iota({2, 2, 2, 2}) |> Nx.argmax(keep_axis: true)
#Nx.Tensor<
  s64
  15
>
```

cc @josevalim @polvalente

nx/lib/nx.ex
```diff
@@ -6881,6 +6881,212 @@ defmodule Nx do
...
+
+      tensor_size == 1 and axis == nil ->
+        if opts[:keep_axis], do: tensor, else: squeeze(tensor)
+
+      tensor_size == 1 and axis != nil ->
+        if opts[:keep_axis], do: tensor, else: squeeze(tensor, axes: [axis])
+
+      axis == nil ->
+        tensor = flatten(tensor)
+        mode_1d(tensor, keep_axis: opts[:keep_axis])
```
jonatanklosko, line 7012: If we flatten, with `keep_axis: true` we will return 1D, instead of the tensor dimensionality. We should optionally reshape into `Tuple.repeat(1, tensor_rank)`.

Also, for 1D we could use `mode_general(tensor, axis: 0)`, no? The implementation is pretty much the same other than general being vectorized, so there's likely no much overhead, and we avoid duplication, wdyt?

jonatanklosko (approved):
One last comment from me :)

josevalim:

PR #969, opened 2022-10-26
Fix Nx.{any,all} with :keep_axes on EXLA

jonatanklosko:
Currently `:keep_axes` fails, because we always try to reshape to a scalar.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -6192,6 +6192,18 @@ defmodule Nx do
         u8[x: 2]
         [0, 1]
       >
+
+  ### Keeping axes
+
+      iex> Nx.all(Nx.tensor([[-1, 0, 1], [2, 3, 4]], names: [:x, :y]), axes: [:y], keep_axes: true)
+      #Nx.Tensor<
+        u8[x: 2][y: 1]
+        [
+          [0],
+          [1]
+        ]
+      >
+
   """
   @doc type: :aggregation
   def all(tensor, opts \\ []) do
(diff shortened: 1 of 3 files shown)


polvalente (approved):

nx/lib/nx.ex
```diff
@@ -6232,6 +6244,18 @@ defmodule Nx do
...
+  ### Keeping axes
+
+      iex> Nx.any(Nx.tensor([[0, 1, 0], [0, 1, 2]], names: [:x, :y]), axes: [:y], keep_axes: true)
+      #Nx.Tensor<
+        u8[x: 2][y: 1]
+        [
+          [1],
+          [1]
+        ]
+      >
```
josevalim, line 6257: We should not use doctests for regression testing. In this case, a doctest is also helpful, but we should also have added a regression test to EXLA. :)

jonatanklosko: I would've added a regular test, but in this case I though the issues was missing doctest, since we test keep axes in other functions. I can also add an EXLA test, but it would be the same?

josevalim: I will push soon :)

I think the point is that we rely on doc tests for convenience, but ideally we would have a test suite that is completely decoupled from docs and reusable across implementations. :)

Especially because in the future EXLA will be in a separate repo and doing joint commits will be harder.

jonatanklosko: Gotcha, sounds good :)

Issue #974, opened 2022-10-29
Telemetry events for XLA compilations


laursisask:
My organization has encountered a problem where there are sometimes spikes in CPU usage and latency with Nx & EXLA. We suspect that one reason this is happening is that some JIT-compiled functions are compiled over and over again.

Currently, EXLA has debug mode which logs out information about XLA compilations, e.g.

```
18:25:45.945 [debug] domain=elixir file=lib/exla/defn.ex line=375  EXLA defn evaluation in 0.7ms
18:25:46.024 [debug] domain=elixir file=lib/exla/defn.ex line=412  EXLA compilation in 79.0ms
18:25:46.024 [debug] domain=elixir file=lib/exla/defn.ex line=271  EXLA device 0 lock in 0.0ms
```

It would be helpful if in addition to these logs there were telemetry events to which applications could listen.

The events could contain measurements for defn evaluation time, compilation time, and the time waiting for EXLA lock. In metadata, could include the reference to function being compiled and the argument shapes.

Then, event handlers could send metrics about how many compilations are happening to observability platforms like Datadog.

If you agree that adding such functionality to exla would fit in the library, then I can open a pull request, and we can discuss lower-level details like event names in there.

josevalim:
Yes, a PR is definitely welcome!

seanmor5:
+1, this would be great!

PR #975, opened 2022-10-30
feat: send telemetry event after XLA compilation

laursisask:
This commit adds telemetry events that are executed when EXLA compiles a function into an executable.

These events can be used to have observability into how many times functions are compiled and how long compiling them takes.

When a function is compiled, an event with name `[:exla, :compilation]`
is executed. The measurements include the following:
- `eval_time` - time it took to call the defn function that is compiled
- `compile_time` - time it took for XLA to compile the instructions into
an executable
- `total_time` - sum of `eval_time` and `compile_time`

This commit is related to the issue
https://github.com/elixir-nx/nx/issues/974

### DIFF
--- a/exla/lib/exla.ex
+++ b/exla/lib/exla.ex
@@ -162,6 +162,18 @@ defmodule EXLA do
   Alternatively, you can pass the `--init` flag to `docker run`, so
   it runs an `init` inside the container that forwards signals and
   reaps processes.
+
+  ## Telemetry events
+
+  EXLA executes a telemetry event every time a function is JIT-compiled.
+  The events are named `[:exla, :compilation]` and include the following
+  measurements, given in microseconds:
+
+    * `:eval_time` - the time spent on turning the function into XLA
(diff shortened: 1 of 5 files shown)


josevalim:
Thank you @laursisask!

I would prefer if we did not change the compilers API. For an initial version, I would prefer if we don't pass metadata and, if any is needed in the future, we can pass metadata as part of `opts`. WDYT?

laursisask:
> Thank you @laursisask!
> 
> I would prefer if we did not change the compilers API. For an initial version, I would prefer if we don't pass metadata and, if any is needed in the future, we can pass metadata as part of `opts`. WDYT?

Sure, we can do that. I made the appropriate changes to the PR.

laursisask:
I also added documentation and a field `total_time` to the measurements.

laursisask:
@josevalim could you take a look at this pull request again?

josevalim (approved):

PR #980, opened 2022-11-06
No loss of precision in `Nx.complex/2`

### LINKED ISSUE #979: Potential loss of precision in Nx.complex/2
When passing two tensors with 64 bit data types to `Nx.complex/2` the result should have type `:c128`. Currently, however, the result is always `:c64` meaning the real and imaginary parts have lost half their bits. 

Generally speaking, I would expect this to succeed:

```elixir
t = Nx.tensor(Complex.new(1, 2), type: :c128)
re = Nx.real(t)
im = Nx.imag(t)
^t = Nx.complex(re, im)
```
But it fails with

```
** (MatchError) no match of right hand side value: #Nx.Tensor<
  c64
  1.0+2.0i
>
```

wfvining:
Ensures that the returned tensor has the same precision for its real and imaginary parts as the lowest precision input by checking precision of inputs, and passing the correct type (`:c64` or `:c128`) to `Nx.Constants.i/1`.

Closes #979

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -5994,7 +5994,7 @@ defmodule Nx do
       Nx.Shared.raise_complex_not_supported("complex", 2)
     end
 
-    t = type(real) |> Nx.Type.merge(type(imag)) |> Nx.Type.to_complex
+    t = type(real) |> Nx.Type.merge(type(imag)) |> Nx.Type.to_complex()
 
     imag
     |> multiply(Nx.Constants.i(type: t))
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -5976,7 +5976,19 @@ defmodule Nx do
 
   ### Examples
 
-      iex> Nx.complex(Nx.tensor(1), Nx.tensor(2))
+      iex> Nx.complex(Nx.tensor(1, type: :f32), Nx.tensor(2, type: :f32))
+      #Nx.Tensor<
+        c64
+        1.0+2.0i
+      >
+
+      iex> Nx.complex(Nx.tensor(1, type: :f64), Nx.tensor(2, type: :f64))
+      #Nx.Tensor<
+        c128
+        1.0+2.0i
+      >
+
+      iex> Nx.complex(Nx.tensor(1, type: :f32), Nx.tensor(2, type: :f64))
       #Nx.Tensor<
         c64
         1.0+2.0i
@@ -5984,7 +5996,7 @@ defmodule Nx do
 
       iex> Nx.complex(Nx.tensor([1, 2]), Nx.tensor([3, 4]))
       #Nx.Tensor<
-        c64[2]
+        c128[2]
         [1.0+3.0i, 2.0+4.0i]
       >
   """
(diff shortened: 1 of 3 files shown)


nx/lib/nx.ex
```diff
@@ -5976,15 +5976,27 @@ defmodule Nx do
 
   ### Examples
 
-      iex> Nx.complex(Nx.tensor(1), Nx.tensor(2))
+      iex> Nx.complex(Nx.tensor(1, type: :f32), Nx.tensor(2, type: :f32))
+      #Nx.Tensor<
+        c64
+        1.0+2.0i
+      >
```
polvalente, line 5983: I wouldn't change this example because it makes for less clarity on the docs. A separate test for this is welcome, though

nx/lib/nx.ex
```diff
@@ -5976,15 +5976,27 @@ defmodule Nx do
...
+
+      iex> Nx.complex(Nx.tensor(1, type: :f64), Nx.tensor(2, type: :f64))
+      #Nx.Tensor<
+        c128
+        1.0+2.0i
+      >
+
+      iex> Nx.complex(Nx.tensor(1, type: :f32), Nx.tensor(2, type: :f64))
```
polvalente, line 5991: These ones should probably be tests instead of examples

nx/lib/nx.ex
```diff
@@ -5976,15 +5976,27 @@ defmodule Nx do
...
       #Nx.Tensor<
         c64
         1.0+2.0i
       >
 
       iex> Nx.complex(Nx.tensor([1, 2]), Nx.tensor([3, 4]))
       #Nx.Tensor<
-        c64[2]
+        c128[2]
```
polvalente, line 5999: Why has this changed?

nx/lib/nx.ex
```diff
@@ -5994,8 +6006,12 @@ defmodule Nx do
       Nx.Shared.raise_complex_not_supported("complex", 2)
     end
 
+    {_, realprecision} = type(real)
+    {_, imagprecision} = type(imag)
+    precision = Kernel.max(64, Kernel.min(realprecision, imagprecision) * 2)
```
polvalente, line 6011: You should use Nx.Type.merge_type in combination with Nx.Type.to_real to get the correctly promoted precision.

This is probably why s64 is getting cast to c128 instead of c64

wfvining: Thanks! I had not seen the `Nx.Type` module. Using the functions from that module makes this much better and fixes the issue with `:s64` that you noted.

nx/lib/nx.ex
```diff
@@ -5994,8 +5994,10 @@ defmodule Nx do
       Nx.Shared.raise_complex_not_supported("complex", 2)
     end
 
+    t = type(real) |> Nx.Type.merge(type(imag)) |> Nx.Type.to_complex
```
polvalente, line 5997:
```suggestion
    t = type(real) |> Nx.Type.merge(type(imag)) |> Nx.Type.to_complex()
```

polvalente (approved):
Thanks!

josevalim:

PR #985, opened 2022-11-15
fix(complex): Nx.as_type

### LINKED ISSUE #983: Casting between `:c64` and `:c128` in binary backend deletes imaginary part
Using as type to convert between `:c64` and `:c128` results in zeroing out the imaginary part of the input when using the binary backend.

```elixir
t = Nx.tensor(Complex.new(1, 2), type: :c64)
Nx.as_type(t, :c128)
```
Produces
```
#Nx.Tensor<
  c128
  1.0+0.0i
>
```

I haven't tested the other backends, but I suspect they are not affected because they do not implement `as_type` directly.

polvalente:
closes #983


polvalente:
I hadn't received the notification for #984! Closing this in favor of that :)

PR #989, opened 2022-11-21
Nx.Serving type == :process

josevalim:
Should be good for review. Only API docs missing and a Nx.Serving.Default implementation.

### DIFF
--- a/nx/lib/nx/serving.ex
+++ b/nx/lib/nx/serving.ex
@@ -353,7 +353,7 @@ defmodule Nx.Serving do
     {%Nx.Serving{process_options: serving_opts} = serving, opts} = Keyword.pop!(opts, :serving)
     {batch_size, opts} = process_option(:batch_size, serving_opts, opts, 1)
     {batch_timeout, opts} = process_option(:batch_timeout, serving_opts, opts, 100)
-    arg = {self(), name, serving, batch_size, batch_timeout}
+    arg = {name, serving, batch_size, batch_timeout}
 
     children = [
       %{
@@ -472,7 +472,8 @@ defmodule Nx.Serving do
   @timeout_message {__MODULE__, :timeout}
 
   @impl true
-  def init({parent, name, serving, batch_size, batch_timeout}) do
+  def init({name, serving, batch_size, batch_timeout}) do
+    [parent | _] = Process.get(:"$ancestors")
     {:ok, module_state} = handle_init(serving.module, :process, serving.arg)
 
     :persistent_term.put(
--- a/nx/lib/nx/serving.ex
+++ b/nx/lib/nx/serving.ex
@@ -110,7 +110,19 @@ defmodule Nx.Serving do
   TODO.
   """
 
+  @doc false
   defstruct [:module, :arg, :client_preprocessing, :client_postprocessing]
+
+  @type metadata() :: term()
+  @type client_preprocessing() :: (term() -> Nx.Batch.t())
+  @type client_postprocessing() :: (Nx.Container.t(), metadata() -> Nx.Batch.t())
+
+  @type t :: %__MODULE__{
+          module: atom(),
+          arg: term(),
+          client_preprocessing: client_preprocessing(),
+          client_postprocessing: client_postprocessing()
+        }
 
   @doc """
   The callback used to initialize the serving.
@@ -121,7 +133,7 @@ defmodule Nx.Serving do
 
   It must return `{:ok, state}`, where the `state` can be any term.
   """
-  @callback init(:inline | :process, arg :: term) :: {:ok, state :: term}
+  @callback init(:inline | :process, arg :: term()) :: {:ok, state :: term()}
 
   @doc """
   Receives a batch and returns a function to execute the batch.
@@ -130,7 +142,7 @@ defmodule Nx.Serving do
   separate process.
   """
   @callback handle_batch(Nx.Batch.t(), state) ::
-              {:execute, (() -> {Nx.Container.t(), term}), state}
+              {:execute, (-> {Nx.Container.t(), metadata()}), state}
             when state: term()
 
   @doc """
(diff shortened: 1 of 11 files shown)


nx/lib/nx/serving.ex
```diff
@@ -176,19 +192,306 @@ defmodule Nx.Serving do
...
+    case acc do
+      [slice] -> {slice, metadata}
+      _ -> {acc |> Enum.reverse() |> Nx.concatenate(axis: @axis), metadata}
     end
   end
 
+  defp receive_batched(total_size, ref, acc, _metadata, name, input) do
+    receive do
+      {^ref, {start, size, output, metadata}} ->
+        slice = Nx.slice_along_axis(output, start, size, axis: @axis)
```
jonatanklosko, line 276: Output may be a container, so we need to slice all the contents.

nx/lib/nx/serving.ex
```diff
@@ -82,25 +82,28 @@ defmodule Nx.Serving do
...
+      {#Nx.Tensor<
          s64[2][2]
          [
            [2, 4],
            [6, 8]
          ]
-       >, :metadata}
+       >,
+       :metadata,
+       :info}
```
jonatanklosko, line 93:
```suggestion
       :client_info}
```

josevalim: Good find, done!

nx/lib/nx/serving.ex
```diff
@@ -176,17 +252,329 @@ defmodule Nx.Serving do
...
+        else
+          funs =
+            output
+            |> Nx.Defn.Composite.reduce(
+              [],
+              &[Nx.slice_along_axis(&1, start, size, axis: @axis) | &2]
+            )
+            |> Enum.reverse()
+
+          receive_batched(total_size - size, ref, [funs | acc], {output, metadata}, name, input)
```
jonatanklosko, line 361: We use the first `receive_batched` clause only here, and it doesn't actually receive. We could check if `total_size - size` is 0 here and call either `receive_batched` or `finalize_batched`?

nx/lib/nx/serving.ex
```diff
@@ -176,17 +252,329 @@ defmodule Nx.Serving do
...
+    }
+
+    {:ok, state, {:continue, {:task_supervisor, parent}}}
+  end
+
+  @impl true
+  def handle_continue({:task_supervisor, parent}, state) do
+    {:ok, task_supervisor} = Supervisor.start_child(parent, Task.Supervisor)
+    {:noreply, %{state | task_supervisor: task_supervisor}}
+  end
```
jonatanklosko, line 424: Why don't we list the task supervisor in the supervisor children right away? To avoid naming it?

jonatanklosko (approved):
Awesome work!!

nx/lib/nx/serving.ex
josevalim: In a way, yes. We would need to generate names for it, which is fine as atoms, but I can see us supporting stuff like `{:via, ..., ...}` in the future, which would break this down.

nx/lib/nx/serving.ex
josevalim: I tried but it only added more nesting :( so I keep the current version for now. If you have ideas, feel free to PR. :)

jonatanklosko: I ended up with this, not sure if it's better, but the flow is probably easier to follow. Whatever you prefer!

```diff
diff --git a/nx/lib/nx/serving.ex b/nx/lib/nx/serving.ex
index 84f58e3..b87b362 100644
--- a/nx/lib/nx/serving.ex
+++ b/nx/lib/nx/serving.ex
@@ -315,50 +315,24 @@ defmodule Nx.Serving do
     ref = :erlang.monitor(:process, pid, alias: :demonitor)
     Process.send(pid, {__MODULE__, :batched_run, ref, batch}, [:noconnect])
 
-    {tensor, metadata} = receive_batched(batch.size, ref, [], nil, name, input)
+    {tensor, metadata} = receive_batched(batch.size, ref, [], name, input)
     handle_postprocessing(postprocessing, tensor, metadata, info)
   end
 
   @axis 0
 
-  defp receive_batched(0, ref, acc, {template, metadata}, _name, _input) do
-    Process.demonitor(ref, [:flush])
-
-    tensors =
-      acc
-      |> Enum.reverse()
-      |> Enum.zip_with(&Nx.concatenate(&1, axis: @axis))
-
-    {output, []} =
-      Nx.Defn.Composite.traverse(template, tensors, fn _template, [tensor | tensors] ->
-        {tensor, tensors}
-      end)
-
-    {output, metadata}
-  end
-
-  defp receive_batched(total_size, ref, acc, _template_metadata, name, input) do
+  defp receive_batched(total_size, ref, acc, name, input) do
     receive do
       {^ref, {start, size, output, metadata}} ->
-        # If we have a single response, slice and return immediately.
-        # Otherwise we collect their contents and build the concatenated result later.
-        if acc == [] and size == total_size do
-          Process.demonitor(ref, [:flush])
-
-          output =
-            Nx.Defn.Composite.traverse(output, &Nx.slice_along_axis(&1, start, size, axis: @axis))
-
-          {output, metadata}
-        else
-          funs =
-            output
-            |> Nx.Defn.Composite.reduce(
-              [],
-              &[Nx.slice_along_axis(&1, start, size, axis: @axis) | &2]
-            )
-            |> Enum.reverse()
-
-          receive_batched(total_size - size, ref, [funs | acc], {output, metadata}, name, input)
+        complete? = total_size == size
+
+        case handle_batched_response(start, size, output, complete?, acc) do
+          {:cont, acc} ->
+            receive_batched(total_size - size, ref, acc, name, input)
+
+          {:halt, tensor} ->
+            Process.demonitor(ref, [:flush])
+            {tensor, metadata}
         end
 
       {:DOWN, ^ref, _, _, reason} ->
@@ -368,6 +342,43 @@ defmodule Nx.Serving do
     end
   end
 
+  # If we have a single response, slice and return immediately.
+  defp handle_batched_response(start, size, output, _complete? = true, _acc = []) do
+    tensor =
+      Nx.Defn.Composite.traverse(output, &Nx.slice_along_axis(&1, start, size, axis: @axis))
+
+    {:halt, tensor}
+  end
+
+  # Otherwise we collect their contents and build the concatenated result later.
+  defp handle_batched_response(start, size, output, complete?, acc) do
+    chunks =
+      output
+      |> Nx.Defn.Composite.reduce(
+        [],
+        &[Nx.slice_along_axis(&1, start, size, axis: @axis) | &2]
+      )
+      |> Enum.reverse()
+
+    acc = [chunks | acc]
+
+    if complete? do
+      tensors =
+        acc
+        |> Enum.reverse()
+        |> Enum.zip_with(&Nx.concatenate(&1, axis: @axis))
+
+      {tensor, []} =
+        Nx.Defn.Composite.traverse(template, tensors, fn _template, [tensor | tensors] ->
+          {tensor, tensors}
+        end)
+
+      {:halt, tensor}
+    else
+      {:cont, acc}
+    end
+  end
+
   ## Process callbacks
 
   require Logger
```

jonatanklosko (approved):

josevalim:

PR #991, opened 2022-11-23
fix(EXLA): SVD returns V instead of transpose(V)

polvalente:


exla/test/exla/backend_test.exs
```diff
@@ -119,7 +119,7 @@ defmodule EXLA.BackendTest do
     # This is not really meant to work in practice,
     # but it does work with the Nx.BinaryBackend so
     # we make it work for EXLA too.
-    defn double(fun), do: double_transform(fun.())
+    defn(double(fun), do: double_transform(fun.()))
```
josevalim, line 122:
```suggestion
    defn double(fun), do: double_transform(fun.())
```

josevalim (approved):

PR #994, opened 2022-11-28
Psuedoinverse function in Nx 

anish-lakkapragada:
`defn` Psuedoinverse function in Nx just as `np.linalg.svd` in numpy. Uses the SVD to compute the psuedoinverse.

Works only for wide matrices with backends that have SVD implemented for wide matrices


.gitignore
```diff
@@ -0,0 +1 @@
+nx/test.ex
```
seanmor5, line 1: I think you have this leftover from debugging!

anish-lakkapragada: Sorry, I just realized that. Newbie PR here. I will create a new PR.

nx/lib/nx/lin_alg.ex
```diff
@@ -1142,6 +1142,49 @@ defmodule Nx.LinAlg do
...
+
+
+  defn pinv(tensor, opts \\ []) do
+
+    opts = keyword!(opts, eps: @default_eps)
+
+    {u, s, vt} = Nx.LinAlg.svd(tensor)
+
+    s_shape = {elem(Nx.shape(s), 0), elem(Nx.shape(u, 1)}
+    adjusted_s = Nx.map(s, fn x -> if Nx.abs(x) < opts[:eps] do 0 else Nx.divide(1, x) end end)
```
seanmor5, line 1182: `Nx.select(Nx.abs(x) < opts[:eps], 0, Nx.divide(1, x))`

.gitignore
seanmor5: There is no need, you can just update your fork and it will reflect here :)

PR #995, opened 2022-11-28
pseudoinverse pull request #2

anish-lakkapragada:
created a second psuedoinverse PR because the first one was too messy. 

defn psuedoinverse function in Nx.LinAlg just as np.linalg.svd in numpy. Uses the SVD to compute the psuedoinverse.

Works only for wide matrices with backends that have SVD implemented for wide matrices


nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,46 @@ defmodule Nx.LinAlg do
...
+
+      iex> pinv = Nx.LinAlg.pinv(Nx.tensor([[1, 1], [3, 4]]))
+      iex> pinv
+      #Nx.Tensor<
+        f32[2][2]
+        [
+          [4.0, -1.0000001192092896],
+          [-3.0, 1.0000001192092896]
+        ]
+      >
```
seanmor5, line 968:
```suggestion
      >

```

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,46 @@ defmodule Nx.LinAlg do
...
+      iex> pinv = Nx.LinAlg.pinv(Nx.tensor([[1, 3], [5, 4], [4, 9]]))
+      iex> pinv
```
seanmor5, line 970:
```suggestion
      iex> Nx.LinAlg.pinv(Nx.tensor([[1, 3], [5, 4], [4, 9]]))
```

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,46 @@ defmodule Nx.LinAlg do
...
+  @doc """
+  Calculates the Moore-Penrose inverse, or the pseudoinverse, of a matrix.
+
+  ## Options
+    * `:eps` - Rounding error threshold used to assume values as 0.
+
+  ## Examples
+
+      iex> pinv = Nx.LinAlg.pinv(Nx.tensor([[1, 1], [3, 4]]))
+      iex> pinv
```
seanmor5, line 961:
```suggestion
      iex> Nx.LinAlg.pinv(Nx.tensor([[1, 1], [3, 4]]))
```

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,46 @@ defmodule Nx.LinAlg do
...
+      >
+  """
+  defn pinv(tensor, opts \\ []) do
+
+    opts = keyword!(opts, eps: @default_eps)
+
+    {u, s, vt} = Nx.LinAlg.svd(tensor)
+
+    s_shape = {elem(Nx.shape(s), 0), elem(Nx.shape(u), 1)}
+    adjusted_s = Nx.map(s, fn x -> Nx.select(Nx.abs(x) < opts[:eps], 0, 1 / x) end)
```
seanmor5, line 986:
```suggestion
    adjusted_s = Nx.select(Nx.abs(x) < opts[:eps], 0, 1 / x)
```

You should basically never use Nx.map, Nx.select works element-wise

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,46 @@ defmodule Nx.LinAlg do
...
+    s_matrix = Nx.broadcast(0, s_shape) |> Nx.put_diagonal(adjusted_s)
+
+    Nx.transpose(vt) |> Nx.dot(s_matrix) |> Nx.dot(Nx.transpose(u))
```
polvalente, line 989:
```suggestion
    adjoint(vt) |> Nx.dot(s_matrix) |> Nx.dot(adjoint(u))
```

The transpose only works for real tensors. A separate test for complex should be added as well (though I don't think we need to clutter examples with this)

nx/lib/nx/lin_alg.ex
anish-lakkapragada: should be fixed in commit 14811dbec97df7007212336b3131042afa8079cc. deleted the extra line

nx/lib/nx/lin_alg.ex
anish-lakkapragada: should be fixed in 14811dbec97df7007212336b3131042afa8079cc. Used Nx.select and no Nx.map

nx/lib/nx/lin_alg.ex
anish-lakkapragada: update: 4a8f75a0e0f9c22f7c87805b12ccd809332e1e9a

nx/lib/nx/lin_alg.ex
anish-lakkapragada: update on latest commit: 4a8f75a0e0f9c22f7c87805b12ccd809332e1e9a

nx/lib/nx/lin_alg.ex
anish-lakkapragada: I've added `adjoint(vt)` but still need to test for complex matrices.

anish-lakkapragada: It appears as though the `svd` at least for the default backend is not implemented for complex inputs.

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,44 @@ defmodule Nx.LinAlg do
...
+      iex> Nx.LinAlg.pinv(Nx.tensor([[1, 3], [5, 4], [4, 9]]))
+      #Nx.Tensor<
+        f32[2][3]
+        [
+          [-0.07312049716711044, 0.3027806580066681, -0.11019568890333176],
+          [0.06900102645158768, -0.1307930052280426, 0.1462409943342209]
+        ]
+      >
+  """
+  defn pinv(tensor, opts \\ []) do
```
josevalim, line 978: Thank you. Let’s rename the function to pseudo_inverse?

We also need to make sure it works batched. If you pass a higher than rank-2 tensor, we should assume that each additional dimension works as a batch. :)

polvalente: The only issue I have with `pseudo_inverse` is that for normal inverse we use `Nx.LinAlg.invert`. Should we change that to `inverse`?

nx/lib/nx/lin_alg.ex
polvalente: Don't worry about this :)
I'm working on #1015 which reimplements SVD as whole, so I can deal with this afterwards

anish-lakkapragada: Ah, I see. I will wait for any other feedback

nx/lib/nx/lin_alg.ex
anish-lakkapragada: for more context, `np.linalg.pinv` is how it is done in numpy

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,44 @@ defmodule Nx.LinAlg do
...
+      >
+
+      iex> Nx.LinAlg.pinv(Nx.tensor([[1, 3], [5, 4], [4, 9]]))
+      #Nx.Tensor<
+        f32[2][3]
+        [
+          [-0.07312049716711044, 0.3027806580066681, -0.11019568890333176],
+          [0.06900102645158768, -0.1307930052280426, 0.1462409943342209]
+        ]
+      >
```
polvalente, line 976:
```suggestion
      iex> Nx.LinAlg.pinv(Nx.tensor(([[0.5, 0], [0, 1], [0.5, 0]]))
      #Nx.Tensor<
        f32[2][3]
        [
          [1.0, 0.0, 1.0],
          [0.0, 1.0, 0.0]
        ]
      >
```

There might be a rounding error or 2, but in general this example might look cleaner for a wide matrix

nx/lib/nx/lin_alg.ex
josevalim: @polvalente should we call it `pseudo_invert` then? or `pinv`? What do you decide?

polvalente: pinv is a really usual name for this function

All of the following langs/libs use it: NumPy/Jax, PyTorch, Julia, MATLAB/Octave

So I think it's ok to go with pinv :)

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,44 @@ defmodule Nx.LinAlg do
...
+      #Nx.Tensor<
+        f32[2][3]
+        [
+          [1.0, 0.0, 1.0],
+          [0.0, 1.0, 0.0]
+        ]
+      >
+  """
+  defn pinv(tensor, opts \\ []) do
+
```
josevalim, line 979:
```suggestion
```

josevalim (approved):

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,43 @@ defmodule Nx.LinAlg do
...
+          [4.0, -1.0000001192092896],
+          [-3.0, 1.0000001192092896]
+        ]
+      >
+
+      iex> Nx.LinAlg.pinv(Nx.tensor([[0.5, 0], [0, 1], [0.5, 0]]))
+      #Nx.Tensor<
+        f32[2][3]
+        [
+          [1.0, 0.0, 1.0],
```
polvalente, line 973:
```suggestion
          [0.9999999403953552, 0.0, 0.9999999403953552],
```

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,43 @@ defmodule Nx.LinAlg do
...
+
+  ## Examples
+
+      iex> pinv = Nx.LinAlg.pinv(Nx.tensor([[1, 1], [3, 4]]))
+      #Nx.Tensor<
+        f32[2][2]
+        [
+          [4.0, -1.0000001192092896],
+          [-3.0, 1.0000001192092896]
+        ]
```
polvalente, line 966:
```suggestion
        [
          [3.9924840927124023, -1.0052788257598877],
          [-3.005120038986206, 1.0071183443069458]
        ]
```

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,43 @@ defmodule Nx.LinAlg do
     )
   end
 
+  @doc """
+  Calculates the Moore-Penrose inverse, or the pseudoinverse, of a matrix.
+
+  ## Options
+    * `:eps` - Rounding error threshold used to assume values as 0.
```
polvalente, line 956:
```suggestion
    * `:eps` - Rounding error threshold used to assume values as 0. Defaults to `1.0e-10`
```

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,43 @@ defmodule Nx.LinAlg do
...
+      #Nx.Tensor<
+        f32[2][3]
+        [
+          [1.0, 0.0, 1.0],
+          [0.0, 1.0, 0.0]
+        ]
+      >
+  """
+  defn pinv(tensor, opts \\ []) do
+    opts = keyword!(opts, eps: @default_eps)
```
polvalente, line 979:
```suggestion
    opts = keyword!(opts, eps: 1.0e-10)
```

polvalente (approved):

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,43 @@ defmodule Nx.LinAlg do
     )
   end
 
+  @doc """
+  Calculates the Moore-Penrose inverse, or the pseudoinverse, of a matrix.
+
+  ## Options
```
polvalente, line 955:
```suggestion
  ## Options
  
```

nx/lib/nx/lin_alg.ex
```diff
@@ -949,6 +949,43 @@ defmodule Nx.LinAlg do
...
+    * `:eps` - Rounding error threshold used to assume values as 0. Defaults to `1.0e-10`
+
+  ## Examples
+
+      iex> pinv = Nx.LinAlg.pinv(Nx.tensor([[1, 1], [3, 4]]))
```
polvalente, line 960:
```suggestion
      iex> Nx.LinAlg.pinv(Nx.tensor([[1, 1], [3, 4]]))
```

PR #996, opened 2022-12-01
chore(torchx): skip broken M1 tests

polvalente:


torchx/config/config.exs
```diff
@@ -1,5 +1,10 @@
 import Config
 
+is_apple_arm64 =
+  :os.type() == {:unix, :darwin} and
+    not String.starts_with?(List.to_string(:erlang.system_info(:system_architecture)), "x86_64")
```
josevalim, line 5:
```suggestion
    not List.starts_with?(:erlang.system_info(:system_architecture), ~c"x86_64")
```

josevalim (approved):

torchx/config/config.exs
polvalente: TIL about `List.starts_with?` (although I should have looked)

josevalim (approved):

PR #1014, opened 2022-12-13
fix windows build

aschiavon91:
Apparently there's some strange relationship between the windows version of runner and the vs build tools version.
this [issue](https://github.com/SCons/scons/issues/3664) helps me to figure out what version is necessary to use

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
 
   win:
     name: Windows
-    runs-on: windows-latest
+    runs-on: windows-2019
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


polvalente:
Thanks! I tried tackling this last week but wasn't successful. Hopefully CI will be happy now :)
This started happening suddenly

aschiavon91:
I guess its  because we were using `windows-latest` runner, maybe it get update something ago, breaking our build.
I hope its gets more consistent now.

polvalente (approved):

polvalente:
> I guess its because we were using `windows-latest` runner, maybe it get update something ago, breaking our build. I hope its gets more consistent now.

Yeap. It's odd that we have to rollback to 2019, but it works!

PR #1021, opened 2022-12-15
docs(nx): specify return values for stateful operations

nickgnd:
Hey 👋 
this tiny PR is the follow-up on this comment https://github.com/elixir-nx/axon/issues/416#issuecomment-1346634166

It updates the docs for `Nx.default_backend/1`, `Nx.Defn.default_options/1` and their corresponding "global" version to highlight that the returned value is the one that was previously set (and not the updated one).

Plus, I added a doctest for the `Nx.Defn.default_options/1` analogously to the one in the `Nx.default_backend/1` docs.

Let me know if it is clear enough or if I can improve anything.

To conclude, I don't know if you follow any specific convention regarding PR titles, commits, etc. so feel free to change everything  as you wish :)

You are great!

Best,
Nicolò

### DIFF
--- a/nx/lib/nx/defn.ex
+++ b/nx/lib/nx/defn.ex
@@ -253,6 +253,15 @@ defmodule Nx.Defn do
 
         config :nx, :#{@app_key}, [compiler: EXLA, client: :cuda]
 
+  The function returns the values that were previously set as default
+  options.
+
+  ## Examples
+
+      iex> Nx.Defn.default_options(compiler: EXLA, client: :cuda)
+      []
+      iex> Nx.Defn.default_options()
+      [compiler: EXLA, client: :cuda]
   """
   def default_options(options) when is_list(options) do
     Process.put(@compiler_key, options) || Application.fetch_env!(:nx, @app_key)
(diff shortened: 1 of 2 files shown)


josevalim (approved):

nx/lib/nx/defn.ex
```diff
@@ -253,6 +253,15 @@ defmodule Nx.Defn do
 
         config :nx, :#{@app_key}, [compiler: EXLA, client: :cuda]
 
+  The function returns the values that were previously set as default
+  options.
+
+  ## Examples
+
+      iex> Nx.Defn.default_options(compiler: EXLA, client: :cuda)
+      []
```
josevalim, line 262:
```suggestion
```

josevalim: We use a different compiler for tests. :)

josevalim:

nx/lib/nx/defn.ex
nickgnd: Ops, i was too naive. sorry, thanks for the correction 🙇

PR #1028, opened 2022-12-16
Respect default EXLA.Backend client when jitting argumentless operations

jonatanklosko:
Given

```elixir
Mix.install(
  [
    {:nx, path: "~/git/nx/nx", override: true},
    {:exla, path: "~/git/nx/exla", override: true}
  ],
  config: [
    exla: [
      clients: [
        host1: [platform: :host],
        host2: [platform: :host]
      ],
      preferred_clients: [:host2]
    ]
  ]
)

Nx.default_backend({EXLA.Backend, client: :host1})
```

This gets allocated on `:host1` as expected, because we respect the default backend and options:

```elixir
Nx.tensor([1])
```

However, this gets allocated on `:host2`:

```elixir
Nx.iota({10})
```

Because we fallback to `EXLA.Client.default_name()` when jitting. This PR changes it such that we respect the default backend options instead.

### DIFF
--- a/exla/lib/exla/backend.ex
+++ b/exla/lib/exla/backend.ex
@@ -101,7 +101,7 @@ defmodule EXLA.Backend do
           ])
         end
 
-        jit(expr_fun, [tensor])
+        jit([], expr_fun, [tensor])
 
       i ->
         expr_fun = fn tensor, start_idx ->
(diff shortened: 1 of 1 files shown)


josevalim:
I believe this is the wrong fix. `iota` and friends receive the `backend_options` and I bet the `:client` is there. We need to pick it up... probably by writing a different version of those functions or adding a special case on `backend_options`.

jonatanklosko:
@josevalim you're right! Let me know if it makes more sense now :)

josevalim (approved):

josevalim:

Issue #1024, opened 2022-12-16
Support batches in SVD


polvalente:
We need to support {..., m, n} tensors for SVD

polvalente:
I have a work in progress on this. Added the issue to keep track of the missing functionality

PR #1036, opened 2022-12-21
fix: pinv edge cases

### LINKED ISSUE #1035: Nx.LinAlg.pinv crashes for zero matrices and similar
@polvalente I was trying to implement your suggestions with `pinv` but ... 😅

polvalente:
closes #1035


polvalente:
@msluszniak any other cases I didn't catch here?

(other than batched tensors with rank > 3 because this needs #1024)

josevalim (approved):

Issue #1005, opened 2022-12-10
Add metrics around Nx.Serving


josevalim:
We should emit events and also have an easy way to compute queue size and similar.

sorentwo:
The Serving module has a handful of public functions, of which `run/2` and `batched_run/2` look most suited toward instrumentation. However, the phrase "compute queue size and similar" leads me to believe the focus here is on `batched_run/2`. Is that accurate?

Also, do you expect to have sub-spans for the `handle_preprocessing`, `receive_batched`, and `handle_postprocessing` invocations?

josevalim:
I think all of those are useful:

1. metrics around run
2. metrics around batched_run
3. metrics around preprocessing and postprocessing

However, those are all client side. I would also like to have metrics on the server side. Especially because I would like to track the queue size. But we can start with the client ones and then add server ones. :)

sorentwo:
Ok, no problem. I'm investigating the server side (which is highly informative) and will take a shot at both client and server telemetry.

PR #1037, opened 2022-12-21
Add telemetry to Nx.Serving

sorentwo:
This begins by adding `telemetry` as an Nx dependency and wraps Serving's internal `batched_run` handler in a span. The span's meta includes simple statistics like queue depth, batch size limits, and other identification.

A few questions:

1. Is there any additional meta that would be helpful? The server's pid? The full stack?
2. Would you prefer a dedicated test for telemetry or include telemetry assertions in more of the server tests?

Once we've nailed this down, I'll move on to client-ide telemetry.

Addresses #1005

### DIFF
--- a/nx/lib/nx/serving.ex
+++ b/nx/lib/nx/serving.ex
@@ -310,17 +310,19 @@ defmodule Nx.Serving do
       client_postprocessing: postprocessing
     } = serving
 
-    meta = %{arg: arg, input: input, module: module}
+    {:ok, state} = handle_init(module, :inline, arg)
+    {%{size: size} = batch, info} = handle_preprocessing(preprocessing, input)
+    {:execute, function, _} = handle_batch(module, batch, state)
 
-    :telemetry.span([:nx, :serving, :run], meta, fn ->
-      {:ok, state} = handle_init(module, :inline, arg)
-      {%{size: size} = batch, info} = handle_preprocessing(preprocessing, input)
-      {:execute, function, _} = handle_batch(module, batch, state)
-      {output, metadata} = handle_executed(module, function.())
-      output = Nx.Defn.Composite.traverse(output, &Nx.slice_along_axis(&1, 0, size, axis: @axis))
+    {output, metadata} =
+      :telemetry.span([:nx, :serving, :execute], %{module: module}, fn ->
+        {output, metadata} = handle_executed(module, function.())
 
-      {handle_postprocessing(postprocessing, output, metadata, info), meta}
-    end)
+        {{output, metadata}, %{metadata: metadata, module: module}}
+      end)
+
+    output = Nx.Defn.Composite.traverse(output, &Nx.slice_along_axis(&1, 0, size, axis: @axis))
+    handle_postprocessing(postprocessing, output, metadata, info)
   end
 
   ## Process API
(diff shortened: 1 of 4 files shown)


nx/lib/nx/serving.ex
```diff
@@ -516,44 +516,48 @@ defmodule Nx.Serving do
...
-
-        # Exact match.
-        true ->
-          state
-          |> server_stack(ref, batch)
-          |> server_execute()
-      end
+    meta = %{count: count, limit: limit, module: state.module, name: state.name}
+
+    :telemetry.span([:nx, :serving, :batched_run], meta, fn ->
```
josevalim, line 521: Unfortunately this event is not very helpful because `server_execute` only starts a task. :( So if we want to measure it from the server, we cannot use span but emulate span, which is somewhat annoying. The other thing is that we would want to include the queue size as a measurement, not as metadata, and we cannot add new measurements to telemetry spans anyway.

So my suggestion, for now, is to start with the client events? And then later we figure out the queue size. Because for queue size we will need to either send it to the client (and use custom spans) or emit additional events on the server (spans or not).

sorentwo: Ok, thanks for the context. I'll roll this back and switch to client events initially.

sorentwo:
@josevalim Reworked to add telemetry only to `preprocessing`, `postprocessing`, and `run`. At least this gets the telemetry dep in, establishes event naming, and makes use of the brand new `attach_event_handlers`.

nx/test/nx/serving_test.exs
```diff
@@ -115,6 +115,34 @@ defmodule Nx.ServingTest do
       assert batch.pad == 0
       assert Nx.Defn.jit_apply(&Function.identity/1, [batch]) == Nx.tensor([[1, 2], [3, 4]])
     end
+
+    @tag :capture_log
```
sorentwo, line 119: `attach_event_handlers` triggers telemetry's "anonymous function" warning, hence capturing the logs. The warning should be silenced/avoided by the test helper. I think the warning would be confusing if you haven't worked with telemetry handlers before.

josevalim: 1.2.1 is out with a fix for that :)

josevalim:
@sorentwo should we avoid emitting the telemetry event if there is no preprocessing or postprocessing callback?

nx/lib/nx/serving.ex
```diff
@@ -310,12 +310,17 @@ defmodule Nx.Serving do
...
 
-    {:ok, state} = handle_init(module, :inline, arg)
-    {%{size: size} = batch, info} = handle_preprocessing(preprocessing, input)
-    {:execute, function, _} = handle_batch(module, batch, state)
-    {output, metadata} = handle_executed(module, function.())
-    output = Nx.Defn.Composite.traverse(output, &Nx.slice_along_axis(&1, 0, size, axis: @axis))
-    handle_postprocessing(postprocessing, output, metadata, info)
+    meta = %{arg: arg, input: input, module: module}
+
+    :telemetry.span([:nx, :serving, :run], meta, fn ->
```
josevalim, line 315: Instead of wrapping the whole thing, what if we wrap only `handle_executed` with an event of `:execute`? It is better as it applies to both batched_run and run.

sorentwo: Sounds good to me!

sorentwo:
> should we avoid emitting the telemetry event if there is no preprocessing or postprocessing callback?

Personally, I'd opt for consistent events, but I'll leave it up to you.

nx/test/nx/serving_test.exs
sorentwo: Indeed. Updating it 👍

nx/lib/nx/serving.ex
sorentwo: Did you mean `handle_batch`? The `handle_executed` function only validates that the result is a two-element tuple.

josevalim: @sorentwo yes, I meant the call to that exact function!

josevalim:
> Personally, I'd opt for consistent events, but I'll leave it up to you.

Let's remove it for now, easy to add later. :)

nx/lib/nx/serving.ex
```diff
@@ -705,24 +712,34 @@ defmodule Nx.Serving do
...
-    case preprocessing.(input) do
-      {%Nx.Batch{} = batch, info} ->
-        {no_empty_batch!(batch), info}
+    meta = %{input: input}
 
-      other ->
-        raise "client_preprocessing function #{inspect(preprocessing)} must return a two element tuple " <>
-                "where the first element is a Nx.Batch and the second is any value. Got: #{inspect(other)}"
-    end
+    :telemetry.span([:nx, :serving, :preprocessing], meta, fn ->
```
sorentwo, line 725: Events are only fired when a callback is present now.

nx/lib/nx/serving.ex
```diff
@@ -313,7 +313,14 @@ defmodule Nx.Serving do
     {:ok, state} = handle_init(module, :inline, arg)
     {%{size: size} = batch, info} = handle_preprocessing(preprocessing, input)
     {:execute, function, _} = handle_batch(module, batch, state)
-    {output, metadata} = handle_executed(module, function.())
+
+    {output, metadata} =
+      :telemetry.span([:nx, :serving, :execute], %{module: module}, fn ->
+        {output, metadata} = handle_executed(module, function.())
```
sorentwo, line 319: The function call and `handle_executed` data check are both wrapped.

nx/lib/nx/serving.ex
```diff
@@ -616,8 +623,16 @@ defmodule Nx.Serving do
     batch = Nx.Batch.merge(batches)
     {:execute, function, module_state} = handle_batch(module, batch, module_state)
 
+    wrapped_function = fn ->
```
sorentwo, line 626: The function is passed to `async_nolink` below, so this additional wrapper is the best solution I see.

josevalim:

PR #1044, opened 2022-12-24
Touch up `Nx.Serving` docs

Benjamin-Philip:


josevalim (approved):

Benjamin-Philip:
The word "serving" is typically used as conjugated from of the verb "to serve".  The only context where it is used as a noun is in the context of cooking i.e. "large serving of rice". Using "serving" in the docs can be somewhat ambiguous and also feels like incorrect English to me. So, I propose that we replace "serving" with "service" in the docs.

 https://github.com/elixir-nx/nx/blob/5c77611d0548505555fbaf34529e2e9a09dceae9/nx/lib/nx/serving.ex#L23

However, the word "service" implicitly suggests that a `Nx.Serving` instance is a long running process. While it is not necessary that a "service" is something long running (like when you service you car), in the context of computers, it usually is. Also, switching to "service" without changing the module name, can also be confusing.

This problem has 2 solutions (that I can think of):

1. We could dedicate a paragraph in the moduledoc explaining the "serving" terminology
2. We could switch to "service" like proposed.

I'm not sure which is a better solution. Once José, Sean and Paulo have chosen a solution, we can proceed with it.

josevalim:
Let’s introduce a note clarifying the use of “serving”. I think service is too ambiguous within and outside the Erlang VM. :)

polvalente:
> The word "serving" is typically used as conjugated from of the verb "to serve". The only context where it is used as a noun is in the context of cooking i.e. "large serving of rice". Using "serving" in the docs can be somewhat ambiguous and also feels like incorrect English to me. So, I propose that we replace "serving" with "service" in the docs.
> 
> https://github.com/elixir-nx/nx/blob/5c77611d0548505555fbaf34529e2e9a09dceae9/nx/lib/nx/serving.ex#L23
> 
> However, the word "service" implicitly suggests that a `Nx.Serving` instance is a long running process. While it is not necessary that a "service" is something long running (like when you service you car), in the context of computers, it usually is. Also, switching to "service" without changing the module name, can also be confusing.
> 
> This problem has 2 solutions (that I can think of):
> 
> 1. We could dedicate a paragraph in the moduledoc explaining the "serving" terminology
> 2. We could switch to "service" like proposed.
> 
> I'm not sure which is a better solution, so I would like José, Sean and Paulo's opinion with which we can proceed accordingly.

@Benjamin-Philip this is a good point to bring up. Just leaving my 2 cents here:

Foreword: Since we're talking about the grammar of "serving", I'll explain how I read Nx.Serving, this isn't necessarily the only correct way.

While I do agree that "serving" by itself as a noun means "portion of food", we can read "Nx.Serving" as an expression of noun (Nx) + noun as adjective (Serving).
Reading it like this, "Serving" acts as an adjective and conveys the idea of serving portions (_batches_) of "Nx". "Nx" would be a metonymy for "Nx Computations", which finally connects to the description of "encapsulates [...] to perform batched [computations]"

Although, I'm not exactly bothered by the current naming, perhaps removing the metonymy would clarify things? `Nx.Serving` -> `Nx.ComputationServing`

Benjamin-Philip:
@josevalim, could you explain what you mean by "by started with the process" in the docs:

https://github.com/elixir-nx/nx/blob/4c04d5ae660ca57a3f477d5e7938d79219c726c8/nx/lib/nx/serving.ex#L225

Benjamin-Philip:
> Although, I'm not exactly bothered by the current naming, perhaps removing the metonymy would clarify things? `Nx.Serving` -> `Nx.ComputationServing`

I'm not in favour of that. I think that would just make 3rd party documentation and references to the `Nx.Serving` inconsistent.

Benjamin-Philip:
One more thing:

Nothing in the docs (except maybe the type signatures) specifies that a serving can have only one pre and post processing hook. I'm think we should mention this in the docs, and warn on overriding.

josevalim:

I merged this one since I have reviewed it twice. Further changes can be sent as separate PRs. :)

PR #1048, opened 2022-12-31
Allow tensors to be given on Nx.tensor/2

josevalim:


josevalim:
I am closing this because this would encourage people to pass lists as options (which are uniquely cached) into `defn`.

josevalim:
Actually, we already have such issue today. Someone can do this:

```elixir
defn foo(opts) do
  Nx.tensor(opts[:weights])
end
```

At the same time, we also want to allow this code:

```elixir
defn foo(...) do
  Nx.tensor([-1, 0, 1])
end
```

Perhaps one option is to force the first argument of `Nx.tensor/2` to be literal inside `defn`.

josevalim:
Two follow up questions from this:

1. Do we also want to have an explicit `Nx.from_list` that explicitly requires the input to be lists? A downside of `from_list` is that it cannot create a scalar (nor does `from_binary` but `from_binary` does not support nesting).

2. Do we want to support binaries in `Nx.tensor`? They would require the `:type` option to be given.

jonatanklosko (approved):
I don't have a strong opinion on whether to allow the tensor as argument, but works fine as a convenience.

I agree that ideally we would only allow literals in defn. The question is if this won't hurt certain cases where we want to convert an option to a scalar of specific type like `Nx.tensor(opts[:x], :s8)`. We can always wrap in a `deftransform`, but just pointing this out.

polvalente:
I'm not sure about the behavior of transferring the tensor if the options are given. That seems like a potential hidden source of problems

josevalim:
@polvalente the other options are doing nothing or raising. I think doing nothing is bad: we either return something in the same backend or raise. Would you prefer raising?

one of the reasons for this change is convenience for code in Scholar and I think it would be convenient to be able to allocated a tensor somewhere regardless if it comes as a list, number, or another tensor.

polvalente:
I think we could raise if the input is a tensor but the backend is different

josevalim:
And what about :type? Should we raise or convert? If convert, Why in one but not the other? 🤔

polvalente:
> And what about :type? Should we raise or convert? If convert, Why in one but not the other? 🤔

Ideally `Nx.tensor(%Nx.Tensor{})` would be a no-op IMO, so raise as well.

In this sense, the implementation would be a convenience for accepting both the tensor and the numerical representation of the same value/type/backend.

seanmor5:
+1 to @polvalente comment

josevalim:
@polvalente @seanmor5 I have implemented your suggestions. My goal was to make sure that the result of `Nx.tensor/2` is consistent regardless if a tensor or list is given. To achieve so, each option has a different behaviour:

* `:type` - we check the tensor type only if a `:type` is given. This makes sense because `Nx.tensor` can return integer, boolean, and floating tensors naturally if a type is not given. So the type is by default based on the input

* `:backend` - if backend is given, we check it. This has two issues: 1. we only check the backend but not the device/client (since device/client are currently backend specific and opaque to Nx). 2. if you don't pass the backend option, we don't check the tensor, which means you can get a tensor not allocated on the default backend. I believe this is reasonable behaviour

* `:names` - the current names in tensor are always discarded but they are checked against the given names, if any, to avoid conflicts

nx/lib/nx.ex
```diff
@@ -436,11 +436,16 @@ defmodule Nx do
...
-  (zero-dimensions), a list of those (the tensor is a vector) or
-  a list of n-lists of those, leading to n-dimensional tensors.
-  The tensor will be allocated in `Nx.default_backend/0`, unless the
-  `:backend` option is given, which overrides the default one.
+  The argument must be one of:
+
+    * a tensor
+    * a number (which means the tensor is scalar/zero-dimensional)
+    * a boolean (also scalar/zero-dimensional)
+    * an arbitraly nested list of numbers and booleans
```
polvalente, line 444:
```suggestion
    * an arbitrarily nested list of numbers and booleans
```

polvalente (approved):

josevalim:

PR #1049, opened 2023-01-03
docs(exla): add nerves section

polvalente:
Adds a section on the Readme for usage with Nerves


exla/README.md
```diff
@@ -44,6 +44,19 @@ EXLA relies on the [XLA](https://github.com/elixir-nx/xla) package to provide th
...
+
+To get EXLA to work on your target Nerves device, you need to add `build_embedded: true` to your app's config.
+
+For example, if you're building an app called `:hello_nx`, you set:
+
+```elixir
+config :hello_nx, build_embedded: true
+```
+
+You also need to set your `XLA_ARCHIVE_URL` variable to the correct `aarch64` archive.
```
polvalente, line 57: Perhaps we can improve the behavior for automatically setting this somehow?
My question would be how to decide on the target architecture, though

jonatanklosko: @polvalente what is the target we currently infer?

jonatanklosko: Ahh, it's because we use the host target at build time, I see.

jonatanklosko: We could add something like `XLA_FULL_TARGET=aarch64-linux-gnu-cpu` and then pass that in mix.exs?

exla/README.md
```diff
@@ -44,6 +44,19 @@ EXLA relies on the [XLA](https://github.com/elixir-nx/xla) package to provide th
...
     * If you have `protoc` installed on your machine, it may conflict with the `protoc` precompiled inside XLA. Uninstall, unlink, or remove `protoc` from your path to continue.
 
+### Usage with Nerves
+
+To get EXLA to work on your target Nerves device, you need to add `build_embedded: true` to your app's config.
+
+For example, if you're building an app called `:hello_nx`, you set:
+
+```elixir
+config :hello_nx, build_embedded: true
```
josevalim, line 54: Why do we need build_embedded?

polvalente: This is read by EXLA's (and Torchx's) mix.exs and is passed to the Makefile so that the dynamic libs are copied to priv instead of linked, which is necessary for Nerves to copy the data to the target image.

This also tells nerves to use a cross-compiler instead of a regular compiler

exla/README.md
polvalente: That sounds like a better alternative!
Either that or a variable for setting the architecture (since we already have one for cpu/cuda/etc)

jonatanklosko: Sounds good! https://github.com/elixir-nx/xla/pull/38

josevalim (approved):

PR #1055, opened 2023-01-09
Add axes to flatten

### LINKED ISSUE #1050: Add `:axes` option to flatten
Right now `Nx.flatten` will flatten all of the axes in a tensor into a single dimension. I propose we add an `:axes` option, which requires a sequence of successive axes to flatten into a single axis. Some examples:

```elixir
Nx.flatten(Nx.tensor([[[1], [2], [3]], [[4], [5], [6]]]), axes: [-2, -3])
```

will flatten the last 2 axes into a single dimension:

```
#Nx.Tensor<
  s64[2][3]
  [
    [1, 2, 3],
    [4, 5, 6]
  ]
>
```

Because it's unclear how to flatten successive dimensions, this should raise:

```elixir
Nx.flatten(Nx.tensor([[[1], [2], [3]], [[4], [5], [6]]]), axes: [0, 2])
```

seanmor5:
Resolves #1050

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -2472,10 +2472,48 @@ defmodule Nx do
         s64[16]
         [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
       >
+
+  You may also pass `:axes` to `Nx.flatten/2`, to specify which consecutive
+  axes to flatten:
+
+      iex> t = Nx.iota({1, 2, 3})
+      #Nx.Tensor<
+        s64[1][2][3]
+        [
+          [
+            [0, 1, 2],
+            [3, 4, 5]
+          ]
+        ]
+      >
+      iex> Nx.flatten(t, axes: [1, 2])
+      #Nx.Tensor<
+        s64[1][6]
+        [
+          [0, 1, 2, 3, 4, 5]
+        ]
+      >
+
+  `:axes` must be consecutive, otherwise it will raise:
+
+      iex> t = Nx.iota({1, 2, 3})
+      #Nx.Tensor<
+        s64[1][2][3]
+        [
+          [
+            [0, 1, 2],
+            [3, 4, 5]
+          ]
+        ]
+      >
+      iex> Nx.flatten(t, axes: [0, 2])
+      ** (ArgumentError) flatten axes must be consecutive
   """
   @doc type: :shape
-  def flatten(tensor) do
-    reshape(tensor, {size(tensor)})
+  def flatten(tensor, opts \\ []) do
+    opts = Keyword.validate!(opts, [:axes])
+    {shape, names} = Nx.Shape.flatten(tensor.shape, tensor.names, opts[:axes])
+    reshape(tensor, shape, names: names)
   end
 
   @doc """
@@ -7016,10 +7054,15 @@ defmodule Nx do
       end
 
     t =
-      if axis do
-        sort(tensor, axis: axis)
-      else
-        tensor |> flatten() |> sort()
+      case {shape, axis} do
+        {{}, _} ->
+          reshape(tensor, {1})
+
+        {_, nil} ->
+          tensor |> flatten() |> sort()
+
+        {_, axis} ->
+          sort(tensor, axis: axis)
       end
 
     axis_size =
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7054,15 +7054,10 @@ defmodule Nx do
       end
 
     t =
-      case {shape, axis} do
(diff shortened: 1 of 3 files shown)


nx/lib/nx/shape.ex
```diff
@@ -556,6 +556,78 @@ defmodule Nx.Shape do
...
+      {{1, 2, 3}, [nil, nil, nil]}
+
+  ### Error cases
+
+      iex> Nx.Shape.flatten({1, 2, 3}, [:batch, nil, nil], [0, 2])
+      ** (ArgumentError) flatten axes must be consecutive
+  """
+  def flatten(shape, names, axes)
+
+  def flatten({}, _, _), do: {{}, []}
```
josevalim, line 581: Shouldn't we raise if you pass axes and the shape does not have said axes? So I think we still need to call `normalize_axes` on this? Perhaps:

```suggestion
  def flatten({}, [], []), do: {{}, []}
```

But perhaps not even worth having this branch because it is not common to flatten scalars?

nx/lib/nx/shape.ex
```diff
@@ -556,6 +556,78 @@ defmodule Nx.Shape do
...
+
```
josevalim, line 582: Shouldn't we raise if you pass axes and the shape does not have said axes? So I think we still need to call `normalize_axes` on this? But I think we should rather remove this clause because the optimizations below will also handle it.

```suggestion
```

nx/lib/nx/shape.ex
```diff
@@ -556,6 +556,78 @@ defmodule Nx.Shape do
...
+    {rev_shape, rev_names, flatten_size} =
+      for {{size, name}, i} <- Enum.with_index(shape_and_names), reduce: {[], [], 1} do
+        {cur_shape, cur_names, cur_size} ->
+          if i in axes do
+            {cur_shape, cur_names, size * cur_size}
+          else
+            {[size | cur_shape], [name | cur_names], cur_size}
+          end
+      end
```
josevalim, line 604: Nitpick: since we already have a loop, might as well "fuse" the `with_index` operation:

```suggestion
    {rev_shape, rev_names, flatten_size, _} =
      for {size, name} <- shape_and_names, reduce: {[], [], 1, 0} do
        {cur_shape, cur_names, cur_size, i} ->
          if i in axes do
            {cur_shape, cur_names, size * cur_size, i + 1}
          else
            {[size | cur_shape], [name | cur_names], cur_size, i + 1}
          end
      end
```

nx/lib/nx/shape.ex
```diff
@@ -556,6 +556,78 @@ defmodule Nx.Shape do
...
+  defp consecutive?([]), do: true
+
+  defp consecutive?([last | next]) do
+    {res, _} =
+      Enum.reduce(next, {true, last}, fn cur, {cons?, last} ->
+        {cons? and last == cur - 1, cur}
+      end)
+
+    res
+  end
```
jonatanklosko, line 629:
```suggestion
  defp consecutive?([]), do: true
  defp consecutive?([_]), do: true
  defp consecutive?([x, y | rest]) when x == y - 1, do: consecutive?([y | rest])
  defp consecutive?(_), do: false
```

josevalim, line 629: I have not tested it but wdyt about:

```suggestion
  defp consecutive?([cur, next | rest]) when next == cur + 1, do: consecutive?([next | rest])
  defp consecutive?([_last]), do: true
  defp consecutive?(_), do: false
```

josevalim: I dropped a similar suggestion. :)  Just note you don't need to check for the empty list!

josevalim: It is just that this function is never called with an empty list!

jonatanklosko: Ah optimising for that, sounds good then! I would make the catch all `[_, _ | _]` just in case then.

nx/lib/nx.ex
```diff
@@ -7016,10 +7054,15 @@ defmodule Nx do
       end
 
     t =
-      if axis do
-        sort(tensor, axis: axis)
-      else
-        tensor |> flatten() |> sort()
+      case {shape, axis} do
+        {{}, _} ->
+          reshape(tensor, {1})
```
polvalente, line 7059: Doesn't this mean that this PR changes the way default flatten works?

I wouldn't expect having to change anywhere thag calls Nx.flatten for adding the new feature

seanmor5: It does, but I think the original behavior was wrong. I would not expect `Nx.flatten` on a scalar to create a new dimension

josevalim (approved):

PR #1056, opened 2023-01-09
fix: render containers in compile error type mismatch

polvalente:


nx/lib/nx/template_backend.ex
```diff
@@ -11,8 +11,12 @@ defmodule Nx.TemplateBackend do
   defstruct []
 
   @impl true
-  def inspect(%Nx.Tensor{}, _opts) do
-    "Nx.TemplateBackend"
+  def inspect(%Nx.Tensor{}, opts) do
+    if opts.custom_options[:add_backend_on_inspect] do
+      "Nx.TemplateBackend"
+    else
+      ""
```
josevalim, line 18:
```suggestion
      Inspect.Algebra.empty()
```

josevalim: Actually, this logic should be on Nx.Tensor, so it applies to all backends.

polvalente: I'll try something out :)

nx/lib/nx/defn/expr.ex
```diff
@@ -1182,32 +1191,17 @@ defmodule Nx.Defn.Expr do
...
-    names = []
-    type = Nx.Type.infer(number)
-    Nx.Type.to_string(type) <> Nx.Shape.to_string(shape, names)
+  defp inspect_as_template(tensor_or_container)
+       when Nx.is_tensor(tensor_or_container) or is_map(tensor_or_container) or
+              is_tuple(tensor_or_container) do
+    Kernel.inspect(Nx.to_template(tensor_or_container),
+      custom_options: [skip_template_backend_header: true]
+    )
   end
```
josevalim, line 1202: I think this would work too:

```suggestion
  defp inspect_as_template(tensor_or_container) do
    Kernel.inspect(Nx.to_template(tensor_or_container),
      custom_options: [skip_template_backend_header: true]
    )
  end
```

polvalente: It raised for atoms

nx/lib/nx/template_backend.ex
josevalim: And this one should still apply!

nx/lib/nx/defn/expr.ex
josevalim: What about:

```
defp inspect_as_template(data) do
  if Nx.Container.impl_for(data) == Nx.Container.Any do
    inspect(data)
  else
    Kernel.inspect(Nx.to_template(tensor_or_container),
      custom_options: [skip_template_backend_header: true]
    )
  end
end
```

?

polvalente: TIL about impl_for

This is much better

nx/test/nx/container_test.exs
```diff
@@ -141,7 +141,7 @@ defmodule Nx.ContainerTest do
       var.a + var.b
     end
 
-    deftransformp assert_fields!(%C{c: %{}, d: :keep}), do: 1
+    deftransformp(assert_fields!(%C{c: %{}, d: :keep}), do: 1)
```
polvalente, line 144:
```suggestion
    deftransformp assert_fields!(%C{c: %{}, d: :keep}), do: 1
```

josevalim (approved):

PR #1062, opened 2023-01-11
Makes the compile library symlink relative

### LINKED ISSUE #1061: EXLA compiled libraries absolute symbolic links
Currently, during compilation, EXLA creates symbolic links to the compiled libraries in these lines:

https://github.com/elixir-nx/nx/blob/main/exla/Makefile#L41-L49

The problem is that they are absolute path symlinks, for example:
```
./_build/test/lib/exla/priv:
lib -> /app/deps/exla/cache/xla_extension/lib
libexla.so -> /app/deps/exla/cache/libexla.so
```

This behavior is different from all other symbolic links created by dependencies, where relative paths are used, such as Phoenix:

```
./_build/test/lib/phoenix:
priv -> ../../../../deps/phoenix/priv
```

When the project folder is moved, the absolute symbolic links lose track of the libraries, breaking the build. 

In our case, our build was breaking because we move all the files between the compilation and the tests execution. It took a while to find these special absolute symbolic links. 

I propose making these symlinks relative the same way as the other symlinks so they are consistent on the "non-release" enviroments.

For the releases, they are already copied since https://github.com/elixir-lang/elixir/pull/12011. Making them relative shouldn't affect them.

thiagopromano:
Closes #1061

### DIFF
--- a/exla/Makefile
+++ b/exla/Makefile
@@ -44,7 +44,8 @@ $(EXLA_SO): $(EXLA_CACHE_SO)
 		cp -a $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
 		cp -a $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
 	else \
-		if ! command -v realpath &> /dev/null; then \
+		if ! [ -x "$(shell which realpath)" ]; then \
+			echo "realpath not found, falling back to absolute symlinks" ; \
 			ln -sf $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
 			ln -sf $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
 		else \
--- a/exla/Makefile
+++ b/exla/Makefile
@@ -44,8 +44,8 @@ $(EXLA_SO): $(EXLA_CACHE_SO)
 		cp -a $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
 		cp -a $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
 	else \
-		ln -sf $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
-		ln -sf $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
+		ln -srf $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
+		ln -srf $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
 	fi
 
 $(EXLA_CACHE_SO): $(XLA_EXTENSION_DIR) $(EXLA_DIR)/exla.cc $(EXLA_DIR)/exla_client.cc $(EXLA_DIR)/exla_client.h $(EXLA_DIR)/exla_nif_util.cc $(EXLA_DIR)/exla_nif_util.h $(EXLA_DIR)/exla_log_sink.h
--- a/exla/Makefile
+++ b/exla/Makefile
@@ -48,8 +48,8 @@ $(EXLA_SO): $(EXLA_CACHE_SO)
 			ln -sf $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
 			ln -sf $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
 		else \
-			ln -sf $(realpath --relative-to=$(EXLA_LIB_DIR) $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
-			ln -sf $(realpath --relative-to=$(EXLA_SO) $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
+			ln -sf $(shell realpath --relative-to=$(PRIV_DIR) $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
+			ln -sf $(shell realpath --relative-to=$(PRIV_DIR) $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
 		fi \
 	fi
 
--- a/exla/Makefile
+++ b/exla/Makefile
@@ -38,16 +38,22 @@ else
 	POST_INSTALL = $(NOOP)
 endif
 
+ifeq (, $(shell which realpath))
+	XLA_EXTENSION_LIB_LINK_PATH = $(abspath $(XLA_EXTENSION_LIB))
+	EXLA_CACHE_SO_LINK_PATH = $(abspath $(EXLA_CACHE_SO))
+else
+	XLA_EXTENSION_LIB_LINK_PATH = $(shell realpath --relative-to=$(PRIV_DIR) $(XLA_EXTENSION_LIB))
+	EXLA_CACHE_SO_LINK_PATH = $(shell realpath --relative-to=$(PRIV_DIR) $(EXLA_CACHE_SO))
+endif
+
 $(EXLA_SO): $(EXLA_CACHE_SO)
 	@ mkdir -p $(PRIV_DIR)
 	@ if [ "${MIX_BUILD_EMBEDDED}" = "true" ]; then \
 		cp -a $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
 		cp -a $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
 	else \
-		xla_extension_path="${shell which realpath && realpath --relative-to=$(PRIV_DIR) $(XLA_EXTENSION_LIB) || echo ${abspath $(XLA_EXTENSION_LIB)} }" ; \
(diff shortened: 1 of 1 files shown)


exla/Makefile
```diff
@@ -44,8 +44,8 @@ $(EXLA_SO): $(EXLA_CACHE_SO)
 		cp -a $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
 		cp -a $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
 	else \
-		ln -sf $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
-		ln -sf $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
+		ln -srf $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
+		ln -srf $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
```
polvalente, line 48: This option isn't available on Mac

josevalim: Unfortunately realpath is only available on the latest macOS Ventura. So my suggestion is to check if there  is a  `realpath` and act accordingly.

exla/Makefile
```diff
@@ -44,8 +44,13 @@ $(EXLA_SO): $(EXLA_CACHE_SO)
...
 	else \
-		ln -sf $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
-		ln -sf $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
+		if ! command -v realpath &> /dev/null; then \
+			ln -sf $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
+			ln -sf $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
+		else \
+			ln -sf $(realpath --relative-to=$(EXLA_LIB_DIR) $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
+			ln -sf $(realpath --relative-to=$(EXLA_SO) $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
+		fi \
```
josevalim, line 53: @thiagopromano can you please check if this version works as expected for you?

thiagopromano: It doesn't actually, it still generates absolute links, I believe that MAKE ends up using `realpath(3)` as said here: https://www.gnu.org/software/make/manual/html_node/File-Name-Functions.html

Anyway, I've just pushed a change that works on both MacOS and Linux, I've tested it locally on both OSs this time 😄 .

thiagopromano: I didn't see you last comment on Venture, I'll add a check there.

exla/Makefile
```diff
@@ -44,8 +44,8 @@ $(EXLA_SO): $(EXLA_CACHE_SO)
 		cp -a $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
 		cp -a $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
 	else \
-		ln -sf $(abspath $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
-		ln -sf $(abspath $(EXLA_CACHE_SO)) $(EXLA_SO) ; \
+		ln -sf $(shell realpath --relative-to $(PRIV_DIR) $(XLA_EXTENSION_LIB)) $(EXLA_LIB_DIR) ; \
```
polvalente, line 47: I wonder if just passing the relative path for `XLA_EXTENSION_LIB` directly would work, since realpath isn't available:

```suggestion
		ln -sf $(PRIV_DIR)/../$(XLA_EXTENSION_LIB) $(EXLA_LIB_DIR) ; \
```

polvalente: I guess it'd depend if PRIV_DIR is a relative path in itself though

thiagopromano:
I'm running some tests with the latest version, it seems that it still creates absolute links on my Linux machine.

exla/Makefile
polvalente: Yeah, I wasn't sure if PRIV_DIR was an abs path

thiagopromano:
The check wasn't working as intended, I've fixed it so it works.

In a system **without** `realpath` with EXLA as a dependency:

```
(base) thiago@LAPTOP-KUNGDKE5:~/workspace/app$ rm -rf _build/dev/lib/exla/ && mix deps.compile exla
==> exla
Using libexla.so from /home/thiago/.cache/xla/exla/elixir-1.14.0-erts-13.1-xla-0.4.3-exla-0.4.1-m24k4pxznvgx477ou5lb3lhhbe/libexla.so
Compiling 21 files (.ex)
warning: function svd/3 required by behaviour Nx.Backend is not implemented (in module EXLA.Backend)
  lib/exla/backend.ex:1: EXLA.Backend (module)

Generated exla app
(base) thiago@LAPTOP-KUNGDKE5:~/workspace/app$ ls -l _build/dev/lib/exla/priv/
total 4
lrwxrwxrwx 1 thiago thiago 65 Jan 12 15:54 lib -> /home/thiago/workspace/app/deps/exla/exla/cache/xla_extension/lib
lrwxrwxrwx 1 thiago thiago 58 Jan 12 15:54 libexla.so -> /home/thiago/workspace/app/deps/exla/exla/cache/libexla.so
```

In a system **with** `realpath` with EXLA as a dependency:

```
(base) thiago@LAPTOP-KUNGDKE5:~/workspace/app$ rm -rf _build/dev/lib/exla/ && mix deps.compile exla
==> exla
Using libexla.so from /home/thiago/.cache/xla/exla/elixir-1.14.0-erts-13.1-xla-0.4.3-exla-0.4.1-m24k4pxznvgx477ou5lb3lhhbe/libexla.so
Compiling 21 files (.ex)
warning: function svd/3 required by behaviour Nx.Backend is not implemented (in module EXLA.Backend)
  lib/exla/backend.ex:1: EXLA.Backend (module)

Generated exla app
(base) thiago@LAPTOP-KUNGDKE5:~/workspace/app$ ls -l _build/dev/lib/exla/priv/
total 0
lrwxrwxrwx 1 thiago thiago 53 Jan 12 15:55 lib -> ../../../../../deps/exla/exla/cache/xla_extension/lib
lrwxrwxrwx 1 thiago thiago 46 Jan 12 15:55 libexla.so -> ../../../../../deps/exla/exla/cache/libexla.so

```



In a system **without** `realpath` working on Nx:

```
(base) thiago@LAPTOP-KUNGDKE5:~/workspace/nx/exla$ rm -rf _build/dev/lib/exla/ && mix compile
Using libexla.so from /home/thiago/.cache/xla/exla/elixir-1.14.0-erts-13.1-xla-0.4.0-exla-0.4.1-uug4gsdjyvckttsk7pttejl3oe/libexla.so
Compiling 21 files (.ex)
Generated exla app
(base) thiago@LAPTOP-KUNGDKE5:~/workspace/nx/exla$ ls -l _build/dev/lib/exla/priv/
total 0
lrwxrwxrwx 1 thiago thiago 54 Jan 12 15:56 lib -> /home/thiago/workspace/nx/exla/cache/xla_extension/lib
lrwxrwxrwx 1 thiago thiago 47 Jan 12 15:56 libexla.so -> /home/thiago/workspace/nx/exla/cache/libexla.so
```

In a system **with** `realpath` working on Nx:
```
(base) thiago@LAPTOP-KUNGDKE5:~/workspace/nx/exla$ rm -rf _build/dev/lib/exla/ && mix compile
Using libexla.so from /home/thiago/.cache/xla/exla/elixir-1.14.0-erts-13.1-xla-0.4.0-exla-0.4.1-uug4gsdjyvckttsk7pttejl3oe/libexla.so
Compiling 21 files (.ex)
Generated exla app
(base) thiago@LAPTOP-KUNGDKE5:~/workspace/nx/exla$ ls -l _build/dev/lib/exla/priv/
total 0
lrwxrwxrwx 1 thiago thiago 38 Jan 12 15:56 lib -> ../../../../../cache/xla_extension/lib
lrwxrwxrwx 1 thiago thiago 31 Jan 12 15:56 libexla.so -> ../../../../../cache/libexla.so
```


I've also tested on my build enviroment and it works as intended. It's a shame that MacOS doesn't have the -r option in `ln` from the start. :)

josevalim (approved):

jonatanklosko:
Mix-installing from GitHub fails for me currently, because `realpath` on macOS doesn't support `--relative-to`.

josevalim:
@jonatanklosko ah, I still have the old macOS, that's why it works here. Perhaps we should check if the OS is not Darwin. Can you please send a PR?

jonatanklosko:
I'm not sure if it's actually reliably available on all linux distributions either:

```shell
$ docker run --rm alpine:latest realpath --relative-to=/ .
/
realpath: --relative-to=/: No such file or directory
```

jonatanklosko:
Perhaps we could pass the relative path on Elixir side and pass as env var, the downside of that would be more path information living outside Makefile.

polvalente:
This SO answer looks nice: https://unix.stackexchange.com/questions/720080/how-do-i-resolve-a-relative-path-in-a-posix-shell-if-readlink-realpath-is-not-av

josevalim:
I’d prefer to compute it in Elixir if doable.

PR #1063, opened 2023-01-12
chore: update complex to 0.4.3

polvalente:


polvalente:
@josevalim should I also change torchx to depend on `path: "../nx"` so we can update it there as well? Or leave it as is and update on the next release?

josevalim:
@polvalente you can do a mix deps.update there and it will be enough. But as-is also works, it will be fixed at release time. Your call.

josevalim (approved):

PR #1065, opened 2023-01-15
feat: add linspace

polvalente:
Brings the implementation from polvalente/nx-signal#3

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -12090,7 +12090,7 @@ defmodule Nx do
   ## Options
 
     * `:n` - The number of samples in the tensor.
-    * `:axis_name` - Optional name for the output axis.
+    * `:name` - Optional name for the output axis.
     * `:type` - Optional type for the output. Defaults to `{:f, 32}`
     * `:endpoint` - Boolean that indicates whether to include `stop`
       as the last point in the output. Defaults to `true`
@@ -12103,9 +12103,9 @@ defmodule Nx do
         [5.0, 5.75, 6.5, 7.25, 8.0]
       >
 
-      iex> Nx.linspace(0, 10, n: 5, endpoint: false)
+      iex> Nx.linspace(0, 10, n: 5, endpoint: false, name: :x)
       #Nx.Tensor<
-        f32[5]
+        f32[x: 5]
         [0.0, 2.0, 4.0, 6.0, 8.0]
       >
 
@@ -12134,7 +12134,7 @@ defmodule Nx do
       ** (ArgumentError) expected n to be a non-negative integer, got: 1.0
   """
   def linspace(start, stop, opts \\ []) do
-    opts = keyword!(opts, [:n, :axis_name, type: {:f, 32}, endpoint: true])
+    opts = keyword!(opts, [:n, :name, type: {:f, 32}, endpoint: true])
 
     n = opts[:n]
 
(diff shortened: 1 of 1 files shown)


nx/lib/nx.ex
```diff
@@ -12084,6 +12084,86 @@ defmodule Nx do
...
+  @doc """
+  Creates a tensor of shape `{n}` with linearly spaced samples between `start` and `stop`
+
+  ## Options
+
+    * `:n` - The number of samples in the tensor.
+    * `:axis_name` - Optional name for the output axis.
+    * `:type` - Optional type for the output. Defaults to `{:f, 32}`
+    * `:endpoint` - Boolean that indicates whether to include `stop`
+                    as the last point in the output. Defaults to `true`
```
josevalim, line 12096:
```suggestion
      as the last point in the output. Defaults to `true`
```

nx/lib/nx.ex
```diff
@@ -12084,6 +12084,86 @@ defmodule Nx do
     apply(impl!(tensor), kind, [out, tensor, opts])
   end
 
+  @doc """
+  Creates a tensor of shape `{n}` with linearly spaced samples between `start` and `stop`
```
josevalim, line 12088:
```suggestion
  Creates a tensor of shape `{n}` with linearly spaced samples between `start` and `stop`.
```

nx/lib/nx.ex
```diff
@@ -12084,6 +12084,86 @@ defmodule Nx do
...
+  ### Error Cases
+
+      iex> Nx.linspace(0, 24, n: nil)
+      ** (ArgumentError) expected n to be a non-negative integer, got: nil
+
+      iex> Nx.linspace(0, 24, n: -1)
+      ** (ArgumentError) expected n to be a non-negative integer, got: -1
+
+      iex> Nx.linspace(0, 24, n: 1.0)
+      ** (ArgumentError) expected n to be a non-negative integer, got: 1.0
```
josevalim, line 12140: One is enough for examples:

```suggestion
      iex> Nx.linspace(0, 24, n: -1)
      ** (ArgumentError) expected n to be a non-negative integer, got: -1
```

polvalente: I'd rather keep `1.0` to highlight we only accept plain integers. Will commit separately

josevalim (commented):
I dropped some minor comments.

It looks good to me my only question is about `:axis_name`. We don't have a similarly named option anywhere. So I am wondering if we should accept `:names` and force it to be a single-element list or do something else.

Perhaps it is easier to simply remove it for now and someone can call `rename` later if necessary.

nx/lib/nx.ex
```diff
@@ -12084,6 +12084,80 @@ defmodule Nx do
...
 
+  @doc """
+  Creates a tensor of shape `{n}` with linearly spaced samples between `start` and `stop`.
+
+  ## Options
+
+    * `:n` - The number of samples in the tensor.
+    * `:axis_name` - Optional name for the output axis.
+    * `:type` - Optional type for the output. Defaults to `{:f, 32}`
+    * `:endpoint` - Boolean that indicates whether to include `stop`
```
seanmor5, line 12095: Maybe this should be "inclusive" or "include_stop"? Endpoint isn't 100% clear imo

josevalim: I like inclusive although I think numpy uses endpoint.

polvalente: What about include_endpoint?

josevalim (approved):

PR #1066, opened 2023-01-17
docs: set linspace type as creation

polvalente:


josevalim:
Not sure if creation as it works with any tensor of multiple dimensions. Perhaps it is more about shape or adding a new dimension that follows the linspace of the other ones.

PR #1067, opened 2023-01-18
Fix relative EXLA symlinks

jonatanklosko:
See https://github.com/elixir-nx/nx/pull/1062#issuecomment-1386171551.

### DIFF
--- a/exla/mix.exs
+++ b/exla/mix.exs
@@ -27,13 +27,8 @@ defmodule EXLA.MixProject do
         "compile.cached_make": &cached_make/1
       ],
       make_env: fn ->
-        cwd_relative_to_priv =
-          Mix.Project.app_path()
-          |> Path.join("priv")
-          |> Path.relative_to_cwd()
-          |> Path.split()
-          |> Enum.map(fn _ -> ".." end)
-          |> Path.join()
+        priv_path = Path.join(Mix.Project.app_path(), "priv")
+        cwd_relative_to_priv = relative_to(File.cwd!(), priv_path)
 
         %{
           "MIX_BUILD_EMBEDDED" => "#{Mix.Project.config()[:build_embedded]}",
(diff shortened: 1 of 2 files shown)


exla/mix.exs
```diff
@@ -26,9 +26,20 @@ defmodule EXLA.MixProject do
...
         "compile.cached_make": &cached_make/1
       ],
-      make_env: %{
-        "MIX_BUILD_EMBEDDED" => "#{Mix.Project.config()[:build_embedded]}"
-      }
+      make_env: fn ->
+        cwd_relative_to_priv =
+          Mix.Project.app_path()
+          |> Path.join("priv")
+          |> Path.relative_to_cwd()
```
jonatanklosko, line 33: This assumes the priv dir is within the current working directory, but it should always be the case, since it's always in `_build`?

josevalim: You can have the build anywhere by using the `build_path` mix.exs configuration :(

polvalente: Perhaps `cd`-ing into `build_path` before calculating `Path.relative_to_cwd` can resolve this?

josevalim: We need to compute the relative path to the deps folder, which can be anywhere in the system, and the build folder, which can also be anywhere. Perhaps what we have to do is to call realpath and, depending if it return status 0 or 1, we act accordingly.

jonatanklosko: I pushed an alternative version that manually computes the relative path. Let me know if that works, otherwise using `realpath` with status check sounds fine.

exla/mix.exs
```diff
@@ -167,4 +173,16 @@ defmodule EXLA.MixProject do
       :filename.basedir(:user_cache, "xla")
     end
   end
+
+  # Returns `path` relative to the `from` directory.
+  defp relative_to(path, from) do
```
jonatanklosko, line 178: Note that this doesn't consider symlinks, but we don't have a top-level priv in exla that mix would symlink to, so this should be fine (?)

josevalim (approved):

PR #1068, opened 2023-01-18
TorchX - Use symbolic link to library if ENV is not set

thiagopromano:
Uses the same idea as in #1067, but doesn't create symlink if the Torch location was set in the `LIBTORCH_DIR` env.


josevalim (approved):

josevalim:

PR #1071, opened 2023-01-22
feat: matrix_rank

kenken-neko:
# Changes
The `matrix_rank` function is added that has the same function as [numpy.linalg.matrix_rank](https://numpy.org/doc/stable/reference/generated/numpy.linalg.matrix_rank.html).
This function returns matrix rank of input M × N matrix using Singular Value Decomposition (SVD) method.

# References

- [Document of `numpy.linalg.matrix_rank`](https://numpy.org/doc/stable/reference/generated/numpy.linalg.matrix_rank.html)
- Mathematical resources, for example,[ this document from Berkeley University (Proposition 1.3.)](https://math.berkeley.edu/~hutching/teach/54-2017/svd-notes.pdf) 

# Confirmations
I ran the same test on `numpy.linalg.matrix_rank` as doctest and confirmed that the same results were obtained.
numpy_tests

### DIFF
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -1806,6 +1806,7 @@ defmodule Nx.LinAlg do
     opts = keyword!(opts, eps: 1.0e-10)
     shape = Nx.shape(a)
     size = Nx.rank(shape)
+
     if size <= 1 do
       raise(
         ArgumentError,
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -1763,9 +1763,14 @@ defmodule Nx.LinAlg do
   @doc """
   Return matrix rank of input M × N matrix using Singular Value Decomposition method.
 
-  Approximate the number of linearly independent rows by calculating the number of singular values greater than `eps * max(singular values) * max(M, N)`.
-  This also appears in Numerical recipes in the discussion of SVD solutions for linear least squares [1].
-  [1] W. H. Press, S. A. Teukolsky, W. T. Vetterling and B. P. Flannery, “Numerical Recipes (3rd edition)”, Cambridge University Press, 2007, page 795.
+  Approximate the number of linearly independent rows by calculating the number
+  of singular values greater than `eps * max(singular values) * max(M, N)`.
+
+  This also appears in Numerical recipes in the discussion of SVD solutions for
+  linear least squares [1].
+
+  [1] W. H. Press, S. A. Teukolsky, W. T. Vetterling and B. P. Flannery,
+  “Numerical Recipes (3rd edition)”, Cambridge University Press, 2007, page 795.
 
   ## Options
     * `:eps` - Rounding error threshold used to assume values as 0. Defaults to `1.0e-10`
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -1759,4 +1759,63 @@ defmodule Nx.LinAlg do
     |> Nx.take_diagonal(offset: offset)
     |> Nx.product(axes: [rank - 2])
   end
+
+  @doc """
+  Return matrix rank of input M × N matrix using Singular Value Decomposition method.
+  Calculate the number of singular values greater than `eps * max(singular values) * max(M, N)` as a rank.
+  This also appears in Numerical recipes in the discussion of SVD solutions for linear least squares [1].
+  [1] W. H. Press, S. A. Teukolsky, W. T. Vetterling and B. P. Flannery, “Numerical Recipes (3rd edition)”, Cambridge University Press, 2007, page 795.
+
+  ## Options
+    * `:eps` - Rounding error threshold used to assume values as 0. Defaults to `1.0e-10`
+
+  ## Examples
+
+      iex> Nx.LinAlg.matrix_rank(Nx.tensor([[1, 2], [3, 4]]))
+      #Nx.Tensor<
+        u64
+        2
+      >
+
+      iex> Nx.LinAlg.matrix_rank(Nx.tensor([[1, 1, 1, 1], [1, 1, 1, 1], [1, 2, 3, 4]]))
+      #Nx.Tensor<
+        u64
+        2
+      >
+
+      iex> Nx.LinAlg.matrix_rank(Nx.tensor([[1, 1, 1], [2, 2, 2], [8, 9, 10], [-2, 1, 5]]))
+      #Nx.Tensor<
+        u64
+        3
+      >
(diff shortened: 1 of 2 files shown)


kenken-neko:
I could not find the reason for the error in `Test / Windows` on Github actions. Please let me know when someone has time.

nx/lib/nx/lin_alg.ex
```diff
@@ -1759,4 +1759,65 @@ defmodule Nx.LinAlg do
...
+        3
+      >
+
+  ## Error cases
+
+      iex> Nx.LinAlg.matrix_rank(Nx.tensor([1, 2, 3]))
+      ** (ArgumentError) tensor must have at least rank 2, got rank 1 with shape {3}
+  """
+  @doc from_backend: false
+  def matrix_rank(a, opts \\ []) do
```
polvalente, line 1798:
```suggestion
  def matrix_rank(a, opts \\ []) do
    # TODO: support batchikg when SVD supports it too
```

nx/lib/nx/lin_alg.ex
```diff
@@ -1759,4 +1759,65 @@ defmodule Nx.LinAlg do
...
+
+    s_max =
+      s
+      |> Nx.reduce_max()
+      |> Nx.to_number()
+
+    # Set tolerance values
+    tol = opts[:eps] * max_dim * s_max
+
+    # Calucate matrix rank
```
polvalente, line 1818:
```suggestion
    # Calculate matrix rank
```

nx/lib/nx/lin_alg.ex
```diff
@@ -1759,4 +1759,65 @@ defmodule Nx.LinAlg do
     |> Nx.take_diagonal(offset: offset)
     |> Nx.product(axes: [rank - 2])
   end
+
+  @doc """
+  Return matrix rank of input M × N matrix using Singular Value Decomposition method.
```
polvalente, line 1764:
```suggestion
  Return matrix rank of input M × N matrix using Singular Value Decomposition method.

```

nx/lib/nx/lin_alg.ex
```diff
@@ -1759,4 +1759,65 @@ defmodule Nx.LinAlg do
...
+  Calculate the number of singular values greater than `eps * max(singular values) * max(M, N)` as a rank.
```
polvalente, line 1765:
```suggestion
  Approximate the number of linearly independent rows by calculating the number of singular values greater than `eps * max(singular values) * max(M, N)`.
```

polvalente (approved):

nx/lib/nx/lin_alg.ex
```diff
@@ -1759,4 +1759,67 @@ defmodule Nx.LinAlg do
     |> Nx.take_diagonal(offset: offset)
     |> Nx.product(axes: [rank - 2])
   end
+
+  @doc """
+  Return matrix rank of input M × N matrix using Singular Value Decomposition method.
+
+  Approximate the number of linearly independent rows by calculating the number of singular values greater than `eps * max(singular values) * max(M, N)`.
+  This also appears in Numerical recipes in the discussion of SVD solutions for linear least squares [1].
+  [1] W. H. Press, S. A. Teukolsky, W. T. Vetterling and B. P. Flannery, “Numerical Recipes (3rd edition)”, Cambridge University Press, 2007, page 795.
```
josevalim, line 1768:
```suggestion
  Approximate the number of linearly independent rows by calculating the number
  of singular values greater than `eps * max(singular values) * max(M, N)`.

  This also appears in Numerical recipes in the discussion of SVD solutions for
  linear least squares [1].

  [1] W. H. Press, S. A. Teukolsky, W. T. Vetterling and B. P. Flannery, “Numerical Recipes (3rd edition)”, Cambridge University Press, 2007, page 795.
```

josevalim (approved):

nx/lib/nx/lin_alg.ex
```diff
@@ -1759,4 +1759,72 @@ defmodule Nx.LinAlg do
...
+      s
+      |> Nx.reduce_max()
+      |> Nx.to_number()
+
+    # Set tolerance values
+    tol = opts[:eps] * max_dim * s_max
+
+    # Calculate matrix rank
+    s
+    |> Nx.greater(tol)
```
polvalente, line 1827:
```suggestion
    s_max = Nx.reduce_max(s)

    # Set tolerance values
    tol = Nx.multiply(opts[:eps] * max_dim, s_max)

    # Calculate matrix rank
    s
    |> Nx.greater(tol)
```

nx/lib/nx/lin_alg.ex
```diff
@@ -1759,4 +1759,72 @@ defmodule Nx.LinAlg do
...
+        3
+      >
+
+  ## Error cases
+
+      iex> Nx.LinAlg.matrix_rank(Nx.tensor([1, 2, 3]))
+      ** (ArgumentError) tensor must have at least rank 2, got rank 1 with shape {3}
+  """
+  @doc from_backend: false
+  def matrix_rank(a, opts \\ []) do
```
polvalente, line 1804:
```suggestion
  defn matrix_rank(a, opts \\ []) do
```

Can we use defn here?

nx/lib/nx/lin_alg.ex
msluszniak: Why we should use `Nx.multiply` instead of * here?

polvalente: If we can convert to defn we shouldn't :)

But as it stands, we need to remove the Nx.to_number call and this in turn necessitates Nx.multiply instead of *

nx/lib/nx/lin_alg.ex
kenken-neko: @polvalente 
Since we use function `svd/1` in function `matrix_rank`, we cannot use `defn` for function `matrix_rank`.
https://github.com/elixir-nx/nx/actions/runs/3982399082/jobs/6826831892

polvalente: What happens if you use the fully qualified Nx.LinAlg.svd instead of calling it locally?

kenken-neko: Thanks for the advice, I did successfully call `Nx.LinAlg.svd` in defn `matrix_rank`.
I also fixed `ArgumentError` in defn instead of `Nx.Shape.matrix_rank` because `Nx.Shape.matrix_rank` cannot be called in defn.  [3fafb9b](https://github.com/elixir-nx/nx/pull/1071/commits/3fafb9b657244e868d594bc4934c3d6f11bfd55d)

josevalim (approved):

PR #1072, opened 2023-01-23
feat: add Nx.reflect

polvalente:
Based on the `reflect` mode for `numpy.pad`

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -12282,14 +12282,14 @@ defmodule Nx do
     end)
   end
 
-  def left_reflect_index_period(n) do
+  defp left_reflect_index_period(n) do
     # Generates the indices for pre-reflecting on the axis
     left = Nx.iota({n - 1}) |> Nx.add(1)
     right = Nx.subtract(n - 2, Nx.iota({n - 1}))
     Nx.concatenate([left, right])
   end
 
-  def right_reflect_index_period(n) do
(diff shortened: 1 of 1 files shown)


nx/lib/nx.ex
```diff
@@ -12196,6 +12196,106 @@ defmodule Nx do
...
   end
 
+  @doc """
+  Pads a tensor along the given axes through periodic reflections.
+
+  ## Options
+
+    * `:padding_config` - A list of tuples in the format `{pre, post}`,
+    which specify the length (0 or greater) of the reflection before and
+    after the tensor along a each axis.
```
josevalim, line 12206:
```suggestion
      which specify the length (0 or greater) of the reflection before and
      after the tensor along a each axis.
```

nx/lib/nx.ex
```diff
@@ -12196,6 +12196,106 @@ defmodule Nx do
...
+
+    axes = axes(tensor)
+
+    if length(axes) != length(padding_config) do
+      raise ArgumentError, "expected to have one padding_config entry each tensor axis"
+    end
+
+    padding_config
+    |> Enum.zip(axes)
+    |> Enum.reduce(tensor, fn {{left_padding, right_padding}, axis}, tensor ->
```
josevalim, line 12246: Use the new zip_reduce!

josevalim (approved):

PR #1073, opened 2023-01-24
Add full_matrices? option to svd

msluszniak:
As mentioned in [#73](https://github.com/elixir-nx/scholar/pull/73) I add `full_matrices?` option.

### DIFF
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -1178,6 +1178,10 @@ defmodule Nx.LinAlg do
     * `:max_iter` - `integer`. Defaults to `100`
       Number of maximum iterations before stopping the decomposition
 
+    * `:full_matrices?` - `boolean`. Defaults to `true`
+      If True, u and vt are of shape (M, M), (N, N). Otherwise,
+      the shapes are (M, K) and (K, N), where K = min(M, N).
+
   Note not all options apply to all backends, as backends may have
   specific optimizations that render these mechanisms unnecessary.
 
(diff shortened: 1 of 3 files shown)


polvalente:
@msluszniak https://github.com/google/jax/blob/main/jax/_src/lax/svd.py

You can use this as a reference for what to change to support this new option

nx/lib/nx/lin_alg.ex
```diff
@@ -1178,6 +1178,10 @@ defmodule Nx.LinAlg do
     * `:max_iter` - `integer`. Defaults to `100`
       Number of maximum iterations before stopping the decomposition
 
+    * `:full_matrices?` - `boolean`. Defaults to `true`
+      If True, u and vt are of shape (M, M), (N, N). Otherwise,
```
polvalente, line 1182:
```suggestion
      If `true`, `u` and `vt` are of shape (M, M), (N, N). Otherwise,
```

polvalente: I would actually default this to `false` if we can, because it generates less output memory footprint and less calculations overall.
@josevalim this would be a breaking change. Thoughts?

nx/lib/nx/shape.ex
```diff
@@ -1825,22 +1825,32 @@ defmodule Nx.Shape do
...
+      [
+        [unchanged_shape, [m, m]],
+        [unchanged_shape, [min(m, n)]],
+        [unchanged_shape, [n, n]]
+      ]
+    else
+      [
+        [unchanged_shape, [m, min(m, n)]],
+        [unchanged_shape, [min(m, n)]],
+        [unchanged_shape, [min(m, n), n]]
```
polvalente, line 1845: Maybe extract `k` as its own variable before the `if`?

nx/lib/nx/lin_alg/svd.ex
```diff
@@ -86,18 +76,62 @@ defmodule Nx.LinAlg.SVD do
...
+
+    {u, s, v}
+  end
+
+  defn svd_non_full(tensor, opts \\ []) do
+    {m, n} = Nx.shape(tensor)
+
+    # The constant `1.15` comes from Yuji Nakatsukasa's implementation
+    # https://www.mathworks.com/matlabcentral/fileexchange/36830-symmetric-eigenvalue-decomposition-and-the-svd?s_tid=FX_rc3_behav
+    {reduced_to_square, q, a} =
```
polvalente, line 94:
```suggestion
    {reduce_to_square, q, a} =
```

nitpick :p

nx/lib/nx/lin_alg/svd.ex
```diff
@@ -86,18 +76,62 @@ defmodule Nx.LinAlg.SVD do
...
+      else
+        {false, tensor, tensor}
+      end
 
-        false ->
-          u
+    {u, s, v} = svd_tall_and_square(a, opts)
+
+    u =
+      if reduced_to_square do
```
polvalente, line 105:
```suggestion
      if reduce_to_square do
```

nx/lib/nx/lin_alg.ex
josevalim: The default for numpy is `true`, so I would consider mirroring that (and nothing breaks).

polvalente (approved):

PR #1077, opened 2023-01-25
feat: add Nx.Random.shuffle

polvalente:
Based on jax.random.permutation, so I can implement jax.random.choice aftewards

### DIFF
--- a/nx/lib/nx/random.ex
+++ b/nx/lib/nx/random.ex
@@ -526,6 +526,80 @@ defmodule Nx.Random do
     end)
   end
 
+  @doc """
+  Randomly shuffles a tensor along an axis
+
+  ## Options
+
+    * `:axis` - the axis along which to shuffle. Defaults to `0`
+    * `:independent` - a boolean that indicates wether the permutations
+      are independent along the given axis. Defaults to `false`
+
+  ## Examples
+
+      iex> key = Nx.Random.key(42)
+      iex> {shuffled, _new_key} = Nx.Random.permutation(key, Nx.iota({3, 4}, axis: 0))
+      iex> shuffled
+      #Nx.Tensor<
+        s64[3][4]
+        [
+          [2, 2, 2, 2],
+          [0, 0, 0, 0],
+          [1, 1, 1, 1]
+        ]
+      >
+
+      iex> key = Nx.Random.key(10)
+      iex> {shuffled, _new_key} = Nx.Random.permutation(key, Nx.iota({3, 4}, axis: 1), independent: true, axis: 1)
+      iex> shuffled
+      #Nx.Tensor<
+        s64[3][4]
+        [
+          [2, 1, 3, 0],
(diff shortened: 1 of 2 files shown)


jonatanklosko:
@polvalente we do have `Nx.shuffle`, so we need to standardize somehow.

polvalente:
@jonatanklosko I'm ok with exposing shuffle here instead of permutation. I just followed the name that jax uses

nx/lib/nx/random.ex
```diff
@@ -526,6 +526,80 @@ defmodule Nx.Random do
...
 
+  @doc """
+  Randomly shuffles a tensor along an axis
+
+  ## Options
+
+    * `:axis` - the axis along which to shuffle. Defaults to `0`
+    * `:independent` - a boolean that indicates wether the permutations
+      are independent along the given axis. Defaults to `false`
+
```
jonatanklosko, line 537:
```suggestion
    * `:axis` - the axis along which to shuffle. Defaults to `0`

    * `:independent` - a boolean that indicates wether the permutations
      are independent along the given axis. Defaults to `false`

```

jonatanklosko:
Yeah, I meant more that we should deprecate `Nx.shuffle`. Also note that id behaves differently when no axis is given.

josevalim:
Good call. We _must_ deprecate shuffle, because it uses random_uniform. I will add a deprecation warning shortly.

josevalim:
Actually, I can't deprecate it because we need this PR :P @polvalente can you add

```
@doc type: :creation, deprecated: "Use Nx.Random.shuffle/2 instead"
```

to the top of shuffle pls?

nx/lib/nx/random.ex
```diff
@@ -526,6 +526,81 @@ defmodule Nx.Random do
     end)
   end
 
+  @doc """
+  Randomly shuffles a tensor along an axis
```
jonatanklosko, line 530:
```suggestion
  Randomly shuffles a tensor elements along an axis.
```

nx/lib/nx/random.ex
```diff
@@ -526,6 +526,81 @@ defmodule Nx.Random do
...
+      >
+  """
+  defn shuffle(key, tensor, opts \\ []) do
+    opts = keyword!(opts, axis: 0, independent: false)
+    axis = opts[:axis]
+
+    if opts[:independent] do
+      shuffle_independent(key, tensor, axis: axis)
+    else
+      {idx, key} = shuffle_independent(key, Nx.iota({Nx.axis_size(tensor, axis)}), axis: axis)
```
jonatanklosko, line 572:
```suggestion
      {idx, key} = shuffle_independent(key, Nx.iota({Nx.axis_size(tensor, axis)}), axis: 0)
```

polvalente: gotta add a test for that

jonatanklosko (approved):

PR #1078, opened 2023-01-26
Fix attach_token in conditionals example

smaximov:


josevalim:

PR #1082, opened 2023-01-26
feat: Nx.Random.choice

polvalente:


nx/lib/nx/constants.ex
```diff
@@ -292,6 +292,19 @@ defmodule Nx.Constants do
...
 
+  @doc """
+  Returns the smallest normal floating point number.
+
+  The smallest normal floating point number is the smallest positive value with
+  1 as the leading bit for the mantissa.
+
+  Accepts the same options as `Nx.tensor/2`
+  """
+  def smallest_normal(type, opts \\ []) do
```
polvalente, line 303: this is the name Jax uses, but I'm not sure if we should use `min_normal` to keep with `min_finite`

nx/lib/nx/random.ex
```diff
@@ -601,6 +601,245 @@ defmodule Nx.Random do
...
+      g = -g - Nx.log(p)
+      idx = g |> Nx.argsort() |> Nx.slice_along_axis(0, n_draws, axis: 0)
+
+      result = Nx.take(tensor, idx, axis: axis)
+      {result, k}
+    end
+  end
+
+  @doc """
+  Sample Gumbel random values with given shape and float dtype.
```
polvalente, line 815: Should I expose this function, or should we keep it hidden until someone actually needs it?

If it should be exposed, I need to add doctests

nx/lib/nx/random.ex
```diff
@@ -601,6 +601,245 @@ defmodule Nx.Random do
...
+
+    {u, k} = uniform(key, Nx.Constants.smallest_normal(type), 1, shape: shape, type: type)
+    result = -Nx.log(-Nx.log(u))
+
+    {result, k}
+  end
+
+  #   _check_shape("gumbel", shape)
+  #   return -jnp.log(-jnp.log(
+  #       uniform(key, shape, dtype, minval=jnp.finfo(dtype).tiny, maxval=1.)))
```
polvalente, line 841:
```suggestion
```

nx/lib/nx/type.ex
```diff
@@ -590,6 +590,30 @@ defmodule Nx.Type do
...
+  def smallest_normal({:bf, 16}) do
+    f32 = smallest_normal({:f, 32})
+
+    if System.endianness() == :little do
+      <<_::16, x::16-bitstring>> = f32
+      x
+    else
+      <<x::16-bitstring, _::16>> = f32
+      x
+    end
```
polvalente, line 608: there must be a better way to do this :p

nx/lib/nx/constants.ex
josevalim: I think we had this in the past as `min_positive` or something like that and we removed it because it was no longer used.

polvalente: https://github.com/elixir-nx/nx/commit/fc9a2e7148a1d54e35fdd6edb38197c924d3e1d0

found it!

Shall I restore it instead?

nx/lib/nx/random.ex
```diff
@@ -601,6 +601,242 @@ defmodule Nx.Random do
...
+      result = Nx.take(tensor, idx, axis: axis)
+      {result, key}
+    else
+      {shuffled, key} = shuffle(key, tensor, axis: axis)
+      result = Nx.slice_along_axis(shuffled, 0, n_draws, axis: axis)
+      {result, key}
+    end
+  end
+
+  defnp choice_probabilities(key, tensor, p, opts \\ []) do
```
josevalim, line 781: It seems options are always required, so I think you should do this:

```suggestion
  defn choice(key, tensor, p, opts) do
```

nx/lib/nx/random.ex
```diff
@@ -601,6 +601,242 @@ defmodule Nx.Random do
...
+      >
+      iex> {result, _key} = Nx.Random.choice(k, t, p, samples: 6, replace: false) # without replacement
+      iex> result
+      #Nx.Tensor<
+        s64[6]
+        [3, 1, 2, 5, 4, 0]
+      >
+  """
+  deftransform choice(key, tensor, p), do: choice_probabilities(key, tensor, p)
+  deftransform choice(key, tensor, p, opts), do: choice_probabilities(key, tensor, p, opts)
```
josevalim, line 727:
```suggestion
  deftransform choice(key, tensor, p, opts) when is_list(opts), do: choice_probabilities(key, tensor, p, opts)
```

Options seem required...

nx/lib/nx/random.ex
```diff
@@ -601,6 +601,242 @@ defmodule Nx.Random do
...
+        [5, 0, 0, 4, 0, 3]
+      >
+      iex> {result, _key} = Nx.Random.choice(k, t, samples: 6, replace: false) # without replacement
+      iex> result
+      #Nx.Tensor<
+        s64[6]
+        [2, 0, 4, 5, 1, 3]
+      >
+  """
+  deftransform choice(key, tensor), do: choice_uniform(key, tensor, [])
```
josevalim, line 664:
```suggestion
```

nx/lib/nx/random.ex
josevalim: Sounds fine by me. Should we also add `_split` versions to all of these functions?

nx/lib/nx/type.ex
josevalim: https://github.com/elixir-nx/nx/commit/fc9a2e7148a1d54e35fdd6edb38197c924d3e1d0

nx/lib/nx/constants.ex
josevalim: Whatever you prefer!

nx/lib/nx/random.ex
polvalente: What does `_split` do?

nx/lib/nx/random.ex
polvalente: indeed they are :)
probabilities are not, so I'll still keep `choice_probabilities/choice_uniform` as separate `defnp`s

polvalente: No need for this actually :)

nx/lib/nx/random.ex
polvalente: @josevalim Actually,  this depends on `shuffle`,  which `split`s the key underneath a few times. Does it still make sense to have `choice_split`?

josevalim: Ah, good point. Try to have a `_split` if we can, but if you cannot, no problem!

polvalente: I don't see a point of having `choice_split` because of `shuffle`, but we can have `gumbel_split`

nx/lib/nx/constants.ex
```diff
@@ -292,6 +292,46 @@ defmodule Nx.Constants do
     Nx.tensor(Complex.new(0, 1), opts)
   end
 
+  @doc """
+  Returns a scalar tensor with the maximum finite value for the given type.
+
+  ## Options
```
polvalente, line 298:
```suggestion
  ## Options
  
```

josevalim (approved):

Issue #1051, opened 2023-01-04
Better documentation around aggregations functions


ndrean:
It would be nice to have the formulas used by the functions or a reference to the ones used as most functions don't have. Some sublibraries like Element-wise have explicit formulas, but none in the Aggregate module for example. For example, the aggregation option is mysterious whether in `Nx.mean` or `Nx.argmax` for example, and the examples alone are not enough to understand.

For example, running examples is not enough to understand what the function is doing.
```

t = Nx.tensor(
  [
    [
      [4, 2, 3], [6, -5, 3]
    ], 
    [
      [3, 1, 2], [4, -10, 3]
    ],
    [
      [4,2,4], [3,3,3]
    ]
  ],
  names: [:x, :y, :z]
)
Nx.argmax(t, axis: :y)

#Nx.Tensor<
  s64[x: 3][z: 3]
  [
    [1, 0, 0],
    [1, 0, 1],
    [0, 1, 0]
 ]

Nx.argmax(t, axis: :x)
#Nx.Tensor<
  s64[y: 2][z: 3]
  [
    [0, 0, 2],
    [0, 2, 0]
  ]

Nx.argmax(t, axis: y)
#Nx.Tensor<
  s64[x: 3][z: 3]
  [
    [1, 0, 0],
    [1, 0, 1],
    [0, 1, 0]
 ]
```

josevalim:
I am not sure because [`argmax` in Numpy does not have any formulas either](https://numpy.org/doc/stable/reference/generated/numpy.argmax.html) and therefore it is unclear if there is even a readily available formula that we could use.

polvalente:
I don't think mathematical notation for axes aggregation would make things any more clear than the examples.
This is especially so because we can aggregate on any axis or set of axes and don't have a lower nor upper bound on how many axes there are, so I don't see how we can have a clear LaTeX notation for that.

We could instead consider a more "written out" explanation (PRs welcome for consideration/discussion!) for whatever feels confusing.

ndrean:
My remark was general. Most if not all aggregate functions have no reference formula (mean, variance, etc) and I believe a formula is still welcome because we don't know what it does, even if it may seem obvious. But think of statiscal formulas with biais. A formula can be shorter than a literal example in such a case, but both are doable.

[...]

PR #1087, opened 2023-01-31
#1051 Better documentation around aggregations functions

ndrean:
#1051

action: 
added under "nx/guides/" a livebook "modoc_agg_nx_01.livemd

### DIFF
--- a/nx/guides/modoc_agg_nx_01.livemd
+++ b/nx/guides/modoc_agg_nx_01.livemd
@@ -1,0 +1,826 @@
+# Livebook companion for Nx
+
+```elixir
+Mix.install([
+  {:nx, "~> 0.4.2"},
+  {:kino, "~> 0.8"}
+])
+
+# import IEx.Helpers
+```
+
+## Aggregations operations on tensors
+
+
+
+```elixir
+import Nx, only: :sigils
+```
+
+A tensor _in computer science_ can be viewed as a list of multi-indexed numbers. You simply wrap your list with `Nx.tensor`. Equivalently, you can use the compact `~V` sigil:
+
+```elixir
+{
+  Nx.tensor([1, 2, 3]),
+  ~V[1 2 3]
+}
+```
+
+You can aggregation functions on any tensor. This means you apply a function to the elements of your tensor. The function can be applied to the elements as a whole, or by "axis".
+
+Let's take a first example with the `Nx.all_close` full-tensor aggregate.  For each element of your tensor, you make a comparison against a given tensor and evaluate how "close" globally these tensors are.
+
+As an application, we use a tensor of rank 1 (a list) and compare the calculus done by hand with the aggregate `Nx.variance` (which is a tensor of rank 0).
+
+```elixir
+l = ~V[0 1 2]
+
+mean = (0 + 1 + 2) / 3
+variance = ((0 - mean) ** 2 + (1 - mean) ** 2 + (2 - mean) ** 2) / 3
+
+nx_variance = Nx.variance(l)
+
+%{
+  hand_variance: variance,
+  nx_variance: Nx.to_number(nx_variance),
+  variance_result_rank: Nx.rank(nx_variance),
+  _nx_close_agg: Nx.all_close(Nx.tensor(variance, type: :f16), nx_variance)
+}
+```
+
+Let's take an example of the aggregate `Nx.weighted_mean` that supports full-tensor and per axis operations. Let's take an example with a tensor of rank 2; these are vectors, covectors and 2D matrices. Let's check how the weighted mean aggregate of a matrix is computed with a 2D tensor of shape`{2,2}` (aka as "square matrix").
+
+```elixir
+m = ~V[1 2 3 4] |> Nx.reshape({2, 2})
+```
+
+We run the full -tensor aggregate, do the same calculations by hand and finally compare them:
+
+```elixir
+full_weights = ~V[4 5 4 5] |> Nx.reshape({2, 2})
+nx_w_avg = Nx.weighted_mean(m, full_weights)
+
+man_w_avg = (1 * 4 + 2 * 5 + 3 * 4 + 4 * 5) / (4 + 5 + 4 + 5)
+
+%{
+  nx_weighted_avg: Nx.to_number(nx_w_avg),
+  manual_weighted_avg: man_w_avg,
+  _nx_close: Nx.all_close(nx_w_avg, Nx.tensor(man_w_avg))
+}
+```
+
+The weighted mean can be run per axis. Let's take an example of the same squared matrix along the **first** axis (`axes: [0]`) with the same weights per axis: you calculate "by column", so aggregate all elements with the second remaining index 1 constant, or equivalently "reduce" along the selected axis (the first index):
+
+```elixir
+axes_weights = ~V[4 5]
+w_avg_x = Nx.weighted_mean(m, axes_weights, axes: [0])
+
+man_w_avg_x = [(1 * 4 + 3 * 5) / (4 + 5), (2 * 4 + 4 * 5) / (4 + 5)]
+
+{
+  w_avg_x,
+  man_w_avg_x,
+  Nx.all_close(w_avg_x, Nx.tensor(man_w_avg_x))
+}
+```
+
+We calculate weighted mean of a square matrix along the **second** axis with a set of different weights per axis: you calculate per row, so you aggregate all elements the the same remaining. index fiexd (the first index), equivalently reduce along the second axis.
+
+```elixir
+x0_weights = ~V[4 5]
+x1_weights = ~V[5 4]
+weights = Nx.concatenate([x0_weights, x1_weights]) |> Nx.reshape({2, 2})
+
+nx_w_avg_y = Nx.weighted_mean(m, weights, axes: [1])
+
+man_w_avg_y = [(1 * 4 + 2 * 5) / (4 + 5), (3 * 5 + 4 * 4) / (4 + 5)]
+
+{
+  nx_w_avg_y,
+  man_w_avg_y,
+  Nx.all_close(nx_w_avg_y, Nx.tensor(man_w_avg_y))
+}
+```
+
+#### Example with higher rank
+
+The example below will be used through the documentation.
+
+Take a list of numbers of length $n=36$, and turn it into a tensor with the shape `{2,3,2,3}`; this is a computational tensor of rank 4. We will _name_ the axes, with `names: [:x, :y, :z, :t]`.
+
+```elixir
+t =
+  ~V[1 2 3 1 -2 3 4 -3 2 -4 3 2 5 -1 3 -5 1 3 4 6 1 4 -6 1 1 2 3 1 -2 3 4 -3 2 -4 3 2]
+  |> Nx.reshape({2, 3, 2, 3}, names: [:x, :y, :z, :t])
+```
+
+We can view this list as a succession of slices into rows given by the shape of the tensor. With the shape `{2,3,2,4}`, you obtain slices of length $2$,then $3$, then $2$ and the last index of length $4$ is conventionally represented as a row. The picture below will help to understand the aggregations.
+
+$$
+\begin{bmatrix} x=0
+  \begin{bmatrix} y=0
+
+\begin{bmatrix}
+  a_{0,0,0,0} & a_{0,0,0,1} & a_{0,0,0,2} \\
+  a_{0,0,1,0} & a_{0,0,1,1} & a_{0,0,1,2} \\
+\end{bmatrix}
+\\ 
+y=1
+\begin{bmatrix} 
+  a_{0,1,0,0} & a_{0,1,0,1} & a_{0,1,0,2} \\
+  a_{0,1,1,0} & a_{0,1,1,1} & a_{0,1,1,2} \\
+\end{bmatrix}
+\\ 
+y=2
+\begin{bmatrix}
+  a_{0,2,0,0} & a_{0,2,0,1} & a_{0,2,0,2} \\
+  a_{0,2,1,0} & a_{0,2,1,1} & a_{0,2,1,2} \\
+\end{bmatrix}
+
+\end{bmatrix}
+  \\
+
+x=1
+
+\begin{bmatrix} y=0
+
+\begin{bmatrix}
+  a_{1,0,0,0} & a_{1,0,0,1} & a_{1,0,0,2}  \\
+  a_{1,0,1,0} & a_{1,0,1,1} & a_{1,0,1,2}  \\
+\end{bmatrix}
+\\ 
+y=1
+\begin{bmatrix}
+  a_{1,1,0,0} & a_{1,1,0,1} & a_{1,1,0,2}  \\
+  a_{1,1,1,0} & a_{1,1,1,1} & a_{1,1,1,2}  \\
+\end{bmatrix}
+\\ 
+y=2
+\begin{bmatrix}
+  a_{1,2,0,0} & a_{1,2,0,1} & a_{1,2,0,2}  \\
+  a_{1,2,1,0} & a_{1,2,1,1} & a_{1,2,1,2}  \\    
+\end{bmatrix}
+
+\end{bmatrix}
+\end{bmatrix}
+
+
+
+\begin{bmatrix} x=0
+  \begin{bmatrix} y=0
+
+```
+\begin{bmatrix}
+  & 1, 2, 3, & \\
+  & 1, -2, 3, &
+\end{bmatrix}
+\\ 
+y=1
+\begin{bmatrix} 
+  & 4, -3, 2 & \\
+  & -4, 3, 2 &
+\end{bmatrix}
+\\ 
+y=2
+\begin{bmatrix}
+  & 5, -1, 3 &\\
+  & -5, 1, 3 &
+\end{bmatrix}
+```
+
+\end{bmatrix}
+  \\
+  x=1
+  \begin{bmatrix} y=0
+
+```
+\begin{bmatrix}
+  & 4, 6, 1 & \\
+  & 4, -6, 1 &
+\end{bmatrix}
+\\ 
+y=1
+\begin{bmatrix}
+  & 1, 2, 3 & \\
+  & 1, -2, 3 &
+\end{bmatrix}
+\\ 
+y=2
+\begin{bmatrix}
+  & 4, -3, 2 & \\
+  & -4, 3, 2 &
+\end{bmatrix}
+```
+
+\end{bmatrix}
+\end{bmatrix}
+$$
+
+<!-- livebook:{"break_markdown":true} -->
+
+Firstly, let's check some characteristics of this tensor: they use full tensor aggregation functions.
+
+```elixir
+%{
+  dimensions: %{
+    x: Nx.axis_size(t, :x),
+    y: Nx.axis_size(t, :y),
+    z: Nx.axis_size(t, :z),
+    t: Nx.axis_size(t, :t)
+  },
+  n: Nx.size(t),
+  shape: Nx.shape(t),
+  rank: Nx.rank(t),
+  analysis: %{
+    most_frequent_nb: Nx.mode(t) |> Nx.to_number(),
+    smallest_element: %{
+      value: Nx.reduce_min(t) |> Nx.to_number(),
+      position: Nx.argmin(t) |> Nx.to_number()
+    },
+    greatest_element: %{
+      value: Nx.reduce_max(t) |> Nx.to_number(),
+      position: Nx.argmax(t) |> Nx.to_number()
+    },
+    no_zero_nb: Nx.all(t) |> Nx.to_number()
+  }
+}
+```
+
+#### Single row aggregation, along an axis
+
+> ❗ We are going to use the key `:axis` below. The following functions `argmin`, `argmax`, `median` and `mode` use `:axis: <symbol> or number` in singular mode. All of the other aggregating functions use `axes: [<symbol> or number]`; these have multi-axis aggregation implemented.
+
+When you aggregate along an axis, you are going to reshape the tensor and to aggregate with a function, in other words perform a reduction. It is important to consider the _ordering_ of the axes.
+
+<!-- livebook:{"break_markdown":true} -->
+
+###### Aggregate along the first axis `axis: :x` (equivalently `axis: 0`).
+
+The shape of our original tensor  `t`  is `[x: 2, y: 3, z: 2, t: 3]`. When we aggregate along the axis `:x`, it collapses and the shape of the resultant tensor `r` is `[y: 3,z: 2,t: 3]`. The rule is that every axis before the selected one will remain in the structure of the tensor as well as every axis after the selected one.
+
+How? You collect from each slice of `x` the elements that have the _same_ remaining indexes "on the right" $y,z,t$. In the case of our tensor, you have 2 `x`-slices so you build a sublist of 2 elements, and then apply the aggregating function on it. This gives you the value of the resulting tensor $r$ at the location $r[y][z][t]$:
+
+$$
+\rm{agg}\big(t[0][y][z][t], t[1][y][z][t]\big) = r[y][z][t]
+$$
+
+The aggregating function used below is `argmin`: the index of the smallest element of the  sublist. In case of several occurences, the `:tie_break` attribute takes the lowest index.
+
+$$
+\begin{bmatrix} x=0
+  \begin{bmatrix} y=0
+
+```
+\begin{bmatrix}
+  & \boxed{1}, 2, 3, & \\
+  & 1, -2, 3, &
+\end{bmatrix}
+\\ 
+y=1
+\begin{bmatrix} 
+  & 4, -3, 2 & \\
+  & -4, 3, 2 &
+\end{bmatrix}
+\\ 
+y=2
+\begin{bmatrix}
+  & 5, -1, 3 &\\
+  & -5, 1, 3 &
+\end{bmatrix}
+```
+
+\end{bmatrix}
+  \\
+  x=1
+  \begin{bmatrix} y=0
+
+```
+\begin{bmatrix}
+  & \boxed{4}, 6, 1 & \\
+  & 4, -6, 1 &
+\end{bmatrix}
+\\ 
+y=1
+\begin{bmatrix}
+  & 1, 2, 3 & \\
+  & 1, -2, 3 &
+\end{bmatrix}
+\\ 
+y=2
+\begin{bmatrix}
+  & 4, -3, 2 & \\
+  & -4, 3, 2 &
+\end{bmatrix}
+```
+
+\end{bmatrix}
+\end{bmatrix}
+
+\to
+  \begin{bmatrix} y=0
+
+```
+\begin{bmatrix}
+  & \boxed{\rm{agg}(1,4)}, \rm{agg}(2,6), \rm{agg}(3, 1)& \\
+  & \rm{agg}(1,4), \rm{agg}(-2,-6), \rm{agg}(3,1) &
+\end{bmatrix}
+\\ 
+y=1
+\begin{bmatrix} 
+  & \rm{agg}(4,1), \rm{agg}(-3,2), \rm{agg}(2,3) & \\
+  & \rm{agg}(-4,1), \rm{agg}(3,-2), \rm{agg}(2,3) &
+\end{bmatrix}
+\\ 
+y=2
+\begin{bmatrix}
+  & \rm{agg}(5,4), \rm{agg}(-1,-3), \rm{agg}(3,2) &\\
+  & \rm{agg}(-5,-4), \rm{agg}(1,3), \rm{agg}(3,2) &
+\end{bmatrix}
+```
+
+\end{bmatrix}
+
+\xrightarrow[]{\rm{argmin}}
+
+\begin{bmatrix}
+
+```
+\begin{bmatrix}
+  & 0,0, 1 & \\
+  & 0, 1,1 &
+\end{bmatrix}
+\\ 
+\begin{bmatrix} 
+  & 1, 0,0 & \\
+  & 0,1,0 &
+\end{bmatrix}
+\\ 
+\begin{bmatrix}
+  & 1, 1,1 &\\
+  & 0, 0,1 &
+\end{bmatrix}
+```
+
+\end{bmatrix}
+$$
+
+```elixir
+Nx.argmin(t, axis: :x)
+```
+
+###### Aggregate along the second axis `axis: :y` (equivalently `axis: 1`) .
+
+The axis `y` will collapse and the shape of the resultant tensor `r` is `[x: 2, z: 2, t: 3]`.
+
+How? The axis $x$ is before the selected one $y$ so it will remain. We therefor apply the procedure as above for _each_ sub $x$-slice.
+
+More precisely, fix the $x$-slice,  and let's consider the $x=1$ one. You collect in each sub `y`-slice the elements with the same remaining indexes "on the right", $z,t$ (and $x$ of course). You apply the aggregation and the result will be the value at location $r[1][z][t]$ of the resultant tensor $r$. So the elements of the slice $x=1$ of the resultant tensor $r$ will be:
+
+$$
+r[x_{=1}][z][t] = \rm{agg}\big([t[x_{=1}][0][z][t], t[x_{=1}][1][z][t], t[x_{=1}][2][z][t]\big)
+$$
+
+You repeat this operation for each $x$-slice.
+
+In the example below, we used again the `argmin` function.
+
+$$
+\begin{bmatrix} x=0
+  \begin{bmatrix} y=0
+
+```
+\begin{bmatrix}
+  & \boxed{1}, 2, 3, & \\
+  & 1, -2, 3, &
+\end{bmatrix}
+\\ 
+y=1
+\begin{bmatrix} 
+  & \boxed{4}, -3, 2 & \\
+  & -4, 3, 2 &
+\end{bmatrix}
+\\ 
+y=2
+\begin{bmatrix}
+  & \boxed{5}, -1, 3 &\\
+  & -5, 1, 3 &
+\end{bmatrix}
+```
+
+\end{bmatrix}
+  \\
+  x=1
+  \begin{bmatrix} y=0
+
+```
+\begin{bmatrix}
+  & 4, 6, 1 & \\
+  & 4, \boxed{-6}, 1 &
+\end{bmatrix}
+\\ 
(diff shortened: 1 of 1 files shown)


nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+    no_zero_nb: Nx.all(t) |> Nx.to_number()
+  }
+}
+```
+
+#### Single row aggregation, along an axis
+
+> ❗ We are going to use the key `:axis` below. The following functions `argmin`, `argmax`, `median` and `mode` use `:axis: <symbol> or number` in singular mode. All of the other aggregating functions use `axes: [<symbol> or number]`; these have multi-axis aggregation implemented.
+
+When you aggregate along an axis, you are going to reshape the tensor and to aggregate with a function, in other words perform a reduction. It is important to consider the _ordering_ of the axes.
```
polvalente, line 253:
```suggestion
When you aggregate along an axis, it's as if you're selecting all elements of a tensor along that axis for each fixed set of indices in other axes, and then performing the aggregation function over the selected subset.
```

"_ordering_" here is confusing. Do you mean the order in which the axes are originally distributed?

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+
+<!-- livebook:{"break_markdown":true} -->
+
+###### Aggregate along the first axis `axis: :x` (equivalently `axis: 0`).
+
+The shape of our original tensor  `t`  is `[x: 2, y: 3, z: 2, t: 3]`. When we aggregate along the axis `:x`, it collapses and the shape of the resultant tensor `r` is `[y: 3,z: 2,t: 3]`. The rule is that every axis before the selected one will remain in the structure of the tensor as well as every axis after the selected one.
```
polvalente, line 259:
```suggestion
The shape of our original tensor  `t`  is `{2, 3, 2, 3}`. When we aggregate along the axis `:x`, it collapses and the shape of the resulting tensor `r` is `{3, 2, 3}`. The rule is that every other axis will remain in the structure while the selected one collapses. If the `keep_axis: true` option was given, the axis would be kept with size `1` instead, resulting in a `{1, 3, 2, 3}`-shaped tensor.
```

The shape doesn't include the axes' names, so either use something similar to the suggestion (with the corrected values), or reword this so as to not imply that we refer to shape as the collection of names+sizes. This is just how we inspect a tensor, but in all other instances shape is just the tuple with the sizes.

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+
+How? You collect from each slice of `x` the elements that have the _same_ remaining indexes "on the right" $y,z,t$. In the case of our tensor, you have 2 `x`-slices so you build a sublist of 2 elements, and then apply the aggregating function on it. This gives you the value of the resulting tensor $r$ at the location $r[y][z][t]$:
+
+$$
+\rm{agg}\big(t[0][y][z][t], t[1][y][z][t]\big) = r[y][z][t]
+$$
+
+The aggregating function used below is `argmin`: the index of the smallest element of the  sublist. In case of several occurences, the `:tie_break` attribute takes the lowest index.
```
polvalente, line 267:
```suggestion
The aggregation function used below is `argmin`: the index of the smallest element of the sublist. In case of several occurrences, the `:tie_break` option takes the lowest index by default.
```

polvalente: Again, I think it would be clearer to the reader if Nx slicing notation is used below instead of Latex rendering

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+```
+
+\end{bmatrix}
+$$
+
+```elixir
+Nx.argmin(t, axis: :x)
+```
+
+###### Aggregate along the second axis `axis: :y` (equivalently `axis: 1`) .
```
polvalente, line 371:
```suggestion
###### Aggregate along the. `:y` axis
```

That title is saying the same thing 3 times in a row

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+
+#### Single row aggregation, along an axis
+
+> ❗ We are going to use the key `:axis` below. The following functions `argmin`, `argmax`, `median` and `mode` use `:axis: <symbol> or number` in singular mode. All of the other aggregating functions use `axes: [<symbol> or number]`; these have multi-axis aggregation implemented.
+
+When you aggregate along an axis, you are going to reshape the tensor and to aggregate with a function, in other words perform a reduction. It is important to consider the _ordering_ of the axes.
+
+<!-- livebook:{"break_markdown":true} -->
+
+###### Aggregate along the first axis `axis: :x` (equivalently `axis: 0`).
```
polvalente, line 257:
```suggestion
###### Aggregate along the `:x` axis
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+\end{bmatrix}
+$$
+
+```elixir
+Nx.argmin(t, axis: :x)
+```
+
+###### Aggregate along the second axis `axis: :y` (equivalently `axis: 1`) .
+
+The axis `y` will collapse and the shape of the resultant tensor `r` is `[x: 2, z: 2, t: 3]`.
```
polvalente, line 373:
```suggestion
The `:y` axis will collapse and the shape of the resulting tensor `r` is `[x: 2, z: 2, t: 3]`.
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+
+How? The axis $x$ is before the selected one $y$ so it will remain. We therefor apply the procedure as above for _each_ sub $x$-slice.
```
polvalente, line 375:
```suggestion
How? The axis $x$ is before the selected one $y$ so it will remain. We therefore apply the procedure as above for _each_ sub $x$-slice.
```

It's not actually for each x slice though. You're actually travelling each arrangement of {x, z, t} possible values. For each of those you take one full slice over `y`. Showing equivalent elixir code might be more clear to the reader:

```elixir
# The code below is conceptually equivalent to how aggregation subsets are taken:
for x <- 0..(Nx.axis_size(tensor, :x) - 1), z <- 0..(Nx.axis_size(tensor, :z) - 1), t <- 0..(Nx.axis_size(tensor, :t) - 1) do
  tensor[[x, 0..-1//1, z, t]]
end
```

And actually showing the result from a code cell might also help. Example:

```elixir
iex(13)> t = Nx.iota({2, 4, 3})
#Nx.Tensor<
  s64[2][4][3]
  [
    [
      [0, 1, 2],
      [3, 4, 5],
      [6, 7, 8],
      [9, 10, 11]
    ],
    [
      [12, 13, 14],
      [15, 16, 17],
      [18, 19, 20],
      [21, 22, 23]
    ]
  ]
>
iex(15)> Nx.sum(t, axes: [1])
#Nx.Tensor<
  s64[2][3]
  [
    [18, 22, 26],
    [66, 70, 74]
  ]
>
iex(16)> for x <- 0..(Nx.axis_size(t, 0) - 1), z <- 0..(Nx.axis_size(t, 2) - 1) do
...(16)> Nx.sum(t[[x, 0..-1//1, z]])
...(16)> end
[
  #Nx.Tensor<
    s64
    18
  >,
  #Nx.Tensor<
    s64
    22
  >,
  #Nx.Tensor<
    s64
    26
  >,
  #Nx.Tensor<
    s64
    66
  >,
  #Nx.Tensor<
    s64
    70
  >,
  #Nx.Tensor<
    s64
    74
  >
]
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+```
+
+\end{bmatrix}
+$$
+
+```elixir
+Nx.argmin(t, axis: 1)
+```
+
+###### Aggregate along the `axis: :z` axis (equivalently `axis: 2`).
```
polvalente, line 478: I don't think showing the aggregation over the next 2 axes brings anything significant to the table.
If you want to show like axes 0 and 2, I'm ok with it, but showing all 4 of them doesn't seem to add anything after the second example.

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+You have the `:tie_break` option to decide how to operate with you have several occurences of the result. It defaults to `tie_break: :low`.
+
+```elixir
+t4 = ~V[2 0 0 0 1]
+
+%{
+  argmin_with_default: Nx.argmin(t4) |> Nx.to_number(),
+  argmin_with_tie_break_high_option: Nx.argmin(t4, tie_break: :high) |> Nx.to_number()
+}
+```
```
polvalente, line 681: This is specific to some aggregation functions (IIRC, argmax and argmin only)

polvalente:
```suggestion
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+
+#### Option `:keep_axis`
+
+Its default value is `false` but when this option set to `keep_axis: true`, you reshape the resulting tensor with a new shape where the dimension of the working axis to $1$. For example, `Nx.argmax(t, axis: :y, keep_axix: true)` will have a shape a `[x: 2,y: 1,z: 2, t: 3]`. This mean that:
```
polvalente, line 685:
```suggestion
Its default value is `false` but when this option set to `keep_axis: true`, instead of being totally collapsed, the aggregation axis will be reduced to size $1$. For example, `Nx.argmin(t, axis: :y, keep_axis: true)` will have a shape `{2, 1, 2, 3}`, as seen below:
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+
+```elixir
+Nx.argmin(t, axis: 1, keep_axis: true) ==
+  Nx.argmin(t, axis: 1) |> Nx.reshape({2, 1, 2, 3}, names: [:x, :y, :z, :t])
```
polvalente, line 689:
```suggestion
Nx.argmin(t, axis: 1, keep_axis: true)
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+  \begin{bmatrix}
+    & 1, 2,0 & \\
+    & 2,0,0 &
+  \end{bmatrix}
+```
+
+\end{bmatrix}
+\end{bmatrix}
+$$
+
```
polvalente, line 739:
```suggestion
```

polvalente: This last Latex section seems superfluous

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+<!-- livebook:{"break_markdown":true} -->
+
+#### Multi-row aggregation
```
polvalente, line 742:
```suggestion
#### Multi-axis aggregation
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+
+<!-- livebook:{"break_markdown":true} -->
+
+Suppose you want to aggregate along the axis `x` and `z`. Then you should get a tensor of shape `[y: 3, t: 3]`. Given an index `x` and `z`, you aggregate all the numbers with the same indexes $(y,t)$.
```
polvalente, line 746:
```suggestion
Suppose you want to aggregate along the axes `x` and `z` for the same tensor still. Aggregation will return a tensor of shape `{1, 2}`. For each arrangement of valid `y` and `t` indices, you select all elements over both the `:x` and `:z` axes, like shown below:
```

Here might be a nice place to also introduce the code-instead-of-formula I suggested.

```elixir
for y <- 0..(Nx.axis_size(tensor, :y) - 1), t <- 0..(Nx.axis_size(tensor, :t) - 1) do
  tensor[[0..-1//1, y, 0..-1//1, t]]
end
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+
+$$
+r[y][t] = \rm{agg}\big( t0[y]0[t], t0[y]1[t], t1[y]0[t], t1[y]1[t]\big)
+
```
polvalente, line 750: Latex formatting here seems broken (no closing delimiter), but I think this won't be kept

polvalente:
```suggestion
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+\to
+
+\begin{bmatrix}
+ \rm{agg}(1,1,4,4),\rm{agg}(2,-2,6,-6),\rm{agg}(3,3,1,1) &\\
+  \rm{agg}(4,-4,1,1), \rm{agg}(-3,3,2,-2), \rm{agg}(2,2,3,3) & \\
+  \rm{agg}(5,-5,4,-4) , \rm{agg}(-1,1,-3,3,  \rm{agg}(3,3,2,2)
+\end{bmatrix}
+$$
+
+<!-- livebook:{"break_markdown":true} -->
```
polvalente, line 812:
```suggestion
<!-- livebook:{"break_markdown":true} -->
```

This seems out of place with the text

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+
+From this, it is easy to undertand what the aggregate returns:
```
polvalente, line 814:
```suggestion
From this, it is easier to understand what the following aggregation calls return:
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+
+```elixir
+Nx.reduce_min(t, axes: [0, 2])
+```
+
+```elixir
+Nx.reduce_max(t, axes: [0, 2])
```
polvalente, line 821:
```suggestion
Nx.reduce_max(t, axes: [:x, :t])
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+```
+
+```elixir
+Nx.mean(t, axes: [0, 2])
```
polvalente, line 825:
```suggestion
Nx.mean(t, axes: [:x, :y])
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+  \rm{agg}(5,-5,4,-4) , \rm{agg}(-1,1,-3,3,  \rm{agg}(3,3,2,2)
+\end{bmatrix}
+$$
+
+<!-- livebook:{"break_markdown":true} -->
+
+From this, it is easy to undertand what the aggregate returns:
+
+```elixir
+Nx.reduce_min(t, axes: [0, 2])
```
polvalente, line 817:
```suggestion
Nx.reduce_min(t, axes: [0, -1])
```

polvalente (commented):
Nice work overall!
I left quite a few comments, but most of those are minor structural things.
To be clear, I'm open to discussion in all comments and suggestions!

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,826 @@
...
+
+## Aggregations operations on tensors
+
+
+
+```elixir
+import Nx, only: :sigils
+```
+
+A tensor _in computer science_ can be viewed as a list of multi-indexed numbers. You simply wrap your list with `Nx.tensor`. Equivalently, you can use the compact `~V` sigil:
```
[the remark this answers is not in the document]
ndrean: I at last took time to read, sorry for the delay. 

Just a remark. The word tensor is ambiguous. It is related to the math we learned at school by the fact that it is multi-indexed (I believe "multi" is important), but that's all.
Ex. No intention of being pedantic at all, but the rank in computer science is just the number of dimensions which of course is not the case in math. Also, you agree that the (math) transpose of a covector (linear form, covariant) is a vector (contravariant), and vice-versa. Here, no, as per the definition of a vector. so I was confused because a covector has shape {n,1}, thus is 2D in computer science whilst not in math.  But again, it's just a matter of definition related to the domain and not important of course.

Maybe some examples to introduce how to build tensors are welcome? Such as:

 <==
We use the function `Nx` to construct tensors. A few examples:
 1. You can pass a number to a tensor and you get a 0D tensor. For example, `Nx(1)` is a 0D tensor.
2. you can build a 1D tensor or "vector" by passing a list of numbers. For example, `Nx([1,2,3])`.  Equivalently, you can use the sigil `~V` as `~V(1 2 3)`. 
3. you can build a matrix or 2D tensor with the sigil `~M`. 
For example a (2,3) matrix
```elixir
~M( 1 2 3
4 5 6)
```
Equivalently, we can use nested lists: `Nx.tensor([[1, 2,3],[4,5,6]])`.
Equivalently, we can reshape a vector: `~V(1 2 3 4 5 6) |>  Nx.reshape({2,3})`.

4. You can pass an arbitrarily nested list of numbers. For example, `Nx.tensor([[[1,2],[3,4]],[[5,6],[7,8]]]))`is a 3D tensor of shape `{2,2,3}`.  Equivalently, you can reshape a vector: `~V(1 2 3 4 5 6 7 8) |> Nx.Reshape({2,2,2})`.
 ==>

nx/guides/modoc_agg_nx_01.livemd
ndrean: + Nx.reduce_max(t, axes: [:x, :z])

polvalente: This was intentional so we have a greater variety of examples. Sorry for not explaining

ndrean: ah ok :)

ndrean:
2 corrections on typos on writing indexes in latex formulas:

###### Aggregate along the last axis, `axis: :t` or (`axis: 3` or `axis: -1`).
...
$$
r[x][y][z] = \rm{agg}\big(t[x][y][z][0],  t[x][y][z][1], t[x][y][z][2]\big)
$$

Multi-row aggregation
...
$$
r[y][t] = \rm{agg}\big( t[0][y][0][t], t[0][y][1][t], t[1][y][0][t], t[1][y][1][t]\big)
$$

Benjamin-Philip:
> [SOLVED By Jonatan] NO INDENTATION !! I can't properly render latex in a markdown cell with nested matrices
> 
> Livebook:
> 
> Screenshot 2023-02-09 at 17 31 50
> 
> vs Github: Screenshot 2023-02-09 at 17 33 53

I'm unable to reproduce the quotes in [katex's demo](https://katex.org/). However,
the incorrect ugly is reproducible in katex:

image

I'm able to reproduce the ugly alignment with regular latex:

```latex
\documentclass[preview, convert={outext=.png}]{standalone}
\usepackage{amsmath}
\usepackage{xcolor}
\begin{document}
\pagecolor{white}
\[
\begin{bmatrix}
x=0 &
\begin{bmatrix}
y=0 &
\begin{bmatrix}
z=0 & a_{0,0,0,0} & a_{0,0,0,1} & a_{0,0,0,2} \\
z=1 & a_{0,0,1,0} & a_{0,0,1,1} & a_{0,0,1,2}
\end{bmatrix}\\
y=1 &
\begin{bmatrix}
z=0 & a_{0,1,0,0} & a_{0,1,0,1} & a_{0,1,0,2} \\
z=1 & a_{0,1,1,0} & a_{0,1,1,1} & a_{0,1,1,2} \\
\end{bmatrix}\\
y=2 & \begin{bmatrix}
z=0 & a_{0,2,0,0} & a_{0,2,0,1} & a_{0,2,0,2} \\
z=1 & a_{0,2,1,0} & a_{0,2,1,1} & a_{0,2,1,2} \\
\end{bmatrix}
\end{bmatrix} \\
x=1 &
\begin{bmatrix}
y=0 &
\begin{bmatrix}
z=0 & a_{1,0,0,0} & a_{1,0,0,1} & a_{1,0,0,2}  \\
z=1 & a_{1,0,1,0} & a_{1,0,1,1} & a_{1,0,1,2} \\
\end{bmatrix}\\
y=1 &
\begin{bmatrix}
z=0 & a_{1,1,0,0} & a_{1,1,0,1} & a_{1,1,0,2}  \\
z=1 & a_{1,1,1,0} & a_{1,1,1,1} & a_{1,1,1,2} \\
\end{bmatrix}\\
y=2 &
\begin{bmatrix}
z=0 & a_{1,2,0,0} & a_{1,2,0,1} & a_{1,2,0,2}  \\
z=1 & a_{1,2,1,0} & a_{1,2,1,1} & a_{1,2,1,2} \\
\end{bmatrix}
\end{bmatrix}
\end{bmatrix}
\equiv
\begin{bmatrix}
\begin{bmatrix}
\begin{bmatrix}
& 1, 2, 3, & \\
& 1, -2, 3, &
\end{bmatrix} \\
\begin{bmatrix}
& 4, -3, 2 & \\
& -4, 3, 2 &
\end{bmatrix}\\
\begin{bmatrix}
& 5, -1, 3 &\\
& -5, 1, 3 &
\end{bmatrix}
\end{bmatrix} \\
\begin{bmatrix}
\begin{bmatrix}
& 4, 6, 1 & \\
& 4, -6, 1 &
\end{bmatrix}\\
\begin{bmatrix}
& 1, 2, 3 & \\
& 1, -2, 3 &
\end{bmatrix}\\
\begin{bmatrix}
& 4, -3, 2 & \\
& -4, 3, 2 &
\end{bmatrix}
\end{bmatrix}
\end{bmatrix}
\]
\end{document}
```

with the following invocation:

```sh
pdflatex --shell-escape foo.tex
```

and the result:

foo.

Quite clearly, something is wrong with how Livebook is using KaTex. They probably need to upgrade their KaTeX version. Feel free to send a PR/create an Issue at [livebook-dev/livebook](https://github.com/livebook-dev/livebook).
I'm looking into how we can improve the alignment in KaTeX. Since the LaTeX packages available in KaTeX are limited, the "usual" solutions like nicematrix, bigstrut et al are not available.

ndrean:
> Feel free to send a PR/create an Issue at [livebook-dev/livebook](https://github.com/livebook-dev/livebook).

In fact, I started a [discussion](https://github.com/livebook-dev/livebook/discussions/1701). Don't put indentations!

To check your latex, I used Katex home page where you can type an expression.

nx/guides/modoc_agg_nx_01.livemd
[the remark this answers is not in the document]
polvalente: I think this whole "what is and how to create a tensor" could be its own guide instead of staying in this one.

> Just a remark. The word tensor is ambiguous. 
> It is related to the math we learned at school by the fact that it is multi-indexed (I believe "multi" is important), but that's all.

Yes, but here we are _defining_ it. Therefore it's not ambiguous anymore in our domain. We can make it more clear by prefacing with "Tensors have many different definitions. In Nx, a tensor is..." or something like that.
I don't believe multi is important, but it's not a hill I'm willing to die on :)

> Ex. No intention of being pedantic at all, but the rank in computer science is just the number of dimensions which of course is not the case in math. 
Yes, but we have to again remember that we're speaking in the confines of the Nx domain language. No need to confuse people by introducing math-speak. This ties in to the next point:

> Also, you agree that the (math) transpose of a covector (linear form, covariant) is a vector (contravariant), and vice-versa. Here, no, as per the definition of a vector. so I was confused because a covector has shape {n,1}, thus is 2D in computer science whilst not in math. But again, it's just a matter of definition related to the domain and not important of course.

None of these definitions really matter to Nx. We can define that a vector is a 1D tensor just for the sake of speaking more easily about things, like we did with matrices being 2D tensors. In general, I think you're coming into this with a "math" mindset. This isn't bad in itself, but you have to remember that the ubiquitous language changes with each domain, and in Nx-land, we don't have the same definitions that you might be used to.

All in all, I think all of this might have a place still by discussing differences and similarities in a more in-depth _separate_ guide. What do you think?

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,748 @@
...
+
+We will use the function `Nx.tensor/2` to [construct tensors](https://hexdocs.pm/nx/Nx.html#module-creating-tensors). A few examples:
+
+* you can pass a number to a tensor and you get a 0D tensor. For example, `Nx.tensor(1)` is a 0D tensor.
+
+```elixir
+Nx.tensor(1)
+```
+
+* you can build a 1D tensor or "vector" by passing a list of numbers. Equivalently, you can use the sigil `~V`. Both parenthesis "(" and brackets "[" are accepted.
```
polvalente, line 27:
```suggestion
* you can build a 1D tensor or vector by passing a list of numbers. Equivalently, you can use the sigil `~V`. Both parenthesis "(" and brackets "[" are accepted.
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,748 @@
...
+t2 = ~V(1 2 3 4 5 6 7 8 9 10 11 12) |> Nx.reshape({2, 2, 3})
+
+{
+  Nx.shape(t1),
+  t1 == t2,
+  t1
+}
+```
+
+#### A note on how to access to values of a tensor
```
polvalente, line 76: tying into my suggestion of splitting into 2 livebooks/guides, I think up until this section would be part of a more general "how to create and operate tensors" guide, and from the "Aggregations..." section onwards would stay in this guide.

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,748 @@
...
+
+{
+  corners,
+  first_row,
+  first_column,
+  reverse_index
+}
+```
+
+## Aggregations operations on tensors
```
polvalente, line 97:
```suggestion
## Aggregation operations on tensors
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,748 @@
...
+
+You can apply [aggregation functions](https://hexdocs.pm/nx/Nx.html#functions-aggregates) on any tensor. The function can be applied to the tensor as a whole or to a subsection of the tensor taken in an axis-wise fashion.
```
polvalente, line 99:
```suggestion
You can apply [aggregation functions](https://hexdocs.pm/nx/Nx.html#functions-aggregates) on any tensor. The functions can be applied to the tensor as a whole or to a subsection of the tensor taken in an axis-wise fashion.
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,748 @@
...
+
+As a first example, let's take a 2D tensor (aka matrix) of shape (2,3). Notice that we can _name_ the axes. The elements of the tensor `t` are the `t[i][j]` when $i$ is the row, and $j$ the column.
```
polvalente, line 101:
```suggestion
As a first example, let's take a 2D tensor of shape `{2, 3}`. Notice that we can _name_ the axes. The elements of the tensor `t` are the `t[i][j]` when $i$ is the row, and $j$ the column.
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,748 @@
...
+
+```elixir
+m = Nx.tensor([[1, 2, 3], [4, 5, 6]], names: [:x, :y])
+```
+
+You can get the maximum number in a tensor with [Nx.reduce_max(matrix)](https://hexdocs.pm/nx/Nx.html#reduce_max/2), returning a scalar (0D tensor). With a reduction, we loose a dimension per axis reduced, and since we did globally, we loose all the dimensions. It should return `Nx.Tensor(6)` here.
```
polvalente, line 107:
```suggestion
You can get the maximum number in a tensor with [Nx.reduce_max(m)](https://hexdocs.pm/nx/Nx.html#reduce_max/2), returning a scalar (0D tensor). With a reduction, we lose a dimension per axis reduced, and since we applied the reduction globally, we lose all the dimensions. It should return `Nx.Tensor(6)` here.
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,748 @@
...
+
+We can get the maximum number for each _row_ with `Nx.reduce_max(matrix, axes: [:y])`, returning a vector (1D tensor) of length 2. Why `:y`? Because for each row, we reduce along the `:y` axis.
+
+We can get the maximum number for each _column_ with `Nx.reduce_max(matrix, axes: [:x])`, returning a vector (1D tensor) of length 3. For each column, we reduce along the axis `:x`.
```
polvalente, line 111:
```suggestion
We can get the maximum number for each _row_ with `Nx.reduce_max(matrix, axes: [:y])`, returning a 1D tensor of length 2. Why `:y`? Because for each row, we reduce along the `:y` axis.

We can get the maximum number for each _column_ with `Nx.reduce_max(matrix, axes: [:x])`, returning a 1D tensor of length 3. For each column, we reduce along the axis `:x`.
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,748 @@
...
+
+```elixir
+max = Nx.reduce_max(m)
+max_x = Nx.reduce_max(m, axes: [:x])
+max_y = Nx.reduce_max(m, axes: [:y])
+{max, max_x, max_y, m}
+```
+
+Let's consider another example with the aggregate [Nx.weighted_mean](https://hexdocs.pm/nx/Nx.html#weighted_mean/3). It supports full-tensor and per axis operations. Let's check how the weighted mean aggregate of a matrix is computed with a 2D tensor of shape (2,2) (aka as "square matrix").
```
polvalente, line 120:
```suggestion
Let's consider another example with the aggregate [Nx.weighted_mean](https://hexdocs.pm/nx/Nx.html#weighted_mean/3). It supports full-tensor and per axis operations. Let's check how the weighted mean aggregate of a matrix is computed with a 2D tensor of shape `{2,2}`.
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,748 @@
...
+
+```elixir
+m = ~M(1 2
+  3 4)
+```
+
+We compute the full -tensor aggregate. The calculations are developped below. We calculate an "array product" (aka [Hadamard product](<https://en.wikipedia.org/wiki/Hadamard_product_(matrices)#:~:text=In%20mathematics%2C%20the%20Hadamard%20product,elements%20i%2C%20j%20of%20the>), a terms by terms product) of our tensor with the tensor of weights, then sum all the elements and divide by the sum of the weights.
```
polvalente, line 127:
```suggestion
We compute the full-tensor aggregate. The calculations are developed below. We calculate an "array product" (aka [Hadamard product](<https://en.wikipedia.org/wiki/Hadamard_product_(matrices)#:~:text=In%20mathematics%2C%20the%20Hadamard%20product,elements%20i%2C%20j%20of%20the>), an element-wise product) of our tensor with the tensor of weights, then sum all the elements and divide by the sum of the weights.
```

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,748 @@
...
+  nx_weighted_avg: Nx.to_number(nx_w_avg),
+  manual_weighted_avg: man_w_avg
+}
+```
+
+The weighted mean can be run per axis. Let's compute it along the **first** axis (`axes: [0]`) with the same weights per axis: you calculate "by column", so you aggregate/reduce along the first axis:
+
+```elixir
+w = ~M(4 5
+  4 5)
```
polvalente, line 148: same issue here

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,748 @@
...
+  w_avg_x,
+  man_w_avg_x
+}
+```
+
+We calculate weighted mean of a square matrix along the **second** axis (`axes: [1]`): you calculate per row, so you aggregate/reduce along the second axis.
+
+```elixir
+w = ~M(4 5
+  5 4)
```
polvalente, line 164: I'd use totally different values instead of just changing the order

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,748 @@
...
+We compute the full -tensor aggregate. The calculations are developped below. We calculate an "array product" (aka [Hadamard product](<https://en.wikipedia.org/wiki/Hadamard_product_(matrices)#:~:text=In%20mathematics%2C%20the%20Hadamard%20product,elements%20i%2C%20j%20of%20the>), a terms by terms product) of our tensor with the tensor of weights, then sum all the elements and divide by the sum of the weights.
+
+```elixir
+w = ~M[4 5
+  4 5]
+
+nx_w_avg = Nx.weighted_mean(m, w)
+
+man_w_avg = (1 * 4 + 2 * 5 + 3 * 4 + 4 * 5) / (4 + 5 + 4 + 5)
+# 46/18
```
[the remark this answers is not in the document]
ndrean: you are right, I was lazy :)

nx/guides/modoc_agg_nx_01.livemd
ndrean: Yes..., useless here

nx/guides/modoc_agg_nx_01.livemd
ndrean: Done!

nx/guides/modoc_agg_nx_01.livemd
```diff
@@ -0,0 +1,669 @@
+# Livebook companion for Nx aggregations
+
+```elixir
+Mix.install([
+  {:nx, "~> 0.4.2"}
```
polvalente, line 5:
```suggestion
  {:nx, "~> 0.5"}
```

nx/guides/modoc_agg_nx_01.livemd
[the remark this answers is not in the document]
ndrean: In computer science, I have only used numpy or sagemth, so code running on CPU, never tensors on GPU! But I agree, it's just a matter of definition.

polvalente (approved):

PR #1088, opened 2023-02-04
Make serving explicitly receive compiler options

josevalim:
The goal is to allow us to device partitioning
in the future that will work transparently.
For example, we could do:

    {Nx.Serving,
     serving: my_serving(),
     partitioned: true}

And, if you have two devices, it will create two
Nx.Serving and automatically partition the load
between them. This will be handy with the distributed
option too, making sure we can spread the load across
multiple GPUs and machines.

### DIFF
--- a/nx/lib/nx/serving.ex
+++ b/nx/lib/nx/serving.ex
@@ -527,9 +527,7 @@ defmodule Nx.Serving do
   @impl true
   def init({name, serving, batch_size, batch_timeout}) do
     [parent | _] = Process.get(:"$ancestors")
-
-    {:ok, module_state} =
-      handle_init(serving.module, :process, serving.arg, serving.defn_options)
+    {:ok, module_state} = handle_init(serving.module, :process, serving.arg, serving.defn_options)
 
     :persistent_term.put(
       persistent_key(name),
--- a/nx/lib/nx/serving.ex
+++ b/nx/lib/nx/serving.ex
@@ -218,7 +218,7 @@ defmodule Nx.Serving do
     :client_preprocessing,
     :client_postprocessing,
     process_options: [],
-    compiler_options: []
+    defn_options: []
   ]
 
   @type metadata() :: term()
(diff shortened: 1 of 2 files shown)


josevalim:
@jonatanklosko please let me know if you think this approach will conflict with Bumblebee in any way.

nx/lib/nx/serving.ex
```diff
@@ -247,18 +262,28 @@ defmodule Nx.Serving do
...
+      when is_function(function, 1) and is_list(compiler_options) do
+    new(Nx.Serving.Default, function, compiler_options)
+  end
+
+  def new(function, compiler_options)
+      when is_function(function, 0) and is_list(compiler_options) do
+    IO.warn(
+      "passing a zero-arity function to Nx.Serving.new is deprecated, " <>
+        "please pass a single arity function that will receive the compiler options"
+    )
```
jonatanklosko, line 284: In this case it's the old `process_options`, so we should set them accordingly?

nx/lib/nx/serving.ex
```diff
@@ -674,13 +715,13 @@ defmodule Nx.Serving do
     {__MODULE__, name}
   end
 
-  defp handle_init(module, type, arg) do
-    case module.init(type, arg) do
+  defp handle_init(module, type, arg, opts) do
+    case module.init(type, arg, opts) do
```
jonatanklosko, line 719: Should we check if /2 is exported and do soft deprecation instead?

jonatanklosko (approved):
Sounds good!

nx/lib/nx/serving.ex
josevalim: Nah, i think it is fine with breaking for now. Most people are likely using the function based version which I will keep as process options.

jonatanklosko: Makes sense!

josevalim:

PR #1090, opened 2023-02-06
Add distributed² serving

josevalim:


nx/lib/nx/serving.ex
```diff
@@ -486,23 +543,83 @@ defmodule Nx.Serving do
   end
 
   @doc """
-  Runs the given `input` on the process given by `name`.
+  Runs the given `input` on the serving given by `name`.
```
josevalim, line 546:
```suggestion
  Runs the given `input` on the serving process given by `name`.
```

nx/lib/nx/serving.ex
```diff
@@ -486,23 +543,83 @@ defmodule Nx.Serving do
...
 
-  The process `name` will batch requests and send a response
-  either when the batch is full or on timeout. See the module
-  documentation for more information.
+  `name` is either an atom representing a local or distributed
+  serving. First it will attempt to dispatch locally, then it
```
josevalim, line 549:
```suggestion
  serving process. First it will attempt to dispatch locally, then it
```

nx/lib/nx/serving.ex
```diff
@@ -486,23 +543,83 @@ defmodule Nx.Serving do
...
+  falls back to the distributed serving. You may specify
+  `{:local, name}` to force a local lookup or `{:distributed, name}`
+  to force a distributed one.
+
+  The `client_preprocessing` callback will be invoked on the `input`
+  which is then sent to the server. The server will batch requests
+  and send a response either when the batch is full or on timeout.
+  Then `client_postprocesing` is invoked on the response. See the
+  module documentation for more information.
```
josevalim, line 558:
```suggestion
  The `client_preprocessing` callback will be invoked on the `input`
  which is then sent to the server. The server will batch requests
  and send a response either when the batch is full or on timeout.
  Then `client_postprocessing` is invoked on the response. See the
  module documentation for more information. In the distributed case,
  the callbacks are invoked in the distributed node, but still outside of
  the serving process.
```

nx/lib/nx/serving.ex
```diff
@@ -562,7 +684,58 @@ defmodule Nx.Serving do
...
+
+        {_, monitor_ref} =
+          Node.spawn_monitor(node(pid), __MODULE__, :__distributed_batched_run__, args)
+
+        receive do
+          {:DOWN, ^monitor_ref, _, _, {^ref, result}} ->
+            result
+
+          {:DOWN, _, _, _, :noproc} ->
+            # TODO: Return no proc for stacked and queued batches on terminate
```
josevalim, line 717:
```suggestion
```

nx/lib/nx/serving.ex
```diff
@@ -172,9 +172,41 @@ defmodule Nx.Serving do
...
+  All `Nx.Serving`s are distributed by default. If the current machine
+  does not have an instance of `Nx.Serving` running, `batched_run/3` will
+  automatically look for one in the cluster. The nodes do not need to run
+  the same code and applications. It is only required that they run the
+  same `Nx` version.
+
+  The load balancing between servings is done randomly, however, the number
+  of partitions are considered. For example, if you have a node with 2 GPUs
+  and another with 4, the latter will receive the double of requests compared
+  to the former.
```
josevalim, line 189:
```suggestion
  The load balancing between servings is done randomly, however, the number
  of partitions are considered if the `partitioned: true` option is also given.
  For example, if you have a node with 2 GPUs and another with 4, the latter
  will receive the double of requests compared to the former.
```

nx/lib/nx/serving.ex
```diff
@@ -172,9 +172,41 @@ defmodule Nx.Serving do
...
+  The load balancing between servings is done randomly, however, the number
+  of partitions are considered if the `partitioned: true` option is also given.
+  For example, if you have a node with 2 GPUs and another with 4, the latter
+  will receive the double of requests compared to the former.
+
+  `batched_run/3` receives an optional `distributed_preprocessing` callback as
+  third argument for preprocessing the input for distributed requests. When
+  using libraries like EXLA or Torchx, the tensor is often allocated in memory
+  inside a third-party library so it is necessary to either transfer or copy
+  the tensor to the binary backend before sending it to another node.
```
seanmor5, line 195: If we support device-to-device transfer's already it would be more efficient to an expose an API for that in the future rather than depend on an additional copy. The NVIDIA implementation would use [NCCL](https://developer.nvidia.com/nccl) which would be more efficient than a copy+comms between nodes.

Not sure how this would best tie in though, and shouldn't block this going out anyway as this is already a pretty big performance win

josevalim: @seanmor5 in this case it may not matter much because we are not really planning to share the load between two gpus and the input data is relatively small (and many times it is not even a tensor yet!).

However, NCCL will play a role on training and pmap. Do you know how we tell XLA that we have a GPU on another _node_?

seanmor5: There is a distributed PjRt client, so I'm guessing when nodes > 1 you need to make use of an API similar to this: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/pjrt/distributed/client.h

nx/lib/nx/serving.ex
```diff
@@ -562,7 +686,57 @@ defmodule Nx.Serving do
...
+    exit({:noproc, {__MODULE__, :distributed_batched_run, [name, input, [retries: 0]]}})
+  end
+
+  defp distributed_batched_run_with_retries!(name, input, retries) do
+    case :pg.get_members(Nx.PG, __MODULE__) do
+      [] ->
+        exit({:noproc, {__MODULE__, :distributed_batched_run, [name, input, [retries: retries]]}})
+
+      entries ->
+        pid = Enum.fetch!(entries, :rand.uniform(length(entries)) - 1)
```
jonatanklosko, line 707:
```suggestion
        pid = Enum.random(entries)
```

nx/lib/nx/serving.ex
```diff
@@ -579,17 +753,21 @@ defmodule Nx.Serving do
...
       %{
         limit: batch_size,
         preprocessing: serving.client_preprocessing,
-        postprocessing: serving.client_postprocessing
+        postprocessing: serving.client_postprocessing,
+        distributed_postprocessing: serving.distributed_postprocessing
       }
     )
 
+    :pg.join(Nx.PG, __MODULE__, List.duplicate(self(), partitions_count))
```
jonatanklosko, line 769: `Nx.Serving.PG` perhaps?

nx/lib/nx/serving.ex
josevalim: My version is a bit more optimized because we know the list is small. The Enum version assumes we don't know the list size.

nx/lib/nx/serving.ex
josevalim: Good call. Will rename after merging. :)

josevalim:

nx/lib/nx/serving.ex
jonatanklosko: Both `Enum.fetch!` and `Enum.random` do `drop_list` with a random number no?

josevalim: Gah, good call. I missed the special case for lists in there. :)

PR #1094, opened 2023-02-13
Add fundamental mathematical constants

### LINKED ISSUE #1092: Add Fundamental Mathematical Constants
I think `Nx` should support some of the most important mathematical constants. In `NumPy` there is support for:

* $e$ - base of the natural logarithm
* $\gamma$ - Euler's constant (Euler–Mascheroni constant)
* $\pi$

Those will be completely sufficient in my opinion.

msluszniak:
Closes #1092 

Except mathematical constants, this PR also adds math notation in docs with the same configuration as in Axon.

### DIFF
--- a/nx/lib/nx/constants.ex
+++ b/nx/lib/nx/constants.ex
@@ -355,7 +355,7 @@ defmodule Nx.Constants do
 
       iex> Nx.Constants.pi({:f, 32})
       #Nx.Tensor<
-        f64
+        f32
         3.1415927410125732
       >
 
--- a/nx/lib/nx/constants.ex
+++ b/nx/lib/nx/constants.ex
@@ -306,7 +306,7 @@ defmodule Nx.Constants do
         f64
         2.2250738585072014e-308
       >
-      
+
       iex> Nx.Constants.smallest_positive_normal({:f, 32})
       #Nx.Tensor<
         f32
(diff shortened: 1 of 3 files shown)


msluszniak:
Maybe it will be better to use `Nx.tensor(pi/e/gamma, type: type, backend: Nx.BinaryBackend)` instead of converting it to binary indirectly, WDYT @polvalente @josevalim?

polvalente:
> Maybe it will be better to use `Nx.tensor(pi/e/gamma, type: type, backend: Nx.BinaryBackend)` instead of converting it to binary indirectly, WDYT @polvalente @josevalim?

I think with the current way we take advantage of the default backend configuration, which makes for better operability, especially in defn

nx/lib/nx/constants.ex
```diff
@@ -333,6 +333,144 @@ defmodule Nx.Constants do
     from_binary(Nx.Type.smallest_positive_normal_binary(type), type, opts)
   end
 
+  @doc ~S"""
+  Returns $\pi$ in f32.
+  """
+  def pi, do: pi({:f, 32}, [])
```
polvalente, line 339: Why isn't type an option?

polvalente: Also, you could declare the function as `def pi(type \\ :f32, opts \\ [])` to avoid having a separate clause, and just document the default type as `{:f, 32}`.

Although it being optional might indicate it should be an option instead

josevalim:

PR #1095, opened 2023-02-13
Adds `to_list/1` to Nx

thiagopromano:
Currently, to create a tensor from a nested list is pretty easy with `Nx.tensor/1`, but the reverse operation can be a little complicated/frustrating.

There are a few options, such as `to_batched/3` and `to_flat_list/2`, but the problem with them is that additional steps are needed to get a nested list that represents the tensor structure back.

For example: https://elixirforum.com/t/converting-tensor-into-nested-list/44352

Because of that, some libraries end up creating their own `to_list` function to solve this, such as [Bumblebee](https://github.com/elixir-nx/bumblebee/blob/v0.1.2/lib/bumblebee/utils/nx.ex#L31) (where I got the code from).

This PR allows for this:
```elixir
iex> [[1, 2], [3, 4]] |> Nx.tensor() |> Nx.to_list()    
[[1, 2], [3, 4]]
```

This probably have already crossed the core developers mind, so what is the general opinion on this?

Tests and docs are still missing  in this PR, but it would contain some info about when `to_batched/3` should be preferred.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -1789,6 +1789,18 @@ defmodule Nx do
     end
   end
 
+  def to_list(tensor) do
+    list = to_flat_list(tensor)
+
+    tensor
+    |> shape()
+    |> Tuple.to_list()
+    |> Enum.drop(1)
+    |> Enum.reverse()
+    |> Enum.reduce(list, &Stream.chunk_every(&2, &1))
+    |> Enum.to_list()
+  end
+
   @doc false
   @deprecated "Use to_batched/3 instead"
   def to_batched_list(tensor, batch_size, opts \\ []) do
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -1789,16 +1789,51 @@ defmodule Nx do
     end
   end
 
-  def to_list(tensor) do
-    list = to_flat_list(tensor)
+  @doc """
+  Converts the tensor into a list reflecting its structure.
 
-    tensor
-    |> shape()
-    |> Tuple.to_list()
-    |> Enum.drop(1)
-    |> Enum.reverse()
-    |> Enum.reduce(list, &Stream.chunk_every(&2, &1))
-    |> Enum.to_list()
+  Negative infinity (-Inf), infinity (Inf), and "not a number" (NaN)
+  will be represented by the atoms `:neg_infinity`, `:infinity`, and
+  `:nan` respectively.
+
+  Note: This function cannot be used in `defn`.
+
+  ## Examples
+
+      iex> Nx.iota({2, 3}) |> Nx.to_list()
+      [
+        [0, 1, 2],
+        [3, 4, 5]
+      ]
+  """
+  @doc type: :conversion
+  def to_list(tensor) do
+    %{type: type, shape: shape} = tensor = to_tensor(tensor)
+    binary = to_binary(tensor, [])
+
+    dims = Tuple.to_list(shape)
+    {list, ""} = chunk(dims, binary, type)
+    list
+  end
+
+  defp chunk([], data, type) do
+    match_types [type] do
+      <<match!(head, 0), tail::binary>> = data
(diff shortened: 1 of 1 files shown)


josevalim:
What is the general use case for such a function?

thiagopromano:
I'd say a general use case is converting a higher dimensional tensor to be used in code that abstracts away Nx.

In my personal example, I've been converting some functions to use Nx instead of erlport+Python, one of these functions worked by receiving strings, tokenizing and returning the a cosine similarity matrix between them.

The last step in the pipeline is turning the tensor matrix into nested lists, which can then used in the business logic. I've been using `Bumblebee.Utils.Nx.to_list/1` for that, but it really fells that it belongs to Nx.

Also, I remember when learning the basics of Nx, I found it frustrating the lack a conversion function that simply converts the tensor structure back to the nested lists structure, which is a common way of creating the tensors. Elixir has always so many functions for converting between types, such as `List.to_tuple` and `Tuple.to_list`, it is strange that the converse of `Nx.tensor/1` doesn't exist natively on Nx.

josevalim:
@jonatanklosko what is your take? Since we have `Nx.to_list` in Bumblebee?

jonatanklosko:
@josevalim the place we use it in bumblebee is when passing model output (tensor) into a tokenizer (lists) to get text back.

IIRC we avoid this conversion to make people think twice before they use `Enum`, but there are valid cases where we want this conversion and it may be surprising that there's no straightforward way. It's kinda like using `String.to_existing_atom`, except when we actually need `String.to_atom`. Also, we already have `to_flat_list` and `flat` is just a limitation.

We've been deferring it, but I'm fine with adding the function, we just need a clear note that this should be used when they actually need a list, otherwise they should consider other `Nx` operations or to_batched_list.

josevalim:
@jonatanklosko there is one issue: what about scalars? They wouldn't return a list in `Nx.to_list`. So maybe we need another name or to reject scalars?

nx/lib/nx.ex
```diff
@@ -1789,6 +1789,18 @@ defmodule Nx do
...
+    list = to_flat_list(tensor)
+
+    tensor
+    |> shape()
+    |> Tuple.to_list()
+    |> Enum.drop(1)
+    |> Enum.reverse()
+    |> Enum.reduce(list, &Stream.chunk_every(&2, &1))
+    |> Enum.to_list()
+  end
```
josevalim, line 1802:
```suggestion
  @doc """
  Converts the tensor into a list reflecting its structure.

  Negative infinity (-Inf), infinity (Inf), and "not a number" (NaN)
  will be represented by the atoms `:neg_infinity`, `:infinity`, and
  `:nan` respectively.

  Note: This function cannot be used in `defn`.

  ## Examples

      iex> Nx.iota({2, 3}) |> Nx.to_list()
      [
        [0, 1, 2],
        [3, 4, 5]
      ]
  """
  @doc type: :conversion
  def to_list(tensor) do
    %{type: type, shape: shape} = tensor = to_tensor(tensor)
    binary = to_binary(tensor, [])

    dims = Tuple.to_list(shape)
    {list, ""} = chunk(dims, binary, type)
    list
  end

  defp chunk([], data, type) do
    match_types [type] do
      <<match!(head, 0), tail::binary>> = data
      {read!(head, 0), tail}
    end
  end

  defp chunk([dim | dims], data, type) do
    chunk_each(dim, data, [], dims, type)
  end

  defp chunk_each(0, data, acc, _dims, _type) do
    {Enum.reverse(acc), data}
  end

  defp chunk_each(dim, data, acc, dims, type) do
    {entry, rest} = chunk(dims, data, type)
    chunk_each(dim - 1, rest, [entry | acc], dims, type)
  end
```

jonatanklosko:
> So maybe we need another name or to reject scalars?

Raising sounds fine to me. We could unify under a single function like `Nx.to_numbers`, but I'm not sure what's the best name.

josevalim:

PR #1097, opened 2023-02-14
Update livebook

msluszniak:


msluszniak (commented):
And I think after new release of Nx it will be worth to change also the line with declaration of pi to use `Nx.Constant.pi`

nx/guides/intro-to-nx.livemd
```diff
@@ -264,7 +264,7 @@ suggest that they would work on primitive values, called scalars.
 Indeed, a tensor can be a scalar:
 
 ```elixir
-pi = Nx.tensor(3.1415, type: :f32)
+pi = Nx.Constants.pi()
```
josevalim, line 267: I think in this case we want to show how to build it by hand, so we should likely revert. :)

josevalim:
```suggestion
pi = Nx.tensor(3.1415, type: :f32)
```

nx/guides/intro-to-nx.livemd
```diff
@@ -2,7 +2,7 @@
 
 ```elixir
 Mix.install([
-  {:nx, "~> 0.2"}
+  {:nx, github: "elixir-nx/nx", override: true, sparse: "nx"}
```
josevalim, line 5:
```suggestion
  {:nx, "~> 0.5"}
```

josevalim:

PR #1101, opened 2023-02-14
Update examples

msluszniak:


exla/examples/mnist.exs
```diff
@@ -1,11 +1,20 @@
+Mix.install([
+  {:nx, "~> 0.5"},
+  {:exla, "~> 0.5"}
+])
```
josevalim, line 4: Those examples are meant to run as `mix run examples/mnist.exs` from the EXLA project root. It is mostly for testing. Can you please remove the `Mix.install` from them and add such note? This means we won't be able to use Scholar though. :(

msluszniak: Ok, I see

josevalim:

PR #1104, opened 2023-02-17
Add top-k

seanmor5:
Adds Top-K to Nx with optimized implementations for TorchX and EXLA.

Top-K is necessary for Beam Search in Bumblebee (as well as contrastive search I believe). EXLA depends on: https://github.com/elixir-nx/xla/pull/39 and a new XLA release which we can bump tomorrow probably


josevalim (approved):

josevalim:
Let's make the callback optional.

josevalim:
Notes:

1. You need to [define it as optional here](https://github.com/elixir-nx/nx/blob/main/nx/lib/nx/backend.ex#L130)
2. You still need to implement Torchx.Backend.top_k :)
3. I will later check if we are invoking the custom op for `defn` but we might not

nx/lib/nx.ex
```diff
@@ -11067,6 +11067,77 @@ defmodule Nx do
...
+    {output_shape, output_names} = Nx.Shape.top_k(shape, names, k)
+
+    out_values = %{tensor | shape: output_shape, names: output_names}
+    out_indices = %{tensor | shape: output_shape, names: output_names, type: {:s, 64}}
+
+    Nx.Shared.optional(:top_k, [tensor, k], {out_values, out_indices}, fn tensor, k ->
+      rank = rank(tensor)
+
+      values = sort(tensor, axis: rank - 1, direction: :desc)
+      indices = argsort(tensor, axis: rank - 1, direction: :desc)
```
polvalente, line 11134: I would implement this like below to avoid double sorting

```suggestion
      indices = argsort(tensor, axis: rank - 1, direction: :desc)
      values = Nx.take_along_axis(tensor, axis: rank - 1)
```

polvalente: Also, why force the "top k" axis to be only the last one instead of making it configurable?

josevalim: Note we would need to make it configurable on both Torchx and EXLA and I am not sure if they support it.

josevalim:

PR #1111, opened 2023-03-01
Fix ArgumentError for complex case in matrix_rank function

kenken-neko:
When a matrix with complex numbers as elements is entered into `matrix_rank` function, the error comment `"Nx.LinAlg.svd/2 is not yet implemented for complex inputs"` is displayed as below.  
svd

Therefore, I corrected it so that the error comment `"Nx.LinAlg.matrix_rank/2 is not yet implemented for complex inputs"` is displayed.


josevalim:

PR #1115, opened 2023-03-06
feat: add Nx.vectorize and support vectorization in binary ops, pred ops and unary math ops

polvalente:


nx/lib/nx.ex
```diff
@@ -3551,6 +3551,165 @@ defmodule Nx do
...
+  end
+
+  def vectorize(%Nx.Tensor{shape: {}}, _name) do
+    raise ArgumentError, "cannot vectorize tensor of rank 0"
+  end
+
+  def vectorize(%Nx.Tensor{names: names, shape: shape, vectorized_axes: vec_axes} = tensor, name) do
+    size = elem(shape, 0)
+    new_shape = Tuple.delete_at(shape, 0)
+    names = Enum.drop(names, 1)
```
josevalim, line 3578:
```suggestion
    names = tl(names)
```

polvalente: tl raises if the list is empty

polvalente: Oh, but I guess it's guaranteed that it's non-empty here

nx/lib/nx.ex
```diff
@@ -3551,6 +3551,165 @@ defmodule Nx do
...
+  defp unvectorize(
+         %T{shape: left_shape, names: left_shape_names} = left,
+         %T{shape: right_shape, names: right_shape_names} = right
+       ) do
+    left_names = Keyword.keys(left.vectorized_axes) |> MapSet.new()
+    right_names = Keyword.keys(right.vectorized_axes) |> MapSet.new()
+
+    left_only = MapSet.difference(left_names, right_names)
+    right_only = MapSet.difference(right_names, left_names)
+    in_both = MapSet.intersection(left_names, right_names)
```
josevalim, line 3609: You can probably do all of this in a single pass. Convert the left (or right) side to a map. Then do a reduce collecting what exists on both, removing from the map, and what exists on one side. What remains on the map is what exists only on the other side.

nx/lib/nx.ex
```diff
@@ -3551,6 +3551,165 @@ defmodule Nx do
...
+
+    # check that common names share the same value
+    for name <- in_both do
```
josevalim, line 3612: This check can be merged into the single reduce above. :)

nx/lib/nx.ex
```diff
@@ -3551,6 +3551,165 @@ defmodule Nx do
...
+    else
+      fun.(tensor)
+    end
+  end
+
+  defp apply_vectorized([left, right], fun) do
+    if left.vectorized_axes != [] or right.vectorized_axes != [] do
+      {left, right, vectorized_names} = unvectorize(left, right)
+      result = fun.(left, right)
+      Enum.reduce(vectorized_names, result, &vectorize(&2, &1))
```
josevalim, line 3707: We should have a (private?) version that vectorize n-dimensions at once!

nx/lib/nx.ex
```diff
@@ -3577,9 +3743,16 @@ defmodule Nx do
   end
 
   defp element_wise_pred_op(left, right, op) do
-    %T{shape: left_shape, names: left_names} = left = to_tensor(left)
-    %T{shape: right_shape, names: right_names} = right = to_tensor(right)
+    left = to_tensor(left)
+    right = to_tensor(right)
+    apply_vectorized([left, right], &unvectorized_element_wise_pred_op(&1, &2, op))
```
josevalim, line 3748: Move the `to_tensor` calls to `apply_vectorized` perhaps?

nx/lib/nx.ex
```diff
@@ -3546,11 +3585,179 @@ defmodule Nx do
...
+  @doc """
+  Transforms the leading dimension in a tensor into a vectorized axis.
+
+  Nested vectorizations will "remove" the leading axis from the shape
+  and append it to the vectorized axes
+
+  {2, 3, 4}
+  |> vectorize(:first) # => {3, 4} vec [first: 2]
+  |> vectorize(:second) # => {4} vec [first: 2, second: 3]
+  """
```
polvalente, line 3606: @josevalim ready for review again. I just need some pointers on what would be the ideal doc here.
Also, will we include a version that accepts a list of axes to vectorize at once as well?

josevalim (approved):
Let's use `__ENV__.function` instead of `__ENV__` to reduce footprint and ship it!

exla/lib/exla/defn.ex
```diff
@@ -1466,7 +1466,7 @@ defmodule EXLA.Defn do
   defp computation_key(op, args) do
     keys =
       Enum.map(args, fn
-        %Nx.Tensor{shape: shape, names: names, type: type} -> {type, shape, names}
+        %Nx.Tensor{shape: shape, names: names, type: type, vectorized_axes: vectorized_axes} -> {type, shape, names, vectorized_axes}
         opts -> opts
       end)
```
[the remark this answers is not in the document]
polvalente: (right below)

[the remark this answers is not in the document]
josevalim: Yes, we would need to have the unvectorized shape. So now I am thinking if we should unvectorize in `Nx.Defn` and here we always check it is an empty list.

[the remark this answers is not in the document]
josevalim: Ping me later :)

exla/test/exla/vectorize_test.exs
```diff
@@ -69,7 +68,13 @@ defmodule Nx.VectorizeTest do
     end
 
     test "Nx.acosh/1 works on vectorized tensor", %{base_math: base_math, vectorized_math: vectorized_math} do
-      fun = EXLA.jit(&(&1 |> Nx.add(1) |> Nx.acosh()))
+      # convert to f64 because there's some funky rounding errors for f32
+      # They are probably due to differences between adding a 3D tensor and a scalar
+      # versus adding a 3D tensor to another 3D tensor (which is what the scalar becomes)
+      # for the vectorized version -- albeit with shape {1, 1, 1})
```
josevalim, line 74: Instead of using f64, maybe we could do `Nx.add(Nx.tensor(1.0))`?

exla/lib/exla/defn.ex
```diff
@@ -368,12 +368,13 @@ defmodule EXLA.Defn do
...
+    {args_key, reverse_args_identifiers} =
       Enum.map_reduce(vars, [], fn var, acc ->
         Nx.Defn.Composite.traverse(var, acc, fn
-          %T{type: type, shape: shape, names: names}, acc ->
-            triplet = {type, shape, names}
-            {triplet, [triplet | acc]}
+          %T{} = t, acc ->
+            %T{type: type, shape: shape, names: names} = Nx.devectorize(t)
+            identifier = {type, shape, names}
+            {identifier, [identifier | acc]}
```
josevalim, line 377: I think we want the vectorized_axes to be part of the cache key (first argument) and the expanded one for the identifier. But you can also have return `{[identifier | vectorized_axes], [identifier | acc]}` and that would also work!

nx/lib/nx.ex
```diff
@@ -3546,11 +3597,316 @@ defmodule Nx do
...
+            ],
+            [
+              [4]
+            ],
+            [
+              [5]
+            ]
+          ]
+        ]
+      >
```
josevalim, line 3645: Can we have a smaller example so it does not take so many lines? Maybe rank=3?

nx/lib/nx.ex
```diff
@@ -3546,11 +3597,316 @@ defmodule Nx do
...
+
+  A vectorized tensor can be thought of as a nested list of tensors
+  that have the shape equal to the tensor's base shape. Operations
+  are then applied in a nested manner accordingly.
```
josevalim, line 3678: I am not sure if this explanation helps. :D  I would try something like:

    A vectorized axis applies each Nx operation to each entry in its dimension.

nx/lib/nx.ex
```diff
@@ -1713,6 +1726,15 @@ defmodule Nx do
   def to_binary(tensor, opts \\ []) do
     opts = keyword!(opts, [:limit])
     tensor = to_tensor(tensor)
+
+    # TODO: add proper implementation
```
josevalim, line 1730: What would be the proper implementation? In any case, I think we can add `flat_size`, which is the size including any vectorized dimension. Maybe that's what you mean?

josevalim (approved):
Beautiful! I just nitpicked the docs and we can ship it!

nx/lib/nx.ex
polvalente: Yes, exactly this. As it stands, `size` will return the base shape's size, so we need to devectorize the tensor.
`flat_size` seems like a nice addition

nx/lib/nx.ex
polvalente: rank=3 seems like a nice compromise

nx/lib/nx.ex
```diff
@@ -3546,11 +3597,316 @@ defmodule Nx do
...
+        ]
+      >
+
+  A vectorized tensor can be thought of as a nested list of tensors
+  that have the shape equal to the tensor's base shape. Operations
+  are then applied in a nested manner accordingly.
+
+  In the following example, notice that you don't need to have the
+  second argument shaped in a way that can be broadcasted
+  (i.e. shape {1, 2} can be broadcasted to {4, 2}, while {2} can't),
```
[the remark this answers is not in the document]
polvalente: Right, I'll fix this :)

I plan on adding `Nx.dot` as an example here as well when we have it implemented

PR #1120, opened 2023-03-15
feat: support vectorized versions for some unary ops

polvalente:


nx/lib/nx.ex
```diff
@@ -5234,21 +5250,12 @@ defmodule Nx do
   """
   @doc type: :element
   def logical_not(tensor) do
```
polvalente, line 5252: @josevalim Should I add docs for all of the functions implemented in this PR, or should we just leave it as a given?
Or maybe just for sort, argsort, top_k and reflect, which can be a bit trickier?

nx/lib/nx.ex
```diff
@@ -11516,32 +11516,35 @@ defmodule Nx do
...
-    opts = keyword!(opts, axis: 0, direction: :asc)
+    opts =
+      keyword!(opts, axis: 0, direction: :asc)
+      |> vectorized_offset_axis_option(tensor)
 
-    direction =
-      case opts[:direction] do
-        :asc ->
-          :asc
+    apply_vectorized([tensor], fn tensor ->
```
josevalim, line 11523: Another option is to receive the offset as argument:

    apply_vectorized_axes([tensor], fn tensor, offset ->

    end)

You only pass the offset if arity is 2. Then you add the offset once you compute the axis. It should be less work than manipulating options.

polvalente: I didn't really see a reason to change the generic function to accommodate 2 functions. Perhaps if more functions appear in the future we can do this change?

polvalente: I still don't see how this avoids manipulating options inside the function

polvalente: Maybe we should have `apply_vectorized(inputs, opts, function)` that fixes some options like `:padding_config`, `:axes` and `:axis` automatically and passes it down to the arity-2 function

nx/lib/nx.ex
```diff
@@ -3847,11 +3847,21 @@ defmodule Nx do
...
+        |> fun.()
+        |> revectorize_and_validate_sizes(vectorized_axes)
+      end
     else
-      fun.(tensor)
+      if is_function(fun, 1) do
+        fun.(tensor)
+      else
+        fun.(tensor, 0)
+      end
```
josevalim, line 3864:
```suggestion
      if is_function(fun, 2) do
        fun.(tensor, 0)
      else
        fun.(tensor)
      end
```

josevalim (approved):

PR #1134, opened 2023-03-21
feat: vectorize linspace and new_axis

polvalente:


nx/lib/nx.ex
```diff
@@ -12905,21 +12966,60 @@ defmodule Nx do
...
+
+        devec = devectorize(out)
+
+        inner_axes = count_up(tuple_size(out.shape), tuple_size(devec.shape) - 1)
+
+        transpose_axes = Enum.reverse(indices_rev) ++ inner_axes
+
+        devec
+        |> transpose(axes: transpose_axes)
+        |> revectorize_and_validate_sizes(vectorized_axes)
```
polvalente, line 13021: This here could be extracted to `transpose_vectorized_axes` or something like that, that acts like transpose, but _only_ on vectorized axes.

nx/lib/nx.ex
```diff
@@ -12905,21 +12966,60 @@ defmodule Nx do
...
-    iota
-    |> multiply(step)
-    |> add(start)
-    |> as_type(opts[:type])
+      out.vectorized_axes == vectorized_axes ->
+        out
+
+      true ->
+        nil
+        # we need to reorder the vectorized axes to the expected order
```
polvalente, line 12997:
```suggestion
        # we need to reorder the vectorized axes to the expected order according to the signature `linspace(start, stop, ...)`
```

nx/lib/nx.ex
```diff
@@ -2762,23 +2762,41 @@ defmodule Nx do
         ]
       >
 
+  ## Vectorized tensors
```
josevalim, line 2765: Perhaps this is a bit too late but, once we add such new sections, it would be nice to have a succinct description of how it behaves. For example:

```suggestion
  ## Vectorized tensors

  The new axis is added after the vectorized axes:

```

nx/lib/nx.ex
polvalente: @josevalim I tested this locally with the changes in #1136 

The only changes to linspace is that we would call `reshape_vectors` on `start` and `stop` instead of the `to_tensor` calls and we extract vectorized_axes from either to build the iota via the `:vectorized_axes` option.

We don't even need `broadcast_vectors` because the other operations already end up taking care of this implicitly for us!

josevalim (approved):

PR #1135, opened 2023-03-21
feat: upgrade to libtorch 2.0

polvalente:


torchx/c_src/torchx.cpp
```diff
@@ -1,5 +1,5 @@
 #include <torch/torch.h>
-#include <ATen/BatchedTensorImpl.h>
+#include <ATen/LegacyBatchedTensorImpl.h>
```
polvalente, line 2: Because the only mandatory change in this file is this, we could in theory still support libtorch 1.12 and therefore earlier CUDA as well. Thoughts?

josevalim: Supporting both for now sounds good to me.

polvalente: Yeah, especially now that I found out that macOS is still only supported for x86 by default.

I'll ping the maintainer for that repo we fetch from to see if we can get 2.0 there too

.github/workflows/ci.yml
```diff
@@ -55,21 +62,28 @@ jobs:
         run: mix test
 
   win:
-    name: Windows
+    name: Windows (${{ matrix.working_directory == 'torchx' && format('torchx({0})', matrix.torch_version) || matrix.working_directory  }}, ${{ matrix.elixir }}, ${{ matrix.otp }})
```
josevalim, line 65:
```suggestion
    name: Windows (${{ matrix.working_directory == 'torchx' && format('torchx({0})', matrix.torch_version) || matrix.working_directory }}, ${{ matrix.elixir }}, ${{ matrix.otp }})
```

josevalim (approved):

PR #1136, opened 2023-03-23
feat: reshape_vectors

polvalente:
This PR adds Nx.reshape_vectors for manipulating collections of vectors.
The PR also proposes an overhaul for the vocabulary in which we abolish
"vectorized tensor" in favor of "vector".

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -699,7 +699,7 @@ defmodule Nx do
   def tensor(arg, opts \\ [])
 
   def tensor(%Nx.Tensor{} = tensor, opts) do
-    opts = keyword!(opts, [:type, :names, :backend])
+    opts = keyword!(opts, [:type, :names, :backend, :vectorized_axes])
 
     tensor =
       if backend = opts[:backend] do
@@ -1281,9 +1281,9 @@ defmodule Nx do
         ]
       >
 
-  ## Vectorized tensors
+  ## Vectors
 
-  If given vectorized axes, are added as leading dimensions to the tensor,
+  If given, vectorized axes, are added as leading dimensions to the tensor,
   effectively broadcasting the base shape along them.
 
       iex> Nx.eye({3}, vectorized_axes: [x: 1, y: 2])
@@ -1555,7 +1555,7 @@ defmodule Nx do
         ]
       >
 
-  ### Vectorized tensors
+  ### Vectors
 
       iex> t = Nx.vectorize(Nx.tensor([[1, 2], [3, 4]]), :x)
       iex> Nx.make_diagonal(t, offset: 1)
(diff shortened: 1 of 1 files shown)


nx/lib/nx.ex
```diff
@@ -699,7 +699,7 @@ defmodule Nx do
   def tensor(arg, opts \\ [])
 
   def tensor(%Nx.Tensor{} = tensor, opts) do
-    opts = keyword!(opts, [:type, :names, :backend])
+    opts = keyword!(opts, [:type, :names, :backend, :vectorized_axes])
```
polvalente, line 702:
```suggestion
    opts = keyword!(opts, [:type, :names, :backend])
```

nx/lib/nx.ex
```diff
@@ -1281,9 +1281,9 @@ defmodule Nx do
         ]
       >
 
-  ## Vectorized tensors
+  ## Vectors
```
josevalim, line 1284: I prefer the previous titled. We already have `~V` each means one dimensional tensor. So calling it "Vectors" here is confusing. :(

nx/lib/nx.ex
```diff
@@ -3985,6 +4001,94 @@ defmodule Nx do
...
+        vectorized =
+          if same_order do
+            revectorize_and_validate_sizes(right, vectorized_axes, true)
+          else
+            reorder_vectorized_axes(right, vectorized_axes, left.vectorized_axes)
+          end
+
+        # we also want to ensure the inner shape is the same
+
+        reshape(vectorized, inner_shape)
```
josevalim, line 4086: Doesn't devectorize_with_axes already reshape?

nx/lib/nx.ex
```diff
@@ -3985,6 +4001,94 @@ defmodule Nx do
...
+    # For all tensors to be compatible, each pair also needs to be compatible
+    # This means that we can do a first pass accumulating axes into
+    # the first tensor, and then a second pass getting them all into their final shapes.
+    left =
+      Enum.reduce(tensors, first, fn right, left ->
+        {left, _, vectorized_axes} = devectorize_with_axes(left, to_tensor(right))
+        revectorize_and_validate_sizes(left, vectorized_axes, true)
+      end)
+
+    left = reshape(left, shape(first))
```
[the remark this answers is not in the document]
polvalente: This is for the non-vectorized shape. Note that it's reshape, not teshape_vectors

nx/lib/nx.ex
polvalente: The problem is that it reshapes in a broadcast able way, so the inner shape for a scalar might become {1} for instance.

At least this is what happened during my tests.
Might be a bug elsewhere though

nx/lib/nx.ex
josevalim: Let's keep it as vectorized tensors because vectors is really ambiguous.

nx/lib/nx.ex
```diff
@@ -3985,6 +4000,121 @@ defmodule Nx do
 
   def devectorize(tensor), do: tensor
 
+  @doc """
+  Reshapes input tensors so that they are all vectorized with the same vectors.
```
josevalim, line 4004: I think the most important here is the order. `reshape_vectors([x, y])` is different from `[y, x]` in the example below. so I would even change the second example to be `[y, x]`instead of the more complex example we have today?

polvalente: Done

josevalim (approved):

PR #1137, opened 2023-03-23
Fix parsing of the NumPy file header

lucaong:
If I save a NumPy array to a `npy` file on NumPy `v1.24.2` with `np.save(my_array)`, and then try to load it with `Nx.load_numpy!`, I get the following error:

```
** (CaseClauseError) no case clause matching: "'descr': '|O', 'fortran_order': False, 'shape': (), }                                                               \n"
    (nx 0.5.2) lib/nx.ex:11517: Nx.parse_header/1
    (nx 0.5.2) lib/nx.ex:11503: Nx.do_numpy_to_tensor/2
```

The issue is in the [`parse_header` private function](https://github.com/elixir-nx/nx/blob/fade169a1cf61c17eaf69754eaa8d99544795f63/nx/lib/nx.ex#L12488), which parses the `npy` file header, expecting the `}` to be the last character in the header. In this case, the `}` is instead followed by several spaces and a newline (see the first line in the error).

This is fixed by trimming trailing whitespace before removing the `}` in `parse_header`:

```elixir
# Before:
header = header |> String.trim("{") |> String.trim("}") |> String.trim(", ")

# Now:
header = header |> String.trim("{") |> String.trim() |> String.trim("}") |> String.trim(", ")
# Trim trailing whitespace here -------^
```

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -12485,7 +12485,7 @@ defmodule Nx do
   end
 
   defp parse_header(header) do
-    header = header |> String.trim("{") |> String.trim("}") |> String.trim(", ")
+    header = header |> String.trim("{") |> String.trim() |> String.trim("}") |> String.trim(", ")
 
     case header do
       "'descr': " <> <<dtype::size(5)-binary>> <> ", 'fortran_order': False, 'shape': " <> shape ->


josevalim:
Hi @lucaong! When reviewing your PR i saw an opportunity to review the existing code and I saw places for improvements. Can you please give de3544c5c97a3851a495568ce858985c78076d0c a try and let me know if it also addresses your issue?

lucaong:
Thank you @josevalim ! Your solution definitely looks cleaner. The tensor file now opens without problems :)

PR #1140, opened 2023-03-23
feat: vectorize broadcast, tile and squeeze

polvalente:


nx/lib/nx.ex
```diff
@@ -3097,33 +3163,137 @@ defmodule Nx do
         ]
       >
 
+  ## Vectorized tensors
+
+  Vectors remain unchanged, and normal broadcast rules apply otherwise.
```
josevalim, line 3168:
```suggestion
  Vectorized axes remain unchanged, and normal broadcast rules apply otherwise.
```

nx/lib/nx.ex
```diff
@@ -3097,33 +3163,137 @@ defmodule Nx do
...
+            [1, 2, 3]
+          ],
+          [
+            [4, 5, 6],
+            [4, 5, 6]
+          ]
+        ]
+      >
+
+  For tensor inputs, the broadcast will only take the shape in consideration.
```
josevalim, line 3187:
```suggestion
  For tensor shapes, the broadcast will only take the shape in consideration.
```

nx/lib/nx.ex
```diff
@@ -3097,33 +3163,137 @@ defmodule Nx do
...
+        vectorized[a: 1]
+        s64[2][y: 3]
+        [
+          [
+            [1, 2, 3],
+            [4, 5, 6]
+          ]
+        ]
+      >
+      iex> Nx.broadcast(a, b, axes: [1], names: [:i, :j])
```
josevalim, line 3209: Are we sure we don't want to vectorize the result with the `:a` as well?

josevalim (approved):

nx/lib/nx.ex
polvalente: I meant tensors as inputs in the second argument. Which also answers the question below regarding vectorizing with `:a` as well.

josevalim: I understood that. But it is a bit unclear because everything has input tensors. If we want to be clearer, we could say “tensors as shapes”?

polvalente: Yeah, that's better!

nx/lib/nx.ex
polvalente: I'm never sure haha!
Though we will have broadcast_vectors, so this behavior allows for a bit more flexibility

josevalim: good point, so let's ship this!

PR #1143, opened 2023-03-24
feat: vectorize Nx.compatible?

polvalente:


nx/lib/nx.ex
```diff
@@ -3586,14 +3586,53 @@ defmodule Nx do
       iex> Nx.compatible?(%{foo: Nx.iota({3, 2})}, %{bar: Nx.iota({3, 2})})
       false
 
+  ## Vectorized tensors
+
+  Same compatibility criteria applies to vectorized tensors, but there's
+  the additional requirement that if a vectorized axis name appears in both
+  tensors, it must appear with the same size, or one of them must be 1.
```
josevalim, line 3593: Vectorized axes must be the same size. We don't consider shape `{1}` to be compatible with `{10}`.

nx/lib/nx.ex
```diff
@@ -3586,14 +3586,53 @@ defmodule Nx do
...
   def compatible?(left, right)
 
-  def compatible?(%T{} = left, %T{} = right) do
-    Nx.Shared.raise_vectorized_not_implemented_yet(left, __ENV__.function)
-    Nx.Shared.raise_vectorized_not_implemented_yet(right, __ENV__.function)
+  def compatible?(
+        %T{type: type, shape: shape, names: l_names, vectorized_axes: l_axes},
+        %T{type: type, shape: shape, names: r_names, vectorized_axes: r_axes}
+      )
+      when l_axes != [] or r_axes != [] do
```
josevalim, line 3610: I think this should pretty much be `l_axes == r_axes`.

nx/lib/nx.ex
```diff
@@ -3586,14 +3586,31 @@ defmodule Nx do
...
   def compatible?(left, right)
 
-  def compatible?(%T{} = left, %T{} = right) do
-    Nx.Shared.raise_vectorized_not_implemented_yet(left, __ENV__.function)
-    Nx.Shared.raise_vectorized_not_implemented_yet(right, __ENV__.function)
+  def compatible?(
+        %T{type: type, shape: shape, names: l_names, vectorized_axes: l_axes},
+        %T{type: type, shape: shape, names: r_names, vectorized_axes: r_axes}
+      ) do
+    l_axes == r_axes and compatible_names?(l_names, r_names)
```
josevalim, line 3610:
```suggestion
        %T{type: type, shape: shape, names: l_names, vectorized_axes: v_axes},
        %T{type: type, shape: shape, names: r_names, vectorized_axes: v_axes}
      ) do
    compatible_names?(l_names, r_names)
```

josevalim (approved):

nx/lib/nx.ex
polvalente: This doesn't work because if axes are different, there's no clause that catches this

PR #1144, opened 2023-03-24
feat: vectorize rename

polvalente:


nx/lib/nx.ex
```diff
@@ -4181,6 +4205,13 @@ defmodule Nx do
     new_shape = Tuple.delete_at(shape, 0)
     names = tl(names)
 
+    for axis_name <- names do
+      if axis_name == name do
+        raise ArgumentError,
+              "cannot use name #{inspect(name)} for vectorized axes because there's already an axis with the same name."
+      end
+    end
```
josevalim, line 4213:
```suggestion
    for ^name <- names do
      raise ArgumentError,
            "cannot use name #{inspect(name)} for vectorized axes because there's already an axis with the same name"
    end
```

josevalim: If you find it too magical, you can do `axis_name <- names, name == axis_name`

polvalente: Nah, this is fine :)

josevalim (approved):

PR #1152, opened 2023-03-24
feat: vectorize (most of) Nx.LinAlg

polvalente:
Vectorizes all of the non-defn functions in Nx.LinAlg with the exception of:

`norm`: one of the inner implementations uses `select/3`, which still needs vectorization
`svd`: The default implementation uses `dot`, which still needs to be vectorized properly.

Also adds more functionality to `Nx.vectorize` so that it can act as an inverse function of `Nx.devectorize` with ease

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -4311,6 +4311,8 @@ defmodule Nx do
           Nx.Tensor.t()
   def vectorize(tensor, name_or_axes)
 
+  def vectorize(%Nx.Tensor{shape: {}} = tensor, []), do: tensor
+
   def vectorize(%Nx.Tensor{shape: {}}, _name) do
     raise ArgumentError, "cannot vectorize tensor of rank 0"
   end
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -4177,8 +4177,10 @@ defmodule Nx do
   Each vectorization removes the leading axes from the shape and appends them to
   the `:vectorized_axes` list for the tensor.
 
-  The vectorization specification can be a list of `t:atom()` or `{t:atom(), t:pos_integer()}`,
-  or a single atom, which will behave the same as a single-element list.
+  The vectorization specification can be a list of atoms or `{atom, pos_integer}`
+  pairs. If a single atom is given, it behaves as a single-element list.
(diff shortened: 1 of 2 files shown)


nx/lib/nx.ex
```diff
@@ -4174,8 +4174,11 @@ defmodule Nx do
   @doc """
   Transforms a tensor into a vectorized tensor.
 
-  Each vectorization removes the leading axis from the shape and appends it to
-  `:vectorized_axes` list for the tensor.
+  Each vectorization removes the leading axes from the shape and appends them to
+  the `:vectorized_axes` list for the tensor.
+
+  The vectorization specification can be a list of `t:atom()` or `{t:atom(), t:pos_integer()}`,
+  or a single atom, which will behave the same as a single-element list.
```
josevalim, line 4181:
```suggestion
  The vectorization specification can be a list of atoms or `{atom, pos_integer}`
  pairs. If a single atom is given, it behaves as a single-element list.
  The atom names the vectorized axes. If a pair is given, we also verify
  that the given size matches the size of the to-be-vectorized axis. 
```

nx/lib/nx.ex
```diff
@@ -4205,6 +4208,25 @@ defmodule Nx do
         ]
       >
 
+  This can also be written in similar ways with list inputs.
+
+  The first way below doesn't validate sizes.
+  The second way, ensures that the second axis has size `3`.
```
josevalim, line 4214:
```suggestion
  You can also vectorize multiple axes at once by passing a list,
  as seen in the examples below. The first example doesn't validate
  sizes. The second ensures the second axis has size `3`.
```

josevalim (approved):

nx/lib/nx/lin_alg.ex
```diff
@@ -1883,4 +2076,27 @@ defmodule Nx.LinAlg do
     # Calculate matrix rank
     Nx.sum(s > tol)
   end
+
+  defp apply_vectorized(tensor, fun) when is_function(fun, 1) do
```
josevalim, line 2080: Shouldn't we also write our own `apply_vectorized` in a way similar to this?

polvalente: I believe we can!
Originally we didn't have the proper tools in place :)

PR #1153, opened 2023-03-25
feat: vectorize select

polvalente:


nx/lib/nx.ex
```diff
@@ -6365,28 +6365,101 @@ defmodule Nx do
...
+          ],
+          [
+            [1, 1, 2],
+            [1, 1, 3]
+          ]
+        ]
+      >
+
+    In the next example, notice that even though the `pred` input
+    is scalar, because we're dealing with vectorize inputs, some
```
josevalim, line 6393:
```suggestion
    is scalar, because we're dealing with vectorized inputs, some
```

nx/lib/nx.ex
```diff
@@ -6365,28 +6365,101 @@ defmodule Nx do
...
 
     {output_shape, output_names} =
-      case pred_shape do
-        {} ->
+      case {vectorized_axes, pred_shape} do
+        {[], {}} ->
           Nx.Shape.binary_broadcast(true_shape, true_names, false_shape, false_names)
 
         _ ->
-          {pred_shape, pred_names}
```
josevalim, line 6387: Should we still match on `{[], _}` and return `{pred_shape, pred_names}`?

nx/lib/nx.ex
```diff
@@ -6401,8 +6474,36 @@ defmodule Nx do
         Nx.Shape.broadcast_axes(false_shape, output_shape)
       )
 
-    out = %{pred | shape: output_shape, type: output_type, names: output_names}
-    impl!(pred, on_true, on_false).select(out, pred, on_true, on_false)
+    out = devectorize(%{pred | shape: output_shape, type: output_type, names: output_names})
+
+    pred = devectorize(pred)
```
josevalim, line 6479: Maybe invert the order of these two operations so you don't have to devectorize `out`?

josevalim, line 6479: Maybe invert the order of these two operations so you don't have to devectorize `out`?

```suggestion
    pred = devectorize(pred)
    out = %{pred | shape: output_shape, type: output_type, names: output_names}
```

?

josevalim (approved):

nx/lib/nx.ex
polvalente: Makes sense. Avoids some work :)

nx/lib/nx.ex
polvalente: We'd have to recalculate `output_shape` if we invert the order

PR #1155, opened 2023-03-25
feat: vectorize dot

polvalente:


nx/lib/nx.ex
```diff
@@ -10504,6 +10534,43 @@ defmodule Nx do
         50.0
       >
 
+  ## Vectorized tensors
+
+  The contracting axes refer to the tensors' shapes
```
josevalim, line 10539:
```suggestion
  The contracting axes refer to the tensors' shapes
  and do not apply to the vectorized axes:
```

nx/lib/nx.ex
```diff
@@ -10694,7 +10799,19 @@ defmodule Nx do
...
+    t1 = devectorize(t1)
+    t2 = devectorize(t2)
+    out = devectorize(out)
+
+    c1 = Enum.map(c1, &(&1 + offset))
+    c2 = Enum.map(c2, &(&1 + offset))
+    b1 = offset_axes ++ Enum.map(b1, &(&1 + offset))
+    b2 = offset_axes ++ Enum.map(b2, &(&1 + offset))
+
+    res = impl!(t1, t2).dot(out, t1, c1, b1, t2, c2, b2)
```
josevalim, line 10812: Let's also wrap this on `if offset > 0` to avoid additional work unless necessary.

josevalim (approved):

PR #1158, opened 2023-03-25
feat: vectorize concatenate

polvalente:
also simplifies some of the implementation for reshape_/broadcast_vectors

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -4180,7 +4180,7 @@ defmodule Nx do
   The vectorization specification can be a list of atoms or `{atom, pos_integer}`
   pairs. If a single atom is given, it behaves as a single-element list.
   The atom names the vectorized axes. If a pair is given, we also verify
-  that the given size matches the size of the to-be-vectorized axis. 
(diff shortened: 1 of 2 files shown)


nx/lib/nx.ex
```diff
@@ -12450,22 +12469,29 @@ defmodule Nx do
...
         t
 
-      [t1 | _] = tensors ->
+      [_ | _] = tensors ->
+        [%T{vectorized_axes: vectorized_axes} | _] = tensors = broadcast_vectors(tensors)
+
+        offset = length(vectorized_axes)
+
+        [t1 | _] =
+          tensors = if vectorized_axes != [], do: Enum.map(tensors, &devectorize/1), else: tensors
```
josevalim, line 12478: couldn't the first tensor not be vectorized but the other ones are?

polvalente: Not in this case, because we're checking after the broadcast_vectors

polvalente: So they all have the same vectorized axes

josevalim (approved):

PR #1160, opened 2023-03-25
feat: vectorize slice/put_slice

polvalente:


nx/lib/nx.ex
```diff
@@ -12096,27 +12157,72 @@ defmodule Nx do
...
+
+    output_names = List.duplicate(nil, offset) ++ output_names
+
+    result =
+      impl!(tensor).put_slice(
+        %{tensor | shape: output_shape_devec, names: output_names, type: output_type},
+        tensor,
+        start_indices,
+        slice
+      )
```
josevalim, line 12223: It feels we can share logic with slice but it is your call if it makes sense. :)

josevalim (approved):

nx/lib/nx.ex
polvalente: I'll merge it as is. Even though there are common sections of the code, I think extracting the would make for less readable code because we'd have to intersperse the extracted call with the non-common sections

PR #1162, opened 2023-03-26
feat: vectorize take, take_along_axis and gather

polvalente:


nx/lib/nx.ex
```diff
@@ -12592,6 +12684,7 @@ defmodule Nx do
         [4, 2, 3]
       >
 
+
```
josevalim, line 12687:
```suggestion
```

nx/lib/nx.ex
```diff
@@ -12609,26 +12702,78 @@ defmodule Nx do
...
+    iota_shape =
+      indices.shape |> Tuple.delete_at(tuple_size(indices.shape) - 1) |> Tuple.append(1)
+
+    offset_axes = (offset - 1)..0//-1
+
+    indices =
+      offset_axes
+      |> Enum.reduce([indices], &[Nx.iota(iota_shape, axis: &1) | &2])
+      |> concatenate(axis: -1)
```
josevalim, line 12770: Just to be safe, let's only do this if `offset != 0`, wdyt?

josevalim: Safe in terms of performance. I know concatenate will simply return itself and offset_axes is zero, but it is good to skip the extra calls anyway. :)

josevalim (approved):

PR #1168, opened 2023-03-27
feat: vectorize svd + norm and some fixes

polvalente:


nx/lib/nx/defn/evaluator.ex
```diff
@@ -415,7 +415,25 @@ defmodule Nx.Defn.Evaluator do
...
+    result =
+      case apply(mod, op, args) do
+        %Nx.Tensor{} = t ->
+          Nx.vectorize(t, ans.vectorized_axes)
+
+        t ->
+          t
+      end
+
+    {result, caches}
```
polvalente, line 436: I don't think this is totally correct, but tests are passing

nx/lib/nx/lin_alg.ex
```diff
@@ -1422,7 +1421,23 @@ defmodule Nx.LinAlg do
...
+    {u, s, vt} =
+      Nx.Shared.optional(:svd, [tensor, opts], output, fn tensor, opts ->
+        {u, s, vt} = Nx.LinAlg.SVD.svd(tensor, opts)
+
+        # THIS IS A BUG, DO NOT MERGE AS IS
+        {
+          Nx.devectorize(u, keep_names: false),
+          Nx.devectorize(s, keep_names: false),
+          Nx.devectorize(vt, keep_names: false)
+        }
```
polvalente, line 1433: Somehow the tuple returned from the SVD defn call is not devectorized, so we have at least this bug and the one I'll point out below to prune before merging

nx/lib/nx/lin_alg/svd.ex
```diff
@@ -24,16 +24,53 @@ defmodule Nx.LinAlg.SVD do
...
+    result = {u, s, vt}
+
+    {u, s, vt} =
+      custom_grad(result, [input_tensor], fn g ->
+        svd_grad(result, input_tensor, g)
+      end)
+
+    u = Nx.devectorize(u, keep_names: false)
+    s = Nx.devectorize(s, keep_names: false)
+    vt = Nx.devectorize(vt, keep_names: false)
```
polvalente, line 47: Somehow each time we put vectorized values into a tuple and then extract them back, the rightmost value is coming back devectorized with `keep_names: true`, so these devectorize calls won't work correctly

polvalente: After we fix this I'll add more tests because we still need to "uncollapse" batch axes after we devectorize here

nx/lib/nx/defn/evaluator.ex
```diff
@@ -426,7 +428,9 @@ defmodule Nx.Defn.Evaluator do
           {Nx.Shared.list_impl!(args), [ans | args]}
       end
 
-    {apply(mod, op, args), caches}
+    result = apply(mod, op, args)
+
+    {result, caches}
```
polvalente, line 433:
```suggestion
    {apply(mod, op, args), caches}
```

nx/lib/nx/defn/evaluator.ex
```diff
@@ -402,6 +402,8 @@ defmodule Nx.Defn.Evaluator do
   defp eval_apply(op, ans, state, caches) do
     {args, caches} = Tree.apply_args(ans, caches, &eval(&1, state, &2))
 
+    ans = Nx.devectorize(ans, keep_names: false)
```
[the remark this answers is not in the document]
josevalim: I believe the vectorized axes is coming from `while`, so we need to address it first. :)

nx/lib/nx/defn/evaluator.ex
```diff
@@ -83,7 +83,10 @@ defmodule Nx.Defn.Evaluator do
     expr = fun.(vars)
     state = %{hooks: hooks, parent_ids: nil, current_ids: nil}
     cache = init_compute_cache(expr, state)
-    {expr, output} = Nx.Defn.Composite.traverse(expr, [], &{Nx.devectorize(&1), [&1 | &2]})
+
```
polvalente, line 86: Extracted these evaluator fixes to #1180

nx/lib/nx.ex
```diff
@@ -2659,16 +2659,21 @@ defmodule Nx do
...
     names = Nx.Shape.named_axes!(new_names, new_shape)
 
     cond do
-      old_shape == new_shape ->
-        %{tensor | names: names}
+      old_shape == new_shape and names == old_names ->
+        # we can't assume that a given backend does not register names for tensor
+        # shapes, so we can only return a no-op if both the shape and the names
+        # are unchanged
+        tensor
```
polvalente, line 2676: @josevalim 
Another fix that smells like I'm fixing the wrong place, even though this one is more justifiable.

Basically, what's happening is that at the `expand_collapsed_axes` wasn't taking effect properly in the case that the shape was the same but I wanted to change the names of the expression.

On the other hand, it does make sense that if we were to register axes' names in a given backend, suddenly receiving a new unknown name would yield a bug, so I do think this is a structural change we need to apply.

josevalim: It is expected that backends do not register names. Names are a Nx only construct?

josevalim:
This PR looks good but I don't think we should merge it for two reasons:

1. We are moving from `cond` to `select`

2. We are adding broadcasting logic to `svd` and I think we should explore something like `Nx.vectorize_above_rank(...)`

In other words, let's continue using this PR as an opportunity to improve other areas of the library. :)

polvalente:
> This PR looks good but I don't think we should merge it for two reasons:
> 
> 1. We are moving from `cond` to `select`
> 

I think this is ok, though. There are other SVD fixes that are a bit blocked by this PR, and I really don't know how easy it'll be to do vectorized cond. 

> 2. We are adding broadcasting logic to `svd` and I think we should explore something like `Nx.vectorize_above_rank(...)`
> 

Keep in mind that most of the broadcasting, vectorization and overall changes to the SVD module are to support batching *via* vectorization, not the other way around like with other functions.

> In other words, let's continue using this PR as an opportunity to improve other areas of the library. :)

nx/lib/nx.ex
polvalente: I didn't know that. Though if I remove this, the name from vectorization leaks, so I think that we're not dealing with names correctly in the Expr evaluation then.

josevalim: Can you provide a small test case?

josevalim:
> I think this is ok, though. There are other SVD fixes that are a bit blocked by this PR, and I really don't know how easy it'll be to do vectorized cond.

Ok! :)

> Keep in mind that most of the broadcasting, vectorization and overall changes to the SVD module are to support batching via vectorization, not the other way around like with other functions.

I understand that but I would like to encapsulate that, as it will be certainly useful for other libraries that needs to do the same, such as SVD and Scholar.

polvalente:
> I understand that but I would like to encapsulate that, as it will be certainly useful for other libraries that needs to do the same, such as Nx.Signal and Scholar.

By that, you mean the functionality to compress all but the last `n` axes into a single vectorized axis and then de-compress them back into the original format?

Any ideas on naming? `contract_into_vectors`/`expand_from_vectors`?

nx/lib/nx.ex
polvalente:
```elixir
iex(10)> defmodule M do
...(10)>     import Nx.Defn
...(10)>     
...(10)>     defn f(%{shape: {m, n}} = x) do
...(10)>         %{shape: {k, m, n}} = devec = Nx.devectorize(x)
...(10)>         
...(10)>         {
...(10)>             Nx.reshape(devec, {k, m, n}, names: [nil, nil, nil]), 
...(10)>             Nx.reshape(devec, {k, m, n}, names: [:y, nil, nil]),
...(10)>             Nx.reshape(devec, {k, 1, m, n}, names: [:y, nil, nil, nil])
...(10)>         }
...(10)>     end
...(10)>     
...(10)>     defn g(%{shape: {m, n}} = x) do
...(10)>         %{shape: {k, m, n}} = devec = Nx.devectorize(x)
...(10)>         
...(10)>         Nx.reshape(devec, {k, m, n}, names: [nil, nil, nil])
...(10)>     end
...(10)> end
warning: redefining module M (current version defined in memory)
  iex:10

warning: variable "m" is unused (if the variable is not meant to be used, prefix it with an underscore)
  iex:13: M.f/1

warning: variable "n" is unused (if the variable is not meant to be used, prefix it with an underscore)
  iex:13: M.f/1

warning: variable "m" is unused (if the variable is not meant to be used, prefix it with an underscore)
  iex:23: M.g/1

warning: variable "n" is unused (if the variable is not meant to be used, prefix it with an underscore)
  iex:23: M.g/1

{:module, M,
 <<70, 79, 82, 49, 0, 0, 12, 28, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 1, 39,
   0, 0, 0, 30, 8, 69, 108, 105, 120, 105, 114, 46, 77, 8, 95, 95, 105, 110,
   102, 111, 95, 95, 10, 97, 116, 116, 114, ...>>, true}
iex(11)>  M.f(Nx.iota({2, 3}, vectorized_axes: [x: 1])) 
{#Nx.Tensor<
   s64[x: 1][2][3]
   [
     [
       [0, 1, 2],
       [3, 4, 5]
     ]
   ]
 >,
 #Nx.Tensor<
   s64[x: 1][2][3]
   [
     [
       [0, 1, 2],
       [3, 4, 5]
     ]
   ]
 >,
 #Nx.Tensor<
   s64[y: 1][1][2][3]
   [
     [
       [
         [0, 1, 2],
         [3, 4, 5]
       ]
     ]
   ]
 >}
iex(12)>  M.g(Nx.iota({2, 3}, vectorized_axes: [x: 1])) 
#Nx.Tensor<
  s64[x: 1][2][3]
  [
    [
      [0, 1, 2],
      [3, 4, 5]
    ]
  ]
>
```


Here you go. I ran this on Nx main. Notice that from the output of `f`, the only case where `:names` took effect was the one in which the shape changed, and therefore the Nx.Defn.Expr backend was able to take notice of the name change.

The output of `g` shows one of the same cases, just as a sanity check that it's not related to the output being a tuple.

Another fix for this would be going into the Nx.Defn.Evaluator and changing the `apply_output` function to also apply the `output` names into the `result` tensor, but I think this is something that would have to be done by every compiler.

polvalente: After changing `apply_output` to the code below (only actual change is the addition of `Nx.rename/2`)

```elixir
defp apply_output({result, _cache}, output) do
    {result, []} =
      Composite.traverse(result, output, fn
        result, [%Nx.Tensor{vectorized_axes: vectorized_axes, names: names} | acc] ->
          corrected_result = result |> Nx.vectorize(vectorized_axes) |> Nx.rename(names)
          {corrected_result, acc}
      end)

    result
  end
```

we get the correct result:

```elixir
iex(24)> t =  M.g(Nx.iota({2, 3}, vectorized_axes: [x: 1])) 
#Nx.Tensor<
  s64[1][2][3]
  [
    [
      [0, 1, 2],
      [3, 4, 5]
    ]
  ]
>
iex(25)> t =  M.f(Nx.iota({2, 3}, vectorized_axes: [x: 1])) 
{#Nx.Tensor<
   s64[1][2][3]
   [
     [
       [0, 1, 2],
       [3, 4, 5]
     ]
   ]
 >,
 #Nx.Tensor<
   s64[y: 1][2][3]
   [
     [
       [0, 1, 2],
       [3, 4, 5]
     ]
   ]
 >,
 #Nx.Tensor<
   s64[y: 1][1][2][3]
   [
     [
       [
         [0, 1, 2],
         [3, 4, 5]
       ]
     ]
   ]
 >}
 ```
 
 In other news, EXLA does work correctly, because it already forces the buffer results into the output holders which carry the axis name info, so perhaps the way to go is to apply the fix in Defn.Evaluator only :)
 
 ```
 iex(3)> t =  M.f(Nx.iota({2, 3}, vectorized_axes: [x: 1]))                     

19:13:02.598 [info] domain=elixir.xla file=tensorflow/compiler/xla/pjrt/tfrt_cpu_pjrt_client.cc line=214  TfrtCpuClient created.
{#Nx.Tensor<
   s64[1][2][3]
   EXLA.Backend<host:0, 0.200170718.1356202002.226606>
   [
     [
       [0, 1, 2],
       [3, 4, 5]
     ]
   ]
 >,
 #Nx.Tensor<
   s64[y: 1][2][3]
   EXLA.Backend<host:0, 0.200170718.1356202002.226607>
   [
     [
       [0, 1, 2],
       [3, 4, 5]
     ]
   ]
 >,
 #Nx.Tensor<
   s64[y: 1][1][2][3]
   EXLA.Backend<host:0, 0.200170718.1356202002.226608>
   [
     [
       [
         [0, 1, 2],
         [3, 4, 5]
       ]
     ]
   ]
 >}
 ```

Comment (josevalim) [reply nx/lib/nx.ex:2676]: Yes, your changes to apply_output are correct!

Comment (josevalim) [reply nx/lib/nx.ex:2676]: We can also mirror EXLA and do `%{output | data: tensor.data}` (i.e. replace the data instead of vectorize+rename).

File: nx/lib/nx/lin_alg/svd.ex [hunk 8]
```diff
@@ -24,22 +24,59 @@ defmodule Nx.LinAlg.SVD do
...
+    m = elem(shape, rank - 2)
+    n = elem(shape, rank - 1)
+
+    collapsed_axes = shape |> Tuple.delete_at(rank - 2) |> Tuple.delete_at(rank - 2)
+
+    u_shape = collapsed_axes |> Tuple.append(m) |> Tuple.append(:auto)
+    s_shape = Tuple.append(collapsed_axes, :auto)
+    vt_shape = Tuple.append(s_shape, n)
+
+    revectorize = collapsed_axes != {}
```
polvalente, line 76: @josevalim the PR is now mostly done! The only thing for which I couldn't find a root cause is that if we remove this, there are some cases in which we get "unexpected vectorized_axes" raises in the evaluator. Could you take a look at this?

PR #1173, opened 2023-03-27
feat: vectorize/devectorize containers

polvalente:


nx/lib/nx.ex
```diff
@@ -4396,6 +4396,39 @@ defmodule Nx do
         ]
       >
 
+  ## Containers
```
josevalim, line 4399: This is a too verbose doctest. I think I would do a basic test with a two element tuple and leave the rest to actual test files. :)

nx/lib/nx.ex
```diff
@@ -4550,7 +4644,25 @@ defmodule Nx do
...
+      when is_struct(tensor, Complex)
+      when is_number(tensor),
+      do: tensor
+
+  def devectorize(container, opts) do
+    {result, nil} =
+      Nx.Container.traverse(container, nil, fn
+        item, _ when is_number(item) or is_struct(item, Complex) ->
+          result = to_tensor(item)
+          {result, nil}
```
josevalim, line 4658: Here we convert numbers and complex to tensors but not in the clause above.

nx/lib/nx.ex
```diff
@@ -4528,9 +4581,50 @@ defmodule Nx do
           ]
         ]
       >
+
+  ## Containers
+
+  Containers are also supported:
```
josevalim, line 4587: Let’s simplify this example too.

nx/lib/nx.ex
```diff
@@ -4486,6 +4523,22 @@ defmodule Nx do
...
+    {result, nil} =
+      Nx.Container.traverse(container, nil, fn
+        item, [] when is_tensor(item) ->
+          to_tensor(item)
+
+        item, _ when item.shape == {} when is_number(item) when is_struct(item, Complex) ->
+          raise ArgumentError, "cannot vectorize tensor of rank 0"
+
+        item, _ ->
+          {vectorize(item, vectorized_axes), nil}
```
josevalim, line 4536: I think we can just call vectorize for all types here and it will work?

nx/lib/nx.ex
josevalim: I would always convert to tensor and then this can be recursive too!

josevalim (approved):

nx/lib/nx.ex
```diff
@@ -4550,7 +4598,21 @@ defmodule Nx do
     %{tensor | shape: output_shape, names: output_names, vectorized_axes: []}
   end
 
-  def devectorize(tensor, _), do: tensor
+  def devectorize(tensor, _)
+      when tensor.vectorized_axes == []
+      when is_struct(tensor, Complex)
+      when is_number(tensor),
+      do: tensor
```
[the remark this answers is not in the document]
polvalente: Yes, just so it works better with containers. I don't want vectorize to be the same because we have the check for scalars there that we don't have here.

An option to consistency between both functions: we can open an exception on vectorize to plain numbers, though, but I would find that behavior a bit strange, or do to_tensor here.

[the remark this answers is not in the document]
polvalente: I'll merge this, but if you prefer one of the alternatives, I can send a separate PR on this

PR #1175, opened 2023-03-29
Some fixes to Nx.Serving's docs

hugobarauna:
I just started learning about how to use `Nx.Serving` by reading its docs. And I may have found some problems in the doc, but I'm not sure how to fix them.

I'm opening this PR to show some of the places where the doc may be outdated.

### DIFF
--- a/nx/lib/nx/serving.ex
+++ b/nx/lib/nx/serving.ex
@@ -111,7 +111,7 @@ defmodule Nx.Serving do
 
       children = [
         {Nx.Serving,
-         serving: Nx.Serving.new(Nx.Serving.new(fn opts ->  Nx.Defn.jit(&MyDefn.print_and_multiply/1, opts) end)),
+         serving: Nx.Serving.new(fn opts ->  Nx.Defn.jit(&MyDefn.print_and_multiply/1, opts) end),
          name: MyServing,
          batch_size: 10,
          batch_timeout: 100}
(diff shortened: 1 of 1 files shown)


nx/lib/nx/serving.ex
```diff
@@ -111,7 +111,7 @@ defmodule Nx.Serving do
 
       children = [
         {Nx.Serving,
-         serving: Nx.Serving.new(Nx.Defn.jit(&print_and_multiply/1)),
+         serving: Nx.Serving.new(Nx.Serving.new(fn opts ->  Nx.Defn.jit(&MyDefn.print_and_multiply/1, opts) end)),
```
josevalim, line 114:
```suggestion
         serving: Nx.Serving.new(fn opts ->  Nx.Defn.jit(&MyDefn.print_and_multiply/1, opts) end),
```

josevalim:

nx/lib/nx/serving.ex
```diff
@@ -19,7 +19,7 @@ defmodule Nx.Serving do
         import Nx.Defn
 
         defnp print_and_multiply(x) do
-          print_value({:debug, x})
+          print_value(x, label: "debug")
```
hugobarauna, line 22: @josevalim there's still a problem with using `print_value` here.

When I tried it on my machine, it printed nothing.

Given [its docs](https://hexdocs.pm/nx/Nx.Defn.Kernel.html#print_value/2):

```
The return value of this function must be part of the output
```

I think that's because it's not the last expression inside the `print_and_multiply` function.

PR #1180, opened 2023-04-03
Vectorize defn while

polvalente:
The changes to cond and while are temporary fixes until we support actual vectorization on those. As it stands, they still need the conditions to be scalars, but the clauses can be vectorized

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -4648,6 +4648,36 @@ defmodule Nx do
       [y: 1, x: 3]
       iex> yv.vectorized_axes
       [y: 1, x: 1]
+
+  Containers are also supported:
+
+      iex> x = Nx.iota({2}, vectorized_axes: [x: 2])
+      iex> y = Nx.iota({3}, vectorized_axes: [y: 2])
+      iex> [broadcast_x, {broadcast_y}] = Nx.reshape_vectors([x, {y}])
+      iex> broadcast_x
+      #Nx.Tensor<
+        vectorized[x: 2][y: 1]
+        s64[2]
+        [
+          [
+            [0, 1]
+          ],
+          [
+            [0, 1]
+          ]
+        ]
+      >
+      iex> broadcast_y
+      #Nx.Tensor<
+        vectorized[x: 1][y: 2]
+        s64[3]
+        [
+          [
+            [0, 1, 2],
+            [0, 1, 2]
+          ]
+        ]
+      >
   """
   @doc type: :shape
   def reshape_vectors(tensors)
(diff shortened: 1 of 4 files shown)


nx/lib/nx.ex
```diff
@@ -4742,72 +4772,136 @@ defmodule Nx do
...
+          ]
+        ]
+      >
   """
   @doc type: :shape
   def broadcast_vectors([t]), do: [to_tensor(t)]
 
-  def broadcast_vectors(tensors) when is_list(tensors) do
-    {devectorized_tensors, target_vectorized_axes, offset} = do_reshape_vectors(tensors)
+  def broadcast_vectors(tensors_or_containers) when is_list(tensors_or_containers) do
```
polvalente, line 4813: As always, we can split these changes into a separate PR. Just added them as needed here

nx/lib/nx/defn/expr.ex
```diff
@@ -129,14 +132,29 @@ defmodule Nx.Defn.Expr do
...
+        _expr, original
+        when is_number(original)
+        when original in [:infinity, :neg_infinity, :nan]
+        when is_struct(original, Complex) ->
+          original
+
+        expr, _original ->
+          Nx.devectorize(expr, keep_names: false)
+      end)
+
```
polvalente, line 157: This doesn't support keeping scalars inside containers yet. I'm not sure how to "zip" containers or to fit a flat list back into a container.

nx/test/nx/defn/evaluator_test.exs
```diff
@@ -620,4 +620,52 @@ defmodule Nx.Defn.EvaluatorTest do
...
+    test "cond" do
+      a = Nx.tensor(3)
+      b = Nx.tensor(2)
+      c = Nx.vectorize(Nx.tensor([[1]]), x: 1, y: 1)
+
+      # This shows that non-vectorized branches still return unvectorized.
+      assert Nx.add(c, 2) == vectorized_cond(1, a, 0, b, c)
+      assert Nx.add(c, 1) == vectorized_cond(0, a, 1, b, c)
+      assert c == vectorized_cond(0, a, 0, b, c)
+    end
```
polvalente, line 664: This test is failing now. Any ideas @josevalim ?

polvalente: apart from this and the container thing above, everything seems to be working correctly

nx/lib/nx/defn/expr.ex
josevalim: I think this is in the wrong place. The code right below is doing flattening and broadcasting and raising a good error message in case of mismatched clauses. Perhaps we could move the devectorization logic in there instead?

nx/test/nx/defn/evaluator_test.exs
josevalim: I think it should fail. All branches should return the same shape because the branches are executed at compile time and, at compile time, you don't know which branch will win. We should probably raise if they are not equally vectorized.

josevalim: For while we have a similar issue. What happens if `initial` is vectorized and its return from the while `body` is not? It means the value is changing and becoming incompatible.

So I think, ultimately, we want to make sure they are compatible, and only devectorize if they are all equally vectorized!

nx/lib/nx/defn/expr.ex
```diff
@@ -166,6 +204,7 @@ defmodule Nx.Defn.Expr do
       expr
       |> Nx.as_type(type)
       |> Nx.broadcast(shape, names: names)
+      |> Nx.devectorize(keep_names: false)
```
polvalente, line 207: @josevalim I found a problem here.
vectorized broadcasting has an "axis completion" algorithm for broadcasting scalars against vectorized tensors.
The easiest way to deal with this is to keep `broadcast_vectors` above instead of `reshape_vectors`.

As the code is commited, it doesn't deal with branches with different vectorized axes, as per the single failing test we have.

Otherwise, I think we would need to replicate most of the logic for broadcast_vectors and some of the logic for broadcast.

WDYT?

nx/lib/nx/defn/expr.ex
```diff
@@ -92,7 +92,10 @@ defmodule Nx.Defn.Expr do
   inspection.
   """
   def metadata(expr, metadata) when is_map(metadata) do
-    case to_container_expr(expr) do
+    expr
+    |> to_container_expr()
+    |> Nx.devectorize()
```
josevalim, line 97: We may be disappearing with the vectorization on this one.

nx/lib/nx/defn/expr.ex
```diff
@@ -213,7 +213,14 @@ defmodule Nx.Defn.Expr do
...
-    args = [flatten_initial, flatten_arg, condition, flatten_body]
+
+    args = [
+      Nx.devectorize(flatten_initial, keep_names: false),
+      flatten_arg,
+      Nx.devectorize(condition, keep_names: false),
+      Nx.devectorize(flatten_body, keep_names: false)
+    ]
+
     flatten_to_composite(initial, context, clauses, &expr(&1, context, :while, args))
```
josevalim, line 224: We need to make sure the results of while are still vectorized if `flatten_initial` or `flatten_body` (they should match) are vectorized.

nx/lib/nx/defn/expr.ex
```diff
@@ -213,7 +213,14 @@ defmodule Nx.Defn.Expr do
   """
   def while(initial, context, arg, condition, body) do
     [flatten_initial, flatten_arg, flatten_body] = clauses = flatten_clauses([initial, arg, body])
-    args = [flatten_initial, flatten_arg, condition, flatten_body]
+
+    args = [
+      Nx.devectorize(flatten_initial, keep_names: false),
+      flatten_arg,
+      Nx.devectorize(condition, keep_names: false),
```
josevalim, line 220: We should raise if the condition is vectorized. The same for the cond condition. We should raise if they are vectorized because the correct is to handle them properly.

nx/lib/nx/defn/expr.ex
```diff
@@ -213,7 +213,14 @@ defmodule Nx.Defn.Expr do
   """
   def while(initial, context, arg, condition, body) do
     [flatten_initial, flatten_arg, flatten_body] = clauses = flatten_clauses([initial, arg, body])
-    args = [flatten_initial, flatten_arg, condition, flatten_body]
+
+    args = [
+      Nx.devectorize(flatten_initial, keep_names: false),
+      flatten_arg,
```
josevalim, line 219: We should vectorize this one.

nx/lib/nx/defn/expr.ex
```diff
@@ -213,7 +213,14 @@ defmodule Nx.Defn.Expr do
   """
   def while(initial, context, arg, condition, body) do
     [flatten_initial, flatten_arg, flatten_body] = clauses = flatten_clauses([initial, arg, body])
```
josevalim, line 215: I am thinking we should no longer change `reshape_vectors` and friend to work on containers and instead move the devectorize to flatten_clauses.

nx/lib/nx/defn/expr.ex
```diff
@@ -152,18 +152,23 @@ defmodule Nx.Defn.Expr do
 
     clauses = Enum.zip(preds, exprs)
 
-    out = flatten_to_composite(out, context, exprs, &expr(&1, context, :cond, [clauses, last]))
+    out =
```
polvalente, line 155: I had to keep this part here because the output vectorized axes change in relation to the input, but the actual vectorization happens inside flatten_to_composite now

nx/lib/nx/defn/expr.ex
polvalente: I don't get what you mean

nx/lib/nx/defn/expr.ex
polvalente: the matching is done on another part of the code, and `flatten_to_composite` ensures the result is revectorized according to the first argument

nx/lib/nx/defn/expr.ex
```diff
@@ -230,16 +251,62 @@ defmodule Nx.Defn.Expr do
     size = tuple_size(head)
     expr = fun.(tuple_out(size))
 
-    {out, {[], ^size}} =
-      Composite.traverse(out, {Tuple.to_list(head), 0}, fn _, {[head | tail], i} ->
-        {expr(head, context, :elem, [expr, i]), {tail, i + 1}}
+    vectorized_axes =
+      out
+      |> Composite.traverse([], fn
```
josevalim, line 256: I think we have a Composite.reduce too.

polvalente: yup :)

nx/lib/nx/defn/expr.ex
```diff
@@ -213,15 +218,31 @@ defmodule Nx.Defn.Expr do
...
-    args = [flatten_initial, flatten_arg, condition, flatten_body]
+
+    Composite.traverse(condition, fn
+      %T{vectorized_axes: [_ | _]} ->
+        raise ArgumentError,
+              "condition for while cannot be vectorized, got: #{inspect(condition)}"
+
+      _ ->
+        :ok
+    end)
```
josevalim, line 229:
```suggestion
    if condition.vectorized_axes != [] do
      raise ArgumentError,
            "condition for while cannot be vectorized, got: #{inspect(condition)}"
    end
```

I beleive condition is always a single expression.

nx/lib/nx/defn/expr.ex
```diff
@@ -213,15 +218,31 @@ defmodule Nx.Defn.Expr do
...
+
+    args = [
+      flatten_initial,
+      flatten_arg,
+      condition,
+      flatten_body
+    ]
+
```
josevalim, line 237:
```suggestion
    args = [flatten_initial, flatten_arg, condition, flatten_body]
```

nx/lib/nx/defn/expr.ex
```diff
@@ -230,16 +251,61 @@ defmodule Nx.Defn.Expr do
...
+                                                                             [axes | axes_tail]} ->
+        head =
+          if axes do
+            Nx.vectorize(head, axes)
+          else
+            head
+          end
+
+        {expr(head, context, :elem, [expr, i]), {tail, i + 1, axes_tail}}
       end)
```
josevalim, line 277:
```suggestion
      Composite.traverse(out, {Tuple.to_list(head), 0, vectorized_axes},
        fn _, {[head | tail], i, [axes | axes_tail]} ->
          head =
            if axes do
              Nx.vectorize(head, axes)
            else
              head
            end

          {expr(head, context, :elem, [expr, i]), {tail, i + 1, axes_tail}}
      end)
```

nx/lib/nx/defn/expr.ex
```diff
@@ -230,16 +251,61 @@ defmodule Nx.Defn.Expr do
...
-      Composite.traverse(out, {Tuple.to_list(head), 0}, fn _, {[head | tail], i} ->
-        {expr(head, context, :elem, [expr, i]), {tail, i + 1}}
+    vectorized_axes =
+      out
+      |> Composite.reduce([], fn
+        %T{vectorized_axes: [_ | _] = axes}, acc ->
+          [Keyword.keys(axes) | acc]
+
+        _t, acc ->
+          [nil | acc]
```
josevalim, line 261: If we use an empty list here, I think we can skip the conditional in the traverse below.

```suggestion
          [[] | acc]
```

polvalente: We can't. There's a check for scalars in vectorize that impedes this.

nx/lib/nx/defn/expr.ex
```diff
@@ -230,16 +251,61 @@ defmodule Nx.Defn.Expr do
...
+          {t, [Keyword.keys(axes) | acc]}
+
+        t, acc ->
+          {t, [nil | acc]}
+      end)
+      |> elem(1)
+      |> Enum.reverse()
+
+    {out, {[], []}} =
+      Composite.traverse(out, {[fun.(head)], vectorized_axes}, fn _,
```
josevalim, line 296: Refactor here too.

nx/lib/nx/defn/expr.ex
```diff
@@ -230,16 +251,61 @@ defmodule Nx.Defn.Expr do
...
+                                                                   [axes | axes_tail]} ->
+        head =
+          if axes do
+            Nx.vectorize(head, axes)
+          else
+            head
+          end
+
+        {head, {tail, axes_tail}}
+      end)
```
josevalim, line 307: Please do the same changes to this branch too (Composite.reduce, remove the conditional, move the `fn` clause inside, etc.

nx/lib/nx/defn/expr.ex
josevalim: Ah.

josevalim (approved):

PR #1185, opened 2023-04-04
feat(grad): window_scatter_min/max

### LINKED ISSUE #161: Add missing gradients

polvalente:
closes #161

Only caveat is that I wasn't able to get reference values from Jax,
but the ideas here are basically the same as the grad for indexed_add,
which is the main implementation for these functions in Torchx.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -8659,7 +8659,7 @@ defmodule Nx do
         f32
         1.5
       >
-      
+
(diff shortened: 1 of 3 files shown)


josevalim (approved):
BEAUTIFUL!!!

PR #1187, opened 2023-04-05
Devectorize parameters in evaluator

polvalente:


nx/test/nx/defn_test.exs
```diff
@@ -2115,6 +2115,50 @@ defmodule Nx.DefnTest do
...
+               |> String.replace(~r/\s/, "")
+
+      assert Nx.Defn.jit_apply(
+               fn x, y ->
+                 [x, y] = Nx.broadcast_vectors([x, y])
+                 {x, y}
+               end,
+               [x, y],
+               compiler: Evaluator
+             ) == 1
```
polvalente, line 2159: @josevalim I ran into this issue while working on the `cond` PR. In short, it seems that the concrete parameter value that's being fetched in the evaluator to be fed to the backend is returning with vectorized axes.

nx/lib/nx/defn/evaluator.ex
```diff
@@ -305,7 +305,7 @@ defmodule Nx.Defn.Evaluator do
               "cannot pass a tensor expression as argument to defn, got: #{inspect(tensor)}"
 
       %Nx.Tensor{} = tensor ->
-        {tensor, caches}
+        {Nx.devectorize(tensor), caches}
```
polvalente, line 308: This fixes the test, but I don't think it's the correct solution, right?

josevalim: Ideally we wouldn't have anything vectorized here. If it is vectorized, it means it is leaking somewhere. But this is helpful. I will take a look!

josevalim:

PR #1188, opened 2023-04-08
feat(vectorize): Nx.Defn.Expr.cond

polvalente:


nx/lib/nx/defn/expr.ex
```diff
@@ -137,18 +137,33 @@ defmodule Nx.Defn.Expr do
...
+    if vectorized_axes == [] do
+      out
+    else
+      {result, []} =
+        Composite.traverse(out, vectorized_axes, fn expr, [axes | tail] ->
+          {Nx.vectorize(expr, axes), tail}
+        end)
+
+      result
+    end
```
josevalim, line 166: We may use this in `while`as well, so I can see we encapsulating this soon. :)

polvalente: yup. we might even be able to absorb it into `flatten_to_composite` somehow

nx/lib/nx.ex
```diff
@@ -4770,20 +4799,30 @@ defmodule Nx do
...
     {devectorized_tensors, canonical, n}
   end
 
-  defp do_reshape_vectors_devectorize(tensors, [], _n), do: tensors
+  defp do_reshape_vectors_devectorize(tensors, [], _n, _make_broadcastable), do: tensors
+
+  defp do_reshape_vectors_devectorize(tensors, canonical_vectorized_axes, n, make_broadcastable) do
+    rank =
+      Enum.reduce(
+        tensors,
```
josevalim, line 4821:
```suggestion
        tl(tensors),
```

nx/lib/nx.ex
```diff
@@ -4625,6 +4625,13 @@ defmodule Nx do
   For vectors with the same name to be compatible, they need to either
   have the same size or one must be of size 1.
 
+  ## Options
+
+    * `:make_broadcastable` - boolean that indicates whether the inner
+      shapes should also be modified. If `true`, 1-sized leading dimensions
```
josevalim, line 4631: It seems we changed the default to `true`? Also, if the default is `true`, perhaps an option `vectors_only: true` would be better?

polvalente: Yes we did, as it is the most common use case and cleans up a lot of the expression graphs.

polvalente: Maybe "keep_inner_shape" instead? vectors only might convey that only tensors that are already vectorized would be changed

josevalim (approved):

nx/lib/nx.ex
```diff
@@ -4749,10 +4772,16 @@ defmodule Nx do
...
   @doc type: :shape
-  def broadcast_vectors([t]), do: [to_tensor(t)]
+  def broadcast_vectors(tensors_or_containers, opts \\ [])
 
-  def broadcast_vectors(tensors) when is_list(tensors) do
-    {devectorized_tensors, target_vectorized_axes, offset} = do_reshape_vectors(tensors)
+  def broadcast_vectors([t], _opts), do: [to_tensor(t)]
+
+  def broadcast_vectors(tensors, opts) when is_list(tensors) do
+    # mode: broadcastable | vectorized_only
```
polvalente, line 4780:
```suggestion
```

nx/lib/nx.ex
```diff
@@ -4749,10 +4774,16 @@ defmodule Nx do
...
```
polvalente, line 4782:
```suggestion
```

PR #1193, opened 2023-04-12
Do not crash when meta is different on struct compile

lukaszsamson:
This happens when building with `parser_options: [columns: true]`

Originally reported in https://github.com/elixir-lsp/elixir-ls/issues/864


cigrainger:
When @josevalim introduces a failing check due to formatting, does that mean the formatter is wrong? 😛

PR #1199, opened 2023-04-18
fix: vectorized Nx.take

### LINKED ISSUE #1192: Batch over vectorized indices in `Nx.take/2`
Currently `take` uses the first row of indices instead of vmapping:

```elixir
t = Nx.tensor([[1, 2], [3, 4]]) |> Nx.vectorize(:x)
idx = Nx.tensor([[1, 0], [0, 1]]) |> Nx.vectorize(:x)
Nx.take(t, idx)
```

```
#Nx.Tensor<
  vectorized[x: 2]
  s64[2]
  [
    [2, 1],
    [4, 3]
  ]
>
```

We have [`batched_take` in Bumblebee](https://github.com/elixir-nx/bumblebee/blob/6c4d88544d7d93a073bfd6f3b2dd9743ca93d245/lib/bumblebee/utils/nx.ex#L249-L278), but I don't think this technique generalizes to also support `:axis` without additional transpose (it relies on batch axes being adjacent to `:axis`, so those can be flattened out).

polvalente:
closes #1192

This basically replicates the algorithm for take-over-gather used in Torchx, so that we extend the still-vectorized indices to be passed onto the vectorized gather implementation.

Implementing directly over take would imply lots of really weird slicing and wouldn't yield much memory footprint improvement over what we have now.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -13321,8 +13321,6 @@ defmodule Nx do
     if tensor.vectorized_axes != [] or indices.vectorized_axes != [] do
       axes_range = axes(tensor)
 
-      indices = if rank(indices) == 0, do: new_axis(indices, 0), else: indices
-
       indices_shape =
         axes_range
         |> Enum.map(fn
@@ -13341,7 +13339,6 @@ defmodule Nx do
           x, _ ->
             x
         end)
-        |> List.flatten()
 
       indices_for_axis =
         indices
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -13293,83 +13293,126 @@ defmodule Nx do
   """
   @doc type: :indexed
   def take(tensor, indices, opts \\ []) when is_list(opts) do
-    tensor = to_tensor(tensor)
-    indices = to_tensor(indices)
-
-    canonical_vectorized_axes = calculate_canonical_vectorized_axes([tensor, indices])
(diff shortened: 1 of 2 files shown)


josevalim (approved):

nx/lib/nx.ex
```diff
@@ -13293,83 +13293,97 @@ defmodule Nx do
...
+      indices = if rank(indices) == 0, do: new_axis(indices, 0), else: indices
 
-        indices =
-          devec_indices
-          |> slice(starts, lengths)
-          |> squeeze(axes: Enum.reverse(indices_squeeze_axes_rev))
+      indices_shape =
+        axes_range
+        |> Enum.flat_map(fn
+          ^axis -> Tuple.to_list(indices.shape)
```
jonatanklosko, line 13329: Perhaps we could use `Tuple.product(indices.shape)` and remove `axis_offset`, since it's all flattened out anyway?

jonatanklosko (approved):

nx/lib/nx.ex
```diff
@@ -13293,83 +13293,89 @@ defmodule Nx do
...
+        |> Enum.with_index(fn
+          _x, ^axis ->
+            1
 
-        {shape, names, indices}
-      end
+          x, _ ->
+            x
+        end)
+        |> List.flatten()
```
jonatanklosko, line 13344:
```suggestion
```

PR #1201, opened 2023-04-18
feat: vectorize conv

polvalente:


nx/lib/nx.ex
```diff
@@ -12448,13 +12448,27 @@ defmodule Nx do
...
+    {_, names} = Enum.split(tensor.names, offset)
+
+    tensor = reshape(tensor, List.to_tuple([Enum.product(batch_axes) | other_axes]), names: names)
+
+    %{shape: k_shape} = kernel = devectorize(kernel)
+    {_, names} = Enum.split(kernel.names, offset)
+    {batch_axes, other_axes} = k_shape |> Tuple.to_list() |> Enum.split(offset + 1)
 
-    Nx.Shape.validate_conv!(input_shape, kernel_shape)
+    kernel = reshape(kernel, List.to_tuple([Enum.product(batch_axes) | other_axes]), names: names)
```
josevalim, line 12468: I think we can encapsulate this in a function, so we share between input and kernel. Also, I understand that the reshape `shape` will match the tensor shape and be a no-op for the common case, but I think it is clearer if we perform this code path only if the tensor is vectorized. :)

josevalim: By this, I mean this code:

```
    %{shape: k_shape} = kernel = devectorize(kernel)
    {_, names} = Enum.split(kernel.names, offset)
    {batch_axes, other_axes} = k_shape |> Tuple.to_list() |> Enum.split(offset + 1)

    Nx.Shape.validate_conv!(input_shape, kernel_shape)
    kernel = reshape(kernel, List.to_tuple([Enum.product(batch_axes) | other_axes]), names: names)
```

It seems to be shared between both input and kernel.

josevalim (approved):

PR #1203, opened 2023-04-20
feat: Nx.revectorize

polvalente:
Adds another utility function for vectorized axes manipulation.
This solves the problem of collapsing and re-expanding vectorized axes.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -4803,14 +4803,22 @@ defmodule Nx do
   @doc """
   Changes the disposition of the vectorized axes of a tensor or `Nx.Container`.
 
+  This function is basically a short-hand for:
+
+      tensor
+      |> Nx.devectorize(keep_names: false)
+      |> Nx.reshape(vectorized_sizes ++ target_shape, names: target_names)
+      |> Nx.vectorize(vectorized_names)
+
   Accepts the `target_axes` keyword list where the total size must match the current total
-  size of the vectorized axes. One of the items can have the `:auto` value.
+  size of the vectorized axes.
+
+  Between `target_axes` and the `:target_shape` option, there can be at most one `:auto` entry.
 
   ### Options
 
-    * `:target_rank` - the non-vectorized rank for the output. Can be used to transfer some of the vectorized axes
-    to the non-vectorized shape. If not given, the target rank will be at most the input tensor rank, and can be
-    reduced depending on how many `target_axes` are given.
+    * `:target_shape` - the (non-vectorized) output shape.
+    * `:target_names` - the names for the output shape.
 
   ### Examples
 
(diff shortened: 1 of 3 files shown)


nx/lib/nx.ex
```diff
@@ -4662,7 +4662,7 @@ defmodule Nx do
       iex> yv.vectorized_axes
       [y: 1, x: 1]
   """
-  @doc type: :shape
+  @doc type: :vectorization
```
polvalente, line 4665: @josevalim I introduced this new category so that people can more easily find these 3 related functions. Let me know if you think we should keep it as :shape still

josevalim: agreed.

nx/lib/nx.ex
```diff
@@ -4800,6 +4800,130 @@ defmodule Nx do
...
+
+    if auto_axes > 1 do
+      raise ArgumentError,
+            "at most one vectorized axis size can be calculated automatically, got: #{auto_axes}"
+    end
+
+    axes_names = Keyword.keys(target_axes)
+
+    {names, devectorized_shape_l} =
+      if target_rank = opts[:target_rank] do
```
josevalim, line 4901: What happens if rank is more than the current tensor rank?

nx/lib/nx.ex
```diff
@@ -4800,6 +4800,130 @@ defmodule Nx do
...
+            [0, 1],
+            [0, 1]
+          ],
+          [
+            [0, 1],
+            [0, 1]
+          ]
+        ]
+      >
+      iex> revec = Nx.revectorize(t, temp: :auto)
```
josevalim, line 4853: Shouldn't the `target_rank` be here in this example. Imagine you have a tensor of rank 4. 1 dimension is vectorized and you need a rank 2 at the end. So you need to create a vectorized tensor with final shape of two.

Then at the end we revectorize it as `Nx.revectorize(result, t.vectorized_axes)`, no?

polvalente: The idea of the example is that we can introduce new axes (which happens in the line below), and `target_rank` can be used to restore the original rank.

The evolution over the example is: 2 vec + rank 1 -> 1 vec + rank 1 -> 2 vec + rank 0 -> 2 vec + rank 1

nx/lib/nx.ex
polvalente: Then the last vectorized axes will be absorbed into the tensor rank.
The idea of this option is to enable a way for this process to happen.
Maybe we can rework the option into `:pop` or something like this, being the number of trailing vectorized axes that will be absorbed into the inner shape.

Though the current interface is a bit easier because it allows us to restore the original tensor shape and axes by passing these fields as the `target_axes` and `:target_rank`

josevalim: sorry, i mean that if you say target_rank 5 and the current devectorized tensor  has rank 2, we have to raise, no?

nx/lib/nx.ex
josevalim: @polvalente I may be missing something, why isn't `Nx.revectorize(result, t.vectorized_axes)` enough to reestablish it?

nx/lib/nx.ex
polvalente: ah, yes

nx/lib/nx.ex
polvalente: Actually, I think we should accept `target_shape` that will be the inner shape of the tensor.

josevalim:
There are two open discussions in this PR. Maybe we should jump into a call and finish ironing out this API?

josevalim (approved):
Beautiful!

PR #1211, opened 2023-05-08
feat: allow vectorized predicate in while expressions

polvalente:


josevalim (approved):
Absolutely beautiful. Just cond and `while <-` missing :D

PR #1213, opened 2023-05-08
feat: support vectorized predicates in cond

polvalente:


nx/test/nx/vectorize_test.exs
```diff
@@ -512,4 +512,195 @@ defmodule Nx.VectorizeTest do
...
+    )
+    |> IO.inspect()
+
+    # the results here differ, which maybe indicates that we do indeed need to broadcast vectorized_axes between vectorized predicates
+
+    assert cond4(
+             Nx.vectorize(~V[0 1 0], :pred1),
+             Nx.vectorize(~V[10 100], :pred2),
+             Nx.vectorize(~V[1 0], :pred2),
+             Nx.vectorize(~V[20 200 2000], :pred1),
```
polvalente, line 695: @josevalim I think this is the last point of contention between our solutions. pred1 and pred2 are indeed orthogonal, but I think that applying them independently might yield hard to predict results.

By doing the explicit broadcast between them, the result is easier to reason about, and we can do this broadcasting only between the already vectorized predicates.

polvalente: The result on the assertion below is the result given the explicit broadcasting.

polvalente: Ok, so after unwrapping this by hand, I decided that we need to at least reshape_vectors the preds together.
What's happening currently is that the first clause is vectorizing to [:pred1, :pred2] while the second clause is vectorizing to [:pred2, :pred1] and then being reshaped, so it yields unexpected results.

`reshape_vectors` takes care of this.

nx/lib/nx/defn/expr.ex
```diff
@@ -129,13 +129,68 @@ defmodule Nx.Defn.Expr do
...
+    {preds, exprs} = Enum.unzip(clauses)
+    {preds, context} = to_exprs(preds)
+
+    if Enum.all?(preds, &(&1.vectorized_axes == [])) do
+      non_vectorized_cond(clauses, last, context)
+    else
+      # We reshape_vectors to ensure that all predicate vectorized axes are encountered
+      # in the same order throughout the clauses. the zip_with below ensures that only
+      # the vectorized predicates are pushed through with this new order
+      reshaped_preds = Nx.reshape_vectors(preds)
```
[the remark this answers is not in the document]
polvalente: Also,.note that this is just reshape, so we're not creating extra elements explicitly, we're just ensuring that vectorized axes always appear in the same order, and this only applies to vectorized clauses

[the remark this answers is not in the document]
josevalim: Got it, ship it!

PR #1214, opened 2023-05-09
feat: Allow non-batched inputs on indexed_add and indexed_put

### LINKED ISSUE #1212: Allow non-batched inputs on indexed_add and indexed_put
Instead of `{n, rank(t)}`, we can give `{rank(t)}`. If a scalar is given, it is the same as `{1}`.

polvalente:
closes #1212

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -7233,7 +7233,7 @@ defmodule Nx do
 
   As a shorthand notation, rank-1 indices can be used for updating a single entry:
 
-      iex> Nx.indexed_add(Nx.tensor([[1], [2]]), Nx.tensor([1, 0]), Nx.tensor(8))
+      iex> Nx.indexed_add(Nx.tensor([[1], [2]]), Nx.tensor([1, 0]), 8)
       #Nx.Tensor<
         s64[2][1]
         [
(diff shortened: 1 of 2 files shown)


nx/lib/nx.ex
```diff
@@ -7387,6 +7409,12 @@ defmodule Nx do
     indexed_op(target, indices, updates, :indexed_put)
   end
 
+  defp indexed_op(target, %Nx.Tensor{shape: {_}} = index, update, op) when is_tensor(update) do
```
polvalente, line 7412: @josevalim in theory we could extend this to accept a list of numbers/tensors as well in the `index`. WDYT?

Would be something like `index = if is_list(index), do: Enum.map(index, &to_tensor/1) |> Nx.concatenate(), else: to_tensor(index)`

josevalim:
Can we also support scalars? Otherwise it looks good to me. I also want to rewrite the docs, to make the simpler path come first, but I can do that after merging. :)

polvalente:
What do you mean by "also support scalars"?

josevalim:
I mean this: `Nx.indexed_put(Nx.tensor([1, 2, 3]), Nx.tensor(1), 10)`.

PR #1216, opened 2023-05-15
feat: vectorize while generator

polvalente:


nx/lib/nx/defn/expr.ex
```diff
@@ -604,6 +604,94 @@ defmodule Nx.Defn.Expr do
...
+          starts = [index_param | List.duplicate(0, tuple_size(target.shape) - 1)]
+
+          updated_target =
+            target
+            |> Nx.put_slice(starts, Nx.devectorize(node) |> Nx.new_axis(0))
+            |> Nx.vectorize(target.vectorized_axes)
+
+          {updated_target, targets}
+        end
+      )
```
josevalim, line 676: Do you think it is worth sharing some of this logic with the other `while` branch? I think many of these `traverse` calls are identical.

polvalente: Yes. The code is mostly copy-paste :)
I just didn't do it before because I wanted to validate the general idea and behavior of the code with you first

josevalim: Yeah, looks good to me. :) I think we can probably share all of its contents and make the "inner_while" bit an argument?

polvalente: I shared a prefix section and a suffix section of the code. the building of the inner_while seemed different enough to me to warrant keeping it in place instead of passing that section as an anonymous function or whatever

josevalim (approved):

PR #1218, opened 2023-05-15
fix: flaky distributed tests

polvalente:


nx/test/test_helper.exs
```diff
@@ -4,7 +4,8 @@ cond do
   :distributed in Keyword.get(ExUnit.configuration(), :exclude, []) ->
     :ok
 
-  Code.ensure_loaded?(:peer) and match?({:ok, _}, Node.start(:"primary@127.0.0.1", :longnames)) ->
+  Code.ensure_loaded?(:peer) and {"", 0} == System.cmd("epmd", ["-daemon"]) and
```
polvalente, line 7: I think the `Code.ensure_loaded?` part had the intention of starting these only if epmd is already up, but that wasn't the case here, and thus the tests weren't skipped locally.

Any ideas on a better fix?

josevalim: The code loading is to check for the OTP version.

PR #1221, opened 2023-05-19
feat: add constant creation helpers

polvalente:
adds u8/1, f32/1 and friends for more easily creating constant tensors

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -897,6 +897,41 @@ defmodule Nx do
     %T{shape: shape, type: type, names: names, data: %Nx.TemplateBackend{}}
   end
 
+  for {type, result} <- [
+        u8: "1",
+        u16: "1",
+        u32: "1",
+        u64: "1",
+        s8: "1",
+        s16: "1",
+        s32: "1",
+        s64: "1",
+        bf16: "1.0",
+        f16: "1.0",
+        f32: "1.0",
+        f64: "1.0",
+        c64: "1.0+0.0i",
+        c128: "1.0+0.0i"
+      ] do
+    @doc """
+    Short-hand function for creating tensor of type #{type}.
+
+    iex> Nx.#{type}(1)
+    #Nx.Tensor<
+      #{type}
+      #{result}
+    >
+
+    iex> Nx.#{type}([1, 1, 1])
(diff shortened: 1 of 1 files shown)


nx/lib/nx.ex
```diff
@@ -897,6 +897,41 @@ defmodule Nx do
...
+
+    iex> Nx.#{type}([1, 1, 1])
+    #Nx.Tensor<
+      #{type}[3]
+      [#{result}, #{result}, #{result}]
+    >
+    """
+    @doc type: :creation
+    def unquote(type)(tensor), do: Nx.tensor(tensor, type: unquote(type))
+  end
```
polvalente, line 933: @josevalim not sure if this should instead be added to Nx.Constants, and if we should enforce that the input is a scalar

josevalim: Oh, I was thinking we would add this to `Nx.Defn.Kernel`. But her is fine. I don't think it belongs to constant.

polvalente:
@josevalim I removed the doctests because I couldn't figure out what was happening with torchx on windows.
I don't think this is a problem because the new functions are basically aliases, but you decide

nx/lib/nx.ex
```diff
@@ -897,6 +897,16 @@ defmodule Nx do
     %T{shape: shape, type: type, names: names, data: %Nx.TemplateBackend{}}
   end
 
+  for t <- [:u8, :u16, :u32, :u64, :s8, :s16, :s32, :s64, :bf16, :f16, :f32, :f64, :c64, :c128] do
```
josevalim, line 900: I would skip the complex ones for now because maybe for them `c64(real, img)` is better? Anyway, we can discuss this later and the below is enough to get started.

```suggestion
  for t <- [:u8, :u16, :u32, :u64, :s8, :s16, :s32, :s64, :bf16, :f16, :f32, :f64] do
```

josevalim (approved):

nx/lib/nx.ex
polvalente: Makes sense. We already have Nx.complex. Worst case scenario people would be doing `complex(f32(real), f32(imag))` or `f32(real) + Nx.Constants.i() * f32(imag)` or variations on that

PR #1230, opened 2023-05-27
feat: support strides in iota

polvalente:


nx/lib/nx.ex
```diff
@@ -1212,18 +1271,25 @@ defmodule Nx do
...
+        |> vectorize(vectorized_axes)
+      else
+        output
+      end
+
+    if strides do
+      window_min(result, Tuple.duplicate(1, length(strides)), strides: strides)
     else
-      output
+      result
```
polvalente, line 1292: This will probably make us hold on to the iota for longer than needed in EXLA or other backends. I implemented it here first for validation on the concept and interface, but perhaps we should do it on each backend.

libtorch arange does provide this, but XLA iota does not (jax does it by building the strided iota from np.arange instead).

Thoughts?

nx/lib/nx.ex
```diff
@@ -1179,10 +1185,55 @@ defmodule Nx do
           ]
         ]
       >
+
+  An iota can also be strided:
+
+      iex> Nx.iota({2, 3, 5}, strides: [1, 2, 2])
```
josevalim, line 1191: I think is confusing? It means we are no longer returning the given shape once strided?

polvalente: I didn't think it was confusing at all haha

I see your point, but this is literally what the arange function is meant to do.

Do you think we should have a separate function instead?

josevalim: @polvalente I was thinking something like `Nx.iota(0..100//5)` and it would literally create said iota. But I don't know how to implement that with the `axis` option. Without axis, it is a matter of doing `+` and `*` afterwards.

josevalim: I guess with axis it is the same?

polvalente: This doesn't solve the case where we want an iota with the same base shape as a given tensor (i.e. Nx.iota(Nx.shape(tensor), ...)), But want to take every nth element of each axis.

I guess given the actual implementation we can just add this trick to a cheatsheet, though

josevalim: What is the use case? The one in the forum is done on a given dimension, no?

josevalim: Wait, ignore me. Don’t we already support slicing with a range? So instead of `Nx.take(tensor, Nx.iota(range))` we do `tensor[range]`?

polvalente: I think the range must have step=1

polvalente: Stepping over multiple dimensions can be useful for decimating images. For instance, taking every other row and every other column to reduce a 128x256 image into 64x128

josevalim: Right, but for that problem I think a window_min/window_mean directly on the image is more natural?

And should we suport stepped ranges on access then?

polvalente: window_min/max/mean doesn't work the same way. For instance, imagine we have the following 3x6 tensor of type u8:

```
[
  [1, 5, 2, 5, 3, 5],
  [4, 5, 5, 5, 6, 5],
  [7, 5, 8, 5, 9, 5]
]
```

and we want to take every 2rd element over the rows, but we want to take every 2nd element of each row to get the following 3x3 tensor of type u8:

```
[
  [1, 2, 3],
  [4, 5, 6],
  [7, 8, 9]
]
```

All of the window_* functions fail, each for a different reason:

```
# window_min fails because it only works for the 1st half of the tensor
iex(3)> Nx.window_min(t, {1, 2}, strides: [1, 2], padding: :valid)
#Nx.Tensor<
  s64[3][3]
  [
    [1, 2, 3],
    [4, 5, 5],
    [5, 5, 5]
  ]
>
# window_max fails in the complimentary way, only working for the 2nd half
iex(4)> Nx.window_max(t, {1, 2}, strides: [1, 2], padding: :valid)
#Nx.Tensor<
  s64[3][3]
  [
    [5, 5, 5],
    [5, 5, 6],
    [7, 8, 9]
  ]
>
# This just doesn't work because it results in a floating point tensor in any case.
# Even if the values were correct, this isn't the proper way to do it due to usage of
# floating-point units during the calculation, which is a major slowdown.
iex(5)> Nx.window_mean(t, {1, 2}, strides: [1, 2], padding: :valid)
#Nx.Tensor<
  f32[3][3]
  [
    [3.0, 3.5, 4.0],
    [4.5, 5.0, 5.5],
    [6.0, 6.5, 7.0]
  ]
>
```

That being said, we can do exactly what I did in the PR, which is to get the iota from the tensor (`iota = Nx.iota(Nx.shape(tensor))` and then applying
`indices = window_min(iota, {1, 2}, strides: [1, 2])` to get the correct result.

In practice, having this as a separate function has a major problem in defn, which is that we can't easily convert from list to tuple or back inside a defn, so the user will be forced to either pass down the value in each form, or to have a deftransformp just for that conversion.

I like the idea of supporting strided slices with Access, and the way to go is exactly this, but I don't think this should be the only way to apply decimation.

Now, given that all of this has basically being talked about the slicing use case, here's an idea: What if we change the semantics of `slice` and `slice_along_axis` where the `length` of the slice is a maximum length, which can be reduced if we pass `:strides` for each slice axis? This would be _instead_ of accepting strides in iota.

We can even document that the resulting length of each axis can be calculated through: `length = ceil(candidate_length / stride)`

polvalente: Actually, `slice` and `slice_along_axis` already have the strides option! So we just need to support strided ranges in Access :)

polvalente: For completeness:

```
iex(1)> iota = Nx.iota({3, 3, 5})
iex(22)> Nx.slice(iota, List.duplicate(0, Nx.rank(iota)), Tuple.to_list(Nx.shape(iota)), strides: List.duplicate(2, Nx.rank(iota)))
#Nx.Tensor<
  s64[2][2][3]
  [
    [
      [0, 2, 4],
      [10, 12, 14]
    ],
    [
      [30, 32, 34],
      [40, 42, 44]
    ]
  ]
>
```

Which still isn't the best because still warrants calling a deftransform if called from within defn (due to List.duplicate and Tuple.to_list), but provides a nice solution for building a strided iota, taking strided slices of a tensor and so on.

polvalente: Now, I have a follow-up question:

Why is it acceptable for the `slice` to have a resulting axis-length which is less than the given length in the strided case, but it's not acceptable or is confusing for the same to happen with `iota`?

josevalim: > Why is it acceptable for the slice to have a resulting axis-length which is less than the given length in the strided case, but it's not acceptable or is confusing for the same to happen with iota?

For me it is because I am passing a length (as a list) to `slice`, not a tuple `shape`. In any case, we should support steps in access. Should we open up a separate issue for that?

For the window functionality, would it make sense to introduce a window_gather/window_take or something?

:heart:

polvalente: Right, that makes sense. I'll open a separate PR supporting steps in Access!

I don't think we need any window functionality after all. What I did there was just a way to achieve strided slicing in the case where the tensor is guaranteed to be fully ordered (i.e. iota)

josevalim: So I am fine with closing this one I think. :)

PR #1236, opened 2023-05-31
Fix comment for the default `:ord` value of `Nx.LinAlg.norm`

kenken-neko:
Minor correction to the comment statement for function `Nx.LinAlg.norm`.


nx/lib/nx/lin_alg.ex
```diff
@@ -164,7 +164,7 @@ defmodule Nx.LinAlg do
 
     * `:axes` - defines the axes upon which the norm will be calculated.
       Applies only on 2-norm for 2-D tensors. Default: `nil`.
-    * `:ord` - defines which norm will be calculated according to the table below. Default: `2`
+    * `:ord` - defines which norm will be calculated according to the table below. Default: `nil`.
```
polvalente, line 167: These actually result in the same thing

polvalente (approved):

PR #1237, opened 2023-06-01
add split function

tiagodavi:
based on this: https://github.com/elixir-nx/scholar/pull/107

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -3265,6 +3265,61 @@ defmodule Nx do
     end)
   end
 
+  @doc ~S"""
+  Split tensor or list into train and test subsets.
+
+  ## Examples
+
+    Split into 80% for training and 20% for testing.
+
+    iex> {_train, _test} = Nx.split(Nx.tensor([[3, 6, 5], [26, 75, 3], [23, 4, 1]]), split: 0.8)
+    {#Nx.Tensor<
+      s64[2][3]
+      [
+        [3, 6, 5],
+        [26, 75, 3]
+      ]
+      >, #Nx.Tensor<
+        s64[1][3]
+        [
+          [23, 4, 1]
+        ]
+      >
+    }
+
+    iex> {_train, _test} = Nx.split([[3, 6, 5], [26, 75, 3], [23, 4, 1]], split: 0.8)
+    {[[3, 6, 5], [26, 75, 3]], [[23, 4, 1]]}
+  """
+  @doc type: :shape
+  def split(tensor, opts \\ [])
+
+  def split(tensor, opts) when is_tensor(tensor) do
+    {train, test} =
+      tensor
+      |> to_list()
+      |> split_op(opts)
+
+    {Nx.tensor(train), Nx.tensor(test)}
+  end
+
+  def split(data, opts) when is_list(data) do
+    split_op(data, opts)
+  end
+
+  defp split_op([_ | _] = features, opts) do
+    split = Keyword.fetch!(opts, :split)
+
+    if is_float(split) and split > 0.0 and split <= 1.0 do
+      num_samples = Enum.count(features)
+
+      split_size = Kernel.floor(split * num_samples)
+
+      Enum.split(features, split_size)
+    else
+      raise ":split must be a float number between 0.0 and 1.0"
+    end
+  end
+
   @doc """
   Broadcasts `tensor` to the given `broadcast_shape`.
 
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -3280,79 +3280,81 @@ defmodule Nx do
 
   ## Examples
 
-    All examples will operate on the same tensor so that it's easier to compare different configurations.
+  All examples will operate on the same tensor so that it's easier to compare different configurations.
 
-    iex> t = Nx.tensor([[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]])
-    iex> {left, right} = Nx.split(t, 2, axis: 0)
-    iex> left
-    #Nx.Tensor<
-      s64[2][4]
-      [
-        [0, 1, 2, 3],
-        [4, 5, 6, 7]
-      ]
-    >
-    iex> right
-    #Nx.Tensor<
-      s64[1][4]
-      [
-        [8, 9, 10, 11]
-      ]
-    >
-    iex> {left, right} = Nx.split(t, 2, axis: 1)
-    iex> left
-    #Nx.Tensor<
-      s64[3][2]
-      [
-        [0, 1],
-        [4, 5],
-        [8, 9]
-      ]
-    >
-    iex> right
-    #Nx.Tensor<
-      s64[3][2]
-      [
-        [2, 3],
-        [6, 7],
-        [10, 11]
-      ]
-    >
-    iex> {left, right} = Nx.split(t, 0.5, axis: 0)
-    iex> left
-    #Nx.Tensor<
-      s64[2][4]
-      [
-        [0, 1, 2, 3],
-        [4, 5, 6, 7]
-      ]
-    >
-    iex> right
-    #Nx.Tensor<
-      s64[1][4]
-      [
-        [8, 9, 10, 11]
-      ]
-    >
-    iex> {left, right} = Nx.split(t, 0.75, axis: 1)
-    iex> left
-    #Nx.Tensor<
-      s64[3][3]
-      [
-        [0, 1, 2],
-        [4, 5, 6],
-        [8, 9, 10]
-      ]
-    >
-    iex> right
-    #Nx.Tensor<
-      s64[3][1]
-      [
-        [3],
-        [7],
-        [11]
-      ]
-    >
+      iex> t = Nx.tensor([[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]])
+      iex> {left, right} = Nx.split(t, 2, axis: 0)
+      iex> left
+      #Nx.Tensor<
+        s64[2][4]
+        [
+          [0, 1, 2, 3],
+          [4, 5, 6, 7]
+        ]
+      >
+      iex> right
+      #Nx.Tensor<
+        s64[1][4]
+        [
+          [8, 9, 10, 11]
+        ]
+      >
+      iex> {left, right} = Nx.split(t, 2, axis: 1)
+      iex> left
+      #Nx.Tensor<
+        s64[3][2]
+        [
+          [0, 1],
+          [4, 5],
+          [8, 9]
+        ]
+      >
+      iex> right
+      #Nx.Tensor<
+        s64[3][2]
+        [
+          [2, 3],
+          [6, 7],
+          [10, 11]
+        ]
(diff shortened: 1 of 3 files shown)


nx/lib/nx.ex
```diff
@@ -3265,6 +3265,61 @@ defmodule Nx do
...
 
+  @doc ~S"""
+  Split tensor or list into train and test subsets.
+
+  ## Examples
+
+    Split into 80% for training and 20% for testing.
+
+    iex> {_train, _test} = Nx.split(Nx.tensor([[3, 6, 5], [26, 75, 3], [23, 4, 1]]), split: 0.8)
+    {#Nx.Tensor<
```
tiagodavi, line 3276: not sure what to do here to solve doctests. I think the problem is because I am returning a tuple.

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,61 @@ defmodule Nx do
...
+    iex> {_train, _test} = Nx.split([[3, 6, 5], [26, 75, 3], [23, 4, 1]], split: 0.8)
+    {[[3, 6, 5], [26, 75, 3]], [[23, 4, 1]]}
+  """
+  @doc type: :shape
+  def split(tensor, opts \\ [])
+
+  def split(tensor, opts) when is_tensor(tensor) do
+    {train, test} =
+      tensor
+      |> to_list()
```
polvalente, line 3299: For tensors you should be using Nx.slice_along_axis instead of converting to a list. 

Also I don't think we should support lists as inputs here

nx/lib/nx.ex
polvalente: Just extract each tuple position to a variable and assert on each value on separate lines! I believe there are lots of examples on this in Nx.LinAlg

nx/test/nx_test.exs
```diff
@@ -2959,4 +2959,65 @@ defmodule NxTest do
...
+      assert length(Nx.to_list(test)) == 39
+    end
+
+    test "Split into 60% for training and 40% for testing with unbalanced data" do
+      tensor = Nx.iota({73, 4})
+      {train, test} = Nx.split(tensor, split: 0.61)
+
+      assert length(Nx.to_list(train)) == 44
+      assert length(Nx.to_list(test)) == 29
+    end
```
polvalente, line 3021:
```suggestion
    test "Split into 60% for training and 40% for testing with unbalanced data" do
      tensor = Nx.iota({99, 4})
      {train, test} = Nx.split(tensor, split: 0.6)

      assert {60, 4} == Nx.shape(train)
      assert {39, 4} == Nx.shape(test)
    end
```

Two-fold review here:

1. (this applies to all other tests) let's assert on the shape instead, to avoid Nx.to_list
2. I think showing what happens with the split being a "nice" value but the length being indivisible by it is better for learning purposes. i.e. I would have no idea what 61% of 73 should be, but it's easy to know that 60% of 100 is 60, and 60% of 99 is either 60 of 59 depending on rounding.

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,38 @@ defmodule Nx do
...
+  """
+  @doc type: :shape
+  def split(tensor, opts \\ [])
+
+  def split(%T{shape: shape} = tensor, opts) do
+    split = Keyword.fetch!(opts, :split)
+
+    if is_float(split) and split > 0.0 and split < 1.0 do
+      rows = elem(shape, 0)
+      split_size = Kernel.floor(split * rows)
```
polvalente, line 3287: why floor instead of ceil here? Maybe we should allow for this to be an option?

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,38 @@ defmodule Nx do
...
+  @doc ~S"""
+  Split tensor into train and test subsets.
+
+  ## Examples
+
+    Split into 80% for training and 20% for testing.
+
+    iex> train = Nx.tensor([[3, 6, 5],[26, 75, 3]])
+    iex> test = Nx.tensor([[23, 4, 1]])
+    iex> {^train, ^test} = Nx.split(Nx.tensor([[3, 6, 5], [26, 75, 3], [23, 4, 1]]), split: 0.8)
```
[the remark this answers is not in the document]
polvalente: I think we should have an easier-to-read test here. Maybe allocate an iota with shape {10} for the first example, and a second example with {10, 2} and axis: 0 and another example with {2, 10} and axis: 1.

This would make it easier for people to read the example and see how things should behave in the nice cases.

And then show what happens for the not-nice case in a single example. Currently we only have the not-nice case as an example

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,58 @@ defmodule Nx do
     end)
   end
 
+  @doc ~S"""
+  Split a tensor into train and test subsets.
+
+  `split` must be a float number between `0.0` and `1.0`.
```
polvalente, line 3271:
```suggestion
  `split` must be a float `0.0 <= split <= 1.0`.
```

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,129 @@ defmodule Nx do
     end)
   end
 
+  @doc ~S"""
+  Split a tensor into train and test subsets.
+
+  `split` split must be an integer greater than zero and less than the length of the tensor.
```
polvalente, line 3271: I think we got some communication noise haha
We want split to be either an integer, in which case it would behave as you have it right now; or a float, which would have the previous behavior.

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,129 @@ defmodule Nx do
...
+        [26, 75, 3, 9]
+      ]
+    >
+    iex> test
+    #Nx.Tensor<
+      s64[rows: 1][columns: 4]
+      [
+        [23, 4, 1, 5]
+      ]
+    >
```
polvalente, line 3353: I think these are good tests, but we don't need them as examples. You can move them to the test file

nx/lib/nx.ex
tiagodavi: ah ok no problem. I will fix it.

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,129 @@ defmodule Nx do
...
+
+  ## Options
+
+    * `:axis` - The axis along which to take the values from. Defaults to `0`.
+
+  ## Examples
+
+    Split a tensor into two separate tensors.
+
+    iex> {train, test} = Nx.split(Nx.tensor([[3, 6, 5], [26, 75, 3], [23, 4, 1]]), 2, axis: 0)
```
polvalente, line 3281: Let's use the first example without passing the axis option.

I think an iota as the input tensor is more easily readable due to the inherent ordering as well.
And we could even have just a rank 1 tensor for the first example

nx/lib/nx.ex
polvalente: (these == the named axis examples)

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,129 @@ defmodule Nx do
...
+      values = elem(shape, axis)
+      size = values - split
+
+      {
+        slice_along_axis(tensor, 0, split, axis: axis),
+        slice_along_axis(tensor, split, size, axis: axis)
+      }
+    else
+      raise "split must be an integer greater than zero and less than the length of the tensor."
+    end
```
polvalente, line 3388: To support split as both integer and float, you can use a cond here instead of if

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,129 @@ defmodule Nx do
...
+          is_integer(axis) and axis in axis_values ->
+            axis
+
+          is_atom(axis) and axis in axis_names ->
+            dimensions = Enum.zip(axis_names, axis_values)
+            dimensions[axis]
+
+          true ->
+            raise ":axis is out of tensor bounds."
+        end
```
polvalente, line 3377: I believe there's a Nx.Shape.normalize_axis function for doing this

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,129 @@ defmodule Nx do
...
+
+      values = elem(shape, axis)
+      size = values - split
```
josevalim, line 3380: What happens when size is negative? Do we want to allow negative splitting (which counts from the back)?

tiagodavi: not sure if we want this behavior.

nx/lib/nx.ex
tiagodavi: I didn't understand how to use normalize the simple way I would like.

tiagodavi:
@polvalente there's some weird test failing
test/nx/serving_test.exs:637

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,124 @@ defmodule Nx do
...
+        end
+
+      remaining_size = values - split_size
+
+      {
+        slice_along_axis(tensor, 0, split_size, axis: axis),
+        slice_along_axis(tensor, split_size, remaining_size, axis: axis)
+      }
+    else
+      raise "split must be a float number between 0.0 and 1.0."
```
polvalente, line 3382:
```suggestion
      raise "split must be a float such that 0.0 <= split <= 1.0."
```

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,132 @@ defmodule Nx do
...
+      s64[3][1]
+      [
+        [3],
+        [7],
+        [11]
+      ]
+    >
+  """
+  @doc type: :indexed
+  def split(tensor, split, opts \\ [])
```
polvalente, line 3354: @tiagodavi @josevalim I've pushed a refactor to collapse both clauses into the same one, mostly to highlight that the slicing is the same, with the only change being how we calculate it.

I've also changed the examples to operate on a written-out iota instead of arbitrary values because it makes it easier to compare results.

We still need to discuss if we want to accept a negative integer split.

It's easy to do it by setting right after `axis_size = `: `split = if is_integer(split) and split < 0, do: axis_size + split, else: split`

polvalente: Though I think that negative split might be confusing in the sense that we don't know if the results will be reversed or swapped in the result tuple (either or both)

josevalim: @polvalente we will base it on `Enum.split` (which means looking up from the end).

nx/lib/nx.ex
```diff
@@ -3265,6 +3265,138 @@ defmodule Nx do
...
+          [3],
+          [7],
+          [11]
+        ]
+      >
+  """
+  @doc type: :indexed
+  def split(tensor, split, opts \\ [])
+
+  def split(tensor, split, opts) do
```
josevalim, line 3362:
```suggestion
  def split(tensor, split, opts) when is_number(split) do
```

nx/lib/nx.ex
josevalim: But I am fine with postponing this for now. :)

polvalente: wdyt?

tiagodavi:
Awesome, thank you folks.

PR #1242, opened 2023-06-03
feat: improve container diffing

polvalente:
Add some coloring so that we can more easily find the diff in compilation errors

### DIFF
--- a/nx/lib/nx/defn.ex
+++ b/nx/lib/nx/defn.ex
@@ -331,13 +331,9 @@ defmodule Nx.Defn do
 
             #{inspect(arg_template)}
 
-            Expected argument:
+            Comparison:
 
-            #{inspect(Enum.fetch!(template_args, pos - 1))}
-
-            Actual argument:
-
-            #{inspect(arg)}
+            #{Nx.Defn.CompilationDiff.build_and_inspect(Enum.fetch!(template_args, pos - 1), arg)}
             """
           end
 
--- a/nx/lib/nx/defn/compilation_diff.ex
+++ b/nx/lib/nx/defn/compilation_diff.ex
@@ -1,0 +1,94 @@
+defmodule Nx.Defn.CompilationDiff do
+  defstruct [:left, :right, :compatible, :nesting]
+
+  defp is_valid_container?(t) do
+    impl = Nx.Container.impl_for(t)
+    not is_nil(impl) and impl != Nx.Container.Any
+  end
+
+  def build(left, right) do
+    l = is_valid_container?(left)
+    r = is_valid_container?(right)
+
+    cond do
+      not l and not r ->
+        %__MODULE__{left: left, right: right, compatible: left == right, nesting: 0}
+
+      not l or not r ->
+        %__MODULE__{left: left, right: right, compatible: false, nesting: 0}
+
+      Nx.Container.impl_for(left) != Nx.Container.impl_for(right) ->
+        %__MODULE__{left: left, right: right, compatible: false, nesting: 0}
+
+      l and r ->
+        {diff, acc} =
+          Nx.Defn.Composite.traverse(left, Nx.Defn.Composite.flatten_list([right]), fn
+            left, [] ->
+              {%__MODULE__{left: left}, :incompatible_sizes}
+
+            left, [right | acc] ->
+              {
+                %__MODULE__{
+                  left: left,
+                  right: right,
+                  compatible: Nx.compatible?(left, right),
+                  nesting: 2
+                },
+                acc
+              }
+          end)
+
+        if acc == :incompatible_sizes do
+          %__MODULE__{left: left, right: right, compatible: false, nesting: 0}
+        else
+          diff
+        end
+    end
+  end
+
+  def build_and_inspect(left, right) do
+    left
+    |> build(right)
+    |> inspect()
+  end
+
+  defimpl Inspect do
+    import Inspect.Algebra
+
+    def inspect(%Nx.Defn.CompilationDiff{left: left, right: nil}, _opts) do
+      inspect_as_template(left)
+    end
+
+    def inspect(%Nx.Defn.CompilationDiff{left: left, compatible: true}, _opts) do
+      inspect_as_template(left)
+    end
+
+    def inspect(%Nx.Defn.CompilationDiff{left: left, right: right, nesting: nesting}, _opts) do
+      nest(
+        concat([
+          line(),
+          IO.ANSI.light_black_background(),
+          IO.ANSI.green(),
+          inspect_as_template(left),
+          line(),
+          IO.ANSI.red(),
+          inspect_as_template(right),
+          line(),
+          IO.ANSI.reset()
+        ]),
+        nesting
+      )
+    end
+
+    defp inspect_as_template(data) do
+      if is_number(data) or is_tuple(data) or
+           (is_map(data) and Nx.Container.impl_for(data) != Nx.Container.Any) do
+        data
+        |> Nx.to_template()
+        |> Kernel.inspect(custom_options: [skip_template_backend_header: true])
+      else
+        inspect(data)
+      end
+    end
+  end
+end
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -439,13 +439,9 @@ defmodule Nx.Defn.Expr do
                 description: """
                 cond/if expects all branches to return compatible tensor types.
 
-                Got mismatching templates:
+                Comparison:
 
-                #{inspect_as_template(first)}
-
-                and
-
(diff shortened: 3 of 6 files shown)


seanmor5 (commented):
😍

polvalente:
Screenshot 2023-06-04 at 1 23 16 AM
Screenshot 2023-06-04 at 1 23 06 AM
Screenshot 2023-06-04 at 1 22 59 AM

Some prints of the current behavior. I'm just not sure if we need to keep the "at position X" part of the message now that we have a visual diff

nx/lib/nx/defn/compilation_diff.ex
```diff
@@ -0,0 +1,96 @@
+defmodule Nx.Defn.CompilationDiff do
```
josevalim, line 1:
```suggestion
defmodule Nx.Defn.TemplateDiff do
```

nx/lib/nx/defn/compilation_diff.ex
```diff
@@ -0,0 +1,96 @@
...
+  defstruct [:left, :right, :compatible, :nesting]
```
josevalim, line 2:
```suggestion
  @moduledoc false
  defstruct [:left, :right, :compatible, :nesting]
```

nx/test/nx/container_test.exs
```diff
@@ -158,7 +158,7 @@ defmodule Nx.ContainerTest do
       var.a + var.b
     end
 
-    deftransformp dot_assert_fields_transform(%C{c: %{}, d: %{}}), do: 1
+    deftransformp(dot_assert_fields_transform(%C{c: %{}, d: %{}}), do: 1)
```
josevalim, line 161: Check your editor :)

nx/lib/nx/defn/compilation_diff.ex
```diff
@@ -0,0 +1,96 @@
...
+
+  defp is_valid_container?(t) do
+    impl = Nx.Container.impl_for(t)
```
josevalim, line 5: You are getting the container several times. Get it once at the top of `build` for both `left` and `right` and then pass it around. :)

josevalim (commented):
Absolutely awesome!

nx/lib/nx/defn/expr.ex
```diff
@@ -439,13 +439,9 @@ defmodule Nx.Defn.Expr do
...
 
-                Got mismatching templates:
+                Comparison:
 
-                #{inspect_as_template(first)}
-
-                and
-
-                #{inspect_as_template(expr)}
+                #{Nx.Defn.CompilationDiff.build_and_inspect(first, expr)}
```
josevalim, line 444: This will be wrong for cond. cond actually allows different shapes and sizes, but not different containers. You need to pass a function that tells you how to compare tensors!

polvalente: Ah right, just like above, right?

nx/lib/nx/defn.ex
```diff
@@ -331,13 +331,9 @@ defmodule Nx.Defn do
 
             #{inspect(arg_template)}
```
polvalente, line 332: Should we still put the "Expected" vs "Argument" part above given the new diff implementation?

nx/test/nx/defn_test.exs
```diff
@@ -2572,7 +2578,7 @@ defmodule Nx.DefnTest do
     deftransformp(multi_clause_bodiless_tf_private(1, y), do: y)
     deftransformp(multi_clause_bodiless_tf_private(x, _y), do: x)
 
-    defn(multi_clause_transform_bodiless3(a), do: multi_clause_bodiless_tf_private(a))
+    defn multi_clause_transform_bodiless3(a), do: multi_clause_bodiless_tf_private(a)
```
polvalente, line 2581: @josevalim apparently there were lot's of defn(...)  already in the code. Should I leave them as they are right now or rollback?

josevalim: We should not change anything unrelated in this commit. :) but ideally all without parents!

nx/lib/nx/defn/expr.ex
```diff
@@ -764,13 +760,9 @@ defmodule Nx.Defn.Expr do
...
 
-        Body matches template:
-
-        #{inspect_as_template(body)}
-
-        and initial argument has template:
+        Comparison:
 
-        #{inspect_as_template(initial)}
+        #{Nx.Defn.TemplateDiff.build_and_inspect(body, initial)}
```
josevalim, line 765: Don’t we lose the context of which one is body and which one is initial? Can we say something like (green is body, red is intitial)?

nx/lib/nx/defn/template_diff.ex
```diff
@@ -0,0 +1,99 @@
...
+        IO.ANSI.reset()
+      ])
+    end
+
+    defp inspect_as_template(data) do
+      if is_number(data) or is_tuple(data) or
+           (is_map(data) and Nx.Container.impl_for(data) != Nx.Container.Any) do
+        data
+        |> Nx.to_template()
+        |> Kernel.inspect(custom_options: [skip_template_backend_header: true])
```
josevalim, line 93: You don’t want to call Kernel.inspect but rather Inspect.Algebra.to_doc or similar. You can put your custom options directly under the custom options field of the `Opts` struct.

nx/lib/nx/defn/template_diff.ex
```diff
@@ -0,0 +1,99 @@
...
+      else
+        inspect(data)
```
josevalim, line 95: Same here, call to_doc!

nx/lib/nx/defn/expr.ex
polvalente: We can add some parameterized labelling.

Something like this:

`build_and_inspect(body, "body", initial, "initial")`

which would print:

```
<<<<< body <<<<<
green thing
==========
red thing
>>>>> initial >>>>>
```

or something similar

josevalim: That looks great to me, your call :)

polvalente: Nice, I'll apply this then. Makes it easier for non-colorized outputs too

polvalente:
Still working on the bureaucracy for the code (tests passing and whatnot), but here's the current error messages for JIT, while and cond.

```elixir
defmodule M do
  import Nx.Defn

  defn container_cond_raise_type_mismatching(x) do
    container = %{a: 1, b: 2}

    cond do
      x == 1 ->
        container

      x == 2 ->
        container

      x == 3 ->
        container

      true ->
        1
    end
  end

  defn while_error(x) do
    while {x, m = %{a: 1, b: 2 * x}}, _ <- 0..5 do
      {Nx.as_type(x, :f32), %{m | a: 2, b: 2.0 * m.b}}
    end
  end

  def jit_error(x, y) do
    f = Nx.Defn.compile(&{&1, &2}, [{Nx.template({}, :s64)}, %{a: Nx.template({}, :s64)}])

    f.(x, y)
  end
end
```

All screenshots are based on errors on the code above.

Cond:
01_cond_error_message

While:
02_while_error_message

JIT:
03_jit_error_message


@josevalim regarding the JIT and Cond screenshots: I noticed that the compiler uses 1-based index for referring to the arguments. Should I also use 1-based indexing for the cond clause?

nx/lib/nx/defn/template_diff.ex
```diff
@@ -0,0 +1,159 @@
...
+    end
+
+    defp inspect_as_template(data) do
+      if is_number(data) or is_tuple(data) or
+           (is_map(data) and Nx.Container.impl_for(data) != Nx.Container.Any) do
+        data
+        |> Nx.to_template()
+        |> to_doc(Inspect.Opts.new(custom_options: [skip_template_backend_header: true]))
+      else
+        inspect(data)
```
polvalente, line 155:
```suggestion
        to_doc(data)
```

nx/lib/nx/defn/template_diff.ex
```diff
@@ -0,0 +1,158 @@
...
+    end
+
+    defp inspect_as_template(data) do
+      if is_number(data) or is_tuple(data) or
+           (is_map(data) and Nx.Container.impl_for(data) != Nx.Container.Any) do
+        data
+        |> Nx.to_template()
+        |> to_doc(Inspect.Opts.new(custom_options: [skip_template_backend_header: true]))
+      else
+        to_doc(data, Inspect.Opts.new([]))
```
josevalim, line 154: You should not create new options. Instead pass the options that you receive on `def inspect` (you are currently discarding them).

polvalente: done :)

josevalim (approved):

PR #1247, opened 2023-06-12
feat(EXLA): Add support for MLIR compilation

seanmor5:
Requires https://github.com/elixir-nx/xla/pull/45

There's a lot of other changes that need to be made as well. Metal requires MLIR and not XLA Op output, so I need to figure out how to bridge those


exla/lib/exla/client.ex
```diff
@@ -170,6 +170,11 @@ defmodule EXLA.Client do
         :tpu ->
           EXLA.NIF.get_tpu_client()
 
+        :metal ->
+          # TODO: Is this really where/how we want to do this?
+          :ok = EXLA.NIF.load_pjrt_plugin("METAL", "/opt/homebrew/lib/python3.10/site-packages/jax_plugins/metal_plugin/pjrt_plugin_metal_14.dylib")
```
polvalente, line 175: I believe this should be linked in the respective release for elixir-nx/xla directly, and then you can have something like this coming from there:

https://github.com/elixir-nx/nx/blob/219c23ecd75f93a7ad714e9b7b1e01d4a0bdf4bd/torchx/c_src/torchx.cpp#L347-L355

polvalente: For Torchx, this flag is defined here:

https://github.com/elixir-nx/nx/blob/219c23ecd75f93a7ad714e9b7b1e01d4a0bdf4bd/torchx/CMakeLists.txt#L46-L48

I assume you'd do something similar for EXLA

josevalim: If we want to keep doing it on Elixir land: we can load the plugin on EXLA.Application. I also agree that ideally we would bundle the plugin with the precompiled XLA binary for macOS.

josevalim: What I mean is that we ship the plugin with XLA and in here we point to XLA priv dir to load it. Would that work?

polvalente (approved):

PR #1254, opened 2023-06-14
feat: support negative Nx.split index

polvalente:


nx/lib/nx.ex
```diff
@@ -3355,18 +3355,36 @@ defmodule Nx do
...
+  def split(tensor, split, opts) when is_number(split) do
     tensor = to_tensor(tensor)
     opts = keyword!(opts, axis: 0)
     axis = Keyword.fetch!(opts, :axis)
 
     axis = Nx.Shape.normalize_axis(tensor.shape, axis, tensor.names)
     axis_size = axis_size(tensor, axis)
 
+    # negative index `-n` is the the same as `axis_size - n`
+    split = if is_integer(split) and split < 0, do: axis_size + split, else: split
```
josevalim, line 3386: I would rather add a clause to the `cond` below.

nx/lib/nx.ex
```diff
@@ -3355,18 +3355,36 @@ defmodule Nx do
...
+      #Nx.Tensor<
+        s64[1]
+        [4]
+      >
   """
   @doc type: :indexed
   def split(tensor, split, opts \\ [])
 
-  def split(tensor, split, opts) do
+  def split(tensor, split, opts) when is_number(split) do
```
josevalim, line 3377: If you add this guard, then you can remove `true ->` from the `cond`, but I think the `cond` is clearer. So I would not change this here.

josevalim (approved):

PR #1255, opened 2023-06-14
Add integration using trapezoidal rule

msluszniak:


nx/lib/nx.ex
```diff
@@ -9377,6 +9377,112 @@ defmodule Nx do
...
+      >
+
+      iex> Nx.trapz(Nx.tensor([[0, 1, 2], [3, 4, 5]]), Nx.tensor([[1, 1, 1], [2, 2, 2]]), axis: 0)
+      #Nx.Tensor<
+        f32[3]
+        [1.5, 2.5, 3.5]
+      >
+  """
+  @doc type: :aggregation, from_backend: false
+  def trapz(y, x, opts \\ []) do
```
polvalente, line 9409: I'd rename to a more explicit name

polvalente: I think the function should receive x and y instead of y and x

nx/lib/nx.ex
```diff
@@ -9377,6 +9377,112 @@ defmodule Nx do
...
+    opts = keyword!(opts, axis: -1)
+    %T{shape: y_shape, names: y_names} = y = to_tensor(y)
+    %T{shape: x_shape} = x = to_tensor(x)
+
+    axis = Nx.Shape.normalize_axis(y_shape, opts[:axis], y_names)
+
+    d =
+      if rank(x) == 0 do
+        x_diff = diff(x)
```
polvalente, line 9418: Isn't diff on a scalar 0?

polvalente: I don't think it makes sense to support rank 0 for x given that the trapezoidal rule needs 2 points for the calculation

polvalente: There should be some kind of compatibility check between x and y as well

nx/lib/nx.ex
```diff
@@ -9377,6 +9377,112 @@ defmodule Nx do
...
+      else
+        diff(x, axis: axis)
+      end
+
+    y_axis_size = elem(y_shape, axis)
+    scaler = divide(d, 2.0)
+
+    slice_along_axis(y, 0, y_axis_size - 1, axis: axis)
+    |> add(slice_along_axis(y, 1, y_axis_size - 1, axis: axis))
+    |> multiply(scaler)
```
polvalente, line 9435: I'd multiply y first to increase numerical stability

nx/lib/nx.ex
```diff
@@ -9377,6 +9377,112 @@ defmodule Nx do
...
+    |> sum(axes: [axis])
```
polvalente, line 9436: Is it desirable to have a keep_axis option?

nx/lib/nx.ex
```diff
@@ -9377,6 +9377,112 @@ defmodule Nx do
...
+    %T{shape: y_shape, names: y_names} = y = to_tensor(y)
+
+    axis = Nx.Shape.normalize_axis(y_shape, opts[:axis], y_names)
+
+    y_axis_size = elem(y_shape, axis)
+    scaler = divide(opts[:dx], 2.0)
+
+    slice_along_axis(y, 0, y_axis_size - 1, axis: axis)
+    |> add(slice_along_axis(y, 1, y_axis_size - 1, axis: axis))
+    |> multiply(scaler)
```
polvalente, line 9482: same consideration about stability here

polvalente (commented):
Looking good overall!

nx/lib/nx.ex
msluszniak: There should be == 1

nx/lib/nx.ex
msluszniak: We can ofc keep the axis, but at this point, I don't know if it would have some practical usage.

nx/lib/nx.ex
msluszniak: Scaler has a dimension `axis` smaller by one than y, so it won't work

nx/lib/nx.ex
msluszniak: Hmm, but in general y in this function is a key argument and originally x is just an optional argument

nx/lib/nx.ex
polvalente: You can still pre-multiply the `y` slices, though.

nx/lib/nx.ex
polvalente: I think we should provide the option anyway. It's too low a hanging fruit for us not to do it, I think.

nx/lib/nx.ex
josevalim: FWIW, scipy also has it as `(y, x)`: https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.trapezoid.html#scipy.integrate.trapezoid

josevalim:
Yeah, no problem. Numpy has a very large API surface though so I think if we can find opportunities to break it apart, we should and this looks like one of them. :)

So let's move it to Scholar!

nx/lib/nx.ex
polvalente: I still find it really confusing haha
But it makes sense if you're talking about having multiple arities.

Maybe a general `integrate(y, x \\ nil, opts \\ [])` function with `:method` as a mandatory option, then I can see `y` coming first

PR #1257, opened 2023-07-14
Update intro-to-nx.livemd

GuiHeurich:
Going through this guide, my expectation was to be able to use the variable `tensor` that I was instructed to create immediately above:

```
create your own {3, 3} tensor with named dimensions
```

But, if I try to reshape this {3, 3} tensor into a {1, 4} tensor, I get an error:

```
(ArgumentError) cannot reshape, current shape {3, 3} is not compatible with new shape {1, 4}
```

My suggestion would be to use a {1, 9} shape here, which would work to reshape the tensor defined above.


josevalim:

PR #1258, opened 2023-07-14
Add triu tril and tri functions

msluszniak:
This one will be helpful in K-fold implementation.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -1459,7 +1459,7 @@ defmodule Nx do
   @doc type: :creation
   def triu(tensor, opts \\ []) do
     opts = keyword!(opts, k: 0)
-    mask = tri(axis_size(tensor, -2), axis_size(tensor, -1), k: opts[:k]-1)
+    mask = tri(axis_size(tensor, -2), axis_size(tensor, -1), k: opts[:k] - 1)
     select(equal(mask, Nx.tensor(1, type: type(mask))), Nx.tensor(0, type: type(tensor)), tensor)
   end
 
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -1424,7 +1424,7 @@ defmodule Nx do
   def tril(tensor, opts \\ []) do
     opts = keyword!(opts, k: 0)
     mask = tri(axis_size(tensor, -2), axis_size(tensor, -1), k: opts[:k])
-    select(equal(mask, Nx.tensor(1, type: type(mask))), tensor, Nx.tensor(0, type: type(tensor)))
+    select(mask, tensor, 0)
   end
 
   @doc """
@@ -1460,7 +1460,7 @@ defmodule Nx do
   def triu(tensor, opts \\ []) do
     opts = keyword!(opts, k: 0)
     mask = tri(axis_size(tensor, -2), axis_size(tensor, -1), k: opts[:k] - 1)
-    select(equal(mask, Nx.tensor(1, type: type(mask))), Nx.tensor(0, type: type(tensor)), tensor)
+    select(mask, 0, tensor)
   end
 
   @doc """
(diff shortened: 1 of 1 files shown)


josevalim:
From my side, we should move this to `Nx.LinAlg`, we need to make sure it works as if it was batched for rank > 2 (including tests) and it raises for rank < 2.

But I would like to hear @polvalente thoughts before. :)

msluszniak:
> From my side, we should move this to `Nx.LinAlg`, we need to make sure it works as if it was batched for rank > 2 (including tests) and it raises for rank < 2.
> 
> But I would like to hear @polvalente thoughts before. :)

Yeah, we definitely need to add tests for batched tensors and rank checks. I'm not sure about `Nx.LinAlg`.

polvalente:
To elaborate more, I think they belong in `Nx` instead of `Nx.LinAlg` because they are just used for data masking. There are lots of Lin. Alg. operations that use them, but they aren't specific definitions of Lin. Alg. itself. Also, `Nx.eye` is a linear algebra definition that is present outside of Nx.LinAlg.

josevalim:
Alright. So let’s go with the changes above and merge it.

nx/lib/nx.ex
```diff
@@ -1391,6 +1391,117 @@ defmodule Nx do
...
+          [4, 5, 6],
+          [7, 8, 9]
+        ]
+      >
+  """
+  @doc type: :creation
+  def tril(tensor, opts \\ []) do
+    opts = keyword!(opts, k: 0)
+    mask = tri(axis_size(tensor, -2), axis_size(tensor, -1), k: opts[:k])
+    select(equal(mask, Nx.tensor(1, type: type(mask))), tensor, Nx.tensor(0, type: type(tensor)))
```
polvalente, line 1427:
```suggestion
    select(mask, tensor, Nx.tensor(0, type: type(tensor)))
```

nx/lib/nx.ex
```diff
@@ -1391,6 +1391,117 @@ defmodule Nx do
...
+          [0, 0, 6],
+          [0, 0, 0]
+        ]
+      >
+  """
+  @doc type: :creation
+  def triu(tensor, opts \\ []) do
+    opts = keyword!(opts, k: 0)
+    mask = tri(axis_size(tensor, -2), axis_size(tensor, -1), k: opts[:k] - 1)
+    select(equal(mask, Nx.tensor(1, type: type(mask))), Nx.tensor(0, type: type(tensor)), tensor)
```
polvalente, line 1463:
```suggestion
    select(mask, 0, tensor)
```

nx/lib/nx.ex
```diff
@@ -1391,6 +1391,117 @@ defmodule Nx do
...
+          [1, 1, 0],
+          [1, 1, 1],
+          [1, 1, 1]
+        ]
+      >
+  """
+  @doc type: :creation
+  def tri(n, m, opts \\ []) do
+    opts = keyword!(opts, k: 0)
+    greater_equal(new_axis(iota({n}), 1), new_axis(subtract(iota({m}), opts[:k]), 0))
```
polvalente, line 1502:
```suggestion
    greater_equal(iota({n, 1}), subtract(iota({0, m}), opts[:k]))
```

nx/lib/nx.ex
```diff
@@ -1391,6 +1391,117 @@ defmodule Nx do
     eye(Nx.shape(tensor), opts)
   end
 
+  @doc """
+  Lower triangle of an array.
```
josevalim, line 1395:
```suggestion
  Lower triangle of a matrix.
```

And we also need to implement and document the batching behaviour. :)

nx/lib/nx.ex
```diff
@@ -1391,6 +1391,182 @@ defmodule Nx do
...
+            [4, 5, 0, 0],
+            [8, 9, 10, 0]
+          ],
+          [
+            [12, 0, 0, 0],
+            [16, 17, 0, 0],
+            [20, 21, 22, 0]
+          ]
+        ]
+      >
```
josevalim, line 1438:
```suggestion
      >

```

josevalim:

PR #1261, opened 2023-07-17
Replace tril_strict(t) with Nx.tril(t, k: -1) in grad.ex

msluszniak:


josevalim:

PR #1264, opened 2023-07-20
Add logsumexp

krstopro:
Added logsumexp function which calculates the logarithm of the sum of the exponentials of the tensor elements in a numerically stable way. This is a useful function when working with log-probabilities and is implemented in numerical software packages; e.g. see SciPy https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.logsumexp.html or PyTorch https://pytorch.org/docs/stable/generated/torch.logsumexp.html. I've already implemented a special case of it in Gaussian Mixture Model module, see https://github.com/elixir-nx/scholar/blob/399f9348b2dee0666d65ad604427ba20b44ff7b9/lib/scholar/cluster/gmm.ex#L360-L376.
Several things I am not sure about:
- Whether doctype should be :aggregation?
- Should there be more unit tests? Those inside docstring seem to be covering everything.
- Dealing with infinities. One step of the function involves subtracting a maximum from the elements along axis. However, if one of the entries is :infinity the result will be :nan instead of :infitity. This is not a desired behaviour. In NumPy this is easily handled by setting all the infinities to zero before subtracting them:
https://github.com/scipy/scipy/blob/cfe80116aaa145061246b8aec0e98248fbefb835/scipy/special/_logsumexp.py#L102-L105
  I am not aware of a convenient way to do this in Nx. This is a potential PR blocker in my opinion, but I would like to hear 
  what you think as well.


msluszniak:
If you want to change all infinities into zeros the set `infinity_mask = tensor == Nx.Constants.infinity(Nx.type(tensor)) or cast type to float before if it is integer and then use `Nx.select` like that `tensor = Nx.select(infinity_mask, 0, tensor)`

nx/lib/nx.ex
```diff
@@ -16464,4 +16464,116 @@ defmodule Nx do
...
+      >
+  """
+  @doc type: :aggregation
+  def logsumexp(tensor, opts \\ []) do
+    opts = keyword!(opts, [:axes, :exp_scaling_factor, :keep_axes])
+    axes = opts[:axes]
+    keep_axes = opts[:keep_axes]
+    max = reduce_max(tensor, axes: axes, keep_axes: true)
+    exponentials = tensor |> subtract(max) |> exp
+    max = if keep_axes, do: max, else: squeeze(max)
```
polvalente, line 16568:
```suggestion
    max = if keep_axes, do: max, else: squeeze(max, axes: axes)
```

nx/lib/nx.ex
```diff
@@ -16464,4 +16464,116 @@ defmodule Nx do
...
+
+    if exp_scaling_factor = opts[:exp_scaling_factor] do
+      multiply(exp_scaling_factor, exponentials)
+    else
+      exponentials
+    end
```
polvalente, line 16574: I'd extract this to a variable. If piped directly into a pipe chain is kinda hard to read

nx/lib/nx.ex
```diff
@@ -16464,4 +16464,116 @@ defmodule Nx do
...
+    |> sum(axes: axes, keep_axes: keep_axes)
+    |> log
```
polvalente, line 16576:
```suggestion
    |> log()
```

nx/lib/nx.ex
krstopro: Agreed.

nx/lib/nx.ex
msluszniak, line 16568: You should call squeeze here with specified axis

polvalente:
> If you want to change all infinities into zeros the set `infinity_mask = tensor == Nx.Constants.infinity(Nx.type(tensor))` or cast type to float before if it is integer and then use `Nx.select` like that `tensor = Nx.select(infinity_mask, 0, tensor)`

We have `Nx.is_infinity` for creating the mask

nx/lib/nx.ex
krstopro: I'm surprised `mix format` didn't add the parentheses here.

nx/lib/nx.ex
polvalente: This is probably missing a test where there are 1-sized axes that are not part of the aggregation

krstopro:
Completely forgot about `Nx.select`. :)
Thanks!

nx/lib/nx.ex
```diff
@@ -16464,4 +16464,116 @@ defmodule Nx do
...
+        ]
+      >
+  """
+  @doc type: :aggregation
+  def logsumexp(tensor, opts \\ []) do
+    opts = keyword!(opts, [:axes, :exp_scaling_factor, :keep_axes])
+    axes = opts[:axes]
+    keep_axes = opts[:keep_axes]
+    max = reduce_max(tensor, axes: axes, keep_axes: true)
+    exponentials = tensor |> subtract(max) |> exp
```
polvalente, line 16567:
```suggestion
    exponentials = tensor |> subtract(max) |> exp()
```

krstopro: I'm almost convinced `mix format` used to add parentheses. Not sure why it didn't happen here as well.

nx/lib/nx.ex
krstopro: Indeed, this a good case for unit test.

krstopro:
Alright, updated the function and now it works with infinities in the intended way.
I suppose now I need to write at least one test for that.

nx/lib/nx.ex
```diff
@@ -16464,4 +16464,123 @@ defmodule Nx do
...
+  """
+  @doc type: :aggregation
+  def logsumexp(tensor, opts \\ []) do
+    type = type(tensor)
+    opts = keyword!(opts, [:axes, :exp_scaling_factor, :keep_axes])
+    axes = opts[:axes]
+    keep_axes = opts[:keep_axes]
+    max = reduce_max(tensor, axes: axes, keep_axes: true)
+    mask = is_infinity(max)
+    max = select(mask, Nx.tensor(0, type: type), max)
```
polvalente, line 16569: I have a question regarding this treatment of infinities.
Does it make sense to actually treat them here, given that the function is conceptually undefined for positive infinities and just becomes log(0) for negative infinities?

krstopro: What I meant is the following. If there is infinity in the tensor, max will be infinity as well. To keep it numerically stable, max is subtracted from the elements, and subtracting infinity from infinity gives NaN. I suppose the intended thing to return would be infinity as you stated.

krstopro: > Also, it would be nice to add a unit-test that's not a doctest for this case specifically, showing what happens if one entry and if all entries on the reduction axis are infinity and/or neg_infinity

Yes, indeed. Will write unit test for that as well as one checking if axes kept are being squeezed.

polvalente: Let's test this with the following tensor, aggregating on axes: [1]. You can just hardcode the result in the actual test.

vals = [:infinity, :neg_infinity, 0]
Nx.tensor(for x <- vals, y <- vals, z <- vals, do: [x, y, z])

krstopro:
Alright, added unit tests that check if the function removes only those axes given with the axes option as well as unit test that checks if the function works with infinities.

nx/lib/nx.ex
```diff
@@ -16464,4 +16464,123 @@ defmodule Nx do
...
+
+  ## Examples
+
+      iex> Nx.logsumexp(Nx.tensor([1, 2, 3, 4, 5, 6]))
+      #Nx.Tensor<
+        f32
+        6.456193447113037
+      >
+
+      iex> Nx.logsumexp(Nx.tensor([1, 2, 3, 4, 5, 6]), exp_scaling_factor: 0.5)
```
msluszniak, line 16490: I would describe the option `exp_scaling_factor` in docs

nx/test/nx_test.exs
```diff
@@ -3129,4 +3129,84 @@ defmodule NxTest do
...
+                 [13.0]
+               ])
+
+      assert Nx.logsumexp(tensor, axes: [2]) ==
+               Nx.tensor([
+                 [11.0, 12.0, 13.0]
+               ])
+    end
+
+    test "works with infinities" do
```
polvalente, line 3150: nice :)

nx/lib/nx.ex
krstopro: Agreed, added a short explanation - let me know what you think. Sorry it took me a while.

msluszniak: Looks good :)

msluszniak (approved):

polvalente (approved):

polvalente:
EXLA is failing due to precision issues. You can skip those doctests in the EXLA test file and copy them over as unit tests instead, so you can adjust the expected result.

krstopro:
> EXLA is failing due to precision issues. You can skip those doctests in the EXLA test file and copy them over as unit tests instead, so you can adjust the expected result.

You mean in `exla/test/backend_test.exs`? Would just changing `@precision_error_doctests` do the job?

polvalente:
> > EXLA is failing due to precision issues. You can skip those doctests in the EXLA test file and copy them over as unit tests instead, so you can adjust the expected result.
> 
> You mean in `exla/test/backend_test.exs`? Would just changing `@precision_error_doctests` do the job?

That's the way, but you also need to copy the skipped tests over so that we still test the functionality in EXLA. Changing the module attribute will only skip the tests.

krstopro:
> > > EXLA is failing due to precision issues. You can skip those doctests in the EXLA test file and copy them over as unit tests instead, so you can adjust the expected result.
> > 
> > 
> > You mean in `exla/test/backend_test.exs`? Would just changing `@precision_error_doctests` do the job?
> 
> That's the way, but you also need to copy the skipped tests over so that we still test the functionality in EXLA. Changing the module attribute will only skip the tests.

I see. I was wondering why the allowed precision errors are so high. ^_^

krstopro:
Should be fine now.

josevalim:
There was some confusion but, to be clear, I have removed the tests in EXLA. It is not an optional callback, so there is no benefit in testing it from EXLA. :) If there is a bug, it will be in one of the underlying operations, never logsumexp.

josevalim:

PR #1268, opened 2023-07-24
Support batch keys in Nx.Serving

josevalim:


nx/lib/nx/serving.ex
```diff
@@ -283,6 +283,45 @@ defmodule Nx.Serving do
...
+      #Nx.Tensor<
+        s64[10]
+        [0, 2, 4, 6, 8, 10, 12, 14, 16, 18]
+      >
+      iex> half_batch = Nx.Batch.concatenate([Nx.iota({10})]) |> Nx.Batch.key(:half)
+      iex> Nx.Serving.run(serving, half_batch)
+      #Nx.Tensor<
+        f32[10]
+        [0.0, 0.5, 1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5]
+      >
```
polvalente, line 318: I think you should also include an example for same batch key but different input shape. The paragraphs above mention this, but the only examples are in the unit tests.

josevalim: It is tricky because the benefit of different shapes only happen in process mode and then the example becomes overly long (and you can't run as easily).

josevalim:

PR #1272, opened 2023-07-27
feat: prune tensor expr inspect results

polvalente:
Adds pruning of in-between expressions so that we
it's easier to analyze the overall behavior of a Nx.Defn.Expr

### DIFF
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -1565,29 +1565,8 @@ defmodule Nx.Defn.Expr do
   end
 
   def inspect(tensor, opts) do
-    {_,
-     %{
-       exprs: exprs,
-       parameters: parameters,
-       length: length,
-       opts: %{limit: limit}
-     }} =
-      recur_inspect(tensor, %{
-        cache: %{},
-        exprs: [],
-        parameters: [],
-        opts: opts,
-        length: 0
-      })
-
-    {exprs, parameters} =
-      if limit < 0 do
-        [h | t] = exprs
-        exprs = [h | Enum.take(t, opts.limit - 1)] ++ [{"...", ""}]
-        {exprs, []}
-      else
-        {exprs, parameters}
-      end
+    {_, %{exprs: exprs, parameters: parameters, length: length}} =
+      recur_inspect(tensor, %{cache: %{}, exprs: [], parameters: [], opts: opts, length: 0})
 
     concat(line(), color("Nx.Defn.Expr", :map, opts))
     |> append_lines(parameters, length + 3)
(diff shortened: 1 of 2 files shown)


nx/lib/nx/defn/expr.ex
```diff
@@ -1565,8 +1565,45 @@ defmodule Nx.Defn.Expr do
...
+        opts.limit == 0 ->
+          [{"...", ""}]
+
+        (len = length(parameters)) > opts.limit ->
+          [h | t] = parameters
+          [h, {"...", ""} | Enum.drop(t, len - opts.limit)]
+
+        true ->
+          parameters
+      end
```
polvalente, line 1606: Currently this prints:

```elixir
Enum.reduce(1..50, param_tensor.(0, 0), &Nx.add(&2, param_tensor.(&1, &1))) |> IO.inspect(limit: 10); nil
#Nx.Tensor<
  s64
  
  Nx.Defn.Expr
  parameter a:0     s64
  parameter b:1     s64
  parameter d:2     s64
  parameter f:3     s64
  parameter h:4     s64
  parameter j:5     s64
  parameter l:6     s64
  parameter n:7     s64
  parameter p:8     s64
  ...               
  parameter dv:50   s64
  ...               
  de = add dc, dd   s64
  dg = add de, df   s64
  di = add dg, dh   s64
  dk = add di, dj   s64
  dm = add dk, dl   s64
  do = add dm, dn   s64
  dq = add do, dp   s64
  ds = add dq, dr   s64
  du = add ds, dt   s64
  dw = add du, dv   s64
>
```

nx/lib/nx/defn/expr.ex
```diff
@@ -1565,8 +1565,45 @@ defmodule Nx.Defn.Expr do
...
+
+    parameters =
+      cond do
+        opts.limit == 0 ->
+          [{"...", ""}]
+
+        (len = length(parameters)) > opts.limit ->
+          [h | t] = parameters
+          [h, {"...", ""} | Enum.drop(t, len - opts.limit)]
```
jonatanklosko, line 1602:
```suggestion
        length(parameters) > opts.limit ->
          [h | t] = parameters
          [h, {"...", ""} | Enum.take(t, -(opts.limit - 1))]
```

josevalim:
@polvalente sorry, I pushed something but I messed up.

polvalente:
> @polvalente it shuold be good now. I used your idea of appending/prepending to the parameters and using the limit as a flag of when to do that once, so we don't need to do a later check. We just need tests and we are good. :)

Awesome! I'll work on testing this :)

nx/test/nx/defn/expr_test.exs
```diff
@@ -392,5 +392,97 @@ defmodule Nx.Defn.ExprTest do
...
+             """
+
+      assert inspect(result, limit: 0) == """
+             #Nx.Tensor<
+               f32
+             \s\s
+               Nx.Defn.Expr
+               ...  \s
+             >\
+             """
```
polvalente, line 485: @josevalim tests are kinda verbose, but I wanted to cover most edge cases

josevalim (approved):

PR #1276, opened 2023-08-07
fix: load numpy 1-byte width arrays

### LINKED ISSUE #1275: Numpy decoding is failing on u8 values
I suspect that the decoding is failing because 1-byte values don't actually have endianness. Numpy seems to be encoding them in a different way somehow.

I tried the following fix:

```
defp parse_type(<<?', byte_order, type, size, ?'>>) do
    byte_order =
      case byte_order do
        ?> ->
          :big

        ?< ->
          :little

        # We can infer any endianness if the data only
        # occupies 1 byte
        ?| when size == ?1 ->
          :big

        # We can't just infer native endianness matches our native endianness
        endianness ->
          raise ArgumentError, "unsupported numpy endianness: #{endianness}"
      end
```

which got rid of the ArgumentError but resulted in a tensor that didn't have the correct shape

polvalente:
closes #1275


nx/lib/nx/serving.ex
```diff
@@ -388,7 +388,7 @@ defmodule Nx.Serving do
   separate process.
   """
   @callback handle_batch(Nx.Batch.t(), partition :: non_neg_integer(), state) ::
-              {:execute, (-> {Nx.Container.t(), metadata()}), state}
+              {:execute, (() -> {Nx.Container.t(), metadata()}), state}
```
josevalim, line 391: You need to format on a more recent Elixir version. :) (v1.15)

josevalim (approved):

PR #1278, opened 2023-08-12
Update XLA

jonatanklosko:
I will release `:xla` once we verify the tests pass on the GPU :)


josevalim (approved):

jonatanklosko:
FTR it fails on a GPU because the linked library is missing some definitions, I will fix and rebuilt.

PR #1281, opened 2023-08-15
Add multivariate normal distribution

krstopro:
Added `multivariate_normal` to `Nx.Random`. Few things I am not sure about:
- Default method is `:cholesky`. Cholesky is also default for `JAX` while SVD is default for `NumPy`.
- The function does not check if the covariance matrix is positive semi-definite (PSD). I was considering implementing this as well, but it seemed a bit of an overkill. It is done in `NumPy`, but not in `JAX`. Also, Cholesky decomposition will fail if covariance is not PSD.


nx/lib/nx/random.ex
```diff
@@ -551,6 +551,181 @@ defmodule Nx.Random do
...
+
+    a =
+      case opts[:method] do
+        :svd ->
+          {u, s, _} = Nx.LinAlg.svd(covariance)
+          u * Nx.sqrt(s)
+
+        :eigh ->
+          {s, u} = Nx.LinAlg.eigh(covariance)
+          u * Nx.sqrt(Nx.abs(s))
```
polvalente, line 648: Why take the sqrt of the eigenvalues here?

nx/lib/nx/random.ex
```diff
@@ -551,6 +551,181 @@ defmodule Nx.Random do
...
+            [-0.8857421875]
+          ]
+        ]
+      >
+
+      iex> key = Nx.Random.key(12)
+      iex> {multivariate_normal, _new_key} = Nx.Random.multivariate_normal(key, Nx.tensor([0, 0]), Nx.tensor([[1, 0], [0, 1]]), shape: {3, 2})
+      iex> multivariate_normal
+      #Nx.Tensor<
+        f32[3][2][2]
```
polvalente, line 608: I'm a bit confused about these output shapes. I guess they are related to the input tensor shapes, but given the rest of the behavior in Nx.Random, I'd expect the :shape option to fully control the output shape

krstopro: The last dimension is equal to the dimension of the distribution (in this case 2, as both mean and covariance have dimensions 2). If shape option is not provided (i.e. `nil`), then the output will be of shape `{dim}`.
If shape option is `{3, 2}`, then the output will be of shape `{3, 2, 2}`.
This is the same as in NumPy and JAX.
https://github.com/numpy/numpy/blob/104addf377577d12ea34547303822369fcb6e44e/numpy/random/_generator.pyx#L3777
https://jax.readthedocs.io/en/latest/_autosummary/jax.random.multivariate_normal.html#jax.random.multivariate_normal

nx/lib/nx/random.ex
polvalente: Makes sense!
You can add a line to the docs that specifies that if covariance isn't positive semi-definite, it will be garbage in, garbage out, and get rid of the abs

nx/lib/nx/random.ex
polvalente: The docs can use some improvement to clarify this. Something along the lines of the first 3 phrases of this answer should suffice :)

krstopro: Agreed!

nx/lib/nx/random.ex
krstopro: Done!

nx/lib/nx/random.ex
krstopro: And done!

krstopro:
@polvalente Alright, updated the function. Sorry it took me a while. I also updated one of the unit tests (the one for properties). Previously, off-diagonal values of the covariance were not estimated and tested. I guess this should be computed using `Nx.covariance` which we don't have at the moment.

nx/lib/nx/random.ex
```diff
@@ -551,6 +551,186 @@ defmodule Nx.Random do
...
+
+  ## Options
+
+    * `:type` - a float type for the returned tensor
+
+    * `:shape` - batch shape of the returned tensor, i.e. the prefix of the result shape exluding the last axis
+
+    * `:names` - the names of the returned tensor
+
+    * `:method` - a decomposition methond used for the covariance. Must be one of :svd, :eigh, and :cholesky.
```
polvalente, line 567:
```suggestion
    * `:method` - a decomposition method used for the covariance. Must be one of :svd, :eigh, and :cholesky.
```

nx/lib/nx/random.ex
```diff
@@ -551,6 +551,186 @@ defmodule Nx.Random do
...
+  @doc """
+  Returns a sample from a multivariate normal distribution with given `mean` and `covariance` (matrix).
+  The function assumes that the covariance is a positive semi-definite matrix.
+  Otherwise, the result will not be normally distributed.
+
+  ## Options
+
+    * `:type` - a float type for the returned tensor
+
+    * `:shape` - batch shape of the returned tensor, i.e. the prefix of the result shape exluding the last axis
```
polvalente, line 563:
```suggestion
    * `:shape` - batch shape of the returned tensor, i.e. the prefix of the result shape excluding the last axis
```

nx/lib/nx/random.ex
```diff
@@ -551,6 +551,186 @@ defmodule Nx.Random do
...
+
+    * `:names` - the names of the returned tensor
+
+    * `:method` - a decomposition methond used for the covariance. Must be one of :svd, :eigh, and :cholesky.
+      Default :cholesky. For singular covariance matrices, use :svd or :eigh.
```
polvalente, line 568:
```suggestion
      Defaults to :cholesky. For singular covariance matrices, use :svd or :eigh.
```

polvalente (approved):

polvalente:
> @polvalente Alright, updated the function. Sorry it took me a while. I also updated one of the unit tests (the one for properties). Previously, off-diagonal values of the covariance were not estimated and tested. I guess this should be computed using `Nx.covariance` which we don't have at the moment.

@krstopro We can add `covariance`. Let's move ahead with the PR as is. If you want to send a follow-up with covariance and replacing where appropriate here, it would be welcome!

krstopro:
> > @polvalente Alright, updated the function. Sorry it took me a while. I also updated one of the unit tests (the one for properties). Previously, off-diagonal values of the covariance were not estimated and tested. I guess this should be computed using `Nx.covariance` which we don't have at the moment.
> 
> @krstopro We can add `covariance`. Let's move ahead with the PR as is. If you want to send a follow-up with covariance and replacing where appropriate here, it would be welcome!

Sure, shouldn't be hard!

polvalente:
> Sure, shouldn't be hard!

Bold statement! :laughing:

PR #1283, opened 2023-08-18
Fix Nx.pad/2 with different backends

jonatanklosko:
🐈


nx/lib/nx.ex
```diff
@@ -4062,7 +4062,7 @@ defmodule Nx do
       shape = Nx.Shape.pad(tensor.shape, padding_config)
 
       out = %{tensor | type: output_type, shape: shape}
-      impl!(tensor).pad(out, tensor, pad_value, padding_config)
+      impl!(tensor, pad_value).pad(out, tensor, pad_value, padding_config)
```
josevalim, line 4065: Maybe we can add a test to EXLA suite?

josevalim (approved):

PR #1284, opened 2023-08-20
Minor improvement for the `matrix_power` function

kenken-neko:
Define a `matrix_power` function to avoid unnecessary processing when `power` is 1.


josevalim:
I believe we don’t need this because the loop in the next clause will be a no-op. :) what do you think?

kenken-neko:
This function has been closed as unnecessary.

Issue #1285, opened 2023-08-26
[Accessibility] Add option to reduce precision when printing tensors to screen


a-alhusaini:
It would make Nx a lot more accessible for screen readers if I could limit the precision of the output when printing tensors to screen.

I didn't know that the numbers Nx was printing to the screen were in scientific notation because I was skipping some of the digits with my screen reader (there is just too many of them)

This happenned here: https://github.com/elixir-nx/axon/issues/526

I propose we add a config option that limits precision when tensors are printed ot the screen. Instead of a tensor that looks like this:

```
#Nx.Tensor<
  f32[1][10]
  EXLA.Backend<host:0, 0.3007448411.1655570452.243887>
  [
    [2.3431260944184697e-23, 4.296128036651581e-11, 1.0659236316176752e-17, 0.9999992847442627, 6.712944780263512e-22, 1.0, 2.8145804840526482e-22, 2.4531429665408666e-10, 9.463095196338145e-9, 2.6594868813845096e-6]
  ]
>
```

We would have
```
#Nx.Tensor<
  f32[1][10]
  EXLA.Backend<host:0, 0.3007448411.1655570452.243887>
  [
    [2.97e-23, 4.81e-11, 1.52e-17, 0.9927, 6.12e-22, 1.0, 2.82e-22, 2.66e-10, 9.45e-9, 2.46e-6]
  ]
>
```

This shortened format would make Nx a lot easier to interact with for blind users (like me)

josevalim:
Hi @a-alhusaini! We should definitely do something because the precision we print is also misleading, f32 doesn't have such a large precision, and maybe fixing that will already improve this issue considerably. See #448.

a-alhusaini:
> the precision we print is also misleading,

I'm surprised I only noticed this when you pointed it out. This leads me to conclude that most Nx users don't really care about the exact numbers in their tensors. Maybe truncating the printed output should be a default instead of a configuration option?

Just an idea I had. Not sure if it is good or not.

josevalim:
The precision it prints has additional digits which we should discard. Maybe nobody cares about the additional digits but they may care about missing digits.

I don't know either, just adding more context. :)

PR #1286, opened 2023-08-26
feat: add custom float precision inspect option

polvalente:
relates to #1285


nx/test/nx/tensor_test.exs
```diff
@@ -129,4 +131,56 @@ defmodule Nx.TensorTest do
       end
     end
   end
+
+  describe "inspect" do
```
polvalente, line 135: @a-alhusaini This PR will introduce a paliative solution for this. The printing format (and maybe the actual option) will change in the future, but it should solve your problem in the short-term.

The changes below contain some examples on the inspected tensors in the "prints with configured precision" test.

For usage in IEx, you can use `IEx.configure(inspect: [custom_options: [nx_precision: 5]])` for example.

nx/lib/nx/backend.ex
```diff
@@ -234,10 +249,68 @@ defmodule Nx.Backend do
...
+
+  defp inspect_value(%Complex{} = val, precision), do: complex_to_string(val, precision)
+
+  defp inspect_value(integer, _) when is_integer(integer), do: Integer.to_string(integer)
+  defp inspect_value(:neg_infinity, _), do: "-Inf"
+  defp inspect_value(:infinity, _), do: "Inf"
+  defp inspect_value(:nan, _), do: "NaN"
+
+  defp float_to_string(float, precision) do
+    str = :erlang.float_to_binary(float, scientific: precision)
```
josevalim, line 266: This actually makes things worse in many scenarios, because it will always enforce an amount of digits, even when shorter representations are available:

```elixir
iex(4)> :erlang.float_to_binary 0.25, [scientific: 6]
"2.500000e-01"
iex(5)> Float.to_string 0.25
"0.25"
```

We would need to prune the result of `Float.to_string` by slightly fixing it afterwards.

nx/lib/nx/backend.ex
```diff
@@ -234,10 +249,85 @@ defmodule Nx.Backend do
...
+    else
+      "-#{abs_s}"
+    end
+  end
+
+  defp complex_to_string_component(n, precision) do
+    # convert number to float
+    n = if is_float(n), do: n, else: n * 1.0
+
+    str = float_to_string(n, precision)
```
josevalim, line 325:
```suggestion
    str = float_to_string(n * 1.0, precision)
```

josevalim: This code means that, if precision is enabled, `1+2i` will always be printed as `1.0+2.0i`.

nx/lib/nx/backend.ex
```diff
@@ -234,10 +249,85 @@ defmodule Nx.Backend do
...
+      end
+
+    integer_part <> "." <> decimal_part <> exponent_part
+  end
+
+  def complex_to_string(%Complex{re: re, im: im}, precision) do
+    re_str = complex_to_string_real(re, precision)
+    im_str = complex_to_string_component(im, precision)
+
+    "#{re_str}#{im_str}i"
```
josevalim, line 295: There is some duplication here. Could this instead be written as:

```suggestion    
    re_str = inspect_value(re, precision)

    case inspect_value(im, precision) do
      "-" <> _ = im_str -> re_str <> im_str <> "i"
      "+" <> _ = im_str -> re_str <> im_str <> "i"
      im_str -> re_str <> "+" <> im_str <> "i"
    end
```

The second clause may not be necessary.

nx/lib/nx/backend.ex
```diff
@@ -234,10 +249,85 @@ defmodule Nx.Backend do
...
-  defp inspect_value(integer) when is_integer(integer), do: Integer.to_string(integer)
-  defp inspect_value(float) when is_float(float), do: Float.to_string(float)
-  defp inspect_value(:neg_infinity), do: "-Inf"
-  defp inspect_value(:infinity), do: "Inf"
-  defp inspect_value(:nan), do: "NaN"
+  defp inspect_value(float, precision) when is_float(float) and is_integer(precision) do
+    float_to_string(float, precision)
+  end
+
+  defp inspect_value(float, nil) when is_float(float), do: Float.to_string(float)
```
josevalim, line 256: Nitpick. For consistency with complex_to_string, move the precision handling inside?

```suggestion
  defp inspect_value(float, precision) when is_float(float), do: float_to_string(float, precision)
```

nx/lib/nx/backend.ex
polvalente: Yes, but complex numbers are always floating point. The only place where they are possibly input as not floats is in the V and M sigils, so this isn't a problem.

nx/lib/nx.ex
```diff
@@ -6133,7 +6133,7 @@ defmodule Nx do
       #Nx.Tensor<
         f64[2][2]
         [
-          [-3.141592653589793, -0.0],
+          [-3.141592653589793, 0.0],
```
josevalim, line 6136: Is this change correct?

polvalente: Yes, it was intentional, for consistency with the custom handling of -0.0 in complex_to_string

josevalim (approved):

PR #1293, opened 2023-08-30
feat: finish MLIR scaffolding for EXLA.Defn

polvalente:


exla/lib/exla/computation.ex
```diff
@@ -70,6 +72,20 @@ defmodule EXLA.Computation do
...
+        %EXLA.MLIR.Function{module: module, xla_return_shape: ret_shape},
+        client,
+        arg_shapes,
+        _opts
+      ) do
+    EXLA.MLIR.Module.compile(
+      module,
+      client,
+      arg_shapes,
+      EXLA.Shape.make_tuple_shape([ret_shape])
```
polvalente, line 85: @josevalim we need to think about how to deal with this. EXLA wraps the result in a tuple automatically, so I just hardcoded this output value.

polvalente, line 85:
```suggestion
      # TO-DO (mlir): do not hardcode this single-item tuple output type
      EXLA.Shape.make_tuple_shape([ret_shape])
```

exla/lib/exla/builder.ex
```diff
@@ -3,25 +3,75 @@ defmodule EXLA.Builder do
...
     %Computation{ref: ref, output_shape: shape}
   end
+
+  def build(%EXLA.MLIR.Value{} = val) do
+    %EXLA.MLIR.Value{function: function, ref: root_ref} =
+      EXLA.MLIR.Value.get_tuple_element(val, 0)
+
+    %EXLA.MLIR.Function{ref: function_ref, module: %EXLA.MLIR.Module{ref: module_ref}} = function
+    :ok = EXLA.NIF.mlir_build(function_ref, root_ref)
+    # EXLA.NIF.dump_mlir_module(module_ref)
```
polvalente, line 74:
```suggestion
```

exla/lib/exla/builder.ex
```diff
@@ -3,25 +3,74 @@ defmodule EXLA.Builder do
...
-  def build(root = %Op{}) do
+  def build(%Op{} = root) do
     shape = EXLA.Op.get_shape(root)
     {:ok, ref} = EXLA.NIF.build(root.builder, root.ref)
     %Computation{ref: ref, output_shape: shape}
   end
+
+  def build(%EXLA.MLIR.Value{} = val) do
+    %EXLA.MLIR.Value{function: function, ref: root_ref} =
+      EXLA.MLIR.Value.get_tuple_element(val, 0)
```
polvalente, line 70:
```suggestion
      # TO-DO: do not hardcode fetching just the first item as the output
      EXLA.MLIR.Value.get_tuple_element(val, 0)
```

exla/lib/exla/builder.ex
```diff
@@ -3,25 +3,75 @@ defmodule EXLA.Builder do
...
+  def build(%Op{} = root) do
     shape = EXLA.Op.get_shape(root)
     {:ok, ref} = EXLA.NIF.build(root.builder, root.ref)
     %Computation{ref: ref, output_shape: shape}
   end
+
+  def build(%EXLA.MLIR.Value{} = val) do
+    %EXLA.MLIR.Value{function: function, ref: root_ref} =
+      # TO-DO: do not hardcode fetching just the first item as the output
+      EXLA.MLIR.Value.get_tuple_element(val, 0)
```
polvalente, line 71: we probably want to deal with this in a different manner

exla/lib/exla/defn.ex
```diff
@@ -844,6 +845,11 @@ defmodule EXLA.Defn do
   @bin_op [:add, :subtract, :multiply, :min, :max, :remainder, :pow, :divide, :atan2] ++
             [:bitwise_and, :bitwise_or, :bitwise_xor, :left_shift]
 
+  defp to_operator(op, [%EXLA.MLIR.Value{} = left, %EXLA.MLIR.Value{} = right], out, state)
+       when op in @bin_op do
+    apply(EXLA.MLIR.Value, op, [left, right])
+  end
```
polvalente, line 851: In theory, we just need to add the MLIR NIFs for all the `@bin_op`s and they'll be ready :)

polvalente: We should leave it to after this and sm-metal are merged to main

exla/lib/exla/computation.ex
josevalim: Is there an issue in always making it a tuple? A single-element tuple has no overhead afaik and it helped address bugs between CPU and GPU.

exla/lib/exla/builder.ex
```diff
@@ -3,25 +3,75 @@ defmodule EXLA.Builder do
...
+
+  defp exla_shape(%Nx.Tensor{} = t) do
+    EXLA.Shape.make_shape(t.type, t.shape)
+  end
+
   def new(name) when is_binary(name) do
     {:ok, ref} = EXLA.NIF.new_builder(name)
-    %Builder{ref: ref, parent: nil, name: name}
+    %__MODULE__{ref: ref, parent: nil, name: name}
   end
```
josevalim, line 55: I would hide this version of `new`.

exla/lib/exla/builder.ex
```diff
@@ -3,25 +3,75 @@ defmodule EXLA.Builder do
...
 
-  def new(builder = %Builder{ref: ref}, name) when is_binary(name) do
+  def new(builder = %__MODULE__{ref: ref}, name) when is_binary(name) do
     {:ok, ref} = EXLA.NIF.create_sub_builder(ref, name)
-    %Builder{ref: ref, parent: builder, name: name}
+    %__MODULE__{ref: ref, parent: builder, name: name}
   end
```
josevalim, line 60: Btw, this is used in nested scopes such as `if`, `cond`. So I imagine we will add another branch in the future. :)

exla/lib/exla/computation.ex
```diff
@@ -70,6 +72,21 @@ defmodule EXLA.Computation do
     }
   end
 
+  def compile(
+        %EXLA.MLIR.Function{module: module, xla_return_shape: ret_shape},
```
josevalim, line 76: Can we just call it `return_shape`?

exla/lib/exla/builder.ex
```diff
@@ -3,25 +3,75 @@ defmodule EXLA.Builder do
...
+    |> Tuple.to_list()
+    |> Enum.map(&mlir_type/1)
+  end
+
+  defp mlir_type(inputs) when is_list(inputs) do
+    Enum.map(inputs, &mlir_type/1)
+  end
+
+  defp mlir_type(%EXLA.Shape{} = shape) do
+    Type.new(shape)
```
josevalim, line 39: I would move the creation of Type.new fully to the C code right now. Have the C function accept EXLA.Shape and convert to Type there, get rid of the Type module. Otherwise it will be confusing keeping both `mlir_ret_type` and `xla_ret_shape`. We just need one type and the easier for now is EXLA.Shape. We can revisit later.

exla/lib/exla/computation.ex
polvalente: now that we're removing MLIR-specific things like MLIR.Type, yes

exla/lib/exla/builder.ex
polvalente: Yes, I only implemented the minimal changes to get things to work. The new branch would have been untested, so I didn't bother

exla/lib/exla/builder.ex
polvalente: hide === defp it?

exla/lib/exla/computation.ex
polvalente: Not exactly an issue. The problem right now is that ret_shape was taken from a tensor-return. I assume that returning a tuple or another container would need something else for actually handling here.

exla/lib/exla/builder.ex
polvalente: I turned it into a defp. Lots of tests needed changing, but I think that's expected.

exla/lib/exla/computation.ex
polvalente: I removed MLIR.Type, and this is now just return_shape :)

exla/lib/exla/builder.ex
polvalente: Done!

josevalim (approved):

PR #1294, opened 2023-08-30
Candlex: an Nx.Backend for candle ML framework

grzuy:
An experimental work-in-progress exploring the possibility of implementing an Nx.Backend for [candle-core](https://crates.io/crates/candle-core) rust crate.

Note that https://github.com/huggingface/candle is still in early development and a moving target.

~Still a lot of Nx.Backend behavior callbacks pending implementation, so plan it continue making progress here to cover as much as possible.~

~No CUDA at the moment, only CPU.~
~UPDATE: Decent CUDA support for a small subset of backend functions.~
~UPDATE2: Decent CUDA support for slightly more than half backend callbacks.~

UPDATE3:

- About ~90 out of 117 `Nx.Backend` callbacks have some form of implementation (complete or partial).
- Device support: Yes, CPU and GPU/CUDA
- Axon's `examples/vision/mnist.exs`  and `examples/generative/fashionmnist_autoencoder.exs` are working using `Candlex.Backend` with similar performance to `Torchx` both in cpu and cuda. Other examples failing because of still some missing callback or callback option.

Pending:

- [x] Consider `rustler_precompiled` to avoid dependency on the rust toolchain
- [ ] Update github actions to run candlex tests
- [ ] Cover more `Nx.Backend` callbacks (tensor operations)
- [ ] Better support for f16 and bf16
- [ ] Make it work with more elaborated Axon model examples

---

Updates to upstream candle as part of this effort:
https://github.com/huggingface/candle/pull/741
https://github.com/huggingface/candle/pull/925
https://github.com/huggingface/candle/pull/988
https://github.com/huggingface/candle/pull/996
https://github.com/huggingface/candle/pull/1037

---

UPDATE 4: Also published in a separate repo: https://github.com/mimiquate/candlex.


candlex/lib/candlex/backend.ex
```diff
@@ -0,0 +1,423 @@
+defmodule Candlex.Backend do
```
polvalente, line 1: This is looking nice. But I'm going to hold off on further reviews for now.

I'm not sure if we should add this to the Nx repo like it is now.

candlex/lib/candlex/backend.ex
```diff
@@ -0,0 +1,423 @@
...
+  # Binary ops
+
+  for op <- [:add, :equal, :greater_equal, :less, :less_equal, :max, :min, :multiply, :subtract] do
+    @impl true
+    def unquote(op)(%T{} = out, %T{} = left, %T{} = right) do
+      {left, right} = maybe_upcast(left, right)
+      {left, right} = maybe_broadcast_bin_args(out.shape, left, right)
+
+      left
+      |> Native.unquote(op)(right)
```
josevalim, line 148: Could you use candle's broadcast_add instead?

candlex/lib/candlex/backend.ex
```diff
@@ -0,0 +1,423 @@
...
+    |> unwrap!()
+    |> to_nx(out)
+  end
+
+  # Binary ops
+
+  for op <- [:add, :equal, :greater_equal, :less, :less_equal, :max, :min, :multiply, :subtract] do
+    @impl true
+    def unquote(op)(%T{} = out, %T{} = left, %T{} = right) do
+      {left, right} = maybe_upcast(left, right)
```
josevalim, line 144: I assume Candle never upcasts automatically?

candlex/lib/candlex/backend.ex
```diff
@@ -0,0 +1,423 @@
...
+      )
+      |> backend_transfer(__MODULE__, [])
+    end
+  end
+
+  @impl true
+  def select(%T{shape: shape} = out, pred, on_true, on_false) do
+    on_true =
+      on_true
+      |> Nx.as_type(Nx.type(out))
```
josevalim, line 120: Prefer to call `|> from_nx |> Native.cast` so you do less wrapping.

candlex/lib/candlex/backend.ex
```diff
@@ -0,0 +1,423 @@
...
+      "Candlex.Backend(#{:erlang.ref_to_list(ref)})",
+      Inspect.Algebra.line(),
+      result
+    ])
+  end
+
+  # defp device_option(_backend_options) do
+  #   # TODO: Support CUDA
+  #   :cpu
+  # end
```
josevalim, line 298: My suggestion: add device support sooner than later, because you need to know the device on binary operations and potentially transfer across devices if candle does not do it automatically.

josevalim (commented):
This is awesome!

I have dropped some comments and ultimately we will only know if we should merge this or not based on how complete `Candle` is. If it is missing many operations, then perhaps it is better to wait, but I don't see any problems from a glance. We will need to discuss between us too.

candlex/lib/candlex/backend.ex
grzuy: Oh, good call.

I'm already using the candle's `broadcast_*` variants for a few binary ops but I never removed broadcast here for those. Will do.

candlex/lib/candlex/backend.ex
grzuy: Correct.

Example of the error we get running `Nx.add(1, 2.2)` if we remove this upcast:

```
Candle Error: dtype mismatch in add, lhs: I64, rhs: F32
```

candlex/lib/candlex/backend.ex
grzuy: Done in 9f7c26d8115d97322cd2624436c41e44b3104d46

mrluc:
👏 🍿

grzuy:
> This is awesome!
> 
> I have dropped some comments and ultimately we will only know if we should merge this or not based on how complete `Candle` is. If it is missing many operations, then perhaps it is better to wait, but I don't see any problems from a glance. We will need to discuss between us too.

Hi guys, updated this pull request description with a summary of the current status of this effort.

candlex/lib/candlex/backend.ex
grzuy: Added support for CUDA.

All current tests in `test/candlex_test.exs` are both green on cpu and cuda for me :heavy_check_mark:

grzuy:
@josevalim @polvalente Got this PR to a point it might make sense to slow down a bit and wait for some feedback from you guys on whether you see a good fit for that inside Nx or release is separtely for now.

Thanks in advance!

grzuy:
While this package is in consideration for inclusion as part of this repo, found it valuable to publish separately too (https://github.com/mimiquate/candlex), so it's easier for anyone to navigate it, try it out, open issues, propose improvements via PRs, etc.

Plan to keep this PR up to date with any updates upstream.

Thanks again!

josevalim:
Hi @grzuy! I did another pass over it and it looks really solid!

However, after working on Nx over the last few days, i believe we should not merge this. Adding new features to Nx already requires us to add code to both EXLA and Torchx. Although both projects are in C++, it is already time consuming. Adding a Rust project means both additional tooling and an additional language, and I am worried it will harm our productivity.

Therefore I suggest to keep it in a separate project (but feel free to submit a PR to the README to link to it). :)

Have a good one!

PR #1296, opened 2023-08-31
feat: add binary ops

polvalente:


exla/c_src/exla/mlir/builder.h
```diff
@@ -24,6 +24,14 @@ class MLIRFunction {
...
+  mlir::Value MinOp(mlir::Value lhs, mlir::Value rhs);
+  mlir::Value MaxOp(mlir::Value lhs, mlir::Value rhs);
+  mlir::Value RemOp(mlir::Value lhs, mlir::Value rhs);
+  mlir::Value PowOp(mlir::Value lhs, mlir::Value rhs);
+  mlir::Value DivOp(mlir::Value lhs, mlir::Value rhs);
+  mlir::Value Atan2Op(mlir::Value lhs, mlir::Value rhs);
+  mlir::Value ShiftLeftOp(mlir::Value lhs, mlir::Value rhs);
 
   // TODO: Somehow protect this resource so we cannot
   // specify more operations after return?
```
josevalim, line 37: Nah, we just don't expose it via the API, it is fine:

```suggestion
```

exla/test/exla/mlir/executable_test.exs
```diff
@@ -24,4 +24,24 @@ defmodule EXLA.MLIR.ExecutableTest do
     expected = Nx.add(t2, t2)
     assert_equal(result, expected)
   end
+
+  describe "binary ops" do
+    @bin_ops [:add, :subtract, :multiply, :pow, :min] ++
+               [:max, :remainder, :atan2]
+    # TO-DO (mlir): test for type casting
```
josevalim, line 31: Maybe set the defn_options for this process to be `compiler: EXLA, compiler_mode: :mlir` and then include the doctests from Nx one by one? The doctests define and jit on the fly, so it will be good to test this. :)

exla/c_src/exla/mlir/builder.cc
```diff
@@ -30,6 +30,48 @@ mlir::Value MLIRFunction::AddOp(mlir::Value lhs, mlir::Value rhs) {
...
+  module_->builder()->setInsertionPointToEnd(&func_->getBody().back());
+  auto op = module_->builder()->create<mlir::mhlo::DivOp>(module_->builder()->getUnknownLoc(), lhs, rhs);
+  return op;
+}
+
+mlir::Value MLIRFunction::Atan2Op(mlir::Value lhs, mlir::Value rhs) {
+  module_->builder()->setInsertionPointToEnd(&func_->getBody().back());
+  auto op = module_->builder()->create<mlir::mhlo::Atan2Op>(module_->builder()->getUnknownLoc(), lhs, rhs);
+  return op;
+}
```
josevalim, line 73: I wonder if it is worth macro-fying those.

Btw, why do we need to define those? Should we instead call these directly from exla.cc?

josevalim (approved):

exla/c_src/exla/mlir/builder.cc
polvalente: I believe this is so we have access to the function `module_` without cluttering `exla.cc`.
Not sure about macros, though. Torchx has a (IMO) really high first-contribution barrier because of them.

PR #1297, opened 2023-09-02
feat: enable Nx.as_type via MLIR

polvalente:


exla/c_src/exla/mlir/builder.cc
```diff
@@ -105,11 +116,71 @@ MLIRModule::MLIRModule() {
...
+    case 11:
       return builder->getF32Type();
+    case 12:
+      return builder->getF64Type();
+    case 16:
+      return builder->getBF16Type();
   }
 }
 
+xla::PrimitiveType MLIRTypeToPrimitiveType(mlir::Type type) {
```
seanmor5, line 146: What do you think is the benefit of still using primitive types?

polvalente: I didn't think much about benefits, more about having to use them for getting the shape information

polvalente: There might be a way of getting around them

polvalente: Yeah, so this is something we'll have to deal with some time in the future:

I tried making it so that we could return a ref to the ShapedTypeComponents object as the shape in case of the MLIR implementation, but I got segfaulted into oblivion. So I'm gonna leave this MLIRTypeToPrimitiveType conversion for usage in the building of the xla::Shape in mlir_get_shape

polvalente: This refactor will probably clean up some other things like the TypeIntToMLIRType function too, but I think we can deal with this later

josevalim (approved):

PR #1324, opened 2023-10-01
feat: add fft2 and ifft2

### LINKED ISSUE #1323: 2D FFTs

polvalente:
Adds Nx.fft2 and Nx.ifft2
Also adds support for :axis option in Nx.fft and Nx.ifft
Fixes an issue with optional callbacks where options weren't being passed properly through to compilers.

closes #1323

### DIFF
--- a/exla/c_src/exla/exla_ops.cc
+++ b/exla/c_src/exla/exla_ops.cc
@@ -1,14 +1,14 @@
-#include "exla_nif_util.h"
 #include "exla_ops.h"
 
-#include "xla/client/xla_builder.h"
-#include "xla/client/xla_computation.h"
-#include "xla/client/lib/math.h"
+#include "exla_nif_util.h"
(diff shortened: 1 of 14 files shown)


polvalente:
Back to draft because I forgot to add the callbacks in EXLA and Torchx. What's there is ready for review, though.

polvalente:
@josevalim It seems that the way we've been dealing with optional callbacks, by not adding them to Nx.Defn.Expr makes it so that we can't add them to compilers.

Any suggestions on how to deal with this? I'm thinking we'll have to rethink the way we identify passthrough.

This isn't too much of a blocker because XLA's FFT is almost equivalent to our default implementation. Only difference is that we'll end up with an extra pair of transpose calls inside EXLA.

josevalim (approved):
Beautiful.

PR #1325, opened 2023-10-02
feat(mlir): sort

polvalente:


polvalente:
SortOp also solves variadic_sort for Nx.argsort, but I'm leaving that to a follow-up PR

PR #1345, opened 2023-10-17
Bug fix: enforce type assumption

GambolingPangolin:
This call reduces a list of tensors before passing to `update_grads`

https://github.com/elixir-nx/nx/blob/main/nx/lib/nx/defn/grad.ex#L222 

Most of the other matches for `update_grads` account for this as needed, but the `:optional` one did not.  This change fixes that.

### DIFF
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -239,6 +239,8 @@ defmodule Nx.Defn.Grad do
   end
 
   defp update_grads(:optional, [_call, expr], _ans, gs, _to_grad_ids, grads) do
+    gs = List.wrap(gs)
+
     {grads, []} =
       Composite.reduce(expr, {grads, gs}, fn child, {grads, [g | gs]} ->
         {Map.update(grads, child.data.id, [g], &[g | &1]), gs}


nx/lib/nx/defn/grad.ex
```diff
@@ -239,6 +239,8 @@ defmodule Nx.Defn.Grad do
   end
 
   defp update_grads(:optional, [_call, expr], _ans, gs, _to_grad_ids, grads) do
+    gs = List.wrap(gs)
```
polvalente, line 242: Could you add a test to the grads test file that regresses this fix?

polvalente (commented):
Thanks!

nx/lib/nx/defn/grad.ex
polvalente: Also, given the call at line 222, probably we should have a guard for lists here and call with a [g] list there.

WDYT?

GambolingPangolin: Sure, happy to add a test.  

I played around with possible changes a bit.  The call on 222 is mostly used for versions of `update_grad` which expect a tensor such as with `op` being `:sum`, with `:optional` being unusual.  So we cannot change the call site 222 without doing a more extensive refactor.  Also, the `:optional` version is called from both branches of the case (of which 222 is one), so we can't change it to take a tensor.

polvalente: Right, then I think this solution makes sense :)

GambolingPangolin:
I ran into this issue when trying to use the `solve` operation.  Looking at the test for `solve`, I saw that it was not using its own local definition.  When I changed the test to use its local definition, the test fails (against the `main` branch) because of exactly the bug addressed in this PR.  So I simply modified the `solve` test.  Also I changed the numbers in the test, to use a non-triangular matrix.   

I did not check the numbers for this test mod yet.  Also, I did not see how to test this change more directly, so I would appreciate a suggestion or hint!

Issue #1346, opened 2023-10-18
Load balancing for distributed Nx.Serving


mntns:
We're planning to run a `Nx.Serving` distributed over a heterogenous cluster that contains mainly CPU-based nodes and would like to route batches preferentially to GPU nodes (using EXLA's CUDA client).

As far as I can tell there is no way to achieve that with the current distribution mechanism, is that correct? The randomized load balancing performs weighted scheduling based on the partition count, but there doesn't seem to be a way to assign weights based on e.g. the machine type (statically) or have dynamic resource-based scheduling.

josevalim:
Yes, there is no static nor dynamic resource-based scheduling. We could allow weights to be assigned, that would be relatively easy, but dynamic routing is trickier on a distributed system. The best I can think of is to fallback to CPU if there is no GPU.

josevalim:
Yes, there is no static nor dynamic resource-based scheduling. We could allow weights to be assigned, that would be relatively easy, but dynamic routing is trickier on a distributed system. The best I can think of is to fallback to CPU if there is no GPU.

mntns:
Thanks for the quick response! I opened a PR that adds simple weights to that effect.

josevalim:
Closing in favor of the PR :)

PR #1348, opened 2023-10-19
Add weight option to Nx.Serving for static load balancing

mntns:
This adds a simple weight option to `Nx.Serving` that is factored into how many times the serving joins the process group, similar to `partition_count`, thereby making the randomized load balancing weighted.

Addresses https://github.com/elixir-nx/nx/issues/1346, at least in regard to static load balancing.

### DIFF
--- a/nx/lib/nx/serving.ex
+++ b/nx/lib/nx/serving.ex
@@ -190,10 +190,15 @@ defmodule Nx.Serving do
   the same code and applications. It is only required that they run the
   same `Nx` version.
 
-  The load balancing between servings is done randomly, however, the number
-  of partitions are considered if the `partitions: true` option is also given.
+  The load balancing between servings is done randomly by default, however,
+  the number of partitions are considered if the `partitions: true` option is also given.
   For example, if you have a node with 2 GPUs and another with 4, the latter
   will receive the double of requests compared to the former.
+
+  Furthermore, the load balancing allows for assigning weights to servings.
+  Similarly to the number of partitions, when running a serving with `weight: 1`
+  and another one with `weight: 2`, the latter will receive double the requests
+  compared to the former.
 
   `batched_run/3` receives an optional `distributed_preprocessing` callback as
   third argument for preprocessing the input for distributed requests. When
@@ -388,7 +393,8 @@ defmodule Nx.Serving do
     :partitions,
     :shutdown,
     :hibernate_after,
-    :spawn_opt
+    :spawn_opt,
+    :weight
   ]
 
   @doc """
@@ -873,6 +879,11 @@ defmodule Nx.Serving do
       The number of partitions will be determined according to your compiler
       and for which host it is compiling. See the module docs for more information
 
+    * `:weight` - weight used for load balancing when running a distributed serving
+      Defaults to `1`. If it is set to a higher number `w`, the serving process
+      will receive, on average, `w` times the number of requests compared to a
+      process that uses `weight: 1`.
+
     * `:shutdown` - the maximum time for the serving to shutdown. This will
       block until the existing computation finishes (defaults to `30_000`ms)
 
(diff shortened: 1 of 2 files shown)


nx/lib/nx/serving.ex
```diff
@@ -1300,7 +1317,7 @@ defmodule Nx.Serving do
       }
     )
 
-    :pg.join(Nx.Serving.PG, __MODULE__, List.duplicate(self(), partitions_count))
+    :pg.join(Nx.Serving.PG, __MODULE__, List.duplicate(self(), weight * partitions_count))
```
josevalim, line 1320: Should we add the `max(1)` here? Also, could someone want the weight to be a float? Should we validate it is an integer?

nx/lib/nx/serving.ex
```diff
@@ -873,6 +879,11 @@ defmodule Nx.Serving do
       The number of partitions will be determined according to your compiler
       and for which host it is compiling. See the module docs for more information
 
+    * `:weight` - weight used for load balancing when running a distributed serving
```
josevalim, line 882: Perhaps

```suggestion
    * `:distribution_weight` - weight used for load balancing when running a distributed serving
```

for the name?

nx/lib/nx/serving.ex
```diff
@@ -873,6 +879,11 @@ defmodule Nx.Serving do
...
+      Defaults to `1`. If it is set to a higher number `w`, the serving process
+      will receive, on average, `w` times the number of requests compared to a
+      process that uses `weight: 1`.
```
josevalim, line 885: We should document this is multiplied by the number of partitions?

nx/lib/nx/serving.ex
mntns: Yes, that's better, changed it :+1:

nx/lib/nx/serving.ex
mntns: Added a sentence in the docstring.

nx/lib/nx/serving.ex
mntns: I added the `max(1, weight)` in `init` instead of `start_link` and added a check for whether the weight is an integer. 

While I could imagine that someone might want to use a float for it, I don't think it works well with the current way of letting the process join multiple times, since the "join count" is a discrete value. The floats could be rescaled approximately to an discrete "join count" (doesn't seem very elegant to me, since it would lead to big PID list, that is later sampled from), but since even with integer weights a desired float value can be approximated by a weight ratio, the current solution seems to allow enough flexibility for me. For example, if I want a node to receive 45% of all requests on average, I can just set `distribution_weight: 9` and, assuming there is one other node, the other distribution weight to `11`.

mntns:
Added a simple test that relies on the implementation detail of how many times the process joins the group :)

nx/lib/nx/serving.ex
```diff
@@ -873,6 +879,13 @@ defmodule Nx.Serving do
       The number of partitions will be determined according to your compiler
       and for which host it is compiling. See the module docs for more information
 
+    * `:distribution_weight` - weight used for load balancing when running
+      a distributed serving. Defaults to `1`.
+      If it is set to a higher number `w`, the serving process will receive,
+      on average, `w` times the number of requests compared to a process that
+      uses `weight: 1`. Note that the weight is multiplied with the number of
```
josevalim, line 886:
```suggestion
      on average, `w` times the number of requests compared to the
      default. Note that the weight is multiplied with the number of
```

nx/lib/nx/serving.ex
```diff
@@ -901,11 +914,18 @@ defmodule Nx.Serving do
     partitions = Keyword.get(opts, :partitions, false)
     batch_keys = Keyword.get(opts, :batch_keys, [:default])
     batch_timeout = Keyword.get(opts, :batch_timeout, 100)
+    weight = opts |> Keyword.get(:distribution_weight, 1)
```
josevalim, line 917:
```suggestion
    weight = Keyword.get(opts, :distribution_weight, 1)
```

nx/lib/nx/serving.ex
```diff
@@ -1300,7 +1323,8 @@ defmodule Nx.Serving do
       }
     )
 
-    :pg.join(Nx.Serving.PG, __MODULE__, List.duplicate(self(), partitions_count))
+    serving_weight = max(1, weight) * partitions_count
```
josevalim, line 1326: In case we want to support floats in the future, this is the sanest default, right?

```suggestion
    serving_weight = max(1, weight * partitions_count)
```

josevalim:

PR #1358, opened 2023-11-08
refactor(torchx): more efficient put_slice

polvalente:
- wip: torchx put slice
- fix: return outer ref


josevalim (approved):
Beautiful. We may be able to simplify `slice` too!

PR #1360, opened 2023-11-09
fix: grad for indexed ops with axes

polvalente:


nx/lib/nx/binary_backend.ex
```diff
@@ -2097,10 +2097,20 @@ defmodule Nx.BinaryBackend do
...
           |> Enum.map(fn {_, i} -> i end)
 
+        diff = Nx.rank(out) - length(inverse_permutation)
+
+        inverse_permutation =
+          if diff > 0 do
+            Enum.to_list(0..(diff - 1)) ++ Enum.map(inverse_permutation, &(&1 + diff))
+          else
+            inverse_permutation
+          end
```
polvalente, line 2112: I'm not _totally_ sure if this is the correct fix. Basically the test for grad(gather) was failing and this fixed it

josevalim: Will we ever hit the else branch?

polvalente: It seems we can. Keeping only the `do` branch works if we change the range to `0..(diff-1)//1` because it seems that for some cases the diff is 0. I think those are cases where the index tensor as many axes as the input tensor.

The `Enum.filter(... in Nx.axes(out))` takes care of the case where the index tensor has fewer axes, and the `do` block takes care of when there are more axes than in the input.

josevalim (approved):

josevalim:

PR #1363, opened 2023-11-10
Exercises 1-20

billylanchantin:
## Description

As discussed in [Slack](https://the-eef.slack.com/archives/C01PU0RDJ8L/p1699644171094749), this PR adds a translation of the first 20 exercises from the Python notebook _100 Numpy Exercises_:

https://www.kaggle.com/code/utsav15/100-numpy-exercises/notebook

## Changes

* Adds `exercises-1-20.livemd` file
* Tweaks the doc grouping in `mix.exs`
    Screen Shot 2023-11-10 at 5 01 14 PM

## Notes

* Most of these first 20 translate 1-1. There were a few that didn't. I left comments (which we'll want to remove before merging).
* Going forward, not all of the exercises are going to make sense for Nx. This is (hopefully) more a jumping off point for community-curated exercises.

### DIFF
--- a/nx/exercises/exercises-1-20.livemd
+++ b/nx/exercises/exercises-1-20.livemd
@@ -12,40 +12,40 @@ https://www.kaggle.com/code/utsav15/100-numpy-exercises/notebook
 
 ## 1-10
 
-1. Install Nx in a Livebook (★☆☆)
+**1. Install Nx in a Livebook (★☆☆)**
 
 ```elixir
 Mix.install([{:nx, "~> 0.6"}])
 ```
 
-1. Create a 1-D tensor of 10 zeros (★☆☆)
+**2. Create a 1-D tensor of 10 zeros (★☆☆)**
 
 ```elixir
 Nx.broadcast(0, {10})
 ```
 
-1. How to find the number of elements in a tensor (★☆☆)
+**3. How to find the number of elements in a tensor (★☆☆)**
 
 ```elixir
 Nx.tensor([[1, 2, 3], [4, 5, 6]])
 |> Nx.size()
 ```
 
-1. How to find the memory size of any tensor (★☆☆)
+**4. How to find the memory size of any tensor (★☆☆)**
 
 ```elixir
 Nx.tensor([[1, 2, 3], [4, 5, 6]])
 |> Nx.byte_size()
 ```
 
-1. How to get the documentation of the `add/2` function from the command line? (★☆☆)
+**5. How to get the documentation of the `add/2` function from the command line? (★☆☆)**
 
 ```elixir
 # I don't know if this is possible.
 # iex> h Nx.add
 ```
 
-1. Create a tensor of zeros of size 10 but the fifth value which is 1 (★☆☆)
+**6. Create a tensor of zeros of size 10 but the fifth value which is 1 (★☆☆)**
 
 ```elixir
 zeros = Nx.broadcast(0, {10})
@@ -53,26 +53,26 @@ index = Nx.tensor([4])
 Nx.indexed_put(zeros, index, 1)
 ```
 
-1. Create a tensor with values ranging from 10 to 49 (★☆☆)
+**7. Create a tensor with values ranging from 10 to 49 (★☆☆)**
 
 ```elixir
 Nx.linspace(10, 49, n: 39, type: :s8)
 ```
 
-1. Reverse a tensor (first element becomes last) (★☆☆)
+**8. Reverse a tensor (first element becomes last) (★☆☆)**
 
 ```elixir
 Nx.iota({10})
 |> Nx.reverse()
 ```
 
-1. Create a 3x3 tensor with values ranging from 0 to 8 (★☆☆)
+**9. Create a 3x3 tensor with values ranging from 0 to 8 (★☆☆)**
 
 ```elixir
 Nx.iota({3, 3})
 ```
 
-1. Find indices of non-zero elements from [1,2,0,0,4,0] (★☆☆)
+**10. Find indices of non-zero elements from [1,2,0,0,4,0] (★☆☆)**
 
 ```elixir
 Nx.tensor([1, 2, 0, 0, 4, 0])
@@ -81,13 +81,13 @@ Nx.tensor([1, 2, 0, 0, 4, 0])
 
 ## 11-20
 
-1. Create a 3x3 identity tensor (★☆☆)
+**11. Create a 3x3 identity tensor (★☆☆)**
 
 ```elixir
 Nx.eye(3)
 ```
 
-1. Create a 3x3x3 tensor with random values (★☆☆)
+**12. Create a 3x3x3 tensor with random values (★☆☆)**
 
 ```elixir
 key = Nx.Random.key(0)
@@ -95,7 +95,7 @@ key = Nx.Random.key(0)
 random
 ```
 
-1. Create a 10x10 tensor with random values and find the minimum and maximum values (★☆☆)
+**13. Create a 10x10 tensor with random values and find the minimum and maximum values (★☆☆)**
 
 ```elixir
 # This felt a little awkward. Is there a better way?
@@ -110,7 +110,7 @@ r |> IO.inspect()
 }
 ```
 
-1. Create a random 1D tensor of size 30 and find the mean value (★☆☆)
+**14. Create a random 1D tensor of size 30 and find the mean value (★☆☆)**
 
 ```elixir
 key = Nx.Random.key(0)
@@ -118,21 +118,21 @@ key = Nx.Random.key(0)
 Nx.mean(r)
 ```
 
-1. Create a 5x5 tensor with 1 on the border and 0 inside (★☆☆)
+**15. Create a 5x5 tensor with 1 on the border and 0 inside (★☆☆)**
 
 ```elixir
 Nx.broadcast(0, {3, 3})
 |> Nx.pad(1, [{1, 1, 0}, {1, 1, 0}])
 ```
 
-1. How to add a border (filled with 0's) around an existing array? (★☆☆)
+**16. How to add a border (filled with 0's) around an existing array? (★☆☆)**
 
 ```elixir
 Nx.broadcast(1, {3, 3})
 |> Nx.pad(0, [{1, 1, 0}, {1, 1, 0}])
 ```
 
-1. What are the results of the following expressions? (★☆☆)
+**17. What are the results of the following expressions? (★☆☆)**
 
 ```elixir
 nan = Nx.Constants.nan()
@@ -142,20 +142,20 @@ IO.inspect(Nx.greater(nan, nan))
 IO.inspect(Nx.subtract(nan, nan))
 ```
 
-1. Create a 5x5 tensor with values 1,2,3,4 just below the diagonal (★☆☆)
+**18. Create a 5x5 tensor with values 1,2,3,4 just below the diagonal (★☆☆)**
 
 ```elixir
 Nx.make_diagonal(Nx.tensor([1, 2, 3, 4]), offset: -1)
 ```
 
-1. Create a 8x8 tensor of 0 and 1 in a checkerboard pattern with 0 as the first element (★☆☆)
+**19. Create a 8x8 tensor of 0 and 1 in a checkerboard pattern with 0 as the first element (★☆☆)**
 
 ```elixir
 Nx.tensor([[0, 1], [1, 0]])
 |> Nx.tile([4, 4])
 ```
 
-1. What is the index (x,y,z) of the 100th element of a 6x7x8 tensor?
+**20. What is the index (x,y,z) of the 100th element of a 6x7x8 tensor?**
 
 ```elixir
 # The numpy solution is to use `unravel_index`. I didn't see an equivalent.
--- a/nx/exercises/exercises-1-20.livemd
+++ b/nx/exercises/exercises-1-20.livemd
@@ -138,8 +138,6 @@ IO.inspect(Nx.multiply(0, nan))
 IO.inspect(Nx.equal(nan, nan))
 IO.inspect(Nx.greater(nan, nan))
 IO.inspect(Nx.subtract(nan, nan))
-IO.inspect(Enum.member?(MapSet.new([nan]), nan))
-IO.inspect(0.3 == 3 * 0.1)
 ```
 
 1. Create a 5x5 tensor with values 1,2,3,4 just below the diagonal (★☆☆)
--- a/nx/exercises/exercises-1-20.livemd
+++ b/nx/exercises/exercises-1-20.livemd
@@ -442,7 +442,7 @@ Nx.subtract(nan, nan)
 
 <!-- livebook:{"break_markdown":true} -->
 
-**19. Create a 8x8 tensor of 0 and 1 in a checkerboard pattern with 0 as the first element. (★☆☆)**
+**19. Create a 8x8 tensor of 0 and 1 in a checkerboard pattern with 0 as the first element using [`Nx.tile`](https://hexdocs.pm/nx/Nx.html#tile/2). (★☆☆)**
 
 ```elixir
 # Add your solution here.
@@ -462,34 +462,38 @@ Nx.subtract(nan, nan)
 
 <!-- livebook:{"break_markdown":true} -->
 
-**20. Find the (x,y,z) index ([row major](https://en.wikipedia.org/wiki/Row-_and_column-major_order)) of the 100th element of a 6x7x8 tensor. (★★☆)**
+**20. Produce the same checkerboard pattern from exercise 19, but _without_ using `Nx.tile`. (★☆☆)**
 
 ```elixir
 # Add your solution here.
+# Hint: try using `Nx.iota` in combination with `Nx.remainder`.
 ```
 
 <details style="margin-left: 32px">
-  <summary style="cursor: pointer; font-weight: bold;">Example solution</summary>
+  <summary style="cursor: pointer; font-weight: bold;">Example solution 1</summary>
   <div style="padding-top: 8px">
 
   ```elixir
-  defmodule Exercise20 do
-    import Nx.Defn
+  t = Nx.iota({8, 1})
 
-    deftransform solve(shape, target) do
-      dims = Tuple.to_list(shape)
-      
-      dims
-      |> Enum.map_reduce({target, Enum.product(dims)}, fn dim, {remaining, step_size} ->
-        step_size = div(step_size, dim)
-        num_steps = div(remaining, step_size)
-        {num_steps, {remaining - num_steps * step_size, step_size}}
-      end)
-      |> then(fn {indices, {0, 1}} -> indices end)
-    end
-  end
+  Nx.transpose(t)
+  |> Nx.add(t)
+  |> Nx.remainder(2)
+  ```
 
-  Exercise20.solve({6, 7, 8}, 100)
+  </div>
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
+
+<details style="margin-left: 32px">
+  <summary style="cursor: pointer; font-weight: bold;">Example solution 2</summary>
+  <div style="padding-top: 8px">
+
+  ```elixir
+  Nx.iota({9, 9})
+  |> Nx.remainder(2)
+  |> Nx.slice([0, 0], [8, 8])
   ```
 
   </div>
--- a/nx/exercises/exercises-1-20.livemd
+++ b/nx/exercises/exercises-1-20.livemd
@@ -12,168 +12,479 @@ https://www.kaggle.com/code/utsav15/100-numpy-exercises/notebook
 
 ## 1-10
 
-**1. Install Nx in a Livebook (★☆☆)**
+**1. Install `Nx` in a Livebook. (★☆☆)**
 
 ```elixir
-Mix.install([{:nx, "~> 0.6"}])
+# Add your solution here.
+:ok
 ```
 
-**2. Create a 1-D tensor of 10 zeros (★☆☆)**
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
+  <summary class="cursor-pointer font-bold mx-2">Example solution</summary>
+  <div class="p-2">
+
+  ```elixir
+  Mix.install([{:nx, "~> 0.6"}])
+  ```
+
+  </div>
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
+
+**2. Create a 1-D tensor of 10 zeros. (★☆☆)**
 
 ```elixir
-Nx.broadcast(0, {10})
+# Add your solution here.
+:ok
 ```
 
-**3. How to find the number of elements in a tensor (★☆☆)**
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
+  <summary class="cursor-pointer font-bold mx-2">Example solution</summary>
+  <div class="p-2">
+
+  ```elixir
+  Nx.broadcast(0, {10})
+  ```
+
+  </div>
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
+
+**3. Find the number of elements in `tensor_03`. (★☆☆)**
 
 ```elixir
-Nx.tensor([[1, 2, 3], [4, 5, 6]])
-|> Nx.size()
+tensor_03 = Nx.tensor([[1, 2, 3], [4, 5, 6]])
+# Add your solution here.
+:ok
 ```
 
-**4. How to find the memory size of any tensor (★☆☆)**
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
+  <summary class="cursor-pointer font-bold mx-2">Example solution</summary>
+  <div class="p-2">
+
+  ```elixir
+  Nx.size(tensor_03)
+  ```
+
+  </div>
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
+
+**4. Find the number of bytes of memory in `tensor_04`. (★☆☆)**
 
 ```elixir
-Nx.tensor([[1, 2, 3], [4, 5, 6]])
-|> Nx.byte_size()
+tensor_04 = Nx.tensor([[1, 2, 3], [4, 5, 6]])
+# Add your solution here.
+:ok
 ```
 
-**5. How to get the documentation of the `add/2` function from the command line? (★☆☆)**
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
+  <summary class="cursor-pointer font-bold mx-2">Example solution</summary>
+  <div class="p-2">
+
+  ```elixir
+  Nx.byte_size(tensor_04)
+  ```
+
+  </div>
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
+
+**5. Use `IEx.Helpers` to print the documentation of the `Nx.add/2` function. (★☆☆)**
 
 ```elixir
-import IEx.Helpers
-h(Nx.add())
+# Add your solution here.
+:ok
 ```
 
-**6. Create a tensor of zeros of size 10 but the fifth value which is 1 (★☆☆)**
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
+  <summary class="cursor-pointer font-bold mx-2">Example solution</summary>
+  <div class="p-2">
+
+  ```elixir
+  import IEx.Helpers
+  h Nx.add
+  ```
+
+  </div>
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
+
+**6. Create a tensor of zeros of size 10 but where the fifth value is 1. (★☆☆)**
 
 ```elixir
-zeros = Nx.broadcast(0, {10})
-index = Nx.tensor([4])
-Nx.indexed_put(zeros, index, 1)
+# Add your solution here.
+:ok
 ```
 
-**7. Create a tensor with values ranging from 10 to 49 (★☆☆)**
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
+  <summary class="cursor-pointer font-bold mx-2">Example solution</summary>
+  <div class="p-2">
+
+  ```elixir
+  zeros = Nx.broadcast(0, {10})
+  index = Nx.tensor([4])
+  Nx.indexed_put(zeros, index, 1)
+  ```
+
+  </div>
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
+
+**7. Create a tensor with values ranging from 10 to 49. (★☆☆)**
 
 ```elixir
-Nx.linspace(10, 49, n: 39, type: :s8)
+# Add your solution here.
+:ok
 ```
 
-**8. Reverse a tensor (first element becomes last) (★☆☆)**
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
+  <summary class="cursor-pointer font-bold mx-2">Example solution</summary>
+  <div class="p-2">
+
+  ```elixir
+  Nx.linspace(10, 49, n: 39, type: :s8)
+  ```
+
+  </div>
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
+
+**8. Reverse `tensor_08` (first element becomes last). (★☆☆)**
 
 ```elixir
-Nx.iota({10})
-|> Nx.reverse()
+tensor_08 = Nx.tensor([2, 4, 6, 8])
+# Add your solution here.
+:ok
 ```
 
-**9. Create a 3x3 tensor with values ranging from 0 to 8 (★☆☆)**
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
+  <summary class="cursor-pointer font-bold mx-2">Example solution</summary>
+  <div class="p-2">
+
+  ```elixir
+  Nx.reverse(tensor_08)
+  ```
+
+  </div>
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
+
+**9. Create a 3x3 tensor with values ranging from 0 to 8. (★☆☆)**
 
 ```elixir
-Nx.iota({3, 3})
+# Add your solution here.
+:ok
 ```
+
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
+  <summary class="cursor-pointer font-bold mx-2">Example solution</summary>
+  <div class="p-2">
+
+  ```elixir
+  Nx.iota({3, 3})
+  ```
+
+  </div>
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
 
 **10a. Given an initial `tensor_10`, build a "mask" of non-zero elements. That is, build a second tensor with the same shape as the original, but that has a 1 wherever the original has a non-zero element and a 0 elsewhere. (★☆☆)**
 
 ```elixir
 tensor_10 = Nx.tensor([1, 2, 0, 0, 4, 0])
-
-mask = Nx.not_equal(tensor_10, 0)
+# Add your solution here.
+:ok
 ```
+
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
+  <summary class="cursor-pointer font-bold mx-2">Example solution</summary>
+  <div class="p-2">
+
+  ```elixir
+  mask = Nx.not_equal(tensor_10, 0)
+  ```
+
+  </div>
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
 
 **10b. Use the mask from 10a to replace each 0 from `tensor_10` with -1. (★☆☆)**
 
 ```elixir
-Nx.select(mask, tensor_10, -1)
+# Add your solution here.
+:ok
 ```
+
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
+  <summary class="cursor-pointer font-bold mx-2">Example solution</summary>
+  <div class="p-2">
+
+  ```elixir
+  Nx.select(mask, tensor_10, -1)
+  ```
+
+  </div>
+</details>
 
 ## 11-20
 
-**11. Create a 3x3 identity tensor (★☆☆)**
+**11. Create a 3x3 identity tensor. (★☆☆)**
 
 ```elixir
-Nx.eye(3)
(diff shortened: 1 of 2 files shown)


nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+Inspired by the Python notebook _100 Numpy Exercises_:
+
+https://www.kaggle.com/code/utsav15/100-numpy-exercises/notebook
+
+## 1-10
+
+1. Install Nx in a Livebook (★☆☆)
+
+```elixir
+Mix.install([{:nx, github: "elixir-nx/nx", sparse: "nx"}])
```
polvalente, line 18: I think we should point to the latest hex version here.

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+```
+
+2. Print the Nx version and the configuration (★☆☆)
+
+```elixir
+version = Application.spec(:nx, :vsn)
+# I'm not sure if there's an equivalent for "configuration"
+```
```
polvalente, line 26: I don't think this is really relevant in Elixir because we always have the package version scoped per project. In Python this is more of a problem due to global packages and virtual envs

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+#    0 |> List.duplicate(10) |> Nx.tensor()
+
+# 2.
+#    [0] |> Nx.tensor() |> Nx.tile([10])
+
+# 3.
+#    i = Nx.iota({10})
+#    Nx.subtract(i, i)
+
+0 |> List.duplicate(10) |> Nx.tensor()
```
polvalente, line 44: Nx.broadcast(0, {10})

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+```
+
+4. How to find the memory size of any tensor (★☆☆)
+
+```elixir
+Nx.byte_size(Nx.tensor([[1, 2, 3], [4, 5, 6]]))
```
polvalente, line 50: Should also mention `flat_size` for vectorized tensors

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+
+```elixir
+# I don't know if this is possible.
+# iex> h Nx.add
+```
+
+6. Create a tensor of zeros of size 10 but the fifth value which is 1 (★☆☆)
+
+```elixir
+zeros = 0 |> List.duplicate(10) |> Nx.tensor()
```
polvalente, line 63: broadcast here

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+7. Create a tensor with values ranging from 10 to 49 (★☆☆)
+
+```elixir
+# 1.
+#    Nx.iota({40}) |> Nx.add(10)
+
+# 2.
+#    Nx.linspace(10, 49, n: 39, type: :s8)
+
+Nx.linspace(10, 49, n: 39, type: :s8)
```
polvalente, line 77: hah funnily enough I thought o iota + add first. This is probably better.

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+```elixir
+Nx.iota({3, 3})
+```
+
+10. Find indices of non-zero elements from [1,2,0,0,4,0] (★☆☆)
+
+```elixir
+Nx.tensor([1, 2, 0, 0, 4, 0])
+|> Nx.not_equal(0)
+```
```
polvalente, line 98: This isn't exactly 1:1 with the Python alternative. Maybe reword the exercise title a bit

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+# This felt a little awkward. Is there a better way?
+
+key = Nx.Random.key(0)
+{r, _} = Nx.Random.normal(key, shape: {3, 3})
+r |> IO.inspect()
+r_1d = Nx.reshape(r, {:auto})
+
+%{
+  min: Nx.to_number(r_1d[Nx.argmin(r)]),
+  max: Nx.to_number(r_1d[Nx.argmax(r)])
```
polvalente, line 128:
```suggestion
  min: Nx.reduce_min(r),
  max: Nx.reduce_max(r)
```

I don't think we should encourage Nx.to_number in this guide

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+}
+```
+
+14. Create a random 1D tensor of size 30 and find the mean value (★☆☆)
+
+```elixir
+key = Nx.Random.key(0)
+{r, _} = Nx.Random.normal(key, shape: {30})
+r |> Nx.mean() |> Nx.to_number()
```
polvalente, line 137:
```suggestion
Nx.mean(r)
```

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+```
+
+15. Create a 5x5 tensor with 1 on the border and 0 inside (★☆☆)
```
polvalente, line 140:
```suggestion
15. Create a 5x5 tensor with 1s on the border and 0s inside (★☆☆)
```

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+
+```elixir
+zeros_3x3 = Nx.make_diagonal(Nx.tensor([0, 0, 0]))
```
polvalente, line 143: I think broadcast here would be more intuitive as it's the solution for the initial exercises

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+Nx.pad(zeros_3x3, 1, [{1, 1, 0}, {1, 1, 0}])
+```
+
+16. How to add a border (filled with 0's) around an existing array? (★☆☆)
+
+```elixir
+ones_3x3 = 1 |> List.duplicate(9) |> Nx.tensor() |> Nx.reshape({3, 3})
```
polvalente, line 150: Nx.broadcast(1, {3, 3})

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+
+17. What are the results of the following expressions? (★☆☆)
+
+```elixir
+nan = Nx.Constants.nan()
+IO.inspect(Nx.multiply(0, nan))
+IO.inspect(Nx.equal(nan, nan))
+IO.inspect(Nx.greater(nan, nan))
+IO.inspect(Nx.subtract(nan, nan))
+IO.inspect(Enum.member?(MapSet.new([nan]), nan))
```
polvalente, line 162: This check isn't a good one. It would probably fail if you used EXLA or Torchx and had 2 different instances of nan.

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+IO.inspect(0.3 == 3 * 0.1)
```
polvalente, line 163: Not sure what this line is doing here

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+```elixir
+Nx.make_diagonal(Nx.tensor([1, 2, 3, 4]), offset: -1)
+```
+
+19. Create a 8x8 tensor of 0s and 1s in a checkerboard pattern with 0 as the first element (★☆☆)
+
+```elixir
+Nx.tensor([0, 0, 0, 0])
+|> Nx.make_diagonal()
+|> Nx.pad(1, [{0, 0, 1}, {0, 0, 1}])
```
polvalente, line 177:
```suggestion
Nx.iota({8, 8}) |> Nx.remainder(2)
```

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,196 @@
...
+target = 100
+
+# This is how I'd do it with Enum.
+dims
+|> Enum.map_reduce({target, Enum.product(dims)}, fn dim, {remaining, step_size} ->
+  step_size = div(step_size, dim)
+  num_steps = div(remaining, step_size)
+  {num_steps, {remaining - num_steps * step_size, step_size}}
+end)
+|> then(fn {indices, {0, 1}} -> indices end)
```
polvalente, line 195: This is an ambiguous question due to row-major vs col-major. Maybe it could be better worded.

In Nx you can do something like this as well:

```elixir
dims = {6, 7, 8}
Enum.map(0..(tuple_size(dims) - 1), fn axis -> (Nx.iota(dims, axis: axis) |> Nx.flatten())[99] end) |> Nx.stack()
```

Although this is probably not that efficient and it actually might be better to wrap your Enum code in a deftransform

polvalente (approved):
Awesome work! I left some comments, but I'm really happy about the direction this guide is taking

polvalente:
I accidentally approved instead of just leaving a general comment haha! This is overall great, but I think we should iterate and converge on the solutions for a bit more before merging

nx/exercises/exercises-1-20.livemd
billylanchantin: That was my first choice, but I got worried about it getting out of date.

Then again, maybe that's a good thing? If exercises are answered under version (say) 0.6, they might no longer be the best practice in later versions.

nx/exercises/exercises-1-20.livemd
billylanchantin: Agreed. I honestly don't even think printing the version is all that useful since you have to opt into a version at the top of the notebook. (Since, ya know, we have sane package management 😉)

This one is a good example of an exercise we may not want to include for whatever reason. Feel free to go through and mark or any others as "not a good candidate" or similar.

nx/exercises/exercises-1-20.livemd
billylanchantin: 🤯 I don't know how I hadn't come across this!

nx/exercises/exercises-1-20.livemd
billylanchantin: Oh good catch. I should do a gather to make it more 1:1.

nx/exercises/exercises-1-20.livemd
billylanchantin: Yeah, the need for the type specification makes it less obvious though.

nx/exercises/exercises-1-20.livemd
billylanchantin: Oh that's better. Though it was still instructive for me to learn about `pad`.

nx/exercises/exercises-1-20.livemd
billylanchantin: Yeah this one took me by far the longest. Is row- vs. col-major determined by the backend? Some config?

Regardless, yeah I think the amount of memory `iota` potentially allocates makes that a bit too much. You only ever need a list with a length equal to the rank of the tensor.

Another question: what do you think about using `defn` and `deftransform`? It's a bit awkward in livebook form since you need to wrap your answer in a `defmodule`.

nx/exercises/exercises-1-20.livemd
billylanchantin: 🤷 Happy to take it out!

nx/exercises/exercises-1-20.livemd
billylanchantin: > I don't think we should encourage Nx.to_number in this guide.

How come? (Not objecting, just curious!)

> Nx.reduce_min

Does this work with rank > 1? I tried but it only aggregated across a single axis.

nx/exercises/exercises-1-20.livemd
billylanchantin: Er nevermind, that doesn't work. It would work with an odd x odd tensor, but not an even x even.

nx/exercises/exercises-1-20.livemd
billylanchantin: Addressed: [e93dbd9](https://github.com/elixir-nx/nx/pull/1363/commits/e93dbd99643fdac8d2d5a31bded8e9aa1950d248)

nx/exercises/exercises-1-20.livemd
billylanchantin: Addressed: [e93dbd9](https://github.com/elixir-nx/nx/pull/1363/commits/e93dbd99643fdac8d2d5a31bded8e9aa1950d248)

nx/exercises/exercises-1-20.livemd
billylanchantin: Addressed: [dba7470](https://github.com/elixir-nx/nx/pull/1363/commits/dba7470db6146e28011ab26e9f423bc723f43b0a)

nx/exercises/exercises-1-20.livemd
billylanchantin: Addressed: [dba7470](https://github.com/elixir-nx/nx/pull/1363/commits/dba7470db6146e28011ab26e9f423bc723f43b0a)

nx/exercises/exercises-1-20.livemd
billylanchantin: Addressed: [dba7470](https://github.com/elixir-nx/nx/pull/1363/commits/dba7470db6146e28011ab26e9f423bc723f43b0a)

nx/exercises/exercises-1-20.livemd
billylanchantin: Addressed: [dba7470](https://github.com/elixir-nx/nx/pull/1363/commits/dba7470db6146e28011ab26e9f423bc723f43b0a)

nx/exercises/exercises-1-20.livemd
billylanchantin: Addressed: [657a3d9](https://github.com/elixir-nx/nx/pull/1363/commits/657a3d9bd66b93f6fbec678c3e0d4502716a2e3e)

nx/exercises/exercises-1-20.livemd
billylanchantin: New approach: [23d5f77](https://github.com/elixir-nx/nx/pull/1363/commits/23d5f77bc5c5282d856f593d500d0bda0051d70f)

Ironically, even though I "learned about `pad`", my solution was _also_ incorrect.

nx/exercises/exercises-1-20.livemd
billylanchantin: Addressed: [bb6771e](https://github.com/elixir-nx/nx/pull/1363/commits/bb6771e34bfe55e012867994dcfe34982b6bef80)

nx/exercises/exercises-1-20.livemd
billylanchantin: Addressed: [bb6771e](https://github.com/elixir-nx/nx/pull/1363/commits/bb6771e34bfe55e012867994dcfe34982b6bef80)

nx/exercises/exercises-1-20.livemd
billylanchantin: Addressed: [55412e6](https://github.com/elixir-nx/nx/pull/1363/commits/55412e644da53b4ef451a19d6ff563158a95aa0c)

nx/exercises/exercises-1-20.livemd
billylanchantin: We may want consecutive exercises for `size`, `flat_size`, and `byte_size`. It'd highlight how vectorization changes the answers.

billylanchantin:
@polvalente Ok, first pass at edits (thanks for reviewing!). A few thoughts:

1. 20 exercises at a time may be too much to iterate on. Perhaps the next PR can just be 10.
2. I resolved the comments that were rote. Usually I'd leave that up to the reviewer, but there are quite a few and I wanted to simplify the next round of edits.
3. The auto-formatter kept clobbering the numbers. Every time I'd make a change, it'd reset all the numbers to `1.`. I've stopped fighting it for now. Hopefully we can think of a solution before the merge.

nx/exercises/exercises-1-20.livemd
billylanchantin: Actually... how _do_ you do this? I keep wanting to do a numpy masked array, but there isn't an equivalent in Nx that I saw.

nx/exercises/exercises-1-20.livemd
polvalente: Yes, I think we should not include this one

nx/exercises/exercises-1-20.livemd
polvalente: It's better to use the Hex version. If it gets out of date, it's more explicit through the version

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,172 @@
...
+
+```elixir
+Nx.eye(3)
+```
+
+1. Create a 3x3x3 tensor with random values (★☆☆)
+
+```elixir
+key = Nx.Random.key(0)
+{random, _} = Nx.Random.normal(key, shape: {3, 3})
```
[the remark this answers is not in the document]
billylanchantin: 🤦

nx/exercises/exercises-1-20.livemd
billylanchantin: [f9c005c](https://github.com/elixir-nx/nx/pull/1363/commits/f9c005cd5be16578574a3df51d0e2a348f379beb)

nx/exercises/exercises-1-20.livemd
billylanchantin: [f9c005c](https://github.com/elixir-nx/nx/pull/1363/commits/f9c005cd5be16578574a3df51d0e2a348f379beb)

nx/exercises/exercises-1-20.livemd
polvalente: Yeah, I think it's a good idea!

nx/exercises/exercises-1-20.livemd
polvalente: Nx is always row-major. I was talking about the working itself.

I think you could introduce `deftransform` here just as an example on how to use this Elixir-computed values inside a `defn`

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,174 @@
...
+```elixir
+Nx.tensor([[1, 2, 3], [4, 5, 6]])
+|> Nx.byte_size()
+```
+
+**5. How to get the documentation of the `add/2` function from the command line? (★☆☆)**
+
+```elixir
+# I don't know if this is possible.
+# iex> h Nx.add
```
polvalente, line 45: This should be possible. In livebook you can use `import IEx.Helpers` to get access to the `h` helper

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,174 @@
...
+```elixir
+key = Nx.Random.key(0)
+{random, _} = Nx.Random.normal(key, shape: {3, 3, 3})
+random
+```
+
+**13. Create a 10x10 tensor with random values and find the minimum and maximum values (★☆☆)**
+
+```elixir
+# This felt a little awkward. Is there a better way?
```
polvalente, line 101: Do you mean that using random felt awkward?

Unfortunately there's not a better way. The interface is as such because `Nx.Random` provides a stateless PRNG framework, which lends itself to being run alongside calculations on GPUs.

billylanchantin: Oh whoops, no this was a holdover from when I wasn't yet using `reduce_max`. I'll drop this comment.

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,174 @@
...
+```elixir
+Nx.iota({3, 3})
+```
+
+**10. Find indices of non-zero elements from [1,2,0,0,4,0] (★☆☆)**
+
+```elixir
+Nx.tensor([1, 2, 0, 0, 4, 0])
+|> Nx.not_equal(0)
+```
```
nickgnd, line 80: Hey 👋 
I was just checking the PR (great work!)

I think this solution does not respect the exercise description because it returns this tensor `[1, 1, 0, 0, 1, 0]` instead of the indices of non-zero elements `[0, 1, 4]`.

This should work, although there might be better way to achieve that 🤔 

```suggestion
```elixir
t = Nx.tensor([1, 2, 0, 0, 4, 0])
to_keep = Nx.not_equal(t, 0)
indices = Nx.iota(Nx.shape(t))

Nx.select(to_keep, indices, -1)
|> Nx.to_list()
|> Enum.reject(& &1 == -1)
```
```

Cheers ✌️

File: nx/exercises/exercises-1-20.livemd [hunk 21]
```diff
@@ -0,0 +1,174 @@
...
+
+```elixir
+Nx.iota({3, 3})
+```
+
+**10. Find indices of non-zero elements from [1,2,0,0,4,0] (★☆☆)**
+
+```elixir
+Nx.tensor([1, 2, 0, 0, 4, 0])
+|> Nx.not_equal(0)
```
nickgnd, line 79: Hey 👋 
I was just checking the PR (great work!)

I think this solution does not respect the exercise description because it returns this tensor `[1, 1, 0, 0, 1, 0]` instead of the indices of non-zero elements `[0, 1, 4]`.

This should work, although there might be better way to achieve that 🤔 

```suggestion
t = Nx.tensor([1, 2, 0, 0, 4, 0])
to_keep = Nx.not_equal(t, 0)
indices = Nx.iota(Nx.shape(t))

Nx.select(to_keep, indices, -1)
|> Nx.to_list()
|> Enum.reject(& &1 == -1)
```

Cheers ✌️

polvalente: I think the exercise should be reworded instead. Getting the indices isn't semantic in Nx because we don't have dynamic shapes. So getting the indices won't be as useful as getting a mask and applying it subsequently with Nx.select. 

The indices by themselves won't be applicable inside defn, because you have to "exit" the defn world to then feed them back in.

billylanchantin: Thanks for the suggestion! Yes I got stuck here too: https://github.com/elixir-nx/nx/pull/1363#discussion_r1390256518

I think `select` is a good option, but I was really hoping to avoid dropping into a list. Otherwise you could just do

```elixir
list
|> Enum.with_index()
|> Enum.filter(fn {x, _} -> x != 0 end)
|> Enum.map(fn {_, i} -> i end)
```

or similar.

billylanchantin: @polvalente I agree. Without dynamic shapes, I just don't think there's a way to do this exercise.

Want me to reword it to have the user return a mask?

nx/exercises/exercises-1-20.livemd
billylanchantin: Oh and FWIW, I actually find the way `Nx.Random` works to make a lot of sense. It's got a very functional programming feel.

nx/exercises/exercises-1-20.livemd
polvalente: Yeah, I think it would be better. Perhaps follow it up with an exercise "b" item that tells them to update the zero positions with some value given the mask

nx/exercises/exercises-1-20.livemd
billylanchantin: [4be316e](https://github.com/elixir-nx/nx/pull/1363/commits/4be316ed841dec5479a7830fa57037ca9f6fe894)

nx/exercises/exercises-1-20.livemd
billylanchantin: [cf7a994](https://github.com/elixir-nx/nx/pull/1363/commits/cf7a9946665c34914131aad458278b30459fddef)

nx/exercises/exercises-1-20.livemd
billylanchantin: Moved here: https://github.com/elixir-nx/nx/pull/1363#discussion_r1390431325

billylanchantin:
Hey so I apologize for the churn here. I realize it makes this PR harder to review.

After some more discussion on slack, I think I'd like to reformat so solutions are wrapped in a `<details>` tag. That way you're not spoiled while you read.

I have these changes locally. If there are no objections, I'd like to push them:

nx-livebook-solution

polvalente:
> Hey so I apologize for the churn here. I realize it makes this PR harder to review.
> 
> After some more discussion on slack, I think I'd like to reformat so solutions are wrapped in a `<details>` tag. That way you're not spoiled while you read.
> 
> I have these changes locally. If there are no objections, I'd like to push them:
> 
> nx-livebook-solution

Go ahead! This PR isn't _that_ large that a refactor like this would make it impossible to review again

billylanchantin:
Ok, highlights:

* Each solution is inside a `<details>` tag. Now you won't get spoiled.
* All exercise wordings use the imperative mood. It felt better than some exercises being questions and others commands.
* I use the phrase "Example solution" for my solutions since there may be more than one. It also leaves it open for us to add "Example solution 1", "Example solution 2", etc. when there are multiple valid approaches.

nx/exercises/exercises-1-20.livemd
billylanchantin: [1475f49](https://github.com/elixir-nx/nx/pull/1363/commits/1475f4910701a8f2ed00dd5f07e67f9417909772)

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,490 @@
...
+        step_size = div(step_size, dim)
+        num_steps = div(remaining, step_size)
+        {num_steps, {remaining - num_steps * step_size, step_size}}
+      end)
+      |> then(fn {indices, {0, 1}} -> indices end)
+    end
+  end
+
+  Exercise20.solve({6, 7, 8}, 100)
+  ```
```
jonatanklosko, line 487:
````suggestion
  ```

  > **Note:** functions defined with `deftransform` can be invoked inside `defn` and they are oftentimes used to compute tensor shapes. Inside `deftransform` you can call any other modules, such as `Enum` and `Tuple`, which is not the possible inside `defn`.
````

File: nx/exercises/exercises-1-20.livemd [hunk 23]
```diff
@@ -0,0 +1,490 @@
...
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
+
+**2. Create a 1-D tensor of 10 zeros. (★☆☆)**
+
+```elixir
+# Add your solution here.
+:ok
+```
```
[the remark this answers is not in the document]
billylanchantin: I went back and forth on this one. ~The `:ok` is a bit unecessary, but it made the outputs all the same, even for those that defined a tensor in the instructions.~

Nevermind. That was the justification, but now I think I _want_ different outputs. If an exercise defines a tensor as part of the instructions, I think it'd be best for the reader to see it.

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,490 @@
...
+
+  </div>
+</details>
+
+<!-- livebook:{"break_markdown":true} -->
+
+**3. Find the number of elements in `tensor_03`. (★☆☆)**
+
+```elixir
+tensor_03 = Nx.tensor([[1, 2, 3], [4, 5, 6]])
```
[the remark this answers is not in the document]
billylanchantin: I went back and forth on this one too. TBH it was less about isolation and more about giving everything a distinct name.

However, if the exercises ever get reordered, it'd be super annoying. I think I'm leaning toward just naming everything `tensor` now.

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,490 @@
...
+
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
+  <summary class="cursor-pointer font-bold mx-2">Example solution</summary>
+  <div class="p-2">
+
+  ```elixir
+  import IEx.Helpers
+  h Nx.add
+  ```
+
```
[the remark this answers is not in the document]
billylanchantin: Yeah this one was another one that got lost in translation a bit. If the intention is to get the user to hover over a function, any thoughts on wording? There isn't any code to write in that case.

(We can always replace this with a more Nx-specific exercise, too.)

nx/exercises/exercises-1-20.livemd
billylanchantin: I think I want to replace this exercise with something simpler. These early exercises are mostly about getting the reader to lookup the relevant function. In Numpy, you simply call `unravel`. Adhering to the original isn't as important as keeping the difficulty consistent IMO.

WDYT? \cc @polvalente

nx/exercises/exercises-1-20.livemd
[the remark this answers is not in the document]
billylanchantin: [9ecf23d](https://github.com/elixir-nx/nx/pull/1363/commits/9ecf23d533a712e8915385f29f73c37a5a797eba)

nx/exercises/exercises-1-20.livemd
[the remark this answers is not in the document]
billylanchantin: [aa93d54](https://github.com/elixir-nx/nx/pull/1363/commits/aa93d548e07cab5daea6e7a76a1e4e45ad4f84b8)

nx/exercises/exercises-1-20.livemd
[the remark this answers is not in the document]
billylanchantin: [57e8be4](https://github.com/elixir-nx/nx/pull/1363/commits/57e8be4eff3f1614ff4ddce198fc3293b7cf99c5)

Screen Shot 2023-11-14 at 10 26 41 AM

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,496 @@
...
+<!-- livebook:{"break_markdown":true} -->
+
+**2. Create a 1-D tensor of 10 zeros. (★☆☆)**
+
+```elixir
+# Add your solution here.
+```
+
+<details class="rounded-lg border-2 border-green-bright-300 p-2">
```
jonatanklosko, line 38: Ah btw, ideally we should not rely on tailwind classes. Tailwind prunes classes at build time, so whether some of them are available depends on whether they are used anywhere in Livebook. This is not critical, but even more relevant concern is that we want to render this using ExDoc and it's not going to translate.

We don't have to be fancy with it, personally I actually prefer this simpler version:

````
<details>
  <summary style="cursor: pointer; font-weight: bold;">Example solution</summary>
  <div style="padding-top: 8px">

  ```elixir
  ...
  ```

  </div>
</details>
````

It doesn't bring too much attention. Currently the green border makes it the first thing my eyes focus on, and we shouldn't disrupt the reading flow :)

Comment (billylanchantin) [reply nx/exercises/exercises-1-20.livemd:38]: > This is not critical, but even more relevant concern is that we want to render this using ExDoc and it's not going to translate.

Oh good point. I was actually choosing Tailwind classes based on what happened to be available, which felt brittle.

> Currently the green border makes it the first thing my eyes focus on, and we shouldn't disrupt the reading flow :)

Ha that's understandable. I even tried to tone it _down_ from the original:

https://the-eef.slack.com/archives/C01PU0RDJ8L/p1699750471235479?thread_ts=1699720134.906519&cid=C01PU0RDJ8L

File: nx/exercises/exercises-1-20.livemd (hunk 22 above)
Comment (billylanchantin) [reply nx/exercises/exercises-1-20.livemd:493]: Exercise 21 in the kaggle notebook is:

> 21. Create a checkerboard 8x8 matrix using the tile function (★☆☆)

Since that was actually my solution to 19, how about for 20:

<blockquote>

20. Produce the same checkerboard pattern from exercise 19, but use [Nx.window_reduce](https://hexdocs.pm/nx/Nx.html#window_reduce/5) instead. (★☆☆)

<details>
<summary>Example solution</summary>

```elixir
Nx.window_reduce(Nx.iota({9, 9}), 0, {2, 2}, [], fn x, _ -> Nx.remainder(x, 2) end)
```

</details>
</blockquote>

(This may be a 2-star since `window_reduce` is a little complicated and it's not obvious how to use it here.)

polvalente: Just use remainder on the iota as I had initially suggested. We shouldn't incentivize the usage of window_reduce because it's far less efficient than alternatives when available

billylanchantin: I mentioned before that doesn't work because the checkerboard is even by even:

```elixir
Nx.iota({8, 8}) |> Nx.remainder(2)
#Nx.Tensor<
  s64[8][8]
  [
    [0, 1, 0, 1, 0, 1, 0, 1],
    [0, 1, 0, 1, 0, 1, 0, 1],
    [0, 1, 0, 1, 0, 1, 0, 1],
    [0, 1, 0, 1, 0, 1, 0, 1],
    [0, 1, 0, 1, 0, 1, 0, 1],
    [0, 1, 0, 1, 0, 1, 0, 1],
    [0, 1, 0, 1, 0, 1, 0, 1],
    [0, 1, 0, 1, 0, 1, 0, 1]
  ]
>
```

You could do a 9x9 and slice I suppose. Is that what you mean?

polvalente: Ah, makes sense. You could do that slicing , or go with a more mathematical approach that relies on broadcasting. Although maybe you could just use tile and skip the other exercise. PS: I haven't checked the solution below, take it with a grain of salt. 

t = Nx.iota({8, 1}) |> Nx.remainder(2)

Nx.transpose(t) |> Nx.add(t) |> Nx.remainder(2)

billylanchantin: Your solution works! :)

So here's what I'm thinking now: we just say "Make a checkerboard _without_ `Nx.tile`." Should get the creative juices flowing and force you to read the docs for ideas.

polvalente: How about this: suggest Nx.tile here, and the version without Nx.tile later on?

billylanchantin: Hm I think we're having a two slightly different conversations.

I'm talking about replacing exercise 20 since Nx doesn't have an `unravel` equivalent. I was suggesting we replace 20 with a twist on 19 since 19 is the one where we ask for a checkerboard with `Nx.tile`.

Sorry if I'm misunderstanding! It may be too late to be on Github.

polvalente: Yeah, to be clear:

I think the first checkerboard exercise should be with tile, and then the next should be "ok, now do it without tile", maybe hinting on broadcasting or something like that!

billylanchantin: [58eb8b1](https://github.com/elixir-nx/nx/pull/1363/commits/58eb8b1e431653316828a863ddfcdc978e31d5d1)

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,500 @@
...
+
+  <div style="padding-top: 16px">
+    <div style="background-color: rgb(62 100 255); color: white; font-weight: bold; border-top-left-radius: 8px; border-top-right-radius: 8px; padding: 8px 16px;">
+      <span style="padding-right: 4px"><i class="ri-lightbulb-line"></i></span>
+      <span>Tip</span>
+    </div>
+    <div style="border-width: 1px; border-color: rgb(62 100 255); border-bottom-left-radius: 8px; border-bottom-right-radius: 8px; padding: 0px 16px;">
+      <p>In Livebook, you can also hover over a function to display its documentation.</p>
+    </div>
+  </div>
```
josevalim, line 133: I would like to go easy on the markup, because that makes contributing to Livebook harder:

```suggestion
  <div style="padding-top: 16px">
    <i>Tip:</i> You can also hover over a function inside Livebook code cells to display its documentation.
  </div>
```

nx/exercises/exercises-1-20.livemd
```diff
@@ -0,0 +1,500 @@
...
+```elixir
+# Add your solution here.
+```
+
+<details style="margin-left: 32px">
+  <summary style="cursor: pointer; font-weight: bold;">Example solution</summary>
+  <div style="padding-top: 8px">
+
+  ```elixir
+  Nx.linspace(10, 49, n: 39, type: :s8)
```
josevalim, line 172: Would a solution with iota feel more natural? `Nx.iota({39}) |> Nx.add(10)`?

josevalim: I think in Numpy they would use arange or similar, so that solution feels more natural. Perhaps we replace the order of this question with 9.

josevalim: To be clear, make 8 become number 7, number 9 become number 8, and number 7 becomes new number 9 using iota as the answer.

josevalim (commented):
Sorry for the delay, I am (still) on holidays. I have dropped just two minor nits and we can ship it!

nx/exercises/exercises-1-20.livemd
billylanchantin: Update: we've dropped this exercise in favor of another checkerboard one.

nx/exercises/exercises-1-20.livemd
billylanchantin: Works for me! It honestly felt like maybe a bit much, but I figured I'd start with a 1:1 translation of the tailwind and see how we felt.

nx/exercises/exercises-1-20.livemd
billylanchantin: [e667514](https://github.com/elixir-nx/nx/pull/1363/commits/e667514a8e1515dbd73cce6dcbde958dfe171fc2)

After reshuffling, I added the iota solution as solution 1 instead of replacing the linspace one (which is now solution 2). The iota one is definitely the more natural solution. But since it requires that extra step, I figured having both approaches was good.

josevalim: Having both is a great call!

josevalim:

PR #1376, opened 2023-11-27
Add `Nx.zeros/2` helper initialisation function

pnezis:
Similar to `np.zeros`, this is pretty common.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -985,6 +985,59 @@ defmodule Nx do
   end
 
   @doc """
+  Creates a tensor with the given shape, filled with zeros.
+
+  ## Options
+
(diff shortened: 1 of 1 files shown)


polvalente (commented):
I don't think we should add this one.
Firstly, broadcast is already there (albeit we could maybe add the type option to it as well)

Secondly, it's much more rare for us to need a zeros tensor in Nx than in Numpy. In Numpy there's this "norm" of preallocation of an output tensor followed by in-place updates, which we don't have in Nx, so you're almost always better off avoiding said preallocation in favor of rewriting your code to not need it.

polvalente:
Ah, I forgot to say thanks for the contribution 😅
Feel free to submit PRs or issues whenever you feel like there's some crucial functiomality missing!

PR #1390, opened 2023-12-05
Add lazy select

msluszniak:
Actually, the lazy select procedure is not as fast as we want to :/

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -9391,6 +9391,74 @@ defmodule Nx do
     end
   end
 
+  def lazy_select(tensor, opts \\ []) do
+    k = opts[:k]
+
+    if rank(tensor) != 1 do
+      raise ArgumentError, "expected a vector, got a #{inspect(tensor)}"
+    end
+
+    if k >= size(tensor) do
+      raise ArgumentError, "k must be less than the size of the tensor"
+    end
+
+    len = size(tensor)
+    selection_len = :math.pow(len, 3 / 4) |> trunc()
+    opts = Keyword.put(opts, :selection_len, selection_len)
+    lazy_select_core(tensor, opts)
+  end
+
+  def lazy_select_core(tensor, opts \\ []) do
+    k = opts[:k]
+    selection_len = opts[:selection_len]
+
+    type = Nx.type(tensor)
+    max_val = Nx.Constants.max(type)
+
+    key = Nx.Random.key(System.system_time())
+    len = size(tensor)
+
+    Enum.reduce_while(1..1000, key, fn _, key ->
+      {candidates, new_key} = Nx.Random.choice(key, tensor, samples: selection_len)
+      candidates = sort(candidates)
+      x = as_type(multiply(divide(k, len), pow(len, divide(3, 4))), :s64)
+
+      low_index = as_type(Nx.max(0, subtract(x, sqrt(len))), :s64)
+      high_index = as_type(Nx.min(selection_len - 1, add(x, sqrt(len))), :s64)
+
+      low = take(candidates, low_index)
+      high = take(candidates, high_index)
+
+      low_pos = sum(greater(low, tensor))
+      high_pos = sum(greater(high, tensor))
+
+      indices_to_append = logical_and(greater_equal(tensor, low), less_equal(tensor, high))
+
+      {agg_index, agg} =
+        Enum.reduce(0..(len - 1), {0, broadcast(max_val, {len})}, fn index,
+                                                                     {curr_agg_index, curr_agg} ->
+          cond do
+            indices_to_append[index] == Nx.tensor(1, type: :u8) ->
+              {curr_agg_index + 1,
+               indexed_put(curr_agg, new_axis(curr_agg_index, -1), tensor[index])}
+
+            true ->
+              {curr_agg_index, curr_agg}
+          end
+        end)
+
+      if logical_and(
+           logical_and(less_equal(low_pos, k), less_equal(k, high_pos)),
+           less_equal(agg_index, 4 * selection_len + 1)
+         ) == Nx.tensor(1, type: :u8) do
+        agg = sort(agg)
+        {:halt, agg[subtract(k, low_pos)]}
+      else
+        {:cont, new_key}
+      end
+    end)
+  end
+
   @doc """
   Returns the mode of a tensor.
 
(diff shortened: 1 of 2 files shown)


josevalim:
Should we make it a defn? Would that improve the performance?

nx/lib/nx.ex
```diff
@@ -9391,6 +9391,74 @@ defmodule Nx do
...
+    type = Nx.type(tensor)
+    max_val = Nx.Constants.max(type)
+
+    key = Nx.Random.key(System.system_time())
+    len = size(tensor)
+
+    Enum.reduce_while(1..1000, key, fn _, key ->
+      {candidates, new_key} = Nx.Random.choice(key, tensor, samples: selection_len)
+      candidates = sort(candidates)
+      x = as_type(multiply(divide(k, len), pow(len, divide(3, 4))), :s64)
```
polvalente, line 9424: k/len * len ** 3/4 is the same as k * len ** (-1/4). Given that len is fixed, you could extract len ** (-1/4) from the loop and do: `x = as_type(Nx.multiply(k, len_prime), :s64)`

nx/lib/nx.ex
```diff
@@ -9391,6 +9391,74 @@ defmodule Nx do
...
+
+      low_index = as_type(Nx.max(0, subtract(x, sqrt(len))), :s64)
+      high_index = as_type(Nx.min(selection_len - 1, add(x, sqrt(len))), :s64)
```
polvalente, line 9427: sqrt(len) could be extracted to a variable

nx/lib/nx.ex
```diff
@@ -9391,6 +9391,74 @@ defmodule Nx do
...
+
+      indices_to_append = logical_and(greater_equal(tensor, low), less_equal(tensor, high))
+
+      {agg_index, agg} =
+        Enum.reduce(0..(len - 1), {0, broadcast(max_val, {len})}, fn index,
+                                                                     {curr_agg_index, curr_agg} ->
+          cond do
+            indices_to_append[index] == Nx.tensor(1, type: :u8) ->
+              {curr_agg_index + 1,
+               indexed_put(curr_agg, new_axis(curr_agg_index, -1), tensor[index])}
```
polvalente, line 9443: I believe you can get a way better performance by accumulating all values of interest first and doing a single indexed_put in the end

nx/lib/nx.ex
```diff
@@ -9391,6 +9391,74 @@ defmodule Nx do
...
+
+            true ->
+              {curr_agg_index, curr_agg}
+          end
+        end)
+
+      if logical_and(
+           logical_and(less_equal(low_pos, k), less_equal(k, high_pos)),
```
polvalente, line 9451: are you sure it should be less_equal here? This looks a lot like it should be logical_not(indices_to_append)

nx/lib/nx.ex
polvalente: I think we can do a nice trick for this.

```elixir
# We want to take all values that obey a given condition (in the example, x % 3 == 0)
# and store them at the beggining of a given tensor.
iex(2)> {values, _} = Nx.Random.randint(Nx.Random.key(42), 0, 30, shape: {10})
{#Nx.Tensor<
   s64[10]
   [3, 3, 11, 12, 4, 20, 17, 12, 5, 24]
 >,
 #Nx.Tensor<
   u32[2]
   [2465931498, 3679230171]
 >}
iex(3)> pred = Nx.remainder(values, 3) |> Nx.equal(0)#Nx.Tensor<
  u8[10]
  [1, 1, 0, 1, 0, 0, 0, 1, 0, 1]
>
# With cumulative_sum and select, we can build an indices tensor that will sequentially store
# all values of interest to the beginning, and throw other values in a "garbage collection"
# extra index (which will be the first index of the output tensor, to be sliced out at the end)
iex(4)> Nx.select(pred, Nx.cumulative_sum(pred), 0)                
#Nx.Tensor<
  s64[10]
  [1, 2, 0, 3, 0, 0, 0, 4, 0, 5]
>
iex(6)> target_indices = Nx.new_axis(Nx.select(pred, Nx.cumulative_sum(pred), 0), 1)
#Nx.Tensor<
  s64[10][1]
  [
    [1],
    [2],
    [0],
    [3],
    [0],
    [0],
    [0],
    [4],
    [0],
    [5]
  ]
>
iex(7)> destination = Nx.concatenate([Nx.tensor([-1]), values])
#Nx.Tensor<
  s64[11]
  [-1, 3, 3, 11, 12, 4, 20, 17, 12, 5, 24]
>
iex(8)> result = Nx.indexed_put(destination, target_indices, values) |> Nx.slice_along_axis(1, Nx.axis_size(destination, 0) - 1, axis: 0)
#Nx.Tensor<
  s64[10]
  [3, 3, 12, 12, 24, 20, 17, 12, 5, 24]
>
iex(9) agg_index = Nx.sum(pred)
#Nx.Tensor<
  u64
  5
>
# which means that only the first 5 items of result are values of interest.
```

polvalente: @josevalim I wonder if we could actually define a type of dynamic tensor where it's just a special container that Nx knows that carries it's effective index, but is initialized to a maximum size.
Then some operations could maybe optimize that out while still carrying through the effective index.
Concatenate would have to be done as a sort of put_slice for instance.

nx/lib/nx.ex
```diff
@@ -9391,6 +9391,74 @@ defmodule Nx do
...
+           less_equal(agg_index, 4 * selection_len + 1)
+         ) == Nx.tensor(1, type: :u8) do
+        agg = sort(agg)
```
jonatanklosko, line 9454: `agg` has the same shape as `tensor` (because even if we care about a subset of elements, we still use the full shape here). So sorting here has the same complexity as sorting `tensor`. Technically it depends on the sorting procedure, since it could be falling into some optimistic case, but that's probably not something we should rely on.

Given that we have the constraint `less_equal(agg_index, 4 * selection_len + 1)`, we could at least do `sort(agg[0..(4 * selection_len)])`. For large `n` this should make a fair difference, though we still fall into the worst case in a sense.

nx/lib/nx.ex
msluszniak: Yeah, but to implement this trick we need dynamic shape

nx/lib/nx.ex
msluszniak: Indeed, there should be greater_equal, but I think it should be logical_and since we want all there conditions to be met

nx/lib/nx.ex
josevalim: Instead of a dynamic shape, we can also consider allocating a tensor with the maximum size that we could potentially use, and then keep another value that tells how much it is being used, since the goal is to stay within `defn` as much as possible. :)

PR #1413, opened 2023-12-07
chore: unify mlir tests

polvalente:


polvalente:
Note: we're not running CI on the MLIR mode yet because if we use it as a default system-wide we still get broken tests. I'll open a separate PR tagginn everything that needs to be skipped in MLIR so that we can start running a CI job for it.

PR #1418, opened 2023-12-08
Add executable serialization

seanmor5:
To make cold starts faster we can save JIT compiled functions to disk. This should help FLAME cold starts.

Not sure what the API should look like for Nx as this may not work across backends/compilers. Maybe `Nx.Defn.load/save` ?

### DIFF
--- a/exla/c_src/exla/exla.cc
+++ b/exla/c_src/exla/exla.cc
@@ -608,6 +608,24 @@ ERL_NIF_TERM start_log_sink(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
     return exla::nif::error(env, "Bad argument count.");
   }
 
+  exla::ExlaExecutable** executable;
+
+  if (!exla::nif::get<exla::ExlaExecutable*>(env, argv[0], executable)) {
+    return exla::nif::error(env, "Unable to get executable.");
+  }
+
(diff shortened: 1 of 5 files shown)


josevalim:
Didn't we try this before as compile_aot and we struggled with it? Or is the difference here that we only serialize to the same environment that we run? Should we wait until we get MLIR/IREE support so we include precompilation too?

Otherwise, maybe I would rather support it as something like `cache: true` to the regular `jit` functions so we automatically serialize and cache the computation for you?

seanmor5:
We did try it before, the difference is that it was only supported on TPU, but now it is also available on GPU. I'm not sure if they have CPU support yet, but at the very least for those with GPUs in production this would be a significant speed up in compilation times if they have a fixed workload

josevalim:
So what do you feel about us automatically caching it? And then we add it as a cache inside EXLA.Defn. So we first look in memory, and then if there is disk cache enabled, we read it from there?

josevalim:
This means we will automatically cache it based on shape and types, instead of leaving that up to the user?

josevalim (approved):
Anyway, this looks good to me. Btw, will this work with MLIR as well? I.e. do both MLIR and plain XLA emit an EXLA executable? If so, ship it, and i can work on a higher-level API. :)

seanmor5:
Yes, both will emit a PjRtLoadedExdcutable, so this should work regardless! I will ship when I can make sure it works on my GPU :)

PR #1419, opened 2023-12-09
Add square

krstopro:
Adding `square` that squares the elements of a tensor. I find it convenient to use with pipeline operator `|>` instead of `Nx.pow(2)`.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -398,7 +398,7 @@ defmodule Nx do
   """
 
   import Nx.Shared
-  import Nx.Defn.Kernel, only: [keyword!: 2]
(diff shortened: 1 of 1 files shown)


josevalim:
I am unsure we need more shortcuts but I'd like to hear everyone's opinion. :)

polvalente:
I don't think `square` is a particularly good one, because of the clash with `Nx.LinAlg.matrix_power(.., 2)`, which is also a "square" of a matrix, but in a matrix multiplication sense.

polvalente:
There's also the issue with complex numbers, where in general things that "square" for reals tend to become `multiply(z, conjugate(z))` for complex

josevalim:
It seems best to avoid this one in particular then. Thanks @krstopro !

PR #1420, opened 2023-12-10
feat: QR decomposition as optional callback

### LINKED ISSUE #1397: Convert Nx.LinAlg.qr to optional callback

polvalente:
Work towards #1397
We have a bug related to `if` with container outputs in MLIR, so we need to fix that before closing the issue.


polvalente:
There's a final bug to resolve on MLIR before merging. The Nx side is done, though.

josevalim (commented):
Beautiful. Do we need to change torchx to hook into qr via the optional callback system?

polvalente:
> Beautiful. Do we need to change torchx to hook into qr via the optional callback system?

Nope! Torchx already has it's own implementation, and didn't call the Nx.Defn.Expr clause for anything, so it doesn't need any changes

polvalente:
@josevalim I decided to deal with the MLIR things on a separate PR. Let me know if aside from that this one is ok to merge!

PR #1429, opened 2023-12-19
Add Nx.Defn.Compiler.to_backend/1

jonatanklosko:
Partitioned servings receive defn options with respective device id and we may want to copy model params to each partitioned device. So far we've been doing `Nx.Defn.jit_apply(&Function.identity/1, [params], defn_options)`, however for EXLA this requires twice the params memory, because input and output use separate buffers.

This PR adds a function to infer backend from compiler options:

```elixir
# Assume :cuda is available

Nx.Defn.Compiler.to_backend(compiler: EXLA, device_id: 1)
# => {EXLA.Backend, [client: :cuda, device_id: 1]}
```

### DIFF
--- a/nx/lib/nx/defn/compiler.ex
+++ b/nx/lib/nx/defn/compiler.ex
@@ -133,13 +133,8 @@ defmodule Nx.Defn.Compiler do
     Process.get(Nx.Defn, false)
   end
 
-  @doc """
-  Returns a backend corresponding to the compiler options.
-
-  The backend matches the backend used for outputs from computations
-  defined by the given compiler.
-  """
-  def to_backend(opts) do
+  @doc false
+  def __to_backend__(opts) do
     {compiler, opts} = Keyword.pop(opts, :compiler, Nx.Defn.Evaluator)
     compiler.__to_backend__(opts)
   end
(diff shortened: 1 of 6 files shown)


nx/lib/nx/defn/compiler.ex
```diff
@@ -124,6 +133,17 @@ defmodule Nx.Defn.Compiler do
     Process.get(Nx.Defn, false)
   end
 
+  @doc """
+  Returns a backend corresponding to the compiler options.
+
+  The backend matches the backend used for outputs from computations
+  defined by the given compiler.
+  """
+  def to_backend(opts) do
```
jonatanklosko, line 142: I wasn't sure if compiler should be the first argument, but this matches the options we pass to `compile`/`jit`.

josevalim: `compile`/`jit` are private though. Should we expose this on `Nx.Defn` instead, which will also merge the given options with the global/local options, and then call the compiler?

jonatanklosko: I meant `Nx.Defn.{compile,jit}`, yeah. Good call on merging! So `Nx.Defn.to_backend`?

josevalim (approved):

PR #1431, opened 2023-12-22
Add an uniq function to prepare to add one_hot_encoder feature

tiagodavi:
Hi Folks, I noticed Nx does not provide a way to do one_hot_encoder or label_encoder, so I though to add those features. I decided to start by adding a function to remove duplicate elements from a vector given we usually want to encode data with a vector of features/labels. 

Do those functions make sense here?

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -8448,8 +8448,7 @@ defmodule Nx do
         ceil: {"ceil", &:math.ceil/1},
         round: {"round (away from zero)", &:erlang.round/1}
       ] do
-    [res1, res2, res3, res4] =
-      Enum.map([-1.5, -0.5, 0.5, 1.5], fn x -> fun.(x) * 1.0 end)
+    [res1, res2, res3, res4] = Enum.map([-1.5, -0.5, 0.5, 1.5], fn x -> fun.(x) * 1.0 end)
 
     @doc """
     Calculates the #{desc} of each element in the tensor.
@@ -16450,8 +16449,7 @@ defmodule Nx do
     apply_vectorized(tensor, fn tensor, offset ->
       shape = Nx.Shape.fft2(tensor.shape)
 
-      opts =
-        Keyword.validate!(opts, [:lengths, axes: [-2, -1], eps: 1.0e-10])
+      opts = Keyword.validate!(opts, [:lengths, axes: [-2, -1], eps: 1.0e-10])
 
       [ax1, ax2] =
         case opts[:axes] do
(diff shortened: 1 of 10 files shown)


tiagodavi:
Not sure why 10 files have been changed.. I think it's due to `mix format`.

polvalente (commented):
I don't think this feature is supportable in Nx because it forces dynamic shapes, which we can't currently support.

nx/lib/nx.ex
```diff
@@ -17271,4 +17269,32 @@ defmodule Nx do
...
+  """
+  @doc type: :creation
+  def uniq(tensor) do
+    rank_limit = 1
+
+    apply_vectorized(tensor, fn t ->
+      if rank(t) == rank_limit do
+        tensor
+        |> to_list()
+        |> Enum.uniq()
```
polvalente, line 17292: This is also not a desirable pattern for inclusion in Nx because it can't be used inside defn.

tiagodavi: Got it. I tried to use MapSets or Common maps, but NX does not support any of this in map/reduce. Which is the recommend way to have "HasMaps" in NX or something I can use to remove duplicate data from the tensor?

polvalente:
> > I don't think this feature is supportable in Nx because it forces dynamic shapes, which we can't currently support.
> 
> Maybe this is good in another library just to provide the encoder similar to what sklearn does.

@msluszniak do we have something similar in Scholar? Maybe this would be useful there.

nx/lib/nx.ex
polvalente: Not in any meaningful way to what you want, basically because any deduping implies dynamic shapes

tiagodavi: Got..it maybe a way would be to create the tensor from the beginning with a flag to remove duplicate data... `Nx.tensor(allow_duplicate: false)`it will create a simple vector without duplications.. from that we can use it for anything we want... including one_hot_encoder and label_encoder.

polvalente: How would this option behave with a tensor that has more than 1 axis?

tiagodavi: Maybe remove duplicate data along all the available axis ?

polvalente:
```
[
  [1, 3, 3],
  [3, 1, 2]
]
```

What happens with this tensor in this case?

I don't see a way to define this deduping operation in a general sense without referring to a single axis.

polvalente: And even in this case, this tensor is still problematic if we dedupe along `axis: 1`, because the first row would end up with 2 elements and the second row would end up with 3 elements.

Given this, I believe that this would only apply to rank 1 tensors, in which case a simple Enum.uniq before Nx.tensor would suffice

tiagodavi:
> > > I don't think this feature is supportable in Nx because it forces dynamic shapes, which we can't currently support.
> > 
> > 
> > Maybe this is good in another library just to provide the encoder similar to what sklearn does.
> 
> @msluszniak do we have something similar in Scholar? Maybe this would be useful there.

Does make sense to add to Scholar instead ?

1 - Accept vectors only ( 1 dimension )
2 - If tensor then convert to list 
3 - If list keep list 
4 - Remove duplication 
5 - Return a tensor one hot encoded 
6 - Add a label encoder afterwards

josevalim:
Scholar has the same goals of being tailored to Nx and run its algorithms in the GPU. Computing the uniqueness of a 1000x1000x1000 tensor in pure Elixir is going to be slow.

msluszniak:
I'm currently working on it, so you don't need to bother yourself about one-hot encoder ;)

PR #1432, opened 2023-12-26
bugfix: Nx.BinaryBackend.Matrix.ts/8 returns corrupted matrix when b has more columns than rows

### LINKED ISSUE #1427: Nx.LinAlg.solve/2 with Nx.BinaryBackend returns tensor with wrong state
`Nx.LinAlg.solve(a, b)` with `Nx.BinaryBackend` returns tensor with wrong state when `b` has more columns than rows.

For example,

```elixir
a = Nx.tensor([
  [1, 1],
  [1, 2]
], type: :f64)

b = Nx.tensor([
  [2, 2, 2],
  [3, 3, 3]
], type: :f64)

x = Nx.LinAlg.solve(a, b)

Nx.shape(x)  # returns {2, 3} as expected

byte_size(x.data.state)  # expected: 48, actual: 32
```

I was trying to spot the issue, but it seems the issue is inside the function `Nx.BinaryBackend.Matrix.ts/8`, which I don't know what it's doing.

`EXLA.Backend` has no such issue.

Aetherus:
closes #1427

### DIFF
--- a/nx/lib/nx/binary_backend/matrix.ex
+++ b/nx/lib/nx/binary_backend/matrix.ex
@@ -7,151 +7,90 @@ defmodule Nx.BinaryBackend.Matrix do
   import Nx.Shared
 
   def ts(a_data, a_type, a_shape, b_data, b_type, b_shape, output_type, input_opts) do
-    transform_a = input_opts[:transform_a]
-    lower_input = input_opts[:lower]
+    # Here's Paulo Valente's proof in the deleted function ts_handle_opts/3.
 
-    # if transform_a != none, the upper will become lower and vice-versa
-    lower = transform_a == :none == lower_input
+    # When lower: false, we need the following procedure
+    # for reusing the lower_triangular engine:
+    #
+    # First, we need to reverse both rows and columns
+    # so we can turn an upper-triangular matrix
+    # into a lower-triangular one.
+    # The result will also be reversed in this case.
+    #
+    # Proof:
+    # For a result [x1, x2, x3.., xn] and a row [a1, a2, a4, ..., an]
+    # we have the corresponding b = a1 * x1 + a2 * x2 + a3 * x3 + ...+ an * xn
+    # Since the addition of a_i * x_i is commutative, by reversing the columns
+    # of a, the yielded x will be reversed.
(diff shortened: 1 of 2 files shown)


nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -7,151 +7,90 @@ defmodule Nx.BinaryBackend.Matrix do
...
       case b_shape do
         {_rows, _cols} ->
-          b_data |> binary_to_matrix(b_type, b_shape) |> ts_handle_opts(opts, :b)
+          b_data |> binary_to_matrix(b_type, b_shape)
 
         {_rows} ->
-          b_data |> binary_to_vector(b_type) |> ts_handle_opts(opts, :b)
+          b_data |> binary_to_vector(b_type)
+      end
+      |> then(& transpose_b? && transpose_matrix(&1) || &1)
```
polvalente:
```suggestion
      |> then(&if(transpose_b?, do: transpose_matrix(&1), else: &1))
```

Stylistically, I find inline ifs instead of using && and || for a ternary selection to be a more clear option because it doesn't require thinking about operator precedence

nx/lib/nx/binary_backend/matrix.ex
```diff
@@ -7,151 +7,90 @@ defmodule Nx.BinaryBackend.Matrix do
...
+    transpose_x? = transpose_b?
+    reverse_x? = reverse_b?
 
     a_matrix =
       a_data
       |> binary_to_matrix(a_type, a_shape)
-      |> ts_transform_a(transform_a)
-      |> ts_handle_opts(opts, :a)
+      |> then(&((transpose_a? && transpose_matrix(&1)) || &1))
+      |> then(&((reverse_a? && reverse_matrix(&1)) || &1))
```
polvalente, line 64: I believe my comment was sent to limbo due to the force push. Please acknowledge it in the main PR tab, as it applies to both these lines

Aetherus: Thank you for your code style suggestion. I changed it to inline if-else.

polvalente (approved):
Thanks!

PR #1433, opened 2023-12-27
Fix building empty MLIR tuples

jonatanklosko:
Currently we infer `%Function{}` from tuple arguments, but if we want to build empty tuple, there are none. And we even build such explicitly here:

https://github.com/elixir-nx/nx/blob/c5d91781b91f63cae4917e75e0dc77507852e608/exla/lib/exla/defn.ex#L721

### DIFF
--- a/exla/lib/exla/mlir/value.ex
+++ b/exla/lib/exla/mlir/value.ex
@@ -76,8 +76,8 @@ defmodule EXLA.MLIR.Value do
     %Value{op | ref: ref}
   end
 
-  def tuple([%Value{function: %Function{} = func} | _rest] = vals) do
-    refs = Enum.map(vals, fn %Value{ref: ref, function: ^func} -> ref end)
+  def tuple(%Function{} = func, vals) when is_list(vals) do
+    refs = Enum.map(vals, fn %Value{ref: ref} -> ref end)
     ref = EXLA.NIF.mlir_tuple(func.ref, refs) |> unwrap!()
     %Value{ref: ref, function: func}
   end
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -540,13 +540,13 @@ defmodule EXLA.Defn do
 
     {token, result} =
       case state.builder do
-        %Function{} ->
+        %Function{} = function ->
           # for MLIR while, the return is variadic
           # like it would have come from Nx.Defn.Composite.flatten_list.
           # We need to collect the returned values into the nested tuples
           # that should have come from the while expr
           [token | results] = mod.while(pred, body, initial)
-          result = collect_container_results(results, initial_arg)
+          result = collect_container_results(function, results, initial_arg)
           {token, result}
 
         _ ->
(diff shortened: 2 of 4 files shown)


exla/lib/exla/mlir/value.ex
```diff
@@ -76,8 +76,8 @@ defmodule EXLA.MLIR.Value do
     %Value{op | ref: ref}
   end
 
-  def tuple([%Value{function: %Function{} = func} | _rest] = vals) do
-    refs = Enum.map(vals, fn %Value{ref: ref, function: ^func} -> ref end)
```
jonatanklosko, line 80: We no longer assert on the same function, but we don't that for the `dynamic_slice` above either. @polvalente is this fine?

josevalim (approved):

exla/lib/exla/mlir/value.ex
josevalim: Yeah, I think it is fine. I think it would be very hard in practice to mix values from different functions.

jonatanklosko: There was a case where the builder function was different than the vals functions, that's why I removed the assertion, but that's a different case?

josevalim: Hrm.... we need to wait for @polvalente but maybe this clause was never actually invoked and appeared by accident?

polvalente: @jonatanklosko I checked out on this branch and put the `Enum.map` assertion back and all tests still passed. I think we can restore it.

Every time I faced a similar problem, the actual fix was elsewhere.

This tends to happen when building cond and we end up using a value from the root scope inside the if-else block (each is built as a separate function).

polvalente (approved):
LGTM, we just need to decide on whether to restore the function assertion

exla/lib/exla/mlir/value.ex
jonatanklosko: It does fail some tests in bumblebee though. Here is a minimal reproduction:

<details><summary>.livemd</summary>

````markdown
# mlir tuple

```elixir
Mix.install([
  {:nx, path: "~/git/nx/nx", override: true},
  {:exla, path: "~/git/nx/exla", override: true}
])

Nx.global_default_backend(EXLA.Backend)
Nx.Defn.global_default_options(compiler: EXLA, compiler_mode: :mlir)
```

## minimized

```elixir
defmodule Test do
  import Nx.Defn

  defn test(tensor, condition) do
    if condition do
      tensor
    else
      {_i, tensor} =
        while {i = 0, tensor}, i < 1 do
          {i + 1, tensor}
        end

      tensor
    end
  end
end
```

```elixir
tensor = Nx.tensor([1.0, 2.0, 3.0, 4.0])
condition = Nx.tensor(true)
Test.test(tensor, condition)
```
````

</details>

polvalente: Yeah, with this test I agree with removing the assertion. Basically what happens is that we build each branch of the cond as a separate function, and then one of the while tuples ends up trying to create a tuple with the root-scoped `tensor` and the else-scoped `i`, which then fails the assertion.

But these extraneous ocurrences are rewritten when we're splicing the cond together (same with while), and then everything ends up ok.

jonatanklosko: Makes sense!

PR #1438, opened 2024-01-28
Add `least_squares` function

kenken-neko:
# Changes
Added the `least_squares` function to obtain the solution of equation `Ax = y` by the least-squares method.
This function performs the same process as `numpy.linalg.lstsq` ([link](https://numpy.org/doc/stable/reference/generated/numpy.linalg.lstsq.html)) to obtain the approximate values `x`.

# Purpose
As shown in example of `numpy.linalg.lstsq` ([link](https://numpy.org/doc/stable/reference/generated/numpy.linalg.lstsq.html#:~:text=returned%20as%20matrices.-,Examples,-Fit%20a%20line)) , it is possible to obtain a solution for extrapolation that can successfully fit multiple point data.

numpy-linalg-lstsq-1

# Supplement
I have already confirmed that the solution of `numpy.linalg.lstsq` is almost the same value as the solution of this `Nx.LinAlg.least_squares`'s doctests.

# Reference
[1] https://www.sci.utah.edu/~gerig/CS6640-F2012/Materials/pseudoinverse-cis61009sl10.pdf

### DIFF
--- a/exla/test/exla/mlir/nx_linalg_doctest_test.exs
+++ b/exla/test/exla/mlir/nx_linalg_doctest_test.exs
@@ -15,7 +15,7 @@ defmodule EXLA.MLIR.NxLinAlgDoctestTest do
     invert: 1,
     matrix_power: 2
   ]
-  @rounding_error_doctests [triangular_solve: 3, eigh: 2, cholesky: 1]
+  @rounding_error_doctests [triangular_solve: 3, eigh: 2, cholesky: 1, least_squares: 3]
 
   @excluded_doctests @function_clause_error_doctests ++
                        @rounding_error_doctests ++
--- a/exla/test/exla/mlir/nx_linalg_doctest_test.exs
+++ b/exla/test/exla/mlir/nx_linalg_doctest_test.exs
@@ -15,7 +15,7 @@ defmodule EXLA.MLIR.NxLinAlgDoctestTest do
     invert: 1,
     matrix_power: 2
   ]
-  @rounding_error_doctests [triangular_solve: 3, eigh: 2, cholesky: 1, least_squares: 3]
+  @rounding_error_doctests [triangular_solve: 3, eigh: 2, cholesky: 1, least_squares: 2]
 
   @excluded_doctests @function_clause_error_doctests ++
                        @rounding_error_doctests ++
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -2214,12 +2214,13 @@ defmodule Nx.LinAlg do
 
     {a1, _a2} = a_shape
     {b1} = b_shape
+
     if a1 != b1 do
       raise(
         ArgumentError,
         "The number of rows of the matrix as the 1st argument and " <>
-        "the number of columns of the vector as the 2nd argument must be the same, " <>
-        "got 1st argument shape #{inspect(a_shape)} and 2nd argument shape #{inspect(b_shape)}"
+          "the number of columns of the vector as the 2nd argument must be the same, " <>
+          "got 1st argument shape #{inspect(a_shape)} and 2nd argument shape #{inspect(b_shape)}"
       )
     end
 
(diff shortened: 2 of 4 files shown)


kenken-neko:
The execution environment is causing slight differences in values and it does not pass the CI test. I would appreciate it if someone could tell me how to deal with this.

polvalente:
> The execution environment is causing slight differences in values and it does not pass the CI test. I would appreciate it if someone could tell me how to deal with this.

I think you need better conditioned matrices, or a different way of setting the tolerance based on the noise added

nx/lib/nx/lin_alg.ex
```diff
@@ -2138,6 +2138,105 @@ defmodule Nx.LinAlg do
     Nx.sum(s > tol)
   end
 
+  @doc """
+  Return the least-squares solution to a linear matrix equation Ax = b.
+
+  ## Examples
+
+      iex> Nx.LinAlg.lstsq(Nx.tensor([[1, 2], [2, 3]]), Nx.tensor([1, 2]))
```
polvalente, line 2146: How about we rename it to least_squares?

kenken-neko:
@polvalente 
> I think you need better conditioned matrices, or a different way of setting the tolerance based on the noise added

Thank you for your kind advice! I will try to fix it.

nx/lib/nx/lin_alg.ex
kenken-neko: I agree with main contoributor's intention to change the name to `least_squares`, since I was just imitating `numpy.linalg.lstsq`. I'll change it to `least_squares`.

polvalente: In general, we try to use the longer name for readability. pinv is one of the few exceptions to this

nx/lib/nx/lin_alg.ex
```diff
@@ -2138,6 +2138,105 @@ defmodule Nx.LinAlg do
...
+      )
+    end
+
+    {a1, _a2} = a_shape
+    {b1} = b_shape
+
+    if a1 != b1 do
+      raise(
+        ArgumentError,
+        "The number of rows of the matrix as the 1st argument and " <>
```
josevalim, line 2221:
```suggestion
        "the number of rows of the matrix as the 1st argument and " <>
```

josevalim (approved):
@polvalente go ahead and merge it if you are happy with it

josevalim:
@kenken-neko, you will have to list least_squares here:

https://github.com/elixir-nx/nx/blob/main/torchx/test/torchx/nx_linalg_doctest_test.exs#L20

and here:

https://github.com/elixir-nx/nx/blob/main/exla/test/exla/mlir/nx_linalg_doctest_test.exs#L18

To avoid precision issues in tests. :)

polvalente (approved):

exla/test/exla/mlir/nx_linalg_doctest_test.exs
```diff
@@ -15,7 +15,7 @@ defmodule EXLA.MLIR.NxLinAlgDoctestTest do
     invert: 1,
     matrix_power: 2
   ]
-  @rounding_error_doctests [triangular_solve: 3, eigh: 2, cholesky: 1]
+  @rounding_error_doctests [triangular_solve: 3, eigh: 2, cholesky: 1, least_squares: 3]
```
josevalim, line 18:
```suggestion
  @rounding_error_doctests [triangular_solve: 3, eigh: 2, cholesky: 1, least_squares: 2]
```

torchx/test/torchx/nx_linalg_doctest_test.exs
```diff
@@ -17,7 +17,8 @@ defmodule Torchx.NxLinAlgDoctestTest do
     solve: 2,
     invert: 1,
     determinant: 1,
-    pinv: 2
+    pinv: 2,
+    least_squares: 3
```
josevalim, line 21:
```suggestion
    least_squares: 2
```

kenken-neko:
@josevalim 
Thank you for your fix error comments and value of `rounding_error_doctests`

josevalim:

PR #1439, opened 2024-02-10
Remove the Nx.take backend callback

### LINKED ISSUE #1366: Remove take_along_axis and take
`take` can directly call gather, no need for optional callbacks.  `take_along_axis` should have an optional callback.

Benjamin-Philip:
This commit removes the Nx.take backend callback, and replaces it with
a Nx.gather based implementation.

Closes #1366.


Benjamin-Philip:
@polvalente, I tried basing this on the EXLA implementation, but it was very cryptic to me. As far as I can understand, XLA's gather gathers slices, taking the starting indices and the slices dimension, and slicing accordingly. However, Nx's gather takes the actual indices of every element, making the XLA implementation incompatible with Nx's gather. Would basing on top of Torchx's current implementation work?

polvalente:
@Benjamin-Philip I believe you can experiment a bit with the `:axes` option from gather to replicate `take`. What I meant by looking into the EXLA implementation is that in [this section](https://github.com/elixir-nx/nx/blob/2a997f31bdb8de2de1657c444126f7d6d9e04fda/exla/lib/exla/defn.ex#L1596-L1597) you see that the `:axis` for `take` is passed as the `:axes` for gather is passed. Basically, both `gather` and `take` use XLA's gather, and so you can find the equivalence there.

I think just passing `:axis` as `:axes` might work without too much of a hassle

Benjamin-Philip:
> I think just passing :axis as :axes might work without too much of a hassle

I already tried that, and got one of the following errors:

```
  test/nx/vectorize_test.exs:447
     ** (ArgumentError) expected indices rank to be at least 1, got: 0
```

```
    test/nx_test.exs:2822
     ** (ArgumentError) expected the last indices dimension size (2) to be less than or equal to the tensor rank (1)
```

```
    test/nx/doctest_test.exs:4
     ** (ArgumentError) errors were found at the given arguments:

       * 3rd argument: out of range
     
     stacktrace:
       :erlang.binary_part(<<1, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 101, 0, 0, 0, 0, 0, 0, 0, 102, 0, 0, 0, 0, 0, 0, 0, 11, 0, 0, 0, 0, 0, 0, 0, 12, 0, 0, 0, 0, 0, 0, 0, 111, 0, ...>>, 56, 32)
       (nx 0.7.0-dev) lib/nx/binary_backend.ex:2004: Nx.BinaryBackend."-gather/3-lbc$^0/2-0-"/9
       (nx 0.7.0-dev) lib/nx/binary_backend.ex:1998: Nx.BinaryBackend.gather/3
       (nx 0.7.0-dev) lib/nx.ex:14478: Nx.gather/3
       (nx 0.7.0-dev) lib/nx.ex:14129: Nx.take/3
       (for doctest at) lib/nx.ex:13946: (test)
```

(first 2 from `Nx.Shape.gather`) on the following implementation:

```elixir
tensor
|> gather(indices, axes: [axis])
|> reshape(inner_shape, names: inner_names)
```
----

From comparing EXLA's [`take`]() and [`gather`]() implementations side by side, I found a discrepancy in a single parameter `offset_dims` (while everything else is the same):

https://github.com/elixir-nx/nx/blob/51c7122e42230179e1686c972d5103545070c64a/exla/lib/exla/defn.ex#L1765
https://github.com/elixir-nx/nx/blob/51c7122e42230179e1686c972d5103545070c64a/exla/lib/exla/defn.ex#L1830-L1831

Which is why I feel they are incompatible (unless of couse I misunderstood something, in which case I am wrong).

---

Alternatively, I could try copy pasting the old Torchx implementation, which should work out of the box.

Benjamin-Philip:
You can find the same discrepancy in `Nx.Shape`:

https://github.com/elixir-nx/nx/blob/bead39ccd703cd9243e985659b2905eea8977862/nx/lib/nx/shape.ex#L1367
https://github.com/elixir-nx/nx/blob/bead39ccd703cd9243e985659b2905eea8977862/nx/lib/nx/shape.ex#L1638-L1646

Benjamin-Philip:
@polvalente, let me know if just copy pasting this will work, or if we should go with some other approach:

https://github.com/elixir-nx/nx/blob/bead39ccd703cd9243e985659b2905eea8977862/torchx/lib/torchx/backend.ex#L340-L384

nx/lib/nx.ex
```diff
@@ -14124,13 +14124,17 @@ defmodule Nx do
...
-        axis
-      )
+      {indices_axes, tensor_axes} = Enum.split(axes(inner_shape), rank(indices))
+      {leading, trailing} = Enum.split(tensor_axes, axis)
+
+      transpose_axes = leading ++ indices_axes ++ trailing
+
+      tensor
+      |> gather(gather_indices, axes: [axis])
+      |> transpose(axes: transpose_axes)
```
polvalente, line 14136: Why is this transpose needed?

nx/lib/nx.ex
```diff
@@ -14124,13 +14124,17 @@ defmodule Nx do
     else
       tensor = devectorize(tensor, keep_names: false)
       indices = devectorize(indices, keep_names: false)
+      gather_indices = reshape(indices, indices |> shape |> Tuple.append(1))
```
polvalente, line 14127:
```suggestion
      gather_indices = new_axis(indices, rank(indices))

**Comment by Benjamin-Philip** [comment]
The transpose you isn't needed for an axis of zero, but for any axis
greater than that, the shape and thus the indices for each value don't
match. We need the transpose to fix that. Have a look at their respective
`Nx.Shape` helper functions if you don't know what I mean.

-- bp

File: nx/lib/nx.ex (hunk 1 above)
Comment (josevalim) [reply nx/lib/nx.ex:14136]: I took a look at this. `take` will retain the shape according to the given axis, this is also true in Python:

```
>>> np.take(n, [0, 1, 0, 1], axis=1)
array([[1, 2, 1, 2],
       [3, 4, 3, 4]])
>>> np.take(n, [0, 1, 0, 1], axis=0)
array([[1, 2],
       [3, 4],
       [1, 2],
       [3, 4]])
```

On the other hand, gather will always preserve the input shape. This means the underlying `offset_dims` in the gather operation is different. So it may be that we cannot really replace take by gather (unless we want to use lose performance when axis != 0, which may be fine). Another option is to extend gather with a new option that controls the output shape but that's debatable.

josevalim:
CI was failing due to a bug on exla, I have pushed a fix.

nx/lib/nx.ex
josevalim: I will merge this and make it optional in a follow up commit.

josevalim:

Benjamin-Philip:
Sounds good.

-- bp

PR #1440, opened 2024-02-10
Make take_along_axis an optional callback

### LINKED ISSUE #1366: Remove take_along_axis and take

Benjamin-Philip:
Makes take_along_axis an optional callback, using the default
implementation for BinaryBackend and EXLA, and a custom one for
Torchx.

Closes #1366.

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -1259,43 +1259,6 @@ defmodule EXLA.Defn do
     Value.dynamic_update_slice(tensor, slice, start_indices, expr_to_typespec(ans))
   end
 
-  defp to_operator(:take_along_axis, [%Value{} = tensor, indices, axis], ans, state) do
-    %{shape: indices_shape} = indices_typespec = Value.get_typespec(indices)
-    indices_rank = tuple_size(indices_shape)
-
-    axes_range = 0..(indices_rank - 1)//1
-
-    index_vector_dim = indices_rank
-    slice_sizes = List.duplicate(1, indices_rank)
-    offset_dims = []
-    collapsed_slice_dims = Enum.to_list(axes_range)
-    start_index_map = Enum.to_list(axes_range)
-
-    new_axis_typespec = Typespec.to_shape(indices_typespec, Tuple.append(indices_shape, 1))
-
-    full_indices_typespec =
-      Typespec.to_shape(indices_typespec, Tuple.append(indices_shape, indices_rank))
-
-    full_indices =
-      axes_range
-      |> Enum.map(fn
-        ^axis -> Value.reshape(indices, new_axis_typespec)
-        axis -> Value.iota(state.builder, axis, new_axis_typespec)
-      end)
-      |> Value.concatenate(indices_rank, full_indices_typespec)
-
-    Value.gather(
-      tensor,
-      full_indices,
-      index_vector_dim,
-      slice_sizes,
-      offset_dims,
-      collapsed_slice_dims,
-      start_index_map,
-      expr_to_typespec(ans)
-    )
-  end
-
   defp to_operator(:gather, [%Value{} = tensor, indices, opts], ans, _state) do
     axes = Keyword.fetch!(opts, :axes)
     tensor_shape = op_shape(tensor)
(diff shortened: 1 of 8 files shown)


Benjamin-Philip:
I'm still yet to implement the default implementation. Hence the failing tests.

exla/lib/exla/defn.ex
```diff
@@ -1243,43 +1243,6 @@ defmodule EXLA.Defn do
...
-    full_indices_typespec =
-      Typespec.to_shape(indices_typespec, Tuple.append(indices_shape, indices_rank))
-
-    full_indices =
-      axes_range
-      |> Enum.map(fn
-        ^axis -> Value.reshape(indices, new_axis_typespec)
-        axis -> Value.iota(state.builder, axis, new_axis_typespec)
-      end)
-      |> Value.concatenate(indices_rank, full_indices_typespec)
```
josevalim, line 1269: FWIW, this is the main transformation you need to make in order to convert take_along_axis into a gather. :) If you do this transformation and call `gather` without the axes option, it should be equivalent to the work done here.

Benjamin-Philip:
@josevalim, I've committed that transformation. I'm having some trouble making the optional callback work in `Defn`, which is why is failing in `Nx.Defn.GradTest` and `Nx.RandomTest`.

nx/lib/nx.ex
```diff
@@ -14312,7 +14312,27 @@ defmodule Nx do
 
     shape = Nx.Shape.take_along_axis(tensor.shape, indices.shape, axis)
 
-    result = impl!(tensor).take_along_axis(%{tensor | shape: shape}, tensor, indices, axis)
+    out = %{tensor | shape: shape}
+
+    result =
+      Nx.Shared.optional(:take_along_axis, [tensor, indices, axis], out, fn tensor,
+                                                                            indices,
+                                                                            axis ->
```
josevalim, line 14320: All arguments to optional must tensors or options, so you can fix it like this:

```suggestion
      Nx.Shared.optional(:take_along_axis, [tensor, indices, [axis: axis]], out, fn
        tensor, indices, _opts ->
```

nx/lib/nx.ex
```diff
@@ -14312,7 +14312,27 @@ defmodule Nx do
...
+          axes_range
+          |> Enum.map(fn
+            ^axis -> reshape(indices, new_axis_shape)
+            axis -> iota(new_axis_shape, axis: axis)
+          end)
+          |> concatenate(axis: rank(indices))
+
+        out = gather(tensor, full_indices)
+
+        %{out | shape: shape}
```
josevalim, line 14334: Interesting, why do you need to change the shape? We should probably avoid doing this!

nx/lib/nx/backend.ex
```diff
@@ -159,6 +158,7 @@ defmodule Nx.Backend do
   @callback all_close(out :: tensor, tensor, tensor, keyword) :: tensor
   @callback top_k(out :: tensor, tensor, keyword) :: tensor
   @callback take(out :: tensor, input :: tensor, indices :: tensor, keyword) :: tensor
+  @callback take_along_axis(out :: tensor, input :: tensor, indices :: tensor, axis) :: tensor
```
josevalim, line 161:
```suggestion
  @callback take_along_axis(out :: tensor, input :: tensor, indices :: tensor, keyword) :: tensor
```

nx/lib/nx.ex
Benjamin-Philip: I just copied it from the deleted impl call without thinking much about it.

josevalim:
@Benjamin-Philip you can also remove the implementations from `expr.ex` and `grad.ex` :)

Benjamin-Philip:
> you can also remove the implementations from expr.ex and grad.ex :)

Tried that, it still fails. I'm committing it though.

nx/lib/nx/backend.ex
Benjamin-Philip: On second thoughts, I'll revert this: since axis is the only option all the implementations directly accepted the axis.

josevalim:
Fixed and pushed manually, thank you!

PR #1442, opened 2024-02-12
Add Nx.fill/2

jonatanklosko:
Very much a subject to comments.

Let's say we want to create a zeroed tensor, shape-compatible with `tensor`. We can do this:

```elixir
Nx.tensor(0, type: Nx.type(tensor)) |> Nx.broadcast(Nx.shape(tensor))
```

We could make this `Nx.full(shape, value, type: type)`, but I am not sure it has enough value.

However, if we want to create a zeroed tensor to exactly match `tensor`, including names and vectorized axes, then it becomes much more cumbersome:

```elixir
[zeros, _] =
  Nx.broadcast_vectors([
    Nx.tensor(0, type: Nx.type(tensor)) |> Nx.broadcast(Nx.shape(tensor), names: Nx.names(tensor)),
    sample_template
  ])

# or

Nx.tensor(0, type: Nx.type(tensor))
|> Nx.broadcast(Nx.shape(Nx.devectorize(tensor)), names: Nx.names(tensor))
|> Nx.vectorize(sample_template.vectorized_axes)
```

To make it clear that we are getting tensor with the exact same shape characteristics, it's useful to think about this operation as replacing every element, or in other words filling the given tensor with a single value, hence `Nx.fill/2`. I am leaning towards tagging it as element-wise, rather than creation, for that exact reason.

There are [`numpy.ndarray.fill`](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.fill.html) and [`torch.Tensor.fill_`](https://pytorch.org/docs/stable/generated/torch.Tensor.fill_.html), both are methods on the tensor and mutate it.

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -16879,8 +16879,15 @@ defmodule Nx do
   Replaces every value in `tensor` with `value`.
 
   The returned tensor has the same shape, names and vectorized axes
-  as the given one. The type defaults to the type of `tensor`, however
-  type promotions apply.
+  as the given one. The type will be computed based on the type of
+  `tensor` and `value`. You can also pass a `:type` option to change
+  this behaviour.
+
+  ## Options
+
+    * `:type` - sets the type of the returned tensor. If one is not
+      given, it is automatically inferred based on the inputs, with
+      type promotions
 
   ## Examples
 
@@ -16904,6 +16911,17 @@ defmodule Nx do
         ]
       >
 
+      iex> tensor = Nx.iota({2, 2})
+      iex> Nx.fill(tensor, 5, type: :u8)
+      #Nx.Tensor<
+        u8[2][2]
+        [
+          [5, 5],
+          [5, 5]
+        ]
+      >
+
+
   ### Type promotions
 
   Type promotions should happen automatically, with the resulting type being the combination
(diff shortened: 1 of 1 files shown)


nx/lib/nx.ex
```diff
@@ -16875,6 +16875,65 @@ defmodule Nx do
...
+      iex> Nx.fill(tensor, 5)
+      #Nx.Tensor<
+        s64[x: 2][2]
+        [
+          [5, 5],
+          [5, 5]
+        ]
+      >
+
+  ### Type promotions
```
jonatanklosko, line 16907: This matches the behaviour of `indexed_put` and `put_slice`, which I believe makes sense. Semantically it's the same as putting the value at every index.

nx/lib/nx.ex
```diff
@@ -16875,6 +16875,65 @@ defmodule Nx do
...
+        f32[2][2]
+        [
+          [5.0, 5.0],
+          [5.0, 5.0]
+        ]
+      >
+
+  """
+  @doc type: :element
+  def fill(tensor, value) do
```
josevalim, line 16924: Should we have accept the dtype as option? Otherwise I can't fill it with `Nx.u8(0)`, or can I?

jonatanklosko: Sounds good, done!

josevalim (approved):

josevalim:

PR #1444, opened 2024-02-15
feat(mlir): initial infeed/outfeed support

polvalente:
Adds initial support for infeed and outfeed in MLIR

Some tests are still skipped due to incomplete/incorrect implementations

Also changes binary ops to receive an explicit builder function instead of requiring both operands to belong in the same function.

### DIFF
--- a/exla/c_src/exla/mlir/builder.cc
+++ b/exla/c_src/exla/mlir/builder.cc
@@ -1063,6 +1063,9 @@ ERL_NIF_TERM ConstantOpImpl(mlir::OpBuilder *builder, mlir::Type type, ErlNifEnv
 ERL_NIF_TERM MLIRFunction::ConstantOp(mlir::Type type, ErlNifEnv *env, ERL_NIF_TERM term, std::optional<std::vector<int64_t>> dims) {
   module_->builder()->setInsertionPointToEnd(&func_->getBody().back());
 
+  if (type.isSignlessInteger(1)) {
+    return ConstantOpImpl<exla::uint8>(module_->builder(), type, env, term, dims);
+  }
   if (type.isUnsignedInteger(8)) {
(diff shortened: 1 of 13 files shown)


exla/lib/exla/defn.ex
```diff
@@ -137,7 +137,153 @@ defmodule EXLA.Defn do
...
+        true
+      )
+
+    [acc] = EXLA.MLIR.Function.get_arguments(builder)
+
+    init =
+      Value.tuple(builder, [
+        Value.infeed(root_token, flag_shape),
+        acc,
+        # TO-DO: figure out the proper constant init
```
polvalente, line 269: @josevalim I expect this to be resolved by working through one of the still-failing tests.
If you have any ideas to solve this now, though, it'd probably be better.

exla/c_src/exla/mlir/builder.cc
```diff
@@ -1063,6 +1063,9 @@ ERL_NIF_TERM ConstantOpImpl(mlir::OpBuilder *builder, mlir::Type type, ErlNifEnv
 ERL_NIF_TERM MLIRFunction::ConstantOp(mlir::Type type, ErlNifEnv *env, ERL_NIF_TERM term, std::optional<std::vector<int64_t>> dims) {
   module_->builder()->setInsertionPointToEnd(&func_->getBody().back());
 
+  if (type.isSignlessInteger(1)) {
+    return ConstantOpImpl<exla::uint8>(module_->builder(), type, env, term, dims);
+  }
   if (type.isUnsignedInteger(8)) {
```
josevalim, line 1069:
```suggestion
  if (type.isUnsignedInteger(8) || type.isSignlessInteger(1)) {
```

josevalim (approved):

PR #1446, opened 2024-02-15
Support a mapping function in Nx.Defn.Kernel.print_value/2

jonatanklosko:
When debugging large intermediate tensors, I usually look at a small slice or summaries such as `Nx.sum`, however `print_value` requires the returned expression to be used as is. This always leads me to tricks such as:

```elixir
x
|> then(fn x -> x + 0 * Nx.sum(print_value(x[[.., 1..3]])) end)
```

I've just realised we can generalize `print_value` to support additional transformation by using `hook_token`. With the new version, the example above would be:

```elixir
x
|> print_value(& &1[[.., 1..3]])
```

### DIFF
--- a/nx/test/nx/defn/evaluator_test.exs
+++ b/nx/test/nx/defn/evaluator_test.exs
@@ -324,7 +324,9 @@ defmodule Nx.Defn.EvaluatorTest do
   defn mapped_inspect(a, b), do: print_value(a + b, fn x -> x * 2 end)
 
   test "print_value/2 with mapping function" do
-    assert ExUnit.CaptureIO.capture_io(fn -> mapped_inspect(1, 2) end) ==
+    assert ExUnit.CaptureIO.capture_io(fn ->
+             assert mapped_inspect(1, 2) == Nx.tensor(3)
+           end) ==
              """
              #Nx.Tensor<
                s64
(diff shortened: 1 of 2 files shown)


nx/test/nx/defn/evaluator_test.exs
```diff
@@ -321,6 +321,18 @@ defmodule Nx.Defn.EvaluatorTest do
              """
   end
 
+  defn mapped_inspect(a, b), do: print_value(a + b, fn x -> x * 2 end)
```
polvalente, line 324: I found this behavior a bit confusing. I expected the print to show the modified version but not actually affect the results.

Otherwise, what's the difference between this and just modifying the results before piping into print_value?

jonatanklosko: It does not affect the result, the assertion is on the printed value :)

jonatanklosko: I will add an assertion on the result too!

polvalente: Ah cool!

polvalente: I misread the test :)

polvalente: This is a great addition then!

polvalente (approved):

josevalim (approved):

josevalim:

PR #1447, opened 2024-02-16
feat(mlir): support tokens in cond

polvalente:
- feat(mlir): support tokens in cond
- chore: remove stray describetag

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -743,8 +743,12 @@ defmodule EXLA.Defn do
 
     case state.builder do
       %Function{} ->
-        # TO-DO(mlir): deal with token
-        {cond_op, cache}
+        if get_token(cache) do
+          {token, results} = cond_op
+          {results, update_token(cache, token)}
+        else
+          {cond_op, cache}
+        end
 
       _ ->
         if get_token(cache) do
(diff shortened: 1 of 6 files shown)


exla/lib/exla/defn.ex
```diff
@@ -2686,28 +2710,64 @@ defmodule EXLA.Defn do
          cache,
          fun
        ) do
-    # TO-DO(mlir): deal with token
-    inputs = Enum.with_index(args, fn arg, idx -> {"p#{idx}", Value.get_shape(arg)} end)
+    if token = get_token(cache) do
+      # TO-DO(mlir): deal with token
```
josevalim, line 2714:
```suggestion
```

exla/lib/exla/defn.ex
```diff
@@ -2686,28 +2710,64 @@ defmodule EXLA.Defn do
...
+          false,
+          true
+        )
 
-    # input function is actually the parent function still, so we need to actually create a new function
-    # with this name on the same module.
-    function = EXLA.Builder.new({module, name}, inputs, out_expr, :mlir, false, true)
+      case function.return_shape do
+        [%{dtype: {:tuple, _}}] ->
+          raise "MLIR cannot return tuple from if branch"
```
josevalim, line 2732: In the future, we should not need this check, because we always flatten it before calling the builder.

josevalim (approved):

exla/lib/exla/defn.ex
polvalente: Yes :)

PR #1448, opened 2024-02-20
refactor(mlir): reduce tuple usage for returns in favor of variadic+flat tuples

polvalente:
- refactor: new compilation for cond
- refactor: while on new workflow
- refactor: token computation
- refactor: sort computation
- refactor: op computation
- refactor: fun_computation
- refactor: stream computation
- refactor: root builder
- refactor: separate builder new fully
- refactor: remove mlir_build NIF

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -2906,10 +2906,8 @@ defmodule EXLA.Defn do
     to_type(value, {:pred, 8})
   end
 
-  defp new_mlir_function(module_and_name, arg_shapes, return_shape, output_tuple?) do
-    in_shape = arg_shapes
-    out_shape = return_shape
-
+  defp new_mlir_function(module_and_name, in_shape, out_shape, output_tuple?)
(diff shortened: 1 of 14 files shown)


exla/lib/exla/defn.ex
```diff
@@ -2939,4 +2892,58 @@ defmodule EXLA.Defn do
...
+      end
+
+    EXLA.Builder.new_mlir(module_and_name, in_shape, out_shape)
+  end
+
+  defp exla_shape(tensors) when is_list(tensors) do
+    Enum.flat_map(tensors, &exla_shape/1)
+  end
+
+  defp exla_shape(tensors) when is_tuple(tensors) do
```
josevalim, line 2914: I am not the biggest fan of such APIs here, it means we usually don't know the shape of the input. Could we rely on only one or the other?

polvalente: What do you mean by "one or the other"?

josevalim (approved):

PR #1449, opened 2024-02-20
feat: add config for compiler mode

polvalente:
- feat: add compiler mode config
- feat: add compiler_mode config

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -581,7 +581,7 @@ defmodule EXLA.Defn do
 
           inputs_and_shapes = Enum.reverse(reverse_inputs_and_shapes)
 
-          mode = options[:compiler_mode] || :xla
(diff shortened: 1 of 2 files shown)


exla/lib/exla/defn.ex
```diff
@@ -581,7 +581,7 @@ defmodule EXLA.Defn do
 
           inputs_and_shapes = Enum.reverse(reverse_inputs_and_shapes)
 
-          mode = options[:compiler_mode] || :xla
+          mode = options[:compiler_mode] || Application.get_env(:exla, :compiler_mode, :mlir)
```
jonatanklosko, line 584: I think having just the config attribute would be fine, but if streamlines a particular use case both are fine.

josevalim: It is temporary anyway, so both are fine. :)

josevalim (approved):

PR #1453, opened 2024-02-23
refactor(mlir): use if regions

polvalente:
Reworks IfOp so that we write to regions directly using nested scopes instead of allocating temporary functions.


exla/c_src/exla/mlir/builder.cc
```diff
@@ -942,29 +944,65 @@ std::vector<mlir::Value> MLIRFunction::IfOp(mlir::Value pred, std::vector<xla::S
...
 
-  implicit_arguments.erase(implicit_arguments.begin());
-  ReplaceBlockArgumentsWithImplicitOperands(if_op.getOperation(), implicit_arguments);
+  mlir::Region &region = true_or_false ? op.getTrueBranch() : op.getFalseBranch();
+  mlir::Block &block = region.emplaceBlock();
+  builder->setInsertionPointToEnd(&block);
 
-  mlir::Operation::result_range result_range = if_op.getResults();
-  std::vector<mlir::Value> results(result_range.begin(), result_range.end());
+  regions.push(&region);
```
josevalim, line 962: You can push the region and then call the usual  “setInsertionPoint()” after.

exla/c_src/exla/mlir/builder.cc
```diff
@@ -942,29 +944,65 @@ std::vector<mlir::Value> MLIRFunction::IfOp(mlir::Value pred, std::vector<xla::S
...
+}
 
-  return results;
+void MLIRFunction::ResetRegion() {
```
josevalim, line 965: Maybe popRegion?

exla/c_src/exla/mlir/builder.cc
```diff
@@ -942,29 +944,65 @@ std::vector<mlir::Value> MLIRFunction::IfOp(mlir::Value pred, std::vector<xla::S
...
+
+//   mlir::Region &falseBody = if_op.getFalseBranch();
+//   auto &onFalseBlocks = on_false->function()->getBody().getBlocks();
+//   falseBody.getBlocks().splice(falseBody.end(), onFalseBlocks);
+
+//   implicit_arguments.erase(implicit_arguments.begin());
+//   ReplaceBlockArgumentsWithImplicitOperands(if_op.getOperation(), implicit_arguments);
+
+//   return results;
+// }
```
josevalim, line 1004: Remove this code?

exla/lib/exla/defn.ex
```diff
@@ -2545,6 +2546,50 @@ defmodule EXLA.Defn do
 
   ## Cond
 
+  defp to_if(pred, on_true, on_false, %{builder: %Function{} = function} = state, cache) do
+    {pred_op, cache} = recur_operator(pred, state, cache)
+    # pred_op = to_type(pred_op, {:pred, 8})
```
josevalim, line 2551: uncomment or delete?

exla/lib/exla/defn.ex
```diff
@@ -2629,6 +2649,49 @@ defmodule EXLA.Defn do
...
+         %{scope_ids: ids} = state,
+         cache
+       ) do
+    {_, ids_args} =
+      Composite.reduce(expr, {%{}, %{}}, fn node, acc ->
+        {_, acc} = collect_args(node, acc, {ids, other_ids})
+        acc
+      end)
+
+    # sorted_ids_args = Enum.sort_by(ids_args, fn {_id, {i, _old, _new}} -> i end)
```
josevalim, line 2664: Delete.

exla/test.exs
josevalim, line 1: Remove this file.

exla/lib/exla/defn.ex
```diff
@@ -2559,33 +2604,8 @@ defmodule EXLA.Defn do
...
-          if get_token(cache) do
-            [token | results] = if_results
-            {token, wrap_tuple_result(function, results, on_true)}
-          else
-            wrap_tuple_result(function, if_results, on_true)
-          end
-
-        {results, cache}
+      %EXLA.MLIR.Function{} ->
+        raise "asdf"
```
josevalim, line 2608: Raise something better or pattern match on the other earlier.

exla/lib/exla/defn.ex
```diff
@@ -2545,6 +2546,50 @@ defmodule EXLA.Defn do
...
+    cache = to_mlir_if_branch(true, node, on_true, true_ids, state, cache)
+
+    # We need to set the outer token back in the cache so that the false branch also
+    # sees the same input token. The branch that executes will either update or return the token as is,
+    # but both branch off of the same token.
+
+    # the output token is `node` as above, if the computation does indeed contain a token.
+
+    cache =
+      to_mlir_if_branch(false, node, on_false, false_ids, state, update_token(cache, in_token))
```
josevalim, line 2581: In this case, I think you can just not return the cache at all.

```suggestion
    to_mlir_if_branch(true, node, on_true, true_ids, state, cache)
    to_mlir_if_branch(false, node, on_false, false_ids, state, cache)
```

exla/lib/exla/defn.ex
```diff
@@ -2629,6 +2649,49 @@ defmodule EXLA.Defn do
...
+    {res, cache} = recur_composite(expr, &cast_pred_to_u8/1, comp_state, cache)
+
+    if token = get_token(cache) do
+      Value.variadic_return([token, res], true)
+    else
+      Value.variadic_return([res], true)
+    end
+
+    Function.reset_region(state.builder)
+    cache
```
josevalim, line 2692: If we indeed don't need to return the cache, we can simply do:

```suggestion
    {res, _cache} = recur_composite(expr, &cast_pred_to_u8/1, comp_state, cache)

    if token = get_token(cache) do
      Value.variadic_return([token, res], true)
    else
      Value.variadic_return([res], true)
    end

    Function.reset_region(state.builder)
```

josevalim (approved):

PR #1454, opened 2024-02-23
fix: use singleton mlir context

polvalente:
With a little bit of back and forth with @jpienaar from OpenXLA,
turns out that MLIRContext manages threading, so using a global singleton
is better for performance.

MLIR is now on perf. parity with XlaOp from my local benchmarks.

This also explains why we saw no significant difference when running a problematic test by itself.


exla/test/test_helper.exs
```diff
@@ -26,6 +26,8 @@ if client.platform == :host and client.device_count == 1 and System.schedulers_o
   )
 end
 
+IO.gets("Press enter to continue... PID: #{System.pid()}}")
+
```
polvalente, line 30:
```suggestion
```

exla/lib/exla/mlir/module.ex
```diff
@@ -11,11 +11,22 @@ defmodule EXLA.MLIR.Module do
...
   @doc """
   Creates a new MLIR module.
   """
   def new() do
-    ref = EXLA.NIF.new_mlir_module() |> unwrap!()
+    context = :persistent_term.get(@context_key, nil)
+
+    unless context do
+      raise "No MLIR context found"
+    end
```
jonatanklosko, line 27: We could do the init lazily:

```elixir
defp get_or_create_context() do
  if context = :persistent_term.get(@context_key, nil) do
    context
  else
    context = EXLA.NIF.new_mlir_context() |> unwrap!()
    :persistent_term.put(@context_key, context)
    context
  end
end
```

In the very unlikely race condition, where `new/0` is called concurrently we may create two contexts, which should be fine.

@josevalim?

polvalente: I changed exactly because there was a race condition

jonatanklosko: Ok, nevermind then :D

exla/lib/exla/mlir/module.ex
```diff
@@ -11,11 +11,22 @@ defmodule EXLA.MLIR.Module do
   alias EXLA.Executable
   alias EXLA.Shape
 
+  @context_key {__MODULE__, :mlir_context}
+  def init do
+    :persistent_term.put(@context_key, EXLA.NIF.new_mlir_context() |> unwrap!())
+  end
```
josevalim, line 17: What if we initialize one context per client and then we shove this in the client?

polvalente: That might work. I'll try it out

polvalente: I tried initializing this inside the client at the C++ level and that led to some race conditions which ended in segfaults.

Seeing as different contexts will eat up more threads, I'm not sure there's any gain here outside of the code seeming more isolated.

polvalente: Actually, I just realized that multi-device is broken in this implementation already, so I'll have to figure something else out

exla/c_src/exla/mlir/builder.cc
```diff
@@ -1315,6 +1309,9 @@ void MLIRModule::LowerPatterns() {
   mlir::stablehlo::populateStablehloToHloPatterns(&patterns, &converter, context());
 
   mlir::applyPartialConversion(module(), target, std::move(patterns));
+
+  target.addLegalDialect<mlir::stablehlo::StablehloDialect>();
+  target.addLegalDialect<mlir::func::FuncDialect>();
```
polvalente, line 1314:
```suggestion
```

exla/lib/exla/mlir/module.ex
```diff
@@ -14,23 +15,35 @@ defmodule EXLA.MLIR.Module do
...
-      )
-      when is_binary(name) do
+  def add_function(module, name, arg_shapes, return_shapes) do
+    create_function(module, name, arg_shapes, return_shapes, false)
+  end
+
+  # @doc """
+  # Creates a new MLIR function with the given name belonging
+  # to the given MLIR module.
+  # """
```
josevalim, line 38:
```suggestion
```

exla/lib/exla/mlir/context_pool.ex
```diff
@@ -0,0 +1,37 @@
+defmodule EXLA.MLIR.ContextPool do
+  @doc false
```
josevalim, line 2:
```suggestion
  @moduledoc false
```

josevalim: Btw, let's `@moduledoc false` all MLIR modules, cause their APIs will certainly change once we remove xla mode.

josevalim (approved):

PR #1456, opened 2024-02-26
chore: reorganize EXLA tests

polvalente:
Reorganizes tests so that we're not testing some things twice
Also changes things so that MLIR is the default engine for tests too


josevalim (approved):

exla/test/exla/executable_test.exs
```diff
@@ -1,29 +1,44 @@
...
+  alias EXLA.BinaryBuffer
+  alias EXLA.DeviceBuffer
+  alias EXLA.Executable
+  alias EXLA.Op
+  alias EXLA.Shape
+  alias EXLA.MLIR.Value
   import EXLAHelpers
 
+  # MLIR actually allows for more than just tuples at the root
+  @tag skip: :mlir
```
josevalim, line 13: I think we can nuke this test then.

PR #1457, opened 2024-02-26
refactor(mlir): variadic infeed

polvalente:


exla/lib/exla/defn/stream.ex
```diff
@@ -88,7 +88,12 @@ defmodule EXLA.Defn.Stream do
 
       data_and_shapes =
         if client.platform == :host do
-          %EXLA.Shape{dtype: {:tuple, shapes}} = send_shape
+          shapes =
+            case send_shape do
+              %EXLA.Shape{dtype: {:tuple, shapes}} -> shapes
+              l when is_list(l) -> l
+            end
```
josevalim, line 95: Let's add a TODO to remove this once we remove EXLA.OP? Otherwise we may forget this.

```suggestion
          # TODO: Remove first-clause once EXLA.OP is removed
          shapes =
            case send_shape do
              %EXLA.Shape{dtype: {:tuple, shapes}} -> shapes
              l when is_list(l) -> l
            end
```

Unless you think we would be able to remove `dtype: {:tuple, _}` altogether?

josevalim (approved):

exla/lib/exla/defn/stream.ex
polvalente: We might be able to get rid of tuples altogether!
Let's leave the suggested to-do for now

PR #1460, opened 2024-02-27
feat(mlir): QR as custom call

polvalente:
In short, during our migration to MLIR, QR got slower due
to being implemented in defn (which can't leverage things like in-place modifications).
This adds a custom call for using Eigen (which is already a dependency of ours) on the CPU.

### DIFF
--- a/exla/test/test_helper.exs
+++ b/exla/test/test_helper.exs
@@ -1,6 +1,10 @@
 target = System.get_env("EXLA_TARGET", "host")
 client = EXLAHelpers.client()
 
+if System.get_env("DEBUG") in ["1", "true"] do
+  IO.gets("Press enter to continue... -- PID: #{System.pid()}")
+end
+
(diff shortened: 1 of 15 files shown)


josevalim (approved):

exla/test/test_helper.exs
```diff
@@ -1,6 +1,10 @@
 target = System.get_env("EXLA_TARGET", "host")
 client = EXLAHelpers.client()
 
+if System.get_env("DEBUG") in ["1", "true"] do
+  IO.gets("Press enter to continue... -- PID: #{System.pid()}")
+end
```
polvalente, line 6: Can we leave this in this time? Or something similar? Basically whenever I need to debug with lldb this is necessary.

josevalim: Yes, leave it :)

PR #1462, opened 2024-03-02
refactor: remove XlaOp workflow

polvalente:
Removes all of the XlaOp workflow
Because I had to touch the Makefile, I also took the opportunity to start speeding recompilations up.


exla/c_src/exla/exla_client.o
josevalim, line 1: Is this file here intentionally? Same below.

exla/lib/exla/computation.ex
```diff
@@ -35,43 +33,6 @@ defmodule EXLA.Computation do
...
-      client: client,
-      ref: ref,
-      output_shape: output_shape,
-      num_replicas: num_replicas,
-      num_partitions: num_partitions,
-      device_id: device_id
-    }
-  end
-
   def compile(
```
josevalim, line 36: We should talk about this. :)

exla/c_src/exla/exla_client.o
polvalente: No, this is a stray file from while I was working on the Makefile :)

exla/lib/exla/mlir/region.ex
```diff
@@ -1,5 +1,5 @@
 defmodule EXLA.MLIR.Region do
-  @moduledoc false
+  @moduledoc "Holds the ref for an MLIR region."
```
josevalim, line 2: I think all MLIR modules should be false for now.

PR #1470, opened 2024-04-06
fix: stop_grad for logsumexp

polvalente:


nx/lib/nx.ex
```diff
@@ -17332,6 +17332,16 @@ defmodule Nx do
     axes = opts[:axes]
     keep_axes = opts[:keep_axes]
     max = reduce_max(tensor, axes: axes, keep_axes: true)
+
+    max =
+      case max do
+        %T{data: %Nx.Defn.Expr{}} = t ->
+          Nx.Defn.Kernel.stop_grad(t)
```
polvalente, line 17339: @josevalim not sure if this should be done differently for compilation purposes

polvalente: I'm gonna merge, but feel free to suggest changes here. I can commit afterwards :)

PR #1473, opened 2024-04-25
feat: add to/from pointer

polvalente:
Enables pointer manipulation for sharing CUDA memory between processes in EXLA.
In theory, IPC can also be supported for host-allocated memory, but this is not implemented yet.
It is also possible to use local (as in, "same process") pointers through these functions,
albeit there's not much gain besides automated testing.


exla/Makefile
```diff
@@ -1,4 +1,4 @@
-# Environment variables passed via elixir_make
+	# Environment variables passed via elixir_make
```
josevalim, line 1:
```suggestion
# Environment variables passed via elixir_make
```

nx/lib/nx.ex
```diff
@@ -16679,6 +16679,74 @@ defmodule Nx do
...
+      Nx.from_pointer(<<5, 6, 7>>, {:s, 64}, {1, 3}, names: [nil, :col], backend_opts: [some: :option, another: :option])
+      #Nx.Tensor<
+        s64[1][col: 3]
+        [
+          [10, 20, 30]
+        ]
+      >
+  """
+  @doc type: :creation
+  def from_pointer(backend, opaque_pointer, type, shape, opts \\ []) when is_atom(backend) do
```
josevalim, line 16716: Sorry, I gave bad advice. The backend should be an option, similar to how it is done in `from_binary`. :) So there is a single option called `:backend`.

Regarding the backend options, we could simplify proxy everything else to the backend. Or we should call it `pointer_options` to avoid confusion.

nx/lib/nx.ex
```diff
@@ -16679,6 +16679,74 @@ defmodule Nx do
...
+  @doc """
+  Returns an opaque pointer for a given tensor.
+
+  Can be used in conjunction with `from_pointer/5` to share the same memory
+  for multiple tensors, as well as for interoperability with other programming
+  languages.
+
+  ## Options
+
+    All options are backend-specific
```
josevalim, line 16732:
```suggestion
  All options are backend-specific.
```

We may also nest everything under `pointer_options` to avoid confusion.

exla/lib/exla/backend.ex
```diff
@@ -82,6 +82,70 @@ defmodule EXLA.Backend do
...
+      error ->
+        error
+    end
+  end
+
+  @impl true
+  def from_pointer(pointer, type, dims, opts) do
+    template = Nx.template(dims, type, names: opts[:names])
+
+    opts = Keyword.validate!(opts[:backend_opts] || [], [:client_name, :device_id, mode: :local])
```
josevalim, line 121: As commented below, this is mixing backend options (such as client name) with pointer options. We probably should treat them separately.

nx/lib/nx/binary_backend.ex
```diff
@@ -161,6 +161,16 @@ defmodule Nx.BinaryBackend do
...
   end
 
+  @impl true
+  def from_pointer(_, _, _, _) do
+    raise ArgumentError, "#{__MODULE__} does not support pointer manipulation"
+  end
+
+  @impl true
+  def to_pointer(_, _) do
+    raise ArgumentError, "#{__MODULE__} does not support pointer manipulation"
```
josevalim, line 171:
```suggestion
    raise ArgumentError, "#{inspect(__MODULE__)} does not support pointer manipulation"
  end

  @impl true
  def to_pointer(_, _) do
    raise ArgumentError, "#{inspect(__MODULE__)} does not support pointer manipulation"
```

nx/lib/nx.ex
polvalente: I don't like backend being an option here because it is mandatory and there's no sensible default we can have.

polvalente: I agree that pointer_options is a better name! Will change both in the next commit

josevalim: If you want to force the backend, then it should mirror the API from backend_transfer, where the backend is either an atom or a two-element tuple. Also note the examples above are incorrect (they are not passing the backend as first argument).

josevalim (approved):

PR #1475, opened 2024-04-30
Remove all handling of EXLA.Shape tuple

jonatanklosko:
As far as I understand we no longer use it anywhere, as of #1463.


exla/Makefile
```diff
@@ -66,7 +66,7 @@ HEADERS = $(EXLA_DIR)/mlir/ops.h $(EXLA_DIR)/mlir/builder.h $(EXLA_DIR)/mlir/cus
 OBJECTS = $(patsubst $(EXLA_DIR)/%.cc,$(EXLA_CACHE_OBJ_DIR)/%.o,$(SOURCES)) $(EXLA_CACHE_OBJ_DIR)/exla_cuda.o
 
 
-NVCC_RESULT := $(shell which nvcc 2> NULL)
```
jonatanklosko, line 69: This creates a NULL file for me :D

polvalente: ah! this is where that was coming from haha!

polvalente (approved):

PR #1477, opened 2024-05-03
Use generic API for creating MLIR operations

jonatanklosko:
Currently we use C++ API for each individual MLIR operation, which involves a separate NIF that delegates to a corresponding "builder" function. Additionally, we need to parse any configuration we pass to the NIF.

This PR adds a single generic `EXLA.NIF.mlir_op(...)`, where we pass the operation name, operands, attributes, etc. This approach allows us to remove ~3k lines of C++, and doesn't even inflate the Elixir code. Attributes and types are passed as strings (same as in the human-readable MLIR syntax), and we call MLIR API to parse them.

One major difference between the two approaches is that most of the C++ operation APIs don't require passing the result types, because there is C++ logic that does inference based on the operands. With the generic operation this is no longer the case (since we don't use a specific operation class). Consequently, we need to explicitly pass the result types. In most cases it's trivial, since we can just take the type from the corresponding `Defn.Expr`; in others it's easy to compute.

MLIR types for tensors effectively consist of both the numeric type and the shape, so I added `EXLA.Typespec` to make it easier to pass this information around. I also entirely removed `EXLA.Shape` and replaced all its usage with `EXLA.Typespec`. `EXLA.Shape` uses a resource, which I don't think makes sense for this kind of small data; in many places we would actually call `EXLA.Shape.make_shape(...)` (NIF), just to pass the type/shape information around, or right before passing it to another NIF. `EXLA.Typespec` can be passed directly to NIFs using a compact representation, and we do that only in a couple places. I intentionally chose the name "typespec", to use different wording from "type" and "shape" (with `EXLA.Shape` it was ambiguous what a `shape` argument/variable means).

Sidenote: most changes related to the operations are in `EXLA.MLIR.Value` and `MLIR.Defn`. A lot of the other changes are related to `EXLA.Shape -> EXLA.Typespec`.


exla/c_src/exla/exla_mlir.cc
```diff
@@ -0,0 +1,145 @@
...
+namespace exla {
+MLIRFunction::MLIRFunction(MLIRModule *module, std::unique_ptr<mlir::func::FuncOp> func)
+    : func_(std::move(func)),
+      module_(module) {}
+
+std::vector<mlir::Value> MLIRFunction::Op(
+    std::string op_name, std::vector<mlir::Value> operands,
+    std::vector<mlir::Type> result_types,
+    std::vector<std::pair<std::string, mlir::Attribute>> attributes,
+    std::vector<mlir::Region *> regions) {
```
jonatanklosko, line 16: This is the magic 🐈

exla/c_src/exla/custom_calls.h
```diff
@@ -1,9 +1,9 @@
-#pragma once
-
-template <typename DataType>
-void qr_cpu_custom_call(void *out[], const void *in[]);
+#ifndef EXLA_MLIR_CUSTOM_CALLS_H_
+#define EXLA_MLIR_CUSTOM_CALLS_H_
```
polvalente, line 2: why the change here?

exla/c_src/exla/exla_mlir.cc
```diff
@@ -0,0 +1,145 @@
...
+  auto builder = module_->builder();
+  auto context = builder->getContext();
+
+  auto types_range = mlir::TypeRange{llvm::ArrayRef<mlir::Type>{result_types}};
+
+  auto named_attributes = std::vector<mlir::NamedAttribute>{};
+  for (auto const &pair : attributes) {
+    auto key = pair.first;
```
polvalente, line 24: key seems to be unused here

polvalente:
```suggestion
```

exla/c_src/exla/exla_mlir.cc
```diff
@@ -0,0 +1,145 @@
...
+    auto new_region = op_state.addRegion();
+    new_region->getBlocks().splice(new_region->end(), region->getBlocks());
+  }
+
+  auto op = builder->create(op_state);
+
+  auto results = op->getResults();
+  auto results_vec = std::vector<mlir::Value>(results.begin(), results.end());
+
+  return results_vec;
```
polvalente, line 54:
```suggestion
  return std::vector<mlir::Value>(results.begin(), results.end());
```

exla/c_src/exla/exla_mlir.cc
```diff
@@ -0,0 +1,145 @@
...
+  return output_string;
+}
+
+mlir::Type MLIRModule::ParseType(std::string string) {
+  auto type = mlir::parseType(string, context_);
+
+  if (type == nullptr) {
+    std::cerr << "Unable to parse MLIR type: " << string << std::endl;
+    exit(1);
+  }
```
polvalente, line 129: Probably better to return this as an error instead of crashing the VM

exla/c_src/exla/exla_mlir.cc
```diff
@@ -0,0 +1,145 @@
...
+}
+
+mlir::Attribute MLIRModule::ParseAttribute(std::string string) {
+  auto attribute = mlir::parseAttribute(string, context_);
+
+  if (attribute == nullptr) {
+    std::cerr << "Unable to parse MLIR attribute: " << string << std::endl;
+    exit(1);
+  }
+
```
polvalente, line 141: Same error consideration here

exla/lib/exla/mlir/value.ex
```diff
@@ -1,320 +1,388 @@
...
-      )
-      |> unwrap!()
+    result_types = typespecs_to_mlir_types([typespec])
 
-    %Value{value | ref: ref}
+    fft_type =
+      case fft_kind do
+        :fft -> attr_fft_type(:fft)
+        :ifft -> attr_fft_type(:ifft)
+      end
```
polvalente, line 378:
```suggestion
    fft_type = attr_fft_type(fft_kind)
```

exla/lib/exla/mlir/value.ex
```diff
@@ -434,206 +542,414 @@ defmodule EXLA.MLIR.Value do
...
-      |> unwrap!()
+    attributes = [
+      dimensions: attr_dense_i64_elements(dimensions)
+    ]
 
-    %Value{ref: ref, function: func}
+    regions = [mapper]
+
+    op(func, "stablehlo.map", inputs, result_types, attributes: attributes, regions: regions)
+    |> hd()
```
polvalente, line 679: Am I right that most of these `|> hd()` should only return a single value?

If so, I'd rather use hard-matching `[result] = ...` for unwrapping `op`.
Or at least, a `|> single_result!()` `defp` that does this, for more assertive code

exla/lib/exla/mlir/value.ex
```diff
@@ -434,206 +542,414 @@ defmodule EXLA.MLIR.Value do
...
+    result_types = [type_tuple([q_result_type, r_result_type])]
+
+    call_target_name =
+      case op_type do
+        {:f, 32} -> "qr_cpu_custom_call_f32"
+        {:f, 64} -> "qr_cpu_custom_call_f64"
+        {:f, 16} -> "qr_cpu_custom_call_f16"
+        {:bf, 16} -> "qr_cpu_custom_call_bf16"
+        type -> raise "QR decomposition not supported for type #{inspect(type)}"
+      end
```
polvalente, line 759: This function should be receiving the `device` for asserting that it is the host. Otherwise the error message is more confusing than it should be.

exla/lib/exla/mlir/value.ex
```diff
@@ -434,206 +542,414 @@ defmodule EXLA.MLIR.Value do
...
+    data =
+      case value do
+        :nan -> type |> Nx.Type.nan_binary() |> native_to_big()
+        :infinity -> type |> Nx.Type.infinity_binary() |> native_to_big()
+        :neg_infinity -> type |> Nx.Type.neg_infinity_binary() |> native_to_big()
+        value -> <<value::float-size(size)-big>>
+      end
+
+    Base.encode16(data)
   end
```
polvalente, line 878: Is MLIR necessarily always big endian?

polvalente: Would be nice to have a comment with a reference on that

exla/lib/exla/typespec.ex
```diff
@@ -0,0 +1,61 @@
...
+  Builds a tensor typespec.
+  """
+  def tensor(type, shape) do
+    %__MODULE__{type: type, shape: shape}
+  end
+
+  @doc """
+  Builds a token typespec.
+  """
+  def token() do
```
polvalente, line 29:
```suggestion
  def token do
```

exla/lib/exla/typespec.ex
```diff
@@ -0,0 +1,61 @@
...
+
+  @doc false
+  def nif_decode({type_charlist, shape}) do
+    %__MODULE__{shape: shape, type: charlist_to_type(type_charlist)}
+  end
+
+  defp charlist_to_type(~c"token"), do: :token
+  defp charlist_to_type(~c"pred"), do: {:pred, 8}
+  defp charlist_to_type(~c"bf16"), do: {:bf, 16}
+  defp charlist_to_type([letter | int]), do: {List.to_atom([letter]), List.to_integer(int)}
```
polvalente, line 56: I think it would be best to have the types explicitly listed here.
When we introduce quantized types, this is bound to be a cryptic and/or hard to find error

polvalente (commented):
This is looking awesome! Just a few minor or stylistic comments :)

josevalim (approved):
Beautiful work, -3k lines speaks for themselves. It looks good to me, once @polvalente is happy, we can ship it :)

exla/c_src/exla/custom_calls.h
jonatanklosko: Just for consistency with the other files, I can change any way around if you have a preference :)

exla/c_src/exla/exla_mlir.cc
```diff
@@ -0,0 +1,145 @@
...
+  for (auto const &pair : attributes) {
+    auto key = pair.first;
+    auto attribute = builder->getNamedAttr(pair.first, pair.second);
+    named_attributes.push_back(attribute);
+  }
+
+  auto operands_range = mlir::ValueRange{
+      llvm::ArrayRef<mlir::Value>(operands.data(), operands.size())};
+  auto attributes_array = llvm::ArrayRef<mlir::NamedAttribute>{
+      llvm::ArrayRef<mlir::NamedAttribute>{named_attributes}};
```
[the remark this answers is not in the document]
jonatanklosko: Definitely, good catch!

exla/c_src/exla/exla_mlir.cc
jonatanklosko: I did that initially, but then moved here to have the check in a single place. But yeah, I will revert. One detail is that `mlir::parseType` also emits a diagnostic, with possibly a more specific parsing error and I'm not sure if there's a way to capture that. That said, I will defer the error to outside NIF :)

exla/lib/exla/typespec.ex
jonatanklosko: Preference :D

exla/lib/exla/typespec.ex
jonatanklosko: Sounds good!

exla/lib/exla/mlir/value.ex
jonatanklosko: Ah, I wanted to be explicit, but then added guards while refactoring, so that should be fine :)

exla/lib/exla/mlir/value.ex
jonatanklosko: Good call, will change it to `|> one!()`!

exla/lib/exla/mlir/value.ex
jonatanklosko: We already match on `:host` here:

https://github.com/elixir-nx/nx/blob/3b3e92ebc733fb0961813679618f608ee501c199/exla/lib/exla/defn.ex#L580-L585

polvalente: Yes, I know, but the user doesn't know that. And then the error message hides this implementation detail

exla/lib/exla/mlir/value.ex
jonatanklosko: I didn't find any explicit reference, initially I tried BinaryBackend with `Nx.to_binary` and then realised it's big endian. The docs on [FloatAttr](https://mlir.llvm.org/docs/Dialects/Builtin/#floatattr) have an example for NaN and infinity, and they are big-endian, so I assume that's always the case.

> where the hexadecimal value is interpreted as bits of the underlying binary representation

Unless stated otherwise, I would interpret this as big-endian too.

But I will add a link to the docs anyway.

exla/lib/exla/mlir/value.ex
jonatanklosko: Actually, this is an optional, so if the clause doesn't match it falls back to the optional implementation. I think it's all good, let me know otherwise!

jonatanklosko:
@polvalente updated!

polvalente (approved):

PR #1482, opened 2024-05-12
Add stack as a callback

### LINKED ISSUE #1441: Implement Nx.stack as a default callback
This will allow EXLA to JIT the whole operation, rather than copying the input tensors twice.

josevalim:
Generally speaking we want to reduce the number of callbacks but stack causes one additional copy of every tensor and it is used very frequently.

One option is to use optional callbacks, but concatenate/stack are the only operations that require a variable number of arguments, which is hard to include as part of the optional/evaluator system, so it its own callback.

Closes #1441.


josevalim:

PR #1489, opened 2024-05-22
feat: initial groundwork for using the IREE compiler/runtime

polvalente:
Before merging, we should discuss how to deal with the (painfully slow) compilation of the IREE compiler that's needed for using it as a C library.

This has pretty good Nx API coverage, but still not production ready.

The makefile should probably be conditioned to not use anything IREE if the runtime is not set to IREE specifically.


exla/c_src/iree_runtime/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
+cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
```
polvalente, line 1: note: This will most likely be extracted to a separate project so that we can build the compiler as a shared library and the runtime as a separate, more lightweight one (this is important for nerves usage).

So, probably, don't worry about adding cmake as a requirement to build exla.

polvalente:
https://github.com/elixir-nx/nx_iree brings the IREE functionality to Nx in a more decoupled way

PR #1495, opened 2024-05-23
feat: use func.return when returning from func.func

polvalente:


exla/lib/exla/mlir/value.ex
```diff
@@ -125,57 +125,71 @@ defmodule EXLA.MLIR.Value do
...
-        is_inf = op(func, "chlo.is_inf", [operand], result_types) |> one!()
-        is_finite = op(func, "stablehlo.is_finite", [operand], result_types) |> one!()
-        is_not_inf = bitwise_not(is_inf, typespec)
-        is_not_finite = bitwise_not(is_finite, typespec)
-        bitwise_and(is_not_inf, is_not_finite, typespec)
+    if out_typespec.type == {:u, 8} do
+      convert(result, out_typespec)
+    else
+      result
     end
```
jonatanklosko, line 193:
```suggestion
    if out_typespec.type == typespec.type do
      result
    else
      convert(result, out_typespec)
    end
```

exla/lib/exla/mlir/value.ex
```diff
@@ -125,57 +125,71 @@ defmodule EXLA.MLIR.Value do
...
+      end
 
-      true ->
-        result_types = typespecs_to_mlir_types([typespec])
-        op(func, "chlo.is_inf", [operand], result_types) |> one!()
+    if out_typespec.type == {:u, 8} do
+      convert(result, out_typespec)
+    else
+      result
     end
```
jonatanklosko, line 157:
```suggestion
    if out_typespec.type == typespec.type do
      result
    else
      convert(result, out_typespec)
    end
```

exla/lib/exla/defn.ex
```diff
@@ -307,7 +307,7 @@ defmodule EXLA.Defn do
     {res, cache} = recur_flatten(expr, state, new_cache(outfeed))
     outfeed = cache |> get_outfeed() |> Outfeed.close(function)
 
-    Value.return(function, res)
+    Value.func_return(function, res)
```
jonatanklosko, line 310: We should also change `to_root_computation` and here: https://github.com/elixir-nx/nx/blob/c7cdd847a0813897d6ddb7ba7b315a5ff8e73f1a/exla/test/support/exla_helpers.ex#L18

jonatanklosko (approved):
Just the comments and ship it!

exla/lib/exla/defn.ex
jonatanklosko: Updated!

jonatanklosko: Yeah, I changed both of these places. The errors were because we actually created the MLIR function with wrong return type in those cases, and it seems that using `stablehlo.return` omitted a validation that `func.return` does trigger :D

PR #1496, opened 2024-05-23
feat: export executable from EXLA

polvalente:


polvalente:
I think that with #1497 we don't really need this one. Closing for now.

PR #1497, opened 2024-05-23
feat: add EXLA.to_mlir_module/2

polvalente:
- feat: add EXLA.to_mlir_module/3
- doc: add doctest


exla/lib/exla/mlir/module.ex
```diff
@@ -71,6 +71,10 @@ defmodule EXLA.MLIR.Module do
     * `:use_spmd` - enables Single-Program Multiple-Data partioning.
       This is set to true if `:num_partitions` is more than one, otherwise is `false`.
 
+    * `:module_compilation` - either `:to_mlir` or `:to_pjrt`. The default is `:to_pjrt`.
```
polvalente, line 74: @josevalim not sure about the option naming here

josevalim: It is private, so it is fine!

exla/lib/exla.ex
```diff
@@ -355,6 +355,43 @@ defmodule EXLA do
...
+          %2 = stablehlo.add %0, %1 : tensor<f32>
+          stablehlo.return %2 : tensor<f32>
+        }
+      }
+      \"\"\"
+  """
+  def to_mlir_module(function, args, options \\ []) do
+    comp_fun = fn _key, callback ->
+      {:ok, {_xla_time, executable, _extra, _outfeed}} = callback.()
+      throw({:mlir_module, executable.ref})
```
josevalim, line 382: Gah, this is relying on internals of another module but I can't think of anything better for now, so ship it.

josevalim (approved):

exla/lib/exla/mlir/module.ex
jonatanklosko: If we return `%EXLA.MLIR.Module{}` in the other function, shouldn't the whole module be public?

josevalim: Good catch. We should return the string only, not the module struct.

exla/lib/exla.ex
```diff
@@ -355,6 +355,43 @@ defmodule EXLA do
...
+  Takes in a function, the templates variables and the compilation options
+  and returns the `EXLA.Executable` struct.
+
+  ## Examples
+
+      iex> fun = fn x, y -> Nx.add(Nx.sin(x), Nx.cos(y)) end
+      iex> args = [1.0, 2.0]
+      iex> module = EXLA.to_mlir_module(fun, args)
+      iex> EXLA.MLIR.Module.as_string(module)
+      ~c\"\"\"
```
jonatanklosko, line 368: Btw. let's make `Module.as_string` return an actual string, not charlist :D

polvalente: Done

exla/lib/exla/mlir/module.ex
polvalente: The public-facing function is now returning the binary directly

PR #1498, opened 2024-05-28
fix: force hessenberg range to be positive

polvalente:
- fix: force hessenberg range to be positive
- chore: make defn private


josevalim (approved):

josevalim:

PR #1499, opened 2024-06-01
Fix argmax/argmin behaviour with NaNs

jonatanklosko:
Currently the behaviour of argmax/min for NaNs is inconsistent with reduce_max/min:

```elixir
iex> Nx.reduce_max(Nx.tensor([1.0, :nan, 2.0]))
#Nx.Tensor<
  f32
  NaN
>
iex> Nx.argmax(Nx.tensor([1.0, :nan, 2.0]))
#Nx.Tensor<
  s64
  2
>
```

This is the case both for EXLA and BinaryBackend. I changed both such that argmax/argmin to point to a NaN in that case. Torchx already adheres to this behaviour.

I specifically changed the implementation such that MLIR aligns exactly with Jax.

### DIFF
--- a/exla/lib/exla/lib.ex
+++ b/exla/lib/exla/lib.ex
@@ -131,7 +131,7 @@ defmodule EXLA.Lib do
         :high -> &Value.greater/3
       end
 
-    # If lhs and rhs are not NaN and x = y, then pick index based on tie_break
+    # If lhs and rhs are equal (and not NaN), then pick index based on tie_break
     pick_lhs_idx =
       Value.bitwise_or(
         pick_lhs_value,
--- a/exla/lib/exla/lib.ex
+++ b/exla/lib/exla/lib.ex
@@ -34,7 +34,7 @@ defmodule EXLA.Lib do
   def argmax(builder, op, type, opts \\ [])
 
   def argmax(%Function{} = builder, %Value{} = op, type, opts) do
-    argmin_or_max(builder, op, false, type, opts)
+    argmin_or_max(builder, op, :max, type, opts)
   end
 
   @doc """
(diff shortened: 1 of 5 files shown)


exla/lib/exla/lib.ex
```diff
@@ -109,27 +109,42 @@ defmodule EXLA.Lib do
...
-
-        :high ->
-          eq? = Value.equal(lhs_value, rhs_value, pred_typespec)
-          id = Value.max(lhs_index, rhs_index, idx_typespec)
-          Value.select(eq?, id, arg_max, idx_typespec)
+        :low -> &Value.less/3
+        :high -> &Value.greater/3
       end
 
+    # If lhs and rhs are not NaN and x = y, then pick index based on tie_break
```
jonatanklosko, line 134:
```suggestion
    # If lhs and rhs are equal (and not NaN), then pick index based on tie_break
```

josevalim:
Note there is this commit: 3d1b594395184e00c1c281b8a813e13fb5c966ac

Does it make sense to have NaN be both the min and the max? Also, can you please run scholar tests on this branch? It does use NaN for sorting quite extensibly.

jonatanklosko:
> Note there is this commit: https://github.com/elixir-nx/nx/commit/3d1b594395184e00c1c281b8a813e13fb5c966ac

I think sorting is different. Since NaNs are not really comparable to anything, they don't have a correct position, so throwing all of them to the end is reasonable.

> Does it make sense to have NaN be both the min and the max?

I think so, one way to view it is that in case of NaNs, there is no difference between min and max, since NaNs are not comparable. It's basically the same as trying to take a sum of the tensor, one NaN makes the whole thing NaN.

> Also, can you please run scholar tests on this branch?

All tests pass.

josevalim (approved):

Issue #1491, opened 2024-05-23
Default integers to 32-bit precision


jonatanklosko:
This way tensors require less memory by default and people can opt-in if they work with really big numbers. We already default to 32-bit precision for floats. Sidenote: Jax does that too and even guards 64-bit precision behind a configuration flag.

PR #1504, opened 2024-06-05
(Experimental) Integrate Metal PjRt plugin

jonatanklosko:
This integrates the PjRt plugin from the [`jax-metal`](https://developer.apple.com/metal/jax/) for running on the Apple GPU. To test it, one can set `client: :mps` on EXLA backend/compiler. Since the plugin is loaded as a separate dynamic library, it can be tested without any changes to XLA (just make sure to remove the `cache/` directory and the EXLA binary cached globally).

Certain computations can already be run, but the plugin is still very much incomplete. This PR is a room for experimentation and is meant to track the plugin progress. I reported a number of issues upstream, comments in the code point to those. In a few places I applied workarounds as temporary solutions or just to avoid VM crashes (segfaults), those are marked with a TODO.

### Issues

For tracking purposes, here is a list of the Metal plugin issues reported upstream:

**Crucial**

* https://github.com/google/jax/issues/21552
* https://github.com/google/jax/issues/21601
* https://github.com/google/jax/issues/21547
* https://github.com/google/jax/issues/21590

**Not implemented**

* https://github.com/google/jax/issues/21387
* https://github.com/google/jax/issues/21384
* https://github.com/google/jax/issues/21554
* https://github.com/google/jax/issues/21389
* Support for complex https://github.com/google/jax/issues/16416
* Support for float-64 https://github.com/google/jax/issues/20938

**Edge cases**

* https://github.com/google/jax/issues/21577
* https://github.com/google/jax/issues/21397
* https://github.com/google/jax/issues/21393
* https://github.com/google/jax/issues/21392

All issues: [link](https://github.com/google/jax/issues?page=1&q=is%3Aissue+is%3Aopen+label%3A%22Apple+GPU+%28Metal%29+plugin%22).

---

*Note: this PR is against the jk-s32 branch, which changes the default integer precision to 32 bits. This is a planned change (#1491), but it's not integrated yet to avoid conflicts with other work in progress.*

### DIFF
--- a/exla/mix.exs
+++ b/exla/mix.exs
@@ -51,9 +51,10 @@ defmodule EXLA.MixProject do
           cuda: [platform: :cuda],
           rocm: [platform: :rocm],
           tpu: [platform: :tpu],
+          mps: [platform: :mps],
           host: [platform: :host]
         ],
-        preferred_clients: [:cuda, :rocm, :tpu, :host]
+        preferred_clients: [:cuda, :rocm, :tpu, :mps, :host]
       ]
     ]
   end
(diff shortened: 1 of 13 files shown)


exla/mix.exs
```diff
@@ -128,11 +129,31 @@ defmodule EXLA.MixProject do
...
 
+  defp download_metal_plugin!(xla_extension_path) do
+    plugin_path = Path.join(xla_extension_path, "lib/pjrt_plugin_metal.dylib")
+
+    wheel_url =
+      "https://files.pythonhosted.org/packages/80/af/ed482a421a868726e7ca3f51ac19b0c9a8e37f33f54413312c37e9056acc/jax_metal-0.1.0-py3-none-macosx_11_0_arm64.whl"
+
+    wheel_path = Path.join(xla_extension_path, "jax_metal.whl")
+
+    {_, 0} = System.shell("wget --output-document=#{wheel_path} #{wheel_url}")
```
NickTaylor-, line 148: `wget` isn't available in macOS by default, and the default PATH in livebook doesn't include Homebrew's /bin (that might be my own setup), so it would be more portable if this was `curl`.

jonatanklosko: Yes, this is just a proof of concept, this would be moved and downloaded as part of elixir-nx/xla :)

jonatanklosko:
(I pushed an update to reduce the diff, in case someone wants to look into another PjRt plugin integration)

PR #1507, opened 2024-06-06
Add initial FP8 support

seanmor5:
This PR adds support for an emulated FP8 in Nx's binary backend and adds support for FP8 operations from XLA. The FP8 type in EXLA is just there: `FP8E5M2` type. They have 3 of these FP8 types that we can configure if we want. 

In the binary backend the logic is in `shared.ex`. We have a function for interpreting erlang float's to a binary representation, and then we convert the binary chunks to f16 chunks and interpret those to f64.

### DIFF
--- a/exla/lib/exla/mlir/value.ex
+++ b/exla/lib/exla/mlir/value.ex
@@ -880,6 +880,7 @@ defmodule EXLA.MLIR.Value do
   defp type_number({:pred, 8}), do: "i1"
   defp type_number({:s, width}), do: "i#{width}"
   defp type_number({:u, width}), do: "ui#{width}"
+  defp type_number({:f, 8}), do: "f8E5M2"
   defp type_number({:f, width}), do: "f#{width}"
   defp type_number({:bf, width}), do: "bf#{width}"
   defp type_number({:c, 64}), do: "complex<f32>"
@@ -926,6 +927,7 @@ defmodule EXLA.MLIR.Value do
         :nan -> type |> Nx.Type.nan_binary() |> native_to_big()
         :infinity -> type |> Nx.Type.infinity_binary() |> native_to_big()
         :neg_infinity -> type |> Nx.Type.neg_infinity_binary() |> native_to_big()
+        value when size == 8 -> Nx.Shared.erlang_float_to_fp8(value)
         value -> <<value::float-size(size)-big>>
       end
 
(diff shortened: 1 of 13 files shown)


exla/lib/exla/mlir/value.ex
```diff
@@ -874,6 +875,7 @@ defmodule EXLA.MLIR.Value do
         :nan -> type |> Nx.Type.nan_binary() |> native_to_big()
         :infinity -> type |> Nx.Type.infinity_binary() |> native_to_big()
         :neg_infinity -> type |> Nx.Type.neg_infinity_binary() |> native_to_big()
+        value when size == 8 -> Nx.Shared.erlang_float_to_fp8(value)
```
josevalim, line 878: We should call `Nx.Shared` from other libs, it is private API. Sorry :(

josevalim: Wouldn't this conversion depend anyway on which type of f8 we pick?

nx/lib/nx/shared.ex
```diff
@@ -225,6 +242,55 @@ defmodule Nx.Shared do
...
+  precision float supported by Erlang.
+  """
+  def fp8_to_erlang_float(<<sign::size(1), exponent_fp8::size(4), mantissa_fp8::size(3)>>) do
+    exponent_fp16 =
+      if exponent_fp8 == 0,
+        do: 0,
+        else: exponent_fp8 - 7 + 15
+
+    mantissa_fp16 = mantissa_fp8 <<< 7
+    <<x::float-16-big>> = <<sign::size(1), exponent_fp16::size(5), mantissa_fp16::size(10)>>
```
josevalim, line 290: We don't need to convert to a binary, do we? We could probably write the formula directly:

  V=(−1)^S×2^(E−bias)×(1+M/8​)

Or something like that?

nx/lib/nx.ex
```diff
@@ -612,6 +612,15 @@ defmodule Nx do
...
 
+  Certain backends and compilers support 8-bit floats. On the binary
+  backend this behavior is emulated:
+
+      iex> Nx.tensor([1, 2, 3], type: :f8)
+      #Nx.Tensor<
+        f8[3]
+        [1.0, 2.0, 3.0]
+      >
+
```
josevalim, line 623:
```suggestion
```

nx/lib/nx.ex
```diff
@@ -51,7 +51,7 @@ defmodule Nx do
 
     * unsigned integers (`u8`, `u16`, `u32`, `u64`)
     * signed integers (`s8`, `s16`, `s32`, `s64`)
-    * floats (`f16`, `f32`, `f64`)
+    * floats (`f8` `f16`, `f32`, `f64`)
```
josevalim, line 54:
```suggestion
    * floats (`f8`, `f16`, `f32`, `f64`)
```

PR #1516, opened 2024-07-26
Fix min / max value for 16 bit signed type

preciz:
Fixes min / max value for 16 bit signed type, plus a small typo fix.

https://github.com/elixir-nx/nx/blob/2f44f4a9bf950ef4cbeb719785d9d9d881690bb9/exla/test/exla/defn/expr_test.exs#L1108-L1110

Thank you for reviewing.


josevalim:

PR #1524, opened 2024-09-03
Default integers to 32-bit precision

jonatanklosko:
Nx and EXLA passes, but there are Torchx segfaults that I need to debug.

### DIFF
--- a/exla/lib/exla.ex
+++ b/exla/lib/exla.ex
@@ -220,7 +220,7 @@ defmodule EXLA do
 
       iex> EXLA.jit(&Nx.add(&1, &1)).(Nx.tensor([1, 2, 3]))
       #Nx.Tensor<
-        s64[3]
+        s32[3]
         [2, 4, 6]
       >
 
@@ -265,7 +265,7 @@ defmodule EXLA do
 
       iex> EXLA.jit_apply(&Nx.add(&1, &1), [Nx.tensor([1, 2, 3])])
       #Nx.Tensor<
-        s64[3]
+        s32[3]
         [2, 4, 6]
       >
 
@@ -278,10 +278,10 @@ defmodule EXLA do
   @doc """
   A shortcut for `Nx.Defn.compile/3` with the EXLA compiler.
 
-      iex> fun = EXLA.compile(&Nx.add(&1, &1), [Nx.template({3}, {:s, 64})])
+      iex> fun = EXLA.compile(&Nx.add(&1, &1), [Nx.template({3}, {:s, 32})])
       iex> fun.(Nx.tensor([1, 2, 3]))
       #Nx.Tensor<
-        s64[3]
+        s32[3]
         [2, 4, 6]
       >
 
@@ -328,7 +328,7 @@ defmodule EXLA do
 
   Now let's invoke it:
 
-      stream = EXLA.stream(&Streamed.sum/2, [Nx.template({}, {:s, 64}), 0])
+      stream = EXLA.stream(&Streamed.sum/2, [Nx.template({}, {:s, 32}), 0])
 
       for i <- 1..5 do
         Nx.Stream.send(stream, i)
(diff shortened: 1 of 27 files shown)


josevalim:
Our torchx is also old. Those may be fixed if we update it.

jonatanklosko:
One VM crash was related to specific dot product with u32:

```elixir
u = Nx.tensor([[[1]], [[2]]])
v = Nx.tensor([[[3]], [[4]]])
Nx.dot(u, [2], [0], v, [2], [0])
```

I changed the default libtorch version from 2.0.0 to 2.1.0, and it's fixed.

The other crashes I fixed by casing in appropriate places.

torchx/lib/torchx/backend.ex
```diff
@@ -799,9 +810,6 @@ defmodule Torchx.Backend do
     {Nx.as_type(left, type), Nx.as_type(right, type)}
   end
 
-  defp maybe_broadcast_bin_args(_out_shape, %{shape: {}} = l, r), do: {from_nx(l), from_nx(r)}
-  defp maybe_broadcast_bin_args(_out_shape, l, %{shape: {}} = r), do: {from_nx(l), from_nx(r)}
```
jonatanklosko, line 803: @polvalente after upgrading to libtroch 2.1.0, I saw a bunch of warnings like this:

```
[W Resize.cpp:35] Warning: An output with one or more elements was resized since it had shape [], which does not match the required output shape [3]. This behavior is deprecated, and in a future PyTorch release outputs will not be resized unless they have zero elements. You can explicitly reuse an out tensor t by resizing it, inplace, to zero elements with t.resize_(0). (function _resize_output_check)
```

An example where this happens is:

```elixir
Nx.logical_and(Nx.tensor(1), Nx.tensor([1, 2, 3]))
```

The warning only happens if the first operand is a scalar. Doing the actual broadcasting removes the warning.

jonatanklosko:
I noticed that `Nx.tensor(0xFFFFFFFF)` now crashes the VM with:

```
libc++abi: terminating due to uncaught exception of type std::runtime_error: value cannot be converted to type int without overflow
```

It was already the case before with `Nx.s32(0xFFFFFFFF)`, it's just that now it's more likely to happen by default.

Perhaps there's a way to catch it and raise an elixir error instead, but that's not related to this PR.

josevalim (approved):

jonatanklosko:
@polvalente feel free to merge, if you are ok with the torchx changes :)

torchx/lib/torchx/backend.ex
polvalente: We do the broadcasting explicitly in StableHLO too, so that's fine I think.

torchx/test/torchx/random_test.exs
```diff
@@ -65,7 +65,7 @@ defmodule Torchx.Nx.RandomTest do
       # Output does not match Nx because of the sign of the remainder.
       distribution_case(:randint_split,
         args: [0, 10, [shape: {5}]],
-        expected: Nx.tensor([3, 2, 6, 0, 0], type: :s64)
+        expected: Nx.tensor([1, 1, 4, 1, 9], type: :s64)
```
polvalente, line 68: I didn't really understand why these values changed, but that's ok.

polvalente (approved):

torchx/test/torchx/random_test.exs
jonatanklosko: It's because in `args: [0, 10, [shape: {5}]]` passed to defn the numbers `0` and `10` are now passed as s32 rather than s64. We could maintain the behaviour with ``args: [Nx.s64(0), Nx.s64(10), [shape: {5}]]``, but changing the test is equally fine.

exla/lib/exla.ex
```diff
@@ -220,7 +220,7 @@ defmodule EXLA do
```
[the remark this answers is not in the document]
jonatanklosko: I think those are unrelated, we just want to pass fixed integers as XLA inputs. We could actually make those unsigned (and change to uint on the c++ side), since all of those are non-negative sizes. Doesn't matter much, since those are just a few values, your call!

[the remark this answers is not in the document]
jonatanklosko: I opened a PR to change to u64 https://github.com/elixir-nx/nx/pull/1526.

PR #1525, opened 2024-09-03
refactor: Nx.Random.key as deftransform

polvalente:


josevalim (approved):

nx/lib/nx/random.ex
```diff
@@ -80,22 +80,18 @@ defmodule Nx.Random do
...
-      iex> Nx.Random.key(Nx.u64(999999999999))
+      iex> Nx.Random.key(999999999999)
       #Nx.Tensor<
         u32[2]
         [232, 3567587327]
       >
   """
-  defn key(seed) do
+  deftransform key(seed) do
     seed = Nx.as_type(seed, :u64)
```
jonatanklosko, line 90: People may pass seed to defn and only then call `Nx.Random.key/1` (we do this in Bumblebee). But instead of documenting it, we can actually validate this:

```suggestion
    seed =
      case seed do
        seed when is_integer(seed) ->
          Nx.u64(seed)

        %Nx.Tensor{} = seed when seed.type == {:u, 64} ->
          seed
        
        other ->
          raise ArgumentError, "expected seed to be an integer or :u64 tensor, got: #{inspect(other)}"
      end

```

@polvalente @josevalim wdyt?

jonatanklosko:
Good call!

nx/lib/nx/random.ex
polvalente: I think this works, although it might make the function a bit less easy to use.

jonatanklosko: @polvalente can you think of an example? I think most people would call it with an integer beforehand. If someone is passing seed to the defn, they probably use a default like `:erlang.system_time()`, which is a big number, so ideally we want to make sure they pass :u64.

polvalente: Not really. Maybe if the key call is deeply nested in the code, but that's more of an edge case I think

jonatanklosko: @josevalim the thing is that the key has 64 bits, so if the user always crops the seed to 32 bits, multiple seeds may result in the same key, so it reduces the entropy.

polvalente: I'll commit this for the correctness. I think this would only actually raise in edge cases anyway.

josevalim: @jonatanklosko I was thinking specifically in the case of timestamps. The low level digits are always increasing while the higher ones are static, so we would be fine.

nx/lib/nx/random.ex
```diff
@@ -80,22 +80,33 @@ defmodule Nx.Random do
...
+    seed =
+      case seed do
+        seed when is_integer(seed) ->
+          Nx.u64(seed)
 
+        %Nx.Tensor{} = seed when seed.type == {:u, 64} ->
+          seed
+
+        %Nx.Tensor{} = seed when seed.type == {:s, 64} ->
+          Nx.bitcast(seed, {:u, 64})
```
josevalim, line 99: Do we really need this clause? No-one should be passing s64 now.

I am sorry, but I don't understand why we don't keep the `seed = Nx.as_type(seed, :u64)` version.

nx/lib/nx/random.ex
jonatanklosko: Fair. But I would have the check anyway, if it turns out annoying, we can always disable and perhaps add some docs instead.

nx/lib/nx/random.ex
josevalim: for me, i would check for integer and convert it, otherwise cast to u64 inside the `key_n` function:

nx/lib/nx/random.ex
```diff
@@ -80,22 +80,33 @@ defmodule Nx.Random do
...
+
+        other ->
+          raise ArgumentError,
+                "expected seed to be an integer, u64 tensor or s64 tensor, got: #{inspect(other)}"
+      end
+
+    key_n(seed)
+  end
+
+  defnp key_n(seed) do
```
josevalim, line 109:
```suggestion
  deftransform key(seed) do
    if is_integer(seed) do    
      key_n(Nx.u64(seed))
    else
      key_n(seed)
    end
  end

  defnp key_n(seed) do
    seed = Nx.as_type(seed, :u64)
```

jonatanklosko: I think requiring :u64 is fine. The point about the validation is to avoid users accidentally lowering the seed space by passing s32 tensors inside defn.

josevalim: I see though that you want to avoid an integer passed as argument, which will be downcast to s32, from becoming a key. Yeah, ignore me.

jonatanklosko: FWIW jax does what you said :) ([ref](https://github.com/google/jax/blob/f90558d317936281d3edae74e05027939a106841/jax/_src/prng.py#L529-L535))

nx/lib/nx/random.ex
josevalim: Or do we also want to do bitcasts for f64?

jonatanklosko: We do logical and with another `:u64`, so that's good.

nx/lib/nx/random.ex
josevalim: I ran Axon suite and it all worked. So I think we will be fine.

josevalim (approved):

nx/lib/nx/random.ex
polvalente: I think this approach is alot simpler too

nx/lib/nx/random.ex
polvalente: I added the `s64` case because there are explicit tests for this, and it has the same entropy as u64. Not sure about floats.

jonatanklosko: Oh I misread the question. I think it's fair to assume seed is always an integer.

No preference on `{:s, 64}`, if there are tests, then yeah, bitcast sounds good.

jonatanklosko (approved):

PR #1526, opened 2024-09-04
Pass custom call sizes as unsigned integers

jonatanklosko:
🐈

### DIFF
--- a/exla/lib/exla/mlir/value.ex
+++ b/exla/lib/exla/mlir/value.ex
@@ -720,14 +720,14 @@ defmodule EXLA.MLIR.Value do
     eigenvecs_dims = Tuple.to_list(eigenvecs_shape)
     eigenvals_dims = Tuple.to_list(eigenvals_shape)
 
-    dim_sizes = constant(func, dim_sizes, Typespec.tensor({:s, 64}, {length(dim_sizes)}))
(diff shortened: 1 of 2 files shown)


exla/lib/exla/mlir/value.ex
```diff
@@ -720,14 +720,14 @@ defmodule EXLA.MLIR.Value do
     eigenvecs_dims = Tuple.to_list(eigenvecs_shape)
     eigenvals_dims = Tuple.to_list(eigenvals_shape)
 
-    dim_sizes = constant(func, dim_sizes, Typespec.tensor({:s, 64}, {length(dim_sizes)}))
-    operand_dims = constant(func, operand_dims, Typespec.tensor({:s, 64}, {length(operand_dims)}))
+    dim_sizes = constant(func, dim_sizes, Typespec.tensor({:u, 64}, {length(dim_sizes)}))
+    operand_dims = constant(func, operand_dims, Typespec.tensor({:u, 64}, {length(operand_dims)}))
```
josevalim, line 724: should it be u32 instead? since we know default to 32?

jonatanklosko: It's unrelated to the default. Here we basically pass parameters to XLA. Though, I think u32 would be enough, since we don't expect tensors with axis size big enough to overflow u32. At the same time, it's a few numbers, so it doesn't really make a difference memory-wise. Whichever you prefer!

josevalim (approved):

PR #1530, opened 2024-09-05
fix: add support for defn compilation in EXLA.to_mlir_module

polvalente:
`EXLA.to_mlir_module` wasn't working when being called from within a Defn compiler.
The new option adds the control needed for this to work.

### DIFF
--- a/exla/lib/exla.ex
+++ b/exla/lib/exla.ex
@@ -383,12 +383,14 @@ defmodule EXLA do
       throw({:mlir_module, executable.ref, used_inputs, outputs})
     end
 
-    opts = [
-      {EXLA, {&EXLA.Defn.LockedCache.run/2, comp_fun}},
-      {:module_compilation, :to_mlir} | options
-    ]
+    opts =
+      Keyword.merge(options, [
+        {EXLA, {&EXLA.Defn.LockedCache.run/2, comp_fun}},
+        module_compilation: :to_mlir,
+        compiler: EXLA
+      ])
 
-    compile(function, args, opts)
+    EXLA.Defn.__compile__(function, args, function, opts)
   catch
     {:mlir_module, ref, used_inputs, output_container} ->
       %{
@@ -397,7 +399,6 @@ defmodule EXLA do
         mlir_module: EXLA.MLIR.Module.as_string(%EXLA.MLIR.Module{ref: ref})
       }
   end
-
 
   @doc """
   Checks if the compilation of function with args is cached.
--- a/exla/lib/exla.ex
+++ b/exla/lib/exla.ex
@@ -360,6 +360,12 @@ defmodule EXLA do
   Takes in a function, the argument templates and the compilation
   options and returns the textual representation of the MLIR module.
 
+  ## Options
+
+    * `:nested_defn_compilation` - a boolean that indicates whether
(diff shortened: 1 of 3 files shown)


exla/test/exla_test.exs
```diff
@@ -17,4 +17,56 @@ defmodule EXLATest do
...
+                   func.func public @main(%arg0: tensor<i32>, %arg1: tensor<i32>) -> tensor<i32> {
+                     %0 = stablehlo.add %arg0, %arg1 : tensor<i32>
+                     return %0 : tensor<i32>
+                   }
+                 }
+                 """
+
+          assert Nx.compatible?(container, Nx.template({}, :s32))
+
+          assert used_inputs == %{0 => nil, 1 => nil}
```
jonatanklosko, line 68: We should returns something more meaningful, like a list or mapset?

exla/lib/exla.ex
```diff
@@ -360,11 +360,18 @@ defmodule EXLA do
   Takes in a function, the argument templates and the compilation
   options and returns the textual representation of the MLIR module.
 
+  ## Options
+
+    * `:nested_defn_compilation` - a boolean that indicates whether
```
jonatanklosko, line 365: Could we check `Nx.Defn.Compiler.defn?()` instead?

polvalente: I tried, but that only gets set inside the runtime fun call, and that's after what I need

exla/test/exla_test.exs
polvalente: Could be MapSet of Map.keys. WDYT?

exla/lib/exla.ex
```diff
@@ -378,19 +385,31 @@ defmodule EXLA do
...
+      EXLA.Defn.__compile__(function, args, function, opts)
+    else
+      Nx.Defn.compile(function, args, opts)
+    end
   catch
-    {:mlir_module, ref} ->
-      EXLA.MLIR.Module.as_string(%EXLA.MLIR.Module{ref: ref})
+    {:mlir_module, ref, used_inputs, output_container} ->
+      %{
+        used_inputs: used_inputs,
```
josevalim, line 409: Definitely a MapSet. :)

exla/lib/exla.ex
josevalim: What about `Nx.Defn.Compiler.current()`?

exla/lib/exla/defn.ex
```diff
@@ -466,7 +467,9 @@ defmodule EXLA.Defn do
                   )
                 end)
 
-              {:ok, {xla_time, executable, inputs_and_typespecs, %{outfeed | infeeds: []}}}
+              {:ok,
+               {xla_time, executable, {inputs_and_typespecs, used_inputs, outputs},
+                %{outfeed | infeeds: []}}}
```
josevalim, line 472: Adding `outputs` and `used_inputs` here means those values will be cached twice. I think for `to_mlir_module`, we don't want to use the cache funs anymore. I would instead check inside `__compile__` if `module_compilation` is set to `:mlir` and have it throw there the exact data that you need.

Alternatively you can create a new EXLA.Defn.MLIRCompiler, that calls `EXLA.Defn`. We would need to make some APIs public.

PR #1535, opened 2024-09-15
fix: broadcast vectors for grad calculation

### LINKED ISSUE #1533: Nx.Defn.Grad returns sum of gradient over vectorized axis?
Thanks for making Nx!

``` elixir
defmodule Foo do
  import Nx.Defn
  defn f(x, y) do
    x + y
  end

  defn f_and_grad(x, y) do
  end
end

Foo.f_and_grad(x, 1)
```

This evaluates to:

```
{#Nx.Tensor<
   vectorized[bar: 2]
   s64
   [1, 2]
 >,
 #Nx.Tensor<
   f32
   2.0
 >}
```

Thanks!

polvalente:
closes #1533

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -5420,9 +5420,13 @@ defmodule Nx do
       {_, [], 0} ->
         fun.(left, right)
 
-      {[devec_left, devec_right], canonical_vectorized_axes, _offset} ->
-        devec_left
-        |> fun.(devec_right)
+      {[devec_left, devec_right], canonical_vectorized_axes, offset} ->
+        leading_names = Keyword.keys(canonical_vectorized_axes)
+        l = %{devec_left | names: leading_names ++ Enum.drop(devec_left.names, offset)}
+        r = %{devec_right | names: leading_names ++ Enum.drop(devec_right.names, offset)}
+
+        l
+        |> fun.(r)
         |> vectorize(canonical_vectorized_axes)
     end
   end
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -4906,18 +4906,11 @@ defmodule Nx do
 
   def devectorize(%T{shape: shape, names: names, vectorized_axes: vectorized_axes} = tensor, opts)
       when vectorized_axes != [] do
-    opts = keyword!(opts, keep_names: true, drop_inner_names: false)
+    opts = keyword!(opts, keep_names: true)
     {vectorized_names, vectorized_sizes} = Enum.unzip(vectorized_axes)
 
     output_shape_l = vectorized_sizes ++ Tuple.to_list(shape)
     output_shape = List.to_tuple(output_shape_l)
-
-    names =
-      if opts[:drop_inner_names] do
-        Enum.map(names, fn _ -> nil end)
-      else
-        names
-      end
 
     output_names =
       if opts[:keep_names] do
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -4906,11 +4906,18 @@ defmodule Nx do
 
   def devectorize(%T{shape: shape, names: names, vectorized_axes: vectorized_axes} = tensor, opts)
       when vectorized_axes != [] do
-    opts = keyword!(opts, keep_names: true)
+    opts = keyword!(opts, keep_names: true, drop_inner_names: false)
     {vectorized_names, vectorized_sizes} = Enum.unzip(vectorized_axes)
 
     output_shape_l = vectorized_sizes ++ Tuple.to_list(shape)
     output_shape = List.to_tuple(output_shape_l)
+
+    names =
+      if opts[:drop_inner_names] do
+        Enum.map(names, fn _ -> nil end)
+      else
+        names
+      end
 
     output_names =
       if opts[:keep_names] do
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -94,13 +94,13 @@ defmodule Nx.Defn.Expr do
   def metadata(expr, metadata) when is_map(metadata) do
     case to_container_expr(expr) do
       %{data: %{context: context}} = res ->
-        expr(res, context, :metadata, [expr, metadata])
+        expr(res, context, :metadata, [Nx.devectorize(expr), metadata])
 
       t when is_tuple(t) ->
         context = elem(t, 0).data.context
 
         tuple(
-          expr(tuple_out(tuple_size(t)), context, :metadata, [expr, metadata]),
+          expr(tuple_out(tuple_size(t)), context, :metadata, [Nx.devectorize(expr), metadata]),
           Tuple.to_list(t)
         )
     end
@@ -1665,11 +1665,11 @@ defmodule Nx.Defn.Expr do
 
   defp counter_to_name(counter), do: [?a + counter]
 
-  defp to_type_shape(%{vectorized_axes: vectorized_axes, type: type, shape: shape}) do
-    axes =
-      Keyword.values(vectorized_axes) ++ Tuple.to_list(shape)
-
-    brackets = Enum.map(axes, &[?[, Integer.to_string(&1), ?]])
+  defp to_type_shape(%{vectorized_axes: [], type: type, shape: shape}) do
+    brackets =
+      shape
+      |> Tuple.to_list()
+      |> Enum.map(&[?[, Integer.to_string(&1), ?]])
 
     IO.iodata_to_binary([Nx.Type.to_string(type) | brackets])
   end
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -1394,6 +1394,11 @@ defmodule Nx.Defn.Expr do
 
   ## Constant helpers and related optimizations
 
+  defp constant(%{vectorized_axes: [_ | _]} = out, number) do
+    out = %{out | names: Enum.map(out.names, fn _ -> nil end)}
+    tensor(Nx.fill(out, number, type: out.type))
+  end
+
   defp constant(%{shape: shape, type: type} = out, number) do
     number =
       cond do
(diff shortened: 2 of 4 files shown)


nx/lib/nx/defn/grad.ex
```diff
@@ -5,12 +5,19 @@ defmodule Nx.Defn.Grad do
...
+      Composite.traverse(to_grad, {%{}, broadcasted_nodes}, fn _to_grad,
+                                                               {ids, [broadcasted_node | nodes]} ->
         to_grad =
-          Expr.metadata(to_grad, %{__MODULE__ => :to_grad})
+          Expr.metadata(broadcasted_node, %{__MODULE__ => :to_grad})
 
-        {to_grad, Map.put(ids, to_grad.data.id, :stop)}
+        ids = Map.put(ids, to_grad.data.id, :stop)
+        {to_grad, {ids, nodes}}
       end)
```
polvalente, line 21: This isn't actually the proper fix. This solves the issue only when there is no closure for a vectorized tensor.
fn y -> Nx.add(x, y) end fails if y is unvectorized but x is vectorized

nx/lib/nx/defn/grad.ex
```diff
@@ -19,36 +18,32 @@ defmodule Nx.Defn.Grad do
...
     {expr, graded}
   end
 
   defp constant(float, shape) do
-    shape = Nx.shape(shape)
-    names = List.duplicate(nil, tuple_size(shape))
-    Expr.constant(%T{shape: shape, type: {:f, 32}, names: names}, float, [])
+    case shape do
+      %T{vectorized_axes: [_ | _]} = t ->
+        Expr.tensor(Nx.fill(t, float, type: :f32))
```
josevalim, line 40: We should probably get rid of the names here too.

I also wonder if should move the check for vectorized_axes to `constant`. Today if someone passes vectorized_axes, Expr.constant is broken. So maybe we should create a tensor if a vectorized axes is given to tensor?

nx/lib/nx/defn/grad.ex
```diff
@@ -338,6 +333,8 @@ defmodule Nx.Defn.Grad do
   @verify_grad Application.compile_env(:nx, :verify_grad, false)
 
   defp update_grads(op, args, ans, g, _to_grad_ids, grads) do
+    args = revectorize_args(args, ans)
```
josevalim, line 336: I would prefer to not revectorized everything on every operation. Is there any chance we could do in `broadcast` only?

polvalente: `[unbroadcast(x, Nx.multiply(g, y), ans), unbroadcast(y, Nx.multiply(g, x), ans)]`

Lines like this one make it so that `g` is vectorized and `y` is unvectorized but has axes with the same name, so things break there.

nx/lib/nx/defn/grad.ex
polvalente: Done!

nx/lib/nx/defn/expr.ex
```diff
@@ -1394,6 +1394,11 @@ defmodule Nx.Defn.Expr do
 
   ## Constant helpers and related optimizations
 
+  defp constant(%{vectorized_axes: [_ | _]} = out, number) do
+    out = %{out | names: Enum.map(out.names, fn _ -> nil end)}
```
josevalim, line 1398: I don't think this part should be done here, we should preserve the names. Sorry for the confusion.

nx/lib/nx/defn/grad.ex
```diff
@@ -1343,9 +1334,77 @@ defmodule Nx.Defn.Grad do
 
   ## General helpers
 
-  defp unbroadcast(%{shape: shape} = x, res, %{shape: shape}), do: {x, res}
+  defp revectorize_args(args, ans) do
```
josevalim, line 1337: Let's only apply this if args has more than one element and there are vectorized axes.


Also please test `x * sin(y)` where y is vectorized.

josevalim (approved):

nx/lib/nx.ex
```diff
@@ -4906,12 +4906,19 @@ defmodule Nx do
 
   def devectorize(%T{shape: shape, names: names, vectorized_axes: vectorized_axes} = tensor, opts)
       when vectorized_axes != [] do
-    opts = keyword!(opts, keep_names: true)
+    opts = keyword!(opts, keep_names: true, drop_inner_names: false)
```
josevalim, line 4909: Revert.

nx/lib/nx/defn/expr.ex
```diff
@@ -94,13 +94,13 @@ defmodule Nx.Defn.Expr do
...
       %{data: %{context: context}} = res ->
-        expr(res, context, :metadata, [Nx.devectorize(expr), metadata])
+        expr(res, context, :metadata, [expr, metadata])
 
       t when is_tuple(t) ->
         context = elem(t, 0).data.context
 
         tuple(
-          expr(tuple_out(tuple_size(t)), context, :metadata, [Nx.devectorize(expr), metadata]),
+          expr(tuple_out(tuple_size(t)), context, :metadata, [expr, metadata]),
```
josevalim, line 103: Revert. devectorize with keep_names.

nx/lib/nx/defn/expr.ex
```diff
@@ -1661,11 +1665,11 @@ defmodule Nx.Defn.Expr do
...
-  defp to_type_shape(%{type: type, shape: shape}) do
-    brackets =
-      shape
-      |> Tuple.to_list()
-      |> Enum.map(&[?[, Integer.to_string(&1), ?]])
+  defp to_type_shape(%{vectorized_axes: vectorized_axes, type: type, shape: shape}) do
+    axes =
+      Keyword.values(vectorized_axes) ++ Tuple.to_list(shape)
+
+    brackets = Enum.map(axes, &[?[, Integer.to_string(&1), ?]])
```
josevalim, line 1672: Should we revert? 🤔

nx/lib/nx/defn/grad.ex
```diff
@@ -93,47 +83,93 @@ defmodule Nx.Defn.Grad do
...
     case nodes do
-      %{^id => _} -> {parents, nodes}
-      %{} -> parents_args(op, t, id, {parents, Map.put(nodes, id, t)})
+      %{^id => _} ->
+        {parents, nodes}
+
+      %{} ->
+        parent_vectorized_axes = compute_arg_vectorized_axes(t, vectorized_axes)
+
+        nodes = Map.put(nodes, id, {Nx.devectorize(t, keep_names: true), parent_vectorized_axes})
```
josevalim, line 101: We should not have anything vectorized here.

nx/lib/nx/defn/grad.ex
```diff
@@ -93,47 +83,93 @@ defmodule Nx.Defn.Grad do
...
     reduce_args(op, t, acc, fn arg, {parents, nodes} ->
       if arg.data.op in @constants do
         {parents, nodes}
       else
+        arg_vectorized_axes = compute_arg_vectorized_axes(t, parent_vectorized_axes)
         parents = Map.update(parents, arg.data.id, [parent_id], &[parent_id | &1])
-        recur_parents_tree(arg, {parents, nodes})
+
+        recur_parents_tree(
+          Nx.devectorize(arg, keep_names: true),
```
josevalim, line 169: We should not need this either.

nx/lib/nx/defn/grad.ex
```diff
@@ -93,47 +83,93 @@ defmodule Nx.Defn.Grad do
...
-  defp recur_parents_tree(%T{data: %Expr{id: id, op: op}} = t, {parents, nodes}) do
+  defp recur_parents_tree(%T{data: %Expr{id: id, op: op}} = t, {parents, nodes}, vectorized_axes) do
     case nodes do
-      %{^id => _} -> {parents, nodes}
-      %{} -> parents_args(op, t, id, {parents, Map.put(nodes, id, t)})
+      %{^id => _} ->
+        {parents, nodes}
+
+      %{} ->
+        parent_vectorized_axes = compute_arg_vectorized_axes(t, vectorized_axes)
```
josevalim, line 99: This may not be necessary.

nx/lib/nx/defn/grad.ex
```diff
@@ -212,6 +265,34 @@ defmodule Nx.Defn.Grad do
...
+      end)
+
+    vectorized_axes ++ Enum.reverse(reversed_inner_axes)
+  end
+
+  defp revectorize_node(node, vectorized_axes) do
+    vectorized_axes = compute_arg_vectorized_axes(node, vectorized_axes)
+
+    node
+    |> Nx.devectorize(keep_names: false)
```
josevalim, line 292: They should all be devectorized.

nx/lib/nx/defn/grad.ex
```diff
@@ -1343,9 +1424,34 @@ defmodule Nx.Defn.Grad do
 
   ## General helpers
 
-  defp unbroadcast(%{shape: shape} = x, res, %{shape: shape}), do: {x, res}
+  defp unbroadcast(x, res, ans) do
```
josevalim, line 1427: Revert these changes.

nx/lib/nx/defn/grad.ex
polvalente: I've changed the `compute_arg_vectorized_axes` function into a function that returns only the names (it's less assertive, but a tad cheaper), and that allows us to not need this remapping here.

It was needed for cases where `t` has [x: 1] and vectorized_axes has [x: 2], for instance -- implicit broadcast situations.

nx/lib/nx/defn/grad.ex
```diff
@@ -213,6 +265,31 @@ defmodule Nx.Defn.Grad do
...
+    reversed_inner_axes =
+      Enum.reduce(names, [], fn name, acc ->
+        if name in parent_names do
+          [name | acc]
+        else
+          acc
+        end
+      end)
+
+    Keyword.keys(vectorized_axes) ++ Enum.reverse(reversed_inner_axes)
```
josevalim, line 284: This is a Enum.filter, no? 

```suggestion
    Keyword.keys(vectorized_axes) ++ Enum.filter(names, & &1 in parent_names)
```

josevalim (approved):
Beautiful!!!!! We can merge this and add the new ops later!

nx/lib/nx/defn/grad.ex
polvalente: Yup. The loop had to iterate over multiple data structures before

PR #1537, opened 2024-09-24
fix: test on stable ordering instead of ref-dependent order

polvalente:


nx/test/nx/defn/tree_test.exs
```diff
@@ -1,5 +1,5 @@
 defmodule Nx.Defn.TreeTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case, async: false
```
polvalente, line 2:
```suggestion
  use ExUnit.Case, async: true
```

josevalim (approved):

PR #1543, opened 2024-10-09
Support cross-compilation for use with Nerves

fhunleth:
This moves the Makefile code that queries the system type and CUDA
availability under a guard for whether things are being cross-compiled.
If cross-compiled, then the settings are hardcoded. This should work in
non-Nerves compilation cases too, but those weren't tested.

### DIFF
--- a/exla/Makefile
+++ b/exla/Makefile
@@ -40,7 +40,28 @@ endif
 
 LDFLAGS = -L$(XLA_EXTENSION_LIB) -lxla_extension -shared
 
-ifeq ($(shell uname -s), Darwin)
+ifeq ($(CROSSCOMPILE),)
+	# Interrogate the system for local compilation
+	UNAME_S = $(shell uname -s)
+
+	NVCC_RESULT := $(shell which nvcc 2> /dev/null)
+	NVCC_TEST := $(notdir $(NVCC_RESULT))
+
+ifeq ($(NVCC_TEST),nvcc)
+	NVCC := nvcc
+	NVCCFLAGS += -DCUDA_ENABLED
+else
+	NVCC := $(CXX)
+	NVCCFLAGS = $(CFLAGS)
+endif
+else
+	# Determine settings for cross-compiled builds like for Nerves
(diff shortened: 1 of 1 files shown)


josevalim (approved):
Fine by me, thank you! But @polvalente should merge it :)

fhunleth:
Thanks!

@polvalente In case you're curious, @lawik has been working with the Whisper model and we've been collaborating on https://github.com/nerves-livebook/nerves_livebook/pull/559. Note that Whisper doesn't work on that PR due to a lack of ffmpeg, but much simpler use seems to be working.

exla/Makefile
```diff
@@ -40,7 +40,28 @@ endif
...
+else
+	NVCC := $(CXX)
+	NVCCFLAGS = $(CFLAGS)
+endif
+else
+	# Determine settings for cross-compiled builds like for Nerves
+	UNAME_S = Linux
+	NVCC := $(CXX)
+	NVCCFLAGS = $(CFLAGS)
+endif
```
polvalente, line 62: I believe `UNAME_S` should be given as an optional env flag. Linux is _probably_ always the target, but I don't think we can assert this for sure. Maybe a different name should be used for the env var as it's now gonna be (optionally) exposed.

I also wonder if we should allow NVCC to be provided via the toolchain for systems like the nvidia jetson. 

```suggestion
else
	# Determine settings for cross-compiled builds like for Nerves
	UNAME_S ?= Linux
	NVCC := $(CXX)
	NVCCFLAGS = $(CFLAGS)
endif
```

polvalente: Actually, given that this is only used for checking if it's mac or not, we can merge as is. Crosscompiling for mac is highly unlikely IMO

polvalente (approved):

exla/Makefile
fhunleth: The only cross-compilation I've run into that hasn't been Linux is for RTOSs and the "Linux" options seem to work. It might be easier to address when it comes up. 

Thanks for merging!

PR #1544, opened 2024-10-11
feat: experimental sharding backend

polvalente:
Adds a proof-of-concept implementation for sharding as an Nx meta-compiler.

In the current proposal, we shard inputs according to an arbitrary slicing configuration, and the compiler then does its best to propagate those slices through to the output. The compiler can then build a separate {args, function, reducer} tuple for each of the output shards, where:

- args: is a version of the input arguments that is sliced according to which data section is required for that specific output data section
- function: is a new compilation of the input function based on the new sliced arguments
- reducer: is a function that is responsible for inserting the result shard into the correct place in an output accumulator tensor.

The following example showcases how we can shard the example function into 4 separate shards. This happens because There are 2 shards for arg1, and only the first axis of arg0 is able to be sharded, due to the other 2 axes being connected to contracting axes in the dot product.

```elixir
arg0_sharding = %{} # inputs are taken to be fully sharded if no specification is given
arg1_sharding = %{4 => [0..0, 1..1]}

Nx.default_backend(Nx.BinaryBackend)

fun = fn l, r ->
  x = Nx.add(l, Nx.tensor([[1]]))
  x = Nx.transpose(x, axes: [0, 2, 1])
  y = Nx.subtract(r, 1)
  y = Nx.squeeze(y, axes: [0, 1])
  Nx.dot(x, [2, 1], y, [1, 0])
end

# fun = &Nx.dot(&1, [1, 2], &2, [1, 0])
# fun = &Nx.add(&1, &2)

inputs = [
  Nx.iota({2, 2, 3}, type: :f32),
  Nx.add(Nx.iota({1, 1, 3, 2, 2}), 10)
]

{output_holder, shards} =
  Nx.Defn.jit_apply(
    fun,
    inputs,
    compiler: Nx.Defn.ShardingCompiler,
    sharding_config: [arg0_sharding, arg1_sharding],
    sharding_compiler: Nx.Defn.Evaluator,
    sharding_compiler_options: []
  )

sharded_result =
  shards
  |> Task.async_stream(fn {arg, fun, caster} ->
    dbg(self())
    {fun.(arg), caster}
  end)
  |> Enum.reduce(output_holder, fn {:ok, {result, caster}}, acc ->
    caster.(result, acc)
  end)
  |> IO.inspect()

# Ensure that the sharded result is the same as the result for the function applied to the unsharded inputs
IO.inspect(Nx.equal(sharded_result, apply(fun, inputs)) |> Nx.all() |> Nx.to_number() |> Kernel.==(1))
```


nx/lib/nx/defn/sharding_compiler.ex
```diff
@@ -0,0 +1,610 @@
...
+            input = input_fn.()
+            shards = shards_by_input_id[input_id]
+            shards_by_axis = Enum.group_by(shards, & &1.axis)
+
+            {_, _, starts_reverse, lengths_reverse} =
+              Enum.reduce(Tuple.to_list(input.shape), {shards_by_axis, 0, [], []}, fn axis_size,
+                                                                                      {shards_by_axis,
+                                                                                       axis,
+                                                                                       starts,
+                                                                                       lengths} ->
```
josevalim, line 69:
```suggestion
              Enum.reduce(Tuple.to_list(input.shape), {shards_by_axis, 0, [], []}, fn
                axis_size, {shards_by_axis, axis, starts, lengths} ->
```

nx/lib/nx/defn/graph_splitter.ex
```diff
@@ -0,0 +1,283 @@
+defmodule Nx.Defn.GraphSplitter do
+  alias Nx.Defn.Composite
+
+  alias Nx.Tensor, as: T
+  alias Nx.Defn.Expr
+  alias Nx.Defn.GraphSplitter.Stage
+
+  @doc """
+  Traverses the expression and splits it into stages.
```
josevalim, line 9: We need more docs, I don't know how to use it.

nx/lib/nx/defn/graph_splitter.ex
```diff
@@ -0,0 +1,283 @@
+defmodule Nx.Defn.GraphSplitter do
```
josevalim, line 1:
```suggestion
defmodule Nx.Defn.Graph do
```

And traverse becomes `split`?

nx/lib/nx/defn/graph_splitter/stage.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Nx.Defn.GraphSplitter.Stage do
```
josevalim, line 1:
```suggestion
defmodule Nx.Defn.Graph.Stage do
```

nitpick: and maybe merge into the file above?

nx/lib/nx/defn/graph_splitter.ex
```diff
@@ -0,0 +1,277 @@
...
+
+          tuple ->
+            {_idx, scope, reverse_results} =
+              tuple
+              |> Tuple.to_list()
+              |> Enum.reduce({0, scope, []}, fn tensor, {idx, scope, results_acc} ->
+                {idx + 1, Map.put(scope, {id, idx}, tensor), [tensor | results_acc]}
+              end)
+
+            {reverse_results, scope}
```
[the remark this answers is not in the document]
polvalente: Yes. Apparently I can't read my own code.

nx/lib/nx/defn/graph.ex
```diff
@@ -1,4 +1,15 @@
 defmodule Nx.Defn.Graph do
+  @moduledoc """
+  A module for splitting Nx.Defn.Expr into stages.
```
josevalim, line 3:
```suggestion
  A module for splitting `Nx.Defn.Expr` into stages.
```

nx/lib/nx/defn/graph.ex
```diff
@@ -1,4 +1,15 @@
...
+
+  This module is used to split an Nx.Defn.Expr into stages, which are then
```
josevalim, line 5:
```suggestion
  This module is used to split an `Nx.Defn.Expr` into stages, which are then
```

josevalim (approved):

PR #1548, opened 2024-10-25
fix: make exla build resilient to stale upgrades

### LINKED ISSUE #1546: "module EXLA.NIF is not available"  What does this mean?
image

Hi.   I am not a genius like Jose Valim.   I do not have 10 years in Docker experience.  I am just a statistician and cannot spend days of my life learning how to compile things.   

Can someone give a clear and concise, plain english description on what is happening here?  It has destroyed my app and I can no longer run it after upgrading.

polvalente:
fixes #1546

### DIFF
--- a/exla/Makefile
+++ b/exla/Makefile
@@ -9,7 +9,7 @@ XLA_INCLUDE_PATH = $(XLA_EXTENSION_DIR)/include
 
 # Cache configuration
 EXLA_CACHE_SO = cache/libexla.so
-EXLA_CACHE_OBJ_DIR = cache/objs
+EXLA_CACHE_OBJ_DIR = cache/$(EXLA_VERSION)/objs
 
 # Private configuration
 EXLA_DIR = c_src/exla
--- a/exla/Makefile
+++ b/exla/Makefile
@@ -8,7 +8,7 @@ XLA_EXTENSION_LIB = $(XLA_EXTENSION_DIR)/lib
 XLA_INCLUDE_PATH = $(XLA_EXTENSION_DIR)/include
 
 # Cache configuration
-EXLA_CACHE_SO = cache/libexla.so
+EXLA_CACHE_SO = cache/$(EXLA_VERSION)/libexla.so
 EXLA_CACHE_OBJ_DIR = cache/$(EXLA_VERSION)/objs
 
 # Private configuration
--- a/exla/README.md
+++ b/exla/README.md
@@ -58,7 +58,7 @@ We recommend following the steps below:
   1. If the error appeared after upgrading EXLA, ensure that you have the proper dependency versions given by [XLA](https://github.com/elixir-nx/xla). Afterwards, compile with `mix compile --clean-libexla-cache --force` to clean up all cached files.
      * Besides the XLA dependency versions, ensuring `gcc` (or your compiler of choice), `libc` and `make` are compatible is also important.
      * Remember to save the compilation logs from this step for further debugging.
-     * It is a good idea to save the `cache/libexla.so` file so that the team can inspect its contents if needed.
+     * It is a good idea to save the `cache/<version>/libexla.so` file so that the team can inspect its contents if needed.
   2. If the error persists, look for the `** (RuntimeError) Failed to load NIF library.` exception on application start-up.
     This exception should provide more information on what's the issue when loading the NIF. Share these logs in an issue on GitHub
     so that the Nx team can investigate further.
--- a/exla/README.md
+++ b/exla/README.md
@@ -48,6 +48,22 @@ EXLA relies on the [XLA](https://github.com/elixir-nx/xla) package to provide th
 
 For cross-compilation, you need to [set your `XLA_TARGET_PLATFORM` variable](https://github.com/elixir-nx/xla#xla_target_platform) to the correct target platform value (i.e. `aarch64-linux-gnu` for the Raspberry Pi 4).
 
+## Troubleshooting
+
+EXLA uses NIFs (C-interface code called from Elixir) for part of its functionality.
+If for any reason these fail to compile or load, troubleshooting can be tricky.
+
+We recommend following the steps below:
+
+  1. If the error appeared after upgrading EXLA, ensure that you have the proper dependency versions given by [XLA](https://github.com/elixir-nx/xla). Afterwards, compile with `mix compile --clean-libexla-cache --force` to clean up all cached files.
+    - Besides the XLA dependency versions, ensuring `gcc` (or your compiler of choice), `libc` and `make` are compatible is also important.
+    - Remember to save the compilation logs from this step for further debugging.
+    - It is a good idea to save the `cache/libexla.so` file so that the team can inspect its contents if needed.
+  2. If the error persists, look for the `** (RuntimeError) Failed to load NIF library.` exception on application start-up.
+    This exception should provide more information on what's the issue when loading the NIF. Share these logs in an issue on GitHub
+    so that the Nx team can investigate further.
+
+
 ## Contributing
 
 ### Building locally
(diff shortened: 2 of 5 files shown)


exla/mix.exs
```diff
@@ -133,7 +134,23 @@ defmodule EXLA.MixProject do
...
+    clean_libexla_cache? = parsed[:clean_libexla_cache] == true
+    force? = parsed[:force] == true
+
+    if force? do
+      Mix.shell().info("Removing cached .o files in cache/#{@version}/objs")
+      File.rm_rf!("cache/#{@version}/objs")
+    end
+
+    if clean_libexla_cache? or force? do
+      Mix.shell().info("Removing cached libexla.so files in cache/libexla.so")
```
jonatanklosko, line 150:
```suggestion
      Mix.shell().info("Removing cached libexla.so file in cache/libexla.so")
```

exla/lib/exla/nif.ex
```diff
@@ -4,7 +4,20 @@ defmodule EXLA.NIF do
...
+    case :erlang.load_nif(path, 0) do
+      :ok ->
+        :ok
+
+      {:error, {reason, text}} ->
+        raise """
+        Failed to load NIF library.
+        Follow the steps in the :exla README Troubleshooting section for more information.
+
+        #{reason}
```
jonatanklosko, line 17:
```suggestion
        #{inspect(reason)}
```

jonatanklosko:
Looks good to me!

exla/README.md
```diff
@@ -48,6 +48,22 @@ EXLA relies on the [XLA](https://github.com/elixir-nx/xla) package to provide th
...
 For cross-compilation, you need to [set your `XLA_TARGET_PLATFORM` variable](https://github.com/elixir-nx/xla#xla_target_platform) to the correct target platform value (i.e. `aarch64-linux-gnu` for the Raspberry Pi 4).
 
+## Troubleshooting
+
+EXLA uses NIFs (C-interface code called from Elixir) for part of its functionality.
+If for any reason these fail to compile or load, troubleshooting can be tricky.
+
+We recommend following the steps below:
+
+  1. If the error appeared after upgrading EXLA, ensure that you have the proper dependency versions given by [XLA](https://github.com/elixir-nx/xla). Afterwards, compile with `mix compile --clean-libexla-cache --force` to clean up all cached files.
```
josevalim, line 58: `mix compile` is not going to work when compiled as a dependency. We would need to introduce a separate mix compiler or a separate task.

josevalim: It won’t work in the sense that it won’t invoke the code you added in the mix.exs file. `mix compile` is only about the local project. Maybe an env var with `mix deps.compile exla` will work.

exla/mix.exs
```diff
@@ -133,7 +134,23 @@ defmodule EXLA.MixProject do
     {:ok, []}
   end
 
-  defp cached_make(_) do
+  defp cached_make(args) do
+    {parsed, _args, _invalid} =
+      OptionParser.parse(args, strict: [clean_libexla_cache: :boolean, force: :boolean])
```
josevalim, line 139: Suggestion to make this a single env var called EXLA_FORCE_REBUILD, as explained above.

polvalente: Env var is probably the simpler alternative :)

exla/README.md
jonatanklosko: @josevalim it may works as `mix deps.clean exla --build --clean-libexla-cache`?

jonatanklosko: err `mix deps.compile exla --force --clean-libexla-cache`

jonatanklosko: But yeah, env var sounds better either way.

josevalim: I am not sure if we pass options forward in all of these cases. Although we should probably make `mix deps.clean` for dependencies extensible, so we actually invoke `clean` in the relevant dep.

exla/Makefile
```diff
@@ -9,7 +9,7 @@ XLA_INCLUDE_PATH = $(XLA_EXTENSION_DIR)/include
 
 # Cache configuration
 EXLA_CACHE_SO = cache/libexla.so
```
josevalim, line 11: Shouldn't we version this one as well? Just in case?

```suggestion
EXLA_CACHE_SO = cache/$(EXLA_VERSION)/libexla.so
```

exla/mix.exs
```diff
@@ -150,14 +167,21 @@ defmodule EXLA.MixProject do
     cached_so = Path.join([xla_cache_dir(), "exla", cache_key, "libexla.so"])
     cached? = File.exists?(cached_so)
```
josevalim, line 168: I think you can mostly remove the changes to this function and instead simply write this:

```suggestion
    cached? = System.get_env("XLA_FORCE_REBUILD") == nil and File.exists?(cached_so)
```

josevalim: This will be enough to force a rebuild and write the new build to cache.

exla/Makefile
polvalente: Shouldn't hurt :)
I'd tried this during development but it had failed. Seems to be working locally now.

exla/mix.exs
polvalente: @josevalim I changed to using env vars, but kept 2 separate ones because there are many instances in which I wanted to leverage the `.o` caching while getting rid of the libexla.so file and cache, as well as instances where I wanted to clean-slate everything.

josevalim: My problem with the above is that, if something is not working, I need to set two env vars to clear everything. So why not a single env var with different values?

  * `XLA_FORCE_REBUILD=1` - remove all
  * `XLA_FORCE_REBUILD=partial` - keep .o

Also, is there any situation where you don't want to cache files to be rewritten? I don't see a reason to skip that as part of these options?

Finally note that we use `XLA` env vars for pretty much everything, since we have `XLA_CACHE_DIR`, etc.

polvalente: I'm ok with having special values for the env var and having a single one.

I disagree with using `XLA` as the prefix, though, as this doesn't have anything to do with `:xla`. `XLA_CACHE_DIR` being used in EXLA is just an artifact of "let's use the same subdir for both XLA and EXLA". The origin of the variable is in `:xla`.

> Also, is there any situation where you don't want to cache files to be rewritten? I don't see a reason to skip that as part of these options?

As for this part, I don't think we are skipping the rewrite.

polvalente: I refactored the `cached?` check so that there's less confusion about whether rewrites will happen.
`cached?` will now be set to false regardless of the file existing if force rebuild is set.

exla/test/exla/device_memory_sharing_test.exs
```diff
@@ -27,14 +27,13 @@ defmodule EXLA.DeviceMemorySharingTest do
   end
 
   @tag :cuda_required
-  test "ipc handles don't crash the runtime when :local mode is selected" do
-    assert {:error, ~c"Invalid pointer size for selected mode."} ==
+  test "invalid ipc handles don't crash the runtime" do
```
polvalente, line 30: This test was breaking locally (I guess I hadn't run it on my desktop since the update :stuck_out_tongue:)

exla/mix.exs
```diff
@@ -133,7 +134,38 @@ defmodule EXLA.MixProject do
...
+          false
+
+        "partial" ->
+          :partial
+
+        "true" ->
+          true
+
+        "1" ->
+          true
```
josevalim, line 153:
```suggestion
         :none

        "0" ->
          :none

        "partial" ->
          :partial

        "true" ->
          :full

        "1" ->
          :full
```

This should force all the code below to be explicit :)

PR #1550, opened 2024-10-29
fix: least_squares implementation

polvalente:
The Nx.LinAlg.least_squares implementation assumed that for square matrices, it should fall back to Nx.LinAlg.solve.

However, the least squares solution is to minimize norm(Ax - b, 2), which isn't necessarily equal to solving Ax = b.
For example, if A = [[1, 2], [2, 4]] and b = [[3], [5]], Nx.LinAlg.solve will not be able to find a solution as the system doesn't have a valid solution. By going the pinv route, the minimal solution will be found.


polvalente:
@msluszniak This might be relevant to Scholar.

josevalim (approved):

PR #1552, opened 2024-10-30
fix: Nx.Random.shuffle repeating a single value in certain cases on GPU

### LINKED ISSUE #1551: `Nx.Random.shuffle/3` fails for large tensors on `cuda` backend
When running the following code on an EXLA backend with a CUDA GPU, through the `livebook:0.14.5-cuda12` image:
```elixir
key = Nx.Random.key(1)
input = Nx.iota({1_000_000})
{output, _new_key} = Nx.Random.shuffle(key, input)
output
```
the result looks like this:
```elixir
#Nx.Tensor<
  s32[1000000]
  EXLA.Backend<cuda:0, 0.477854050.2664562754.2456>
  [163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, 163918, ...]
>
```
The bug also occurs for smaller tensors, but with a lesser frequency. The deciding factor is the size of the axis used for the shuffle, the other dimensions of the tensor do not seem to be relevant. It seems to start happening with an axis size around `100,000`, and is guaranteed after `1,000,000`.

Moreover, about half the time the execution never completes and the Livebook runtime has to be restarted.

This has been observed on two different machines, one with an RTX4090 graphics card and one with a GTX1070ti. The bug did not occur during testing on the CPU on the same machines.

polvalente:
fixes #1551

In short, what was happening is that we were not testing a specific edge case that resulted in an invalid-shaped tensor for the `sort_keys` sub-input in `sort_key_val`, that in turn yielded an invalid shuffle when the input is 1D.

### DIFF
--- a/exla/lib/exla/mlir/value.ex
+++ b/exla/lib/exla/mlir/value.ex
@@ -54,12 +54,17 @@ defmodule EXLA.MLIR.Value do
   }
 
   for {op, direction} <- @bin_comparison_ops do
-    def unquote(op)(%Value{function: func} = lhs, %Value{function: func} = rhs, typespec) do
-      compare_and_return_bool(func, lhs, rhs, typespec, unquote(direction))
+    def unquote(op)(
+          %Value{function: func} = lhs,
+          %Value{function: func} = rhs,
+          typespec,
+          opts \\ []
+        ) do
+      compare_and_return_bool(func, lhs, rhs, typespec, unquote(direction), opts[:total_order])
     end
   end
 
-  defp compare_and_return_bool(func, lhs, rhs, typespec, direction) do
+  defp compare_and_return_bool(func, lhs, rhs, typespec, direction, total_order? \\ false) do
     %{type: lhs_type} = get_typespec(lhs)
     %{type: rhs_type} = get_typespec(rhs)
 
@@ -69,7 +74,14 @@ defmodule EXLA.MLIR.Value do
           [compare_type: attr_comparison_type(:float)]
 
         Nx.Type.float?(lhs_type) or Nx.Type.float?(rhs_type) ->
-          [compare_type: attr_comparison_type(:float)]
+          attr =
+            if total_order? do
+              attr_comparison_type(:totalorder)
+            else
+              attr_comparison_type(:float)
+            end
+
+          [compare_type: attr]
 
         true ->
           []
@@ -1069,7 +1081,7 @@ defmodule EXLA.MLIR.Value do
   defp attr_comparison_direction(value) when value in [:eq, :lt, :le, :gt, :ge, :ne],
     do: attr_enum("stablehlo", "comparison_direction", value)
 
-  defp attr_comparison_type(value) when value in [:float, :totalorder, :notype],
+  defp attr_comparison_type(value) when value in [:float, :totalorder],
     do: attr_enum("stablehlo", "comparison_type", value)
 
   defp attr_precision(value) when value in [:default, :high, :highest],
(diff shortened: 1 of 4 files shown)


nx/lib/nx/random.ex
```diff
@@ -821,16 +827,25 @@ defmodule Nx.Random do
...
 
-  defnp sort_key_val(tensor, sort_keys, opts \\ []) do
+  deftransformp sort_key_val(tensor, sort_keys, opts \\ []) do
     idx = Nx.argsort(sort_keys, axis: opts[:axis])
-    Nx.take_along_axis(tensor, idx, axis: opts[:axis])
+
+    if opts[:independent] do
+      # We need to use take_along_axis in the independent case because
+      # the sort_keys tensor has the same shape as the input tensor.
+      Nx.take_along_axis(tensor, idx, axis: opts[:axis])
```
jonatanklosko, line 843: I think `take_along_axis` with 1D tensors is supposed to work just fine, the issue must be something else.

jonatanklosko:
I'm still trying to narrow it down, but the issue must be somewhere in the combination of `random_bits` and `take_along_axis`. For example, if we make this change:

```diff
-sort_keys = random_bits(keys[1], shape: tensor.shape)
+sort_keys = randint_split(keys[1], 0, uint32max, shape: tensor.shape)
```

It seems to work just fine.

jonatanklosko:
@polvalente actually, this also fixes it:

```diff
-sort_keys = random_bits(keys[1], shape: tensor.shape)
+sort_keys = random_bits(keys[1], shape: tensor.shape) |> Nx.as_type(:s32)
```

nx/lib/nx/random.ex
polvalente: I wonder if this is actually a bug in EXLA's take_along_axis somehow.

The generated indices from argsort are returning with sensible values, but somehow these aren't being applied properly by take_along_axis

exla/lib/exla/mlir/value.ex
```diff
@@ -66,19 +66,16 @@ defmodule EXLA.MLIR.Value do
...
         Nx.Type.complex?(lhs_type) or Nx.Type.complex?(rhs_type) ->
-          attr_comparison_type(:float)
+          [compare_type: attr_comparison_type(:float)]
 
         Nx.Type.float?(lhs_type) or Nx.Type.float?(rhs_type) ->
-          attr_comparison_type(:float)
+          [compare_type: attr_comparison_type(:float)]
 
         true ->
-          attr_comparison_type(:notype)
```
jonatanklosko, line 75: Let's also remove it here:

https://github.com/elixir-nx/nx/blob/08dc89336c5b15c42c2132b44f5aa28f93fd13ee/exla/lib/exla/mlir/value.ex#L1072

jonatanklosko (approved):

exla/test/exla/random_test.exs
```diff
@@ -39,4 +39,23 @@ defmodule EXLA.NxRandomTest do
...
+
+  @tag :cuda_required
+  test "regression on single-dimensional and multi-dimensional Random.shuffle" do
+    # these are put in the process dictionary, so it's thread-safe to do this
+    Nx.default_backend({EXLA.Backend, client: :cuda})
+    Nx.Defn.default_options(compiler: EXLA, client: :cuda)
+    key = Nx.Random.key(127)
+
+    t1 = Nx.iota({2, 100})
+    t2 = Nx.iota({100})
```
jonatanklosko, line 51: Does the test fail with the previous version? (the tensor is small)

polvalente: Nah. I'm gonna remove this, specially now that we found the actual culprit

jonatanklosko (approved):
:shipit:

PR #1553, opened 2024-10-30
fix: exla rpath

polvalente:


exla/Makefile
```diff
@@ -61,16 +61,24 @@ else
 	NVCCFLAGS = $(CFLAGS)
 endif
 
+
+ifeq ($(MIX_BUILD_EMBEDDED), true)
+	XLA_EXTENSION_LIB_RPATH = "xla_extension/lib"
+else
+	XLA_EXTENSION_LIB_RPATH = "../xla_extension/lib"
```
polvalente, line 68: small fix needed due to the addition of the version to the EXLA lib compile path

josevalim (approved):

jonatanklosko (approved):

polvalente:
There's some weird interplay between how EXLA looks up the rpath if it's the main app or if it's a library.
Just copying over the unversioned cache .so seems to be the simplest solution.

This probably doesn't solve the problem of dealing with a stale libexla.so, but the new env var, even in partial mode, will deal with rebuilding everything for a new version

exla/mix.exs
```diff
@@ -143,6 +143,9 @@ defmodule EXLA.MixProject do
         "0" ->
           :none
 
+        "false" ->
```
jonatanklosko, line 146: I would do this:

```elixir
case System.get_env("EXLA_FORCE_REBUILD") do
  value when value in [nil, "0", "false"] ->
    :none

  "partial" ->
    :partial

  value when value in ["true", "1"] ->
    :full

  other ->
    ...
end
```

A cond also works.

polvalente: I think the readability is either equal or worse in that alternative, given it's a small set of possible values

jonatanklosko: With `cond` readability should be just as good as today and it should be easier to scan through. Your call :)

josevalim: Agreed this would suit cond perfectly.

josevalim (commented):
There is an issue here in that the .so can be huge, back then it was 300MB, so duplicating that in disk can be too much.

josevalim:
Maybe we keep the .o versioned and not the final one. This should be enough for make to trigger a rebuild whenever the version changes.

polvalente:
> There is an issue here in that the .so can be huge, back then it was 300MB, so duplicating that in disk can be too much.
I checked and it was like 2MB or something.

I'll keep libexla.so unversioned for simplicity then.
The main problem was stale intermediate .o, and that's already solved

exla/Makefile
```diff
@@ -71,6 +71,7 @@ else
 	LDFLAGS += -Wl,-rpath,'$$ORIGIN/xla_extension/lib'
 endif
 
+
```
josevalim, line 74:
```suggestion
```

exla/Makefile
```diff
@@ -98,7 +99,7 @@ $(EXLA_CACHE_OBJ_DIR)/%.o: $(EXLA_DIR)/%.cc $(HEADERS)
 	$(CXX) $(CFLAGS) -c $< -o $@
 
 $(EXLA_CACHE_SO): $(XLA_EXTENSION_DIR) $(OBJECTS)
-	$(CXX) $(OBJECTS) -o $(EXLA_CACHE_SO) $(LDFLAGS)
+	$(CXX) $(OBJECTS) -o $(EXLA_VERSIONED_CACHE_SO) $(LDFLAGS)
```
josevalim, line 102:
```suggestion
	$(CXX) $(OBJECTS) -o $(EXLA_CACHE_SO) $(LDFLAGS)
```

PR #1555, opened 2024-10-30
Fix documentation formatting

mooreryan:
The docs formatting for `Nx.all_close` was missing leading spaces for some of the iex examples, causing the docs to render like this:

before

---

Fixing the spacing issues makes it render like so:

after


josevalim:

PR #1557, opened 2024-11-08
feat: add shard execution workflow

polvalente:
Adds the initial version of the process communication structure for sharded execution.

Does not handle container outputs for the sharded function yet,
and also does not yet bring everything together into the compiler jit function.

### DIFF
--- a/nx/lib/nx/application.ex
+++ b/nx/lib/nx/application.ex
@@ -4,6 +4,7 @@ defmodule Nx.Application do
 
   def start(_type, _args) do
     children = [
+      Nx.Defn.ShardingCompiler.ShardRegistry,
       %{id: Nx.Serving.PG, start: {:pg, :start_link, [Nx.Serving.PG]}},
       {Nx.HiddenServing, Nx.Serving.PG}
     ]
--- a/nx/lib/nx/defn/sharding_compiler/shard_execution/argument_provider.ex
+++ b/nx/lib/nx/defn/sharding_compiler/shard_execution/argument_provider.ex
@@ -1,0 +1,21 @@
+defmodule Nx.Defn.ShardingCompiler.ShardExecution.ArgumentProvider do
+  use GenServer
+
+  alias Nx.Defn.ShardingCompiler.ShardRegistry
+
+  def init(data) do
+    {:ok, data}
+  end
+
+  def start_link([data, idx, section_id]) do
+    GenServer.start_link(__MODULE__, data, name: via_tuple(idx, section_id))
+  end
+
+  defp via_tuple(idx, section_id) do
+    {:via, Registry, {ShardRegistry, {nil, idx, section_id}}}
+  end
+
+  def handle_call(:get, _from, data) do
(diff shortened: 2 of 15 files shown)


nx/lib/nx/application.ex
```diff
@@ -4,6 +4,7 @@ defmodule Nx.Application do
 
   def start(_type, _args) do
     children = [
+      Nx.Defn.ShardingCompiler.ShardRegistry,
```
polvalente, line 7: I'm not sure we want to have this here, actually.
In fact, I think we might actually want to go with gen_stage for the execution, since the whole "chain of processes producing data to one another" smells a lot like gen_stage.

Thoughts?

josevalim: We should talk about it but I doubt GenStage will be helpful here. One of the biggest pitfalls in GenStage is that people move data around too much, when they should not. It is cheaper to move computations than to move data.

nx/lib/nx/defn/sharding_compiler/shard_execution/argument_provider.ex
```diff
@@ -0,0 +1,21 @@
+defmodule Nx.Defn.ShardingCompiler.ShardExecution.ArgumentProvider do
```
josevalim, line 1: Why do you need several providers instead of one that receives the index and returns the relevant one?

nx/lib/nx/application.ex
josevalim: Plus the whole demand approach is unnecessary here. Here is either pending or done (like a promise), no?

nx/lib/nx/defn/sharding_compiler/shard_execution/argument_provider.ex
josevalim: Hrm, this may actually be the simplest, so nevermind.

josevalim:
Could we fully decouple the workflow definition and execution from Nx? Ideally we would have a workflow like this:


```
workflow = %{
  0 => %{
    code: &foo(&1, &2, ...),
    args: [1, 2]
  },
  1 =>  %{
    code: &bar(&1),
    args: [2]
  },
  2 => %{
    code: &baz/0,
    args: []
  }
}
```

And then we pass this to a ProcessExecutor which is completely independent of Nx and tensors. You could also have a Nx executor, but the overall idea is that the Executor should worry about resources and not necessarily tensors (except the resources the tensors are located).

polvalente:
> Could we fully decouple the workflow definition and execution from Nx? Ideally we would have a workflow like this:

We talked about this, but just to register here, I believe this is possible, but probably easier to generalize after we have things ready.

Each workflow step needs to be able to determine:

1. The input dependency mapping that tells the workflow from where to fetch each argument (which is represented here as simple numbers, but also should represent indices as we need to represent container outputs)
2. Some way to check whether a given dependency has produced its output.
3. A callback for fetching the output given the requested input section

All of this is mostly already present in the current code, with the coupling happening mostly on how we identify the dependencies based on data section ids.

PR #1558, opened 2024-11-12
Link to versioned XLA docs

jonatanklosko:
See https://github.com/elixir-nx/xla/pull/103.

### DIFF
--- a/exla/lib/exla.ex
+++ b/exla/lib/exla.ex
@@ -14,7 +14,7 @@ defmodule EXLA do
 
   > #### Version requirements {: .info}
   >
-  > For precise requirements, such as CUDA and cuDNN versions, see the `XLA`.
+  > For precise requirements, such as CUDA and cuDNN versions, see `XLA` docs.
 
   ## Configuration
 
--- a/exla/lib/exla.ex
+++ b/exla/lib/exla.ex
@@ -6,11 +6,15 @@ defmodule EXLA do
 
   ## XLA binaries
 
-  EXLA relies on the [XLA](https://github.com/elixir-nx/xla) package to
-  provide the necessary XLA binaries. Whenever possible it tries to download
-  precompiled builds, but you may need to build from source if there is no
-  version matching your target environment. For more details, including
-  GPU/TPU support see [the usage section](https://github.com/elixir-nx/xla#usage).
+  EXLA relies on the `XLA` package to provide the necessary XLA binaries.
+  Whenever possible it tries to download precompiled builds, but you may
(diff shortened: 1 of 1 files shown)


exla/lib/exla.ex
```diff
@@ -6,11 +6,15 @@ defmodule EXLA do
...
-  GPU/TPU support see [the usage section](https://github.com/elixir-nx/xla#usage).
+  EXLA relies on the `XLA` package to provide the necessary XLA binaries.
+  Whenever possible it tries to download precompiled builds, but you may
+  need to build from source if there is no version matching your target
+  environment. For more details, including GPU/TPU support and requirements
+  see the `XLA` docs.
+
+  > #### Version requirements {: .info}
+  >
+  > For precise requirements, such as CUDA and cuDNN versions, see the `XLA`.
```
jonatanklosko, line 17: It basically repeats the last line above, but I think it's worth highlighting, since the reader is likely to look for this.

josevalim:
```suggestion
  > For precise requirements, such as CUDA and cuDNN versions, see the `XLA` module.
```

exla/lib/exla.ex
```diff
@@ -6,11 +6,15 @@ defmodule EXLA do
 
   ## XLA binaries
 
-  EXLA relies on the [XLA](https://github.com/elixir-nx/xla) package to
-  provide the necessary XLA binaries. Whenever possible it tries to download
-  precompiled builds, but you may need to build from source if there is no
-  version matching your target environment. For more details, including
-  GPU/TPU support see [the usage section](https://github.com/elixir-nx/xla#usage).
+  EXLA relies on the `XLA` package to provide the necessary XLA binaries.
```
josevalim, line 9: This is going to link to the module itself, which is not what we want. Should we explicitly link to https://hexdocs.pm/xla instead?

josevalim: Ignore me, I saw the other PR.

exla/lib/exla.ex
josevalim, line 17:
```suggestion
  > For precise requirements, such as CUDA and cuDNN versions, see `XLA` docs.
```

josevalim (approved):

exla/lib/exla.ex
jonatanklosko: We actually do want to link to the module, so that the URL has the relevant XLA version. With https://github.com/elixir-nx/xla/pull/103 the XLA module docs include all the relevant information. WDYT?

PR #1559, opened 2024-11-13
docs: add guide on using Complex numbers inside defn

ndrean:
First commit


nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+Its norm is $ \sqrt{2}$, and angle is $\pi/4$.
+
+```elixir
+sqrt_2 = :math.sqrt(2)
+pi_4 = :math.pi()/4
+
+{
+  Complex.abs(one) == sqrt_2,
+  Complex.phase(one) == pi_4
+}
```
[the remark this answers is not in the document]
polvalente: If you want, you can save the livebook with the outputs instead

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+  defn source(t) do
+    f1 = 5; f2 = 20;
+    Nx.sin(2 * pi() * f1 * t ) + 1/2 * Nx.sin(2 * pi() * f2 * t)
+  end
+
+  defn sample(opts) do
+    fs = opts[:fs]
+    duration = opts[:duration]
+    sampling = Nx.linspace(0, duration, n: fs, endpoint: false, type: {:f, 32})
+    source(sampling)
```
polvalente, line 142:
```suggestion
    t = Nx.linspace(0, duration, n: trunc(duration * fs), endpoint: false, type: {:f, 32})
    source(t)
```

duration is in seconds, fs is in samples/second. Therefore, the number of samples should be `duration * fs`, rounded off

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+
+```elixir
+fs = 50; duration= 1
+
+sample = Signal.sample(fs: fs, duration: 1)
+```
+
+### Analyse the signal with DFT
+
+The DFT algorithm will **approximates the original signal**. It gives us the following information:
```
polvalente, line 157:
```suggestion
Because our signal contains many periods of the underlying function, the DFT results will contain some noise. This noise stem both from the fact that we're likely cutting of the signal before it repeats, and from the fact that we have a specific frequency resolution which ends up grouping our individual components into frequency bins - the latter isn't really a problem as we have chosen `fs` to be [fast enough](https://en.wikipedia.org/wiki/Nyquist%E2%80%93Shannon_sampling_theorem).
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+
+> The number at the index $i$ of the DFT results gives an approximation of the amplitude and phase of the sampled signal at the frequency $i$.
+
```
polvalente, line 160:
```suggestion

```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+In other words, doing `Nx.fft(sample)` returns a list of numbers indexed by the frequency.
```
polvalente, line 161:
```suggestion
In other words, `Nx.fft(sample)` returns a sequence of complex numbers separated in frequency bins, representing the absolute value and phase for each frequency.
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+
+These are complex numbers as they carry the amplitude and phase information.
+
```
polvalente, line 164:
```suggestion
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+dft = Nx.fft(sample)
+```
+
+Our frequency resolution - the accuracy of the DFT algorithm - is limited by the Nyquist rate, which is half the sampling rate, thus 25 Hz.
+
+Check [Nyquist](https://en.wikipedia.org/wiki/Nyquist_rate) to read more.
+
+We will therefor limit our study points to the indices below this threshold value by slicing the tensor, thus retain the 25 first ones.
+
+We are interested in the amplitudes only here. This is calculated by taking the norm of each number (`Nx.abs`).
```
polvalente, line 175:
```suggestion
We will limit our study points to the first half of the `dft` because it is symmetrical on the upper half.
The phase doesn't really matter to us, so we'll use `Nx.abs` to obtain the absolute values at each point.
```

frequency resolution in general refers to the frequency delta between frequency bins, not the maximum representable unaliased bandwidth (which is related to the nyquist frequency/nyquist rate)

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+
+```elixir
+n = Nx.size(dft)
+
+# Nyquist threshold:
+max_freq_index = div(n, 2)
+
+# thus we keep only the meaningfull values and calculate the amplitudes
+amplitudes =  Nx.abs(dft)[0..max_freq_index]
```
polvalente, line 184:
```suggestion
max_freq_index = div(n, 2)

amplitudes =  Nx.abs(dft)[0..max_freq_index]
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+
+# the frequency bins, "n" of them spaced by fs/n=1 unit:
```
polvalente, line 186:
```suggestion
# the frequency bins, "n" of them spaced by fs / fft_length Hz:
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+|> VegaLite.encode_field(:y, "amplitudes",
+  type: :quantitative,
+  title: "amplitutde",
+  scale: [domain: [0, 30]]
+)
+```
+
+We see the peaks at 5Hz, 20Hz with their amplitudes (the second is half the first).
+
+This is indeed our incomming signal 🎉
```
polvalente, line 211:
```suggestion
This is indeed our synthesized signal 🎉
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+  scale: [domain: [0, 50]]
+)
+|> VegaLite.encode_field(:y, "amplitudes",
+  type: :quantitative,
+  title: "amplitutde",
+  scale: [domain: [0, 30]]
+)
+```
+
+We see the peaks at 5Hz, 20Hz with their amplitudes (the second is half the first).
```
polvalente, line 209:
```suggestion
We see the peaks at 5Hz, 20Hz with their amplitudes these aren't necessarily `1` and `1/2` because of mathematical details of the DFT, but it's important to note that the first is twice the second, as expected.
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+frequencies = NxSignal.fft_frequencies(fs, fft_length: n)[0..max_freq_index]
+
+data1 = %{
+  frequencies: Nx.to_list(frequencies),
+  amplitudes: Nx.to_list(amplitudes)
+}
+
+VegaLite.new(width: 700, height: 300)
+|> VegaLite.data_from_values(data1)
+|> VegaLite.mark(:bar)
```
polvalente, line 196:
```suggestion
|> VegaLite.mark(:bar, tooltip: true)
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+  v = Nx.to_list(values)
+  v0 = Enum.at(v,0)
+  Enum.zip(v, Nx.to_list(indices))
+  |> Enum.filter(fn {v,_idx} -> v >=  v0 * threshold end)
+  |> Enum.unzip()
+  |> elem(1)
+
+end
+
+IO.puts("The top frequencies are: #{inspect(filtered_top_frequencies.(values, indices, 0.001 ))}")
```
polvalente, line 228:
```suggestion
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+  title: "amplitutde",
+  scale: [domain: [0, 30]]
+)
+```
+
+We see the peaks at 5Hz, 20Hz with their amplitudes (the second is half the first).
+
+This is indeed our incomming signal 🎉
+
+We can confirm this visual inspection with a peek into our data. We use `Nx.top_k` function.
```
polvalente, line 213:
```suggestion
We can confirm this visual inspection with a peek into our data. We use `Nx.top_k` function to get the top `5` values. Note that only the first 2 are actually significant, while the rest is part of the DFT noise floor.
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+  |> Enum.filter(fn {v,_idx} -> v >=  v0 * threshold end)
+  |> Enum.unzip()
+  |> elem(1)
+
+end
+
+IO.puts("The top frequencies are: #{inspect(filtered_top_frequencies.(values, indices, 0.001 ))}")
+```
+
+### Visualize the original signal and the IFFT reconstructed
```
polvalente, line 231:
```suggestion
### Visualizing the original signal and the Inverse Discrete Fourier Transform
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+
+Let's visualize our incoming signal over 400ms. This correspond to 2 periods of our 5Hz signal.
```
polvalente, line 233:
```suggestion
Let's visualize our incoming signal over 400ms. This correspond to 2 periods of our 5Hz component and 8 periods of our 20Hz component.
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+
+We compute 200 points to have a smooth curve, thus every (400/200=) 2ms.
```
polvalente, line 235:
```suggestion
We'll sample 200 equidistant points to so we can see a smooth curve (i.e. every 400ms/200 = 2ms, or a sampling frequency of 500Hz).
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+
+We also add the "reconstructed" signal via the **Inverse Fast Fourrier Transform** available as `Nx.ifft`.
```
polvalente, line 237:
```suggestion
We also add the reconstructed signal via the **Inverse Discrete Fourier Transform** available as `Nx.ifft`.
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+
+This gives us 50 values  spaced by (1000 / 50) = 20ms (we sampled as 50Hz during 1s).
```
polvalente, line 239:
```suggestion
This gives us 50 values spaced by 1000ms / 50 = 20ms.
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+
+We display them as a bar chart next to the "smooth" real signal.
```
polvalente, line 241:
```suggestion
Below, we display them as a bar chart under the line representing the ideal signal.
```

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+  |> VegaLite.mark(:line)
+  |> VegaLite.encode_field(:x, "x", type: :quantitative, title: "time (ms)", scale: [domain: [0, 0.4]])
+  |> VegaLite.encode_field(:y, "y", type: :quantitative, title: "signal"),
+  VegaLite.new()
+  |> VegaLite.mark(:bar)
+  |> VegaLite.encode_field(:x, "x", type: :quantitative, scale: [domain: [0, 0.4]])
+  |> VegaLite.encode_field(:y, "yr", type: :quantitative, title: "reconstructed"),
+])
+|> VegaLite.resolve(:scale, y: :independent)
+
```
polvalente, line 294:
```suggestion
#----------- REAL SIGNAL
# compute 200 points of the "real" signal during 2/5=400ms (twice the main period)

t = Nx.linspace(0, 0.4, n: trunc(0.4 * 500))
sample = Signal.source(t)

#----------- RECONSTRUCTED IFFT
yr = Nx.ifft(dft) |> Nx.real()
fs = 50
tr = Nx.iota(yr.shape) |> Nx.divide(fs)

idx = Nx.less_equal(tr, 0.4)
xr = Nx.select(idx, tr, :nan)
yr = Nx.select(idx, yr, :nan)
#----------------


data = %{
  x: Nx.to_list(t),
  y: Nx.to_list(sample)
}

data_r = %{
  yr: Nx.to_list(yr),
  xr: Nx.to_list(xr)
}

VegaLite.new(width: 600, height: 300)
|> VegaLite.layers([
  VegaLite.new()
  |> VegaLite.data_from_values(data)
  |> VegaLite.mark(:line, tooltip: true)
  |> VegaLite.encode_field(:x, "x", type: :quantitative, title: "time (ms)", scale: [domain: [0, 0.4]])
  |> VegaLite.encode_field(:y, "y", type: :quantitative, title: "signal")
  |> VegaLite.encode_field(:order, "x"),
  VegaLite.new()
  |> VegaLite.data_from_values(data_r)
  |> VegaLite.mark(:bar, tooltip: true)
  |> VegaLite.encode_field(:x, "xr", type: :quantitative, scale: [domain: [0, 0.4]])
  |> VegaLite.encode_field(:y, "yr", type: :quantitative, title: "reconstructed")
  |> VegaLite.encode_field(:order, "xr")
])
```

polvalente: In this approach you plot 2 separate sets of data onto the same graph. I believe this is a cleaner approach that shows that the math itself ensures things coincide.

polvalente (commented):
This is going in a great direction. We might want to separate the DFT part onto an NxSignal livebook instead, but let's keep working on this as a whole here before we decide!

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
+# Discrete Fourrier Transform with Nx
+
+```elixir
+Mix.install(
+  [
+    {:nx, "~> 0.9.1"},
+    {:kino, "~> 0.14.2"},
+    {:kino_vega_lite, "~> 0.1.11"},
+    {:exla, "~> 0.9.1"},
```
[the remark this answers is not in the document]
ndrean: It was an excuse to put the correct way to use EXLA as a backend. Up to you!

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
+# Discrete Fourrier Transform with Nx
```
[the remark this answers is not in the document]
ndrean: And I studied where he taught... no excuse, really.

nx/guides/advanced/complex_fft.livemd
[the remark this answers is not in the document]
ndrean: Yes good idea

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+
+Nx.abs(t)
+```
+
+We also have the imaginary number $i$ defined within `Nx.Constants`. Once imported, we can use it:
+
+```elixir
+i = Nx.Constants.i()
+
+Nx.add(1 , i)
```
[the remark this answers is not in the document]
ndrean:
```elixir
import Nx.Constants

Nx.add(1, i())
```

nx/guides/advanced/complex_fft.livemd
ndrean: We'll sample 200 equidistant points to have a smooth curve,   every 400ms/200 = 2ms,  or equivalently with a sampling frequency of 500Hz.

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+and compute its norm and phase:
+
+```elixir
+{
+  Nx.abs(t_one),
+  Nx.phase(t_one)
+}
+```
+
+These are "tensor" aware functions.
```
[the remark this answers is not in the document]
polvalente: You can add an extra entry to the stack below and point this out, that the functions will behave similarly to real tensors, applying element-wise whenever possible

nx/guides/advanced/complex_fft.livemd
ndrean: Strange, I tried to superpose 2 sets with different X's, didn't seem to work, but yes in fact.

Now, just stepping back and expressing my POV, the current primitives are no good. I mean, no one has time to spend with 15 LOC. Probably `Tucan` should be used or improved, but you should not need more most 4 LOC with this functional aproach.

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+
+<!-- livebook:{"break_markdown":true} -->
+
+### Build the signal
+
+Let's build a signal that we want to understand.
+
+It will be the sum of two sinusoidal signals, one at 5Hz, and one at 20Hz with the corresponding amplitudes (1, 0.5).
+
+$f(t) = \sin(2\pi*5*t) + \frac12 \sin(2\pi*20*t)$
```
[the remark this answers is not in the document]
ndrean: $\frac{1}{2}$ does not render for you?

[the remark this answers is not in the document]
polvalente: frac12 renders, but I think the usual syntax is clearer instead of using the special case

nx/guides/advanced/complex_fft.livemd
ndrean: Useless? maybe...

polvalente: Ah, I forgot to explain. I think just discussing over the output is better than inducing readers into falling back to Elixir at the first sign of trouble, if it makes sense. Sure, you can't determine things automatically, but even your solution uses prior knowledge anyway

nx/guides/advanced/complex_fft.livemd
ndrean: the second is half the first <=> the first is twice the second, as expected.  ??

nx/guides/advanced/complex_fft.livemd
ndrean: You mean "synthesized" by  the "original" signal?

nx/guides/advanced/complex_fft.livemd
polvalente: VegaLite is meant to be lower level and be a 1-1 mapping wuth the JS linrary.

My suggestion works because each layer is based on a separate set of values. I couldn't get it to work with everything in the same layer, but there is a way to do it I think.

You can refactor to use Tucan instead, as it's meant to have a higher level of abstraction

nx/guides/advanced/complex_fft.livemd
ndrean: ah, I see, yes of course, I used a wrong wording.  But yes, keep this short.

nx/guides/advanced/complex_fft.livemd
ndrean: Never used Tucan, lets see.

nx/guides/advanced/complex_fft.livemd
ndrean: hehehe, trouble, kind of! Just that was tempted to use `top_k` but needs filtering.

nx/guides/advanced/complex_fft.livemd
ndrean: Not sure to understand. What is fft_length? An fft just computes, for each of his frequency steps, the effect of  "rotating" all the points of sample slightly differently. So the total number of frequencies "inspected" are given by the number of points of the sample.

nx/guides/advanced/complex_fft.livemd
ndrean: If you change above, yes.

nx/guides/advanced/complex_fft.livemd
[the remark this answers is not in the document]
ndrean: Everyone does this I believe, but I'm fine!

nx/guides/advanced/complex_fft.livemd
polvalente: The DFT will by default have as many points as the input signal. However, you can zero-pad the signal to obtain an interpolated version of the DFT spectrum with smaller frequency bins.

This does not increase the intrinsic resolution of the transform, as that is fixed upon sampling as per the time-frequency uncertainty principle

polvalente: FFT implementations also benefit of this padding when it's used to yield a power of 2 input for the underlying algorithm

nx/guides/advanced/complex_fft.livemd
```diff
@@ -0,0 +1,297 @@
...
+end
+
+Nx.equal(Example.rotate(one), Nx.complex(-1,1))
+```
+
+## Discrete Fourrier Transform
+
+Suppose we have a periodic signal, and we want to sample it to understand the signal.
+
+We will run a Fast Fourrier Transform. It transforms a time-domain function into the frequency domain. Let's see how.
```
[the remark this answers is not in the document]
ndrean: A signal is represented by ,a sequence $[(t_1, f(t_1), \dots, (t_n, f(t_n)]$ which we call samples. 

~~A signal being periodic means that it contains a pattern that repeats after a given number of samples.~~

Then I suppose that a potential reader is aware of what spectral analysis is about, but probably not about the technical limitations on how to use this tool. So I would not be verbose about what DFT is. You added something valuable about the Shannon thing, if you are not versed in such technics.

nx/guides/advanced/complex_fft.livemd
polvalente: Yes. As we're synthesizing it instead of sampling from a continuous-time signal. Incoming didn't really resonate with me, and I think original is not really clear.

I'm open to suggestions, and I'm ok with using "original" as well, if you think synthesized is too technical

nx/guides/advanced/complex_fft.livemd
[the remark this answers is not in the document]
polvalente: I think we shouldn't assume the reader is knowledgeable in spectral analysis here, nor signal processing.

The signal is really just a sequence of points. You can apply Fourier analysis to spatially distributed data if you want. The time isn't important there.

ndrean:
I wonder why FFT and conv are in Nx and not in NxSignal ??

Now that this aperitif is almost done, I feel it is more interesting to play with the `convolution`, which most probably uses FFT, I will (try to) check. I read the doc, pretty terse. You have things like padding and sliding windows with strides but not sure how it wroks. I would build a little example for example applied on a simple and small image and use 3 or 4 kernels (say horizontal edge, vertical edge, gaussian blur). I don't think its huge, not even sure if the image will render correctly. Is this interesting for you? I may need examples to see and test what this function does exactly. MAybe if you point me directly where they are, could save me time.

nx/guides/advanced/complex_fft.livemd
ndrean: Yes, I used duration = 1 I guess.

polvalente:
> I wonder why FFT and conv are in Nx and not in NxSignal ??
> 
> Now that this aperitif is almost done, I feel it is more interesting to play with the `convolution`, which most probably uses FFT, I will (try to) check. I read the doc, pretty terse. You have things like padding and sliding windows with strides but not sure how it wroks. I would build a little example for example applied on a simple and small image and use 3 or 4 kernels (say horizontal edge, vertical edge, gaussian blur). I don't think its huge, not even sure if the image will render correctly. Is this interesting for you? I may need examples to see and test what this function does exactly. MAybe if you point me directly where they are, could save me time.

Awesome. I will ping you regarding convolutional filters, as I've just given a talk on that! This is the repo if you want to take a look already: https://github.com/polvalente/code-beam-lite-nyc-2024

Regarding fft and conv being in Nx, ultimately the reason is that they are present in the XLA API, really.

And NxSignal is a library that currently only implements things through calling Nx itself. Think of Nx as being analogous to NumPy, while NxSignal is a subsection of SciPy.

Finally, conv is an operation in itself and can be implemented in the time domain. Nx doesn't build it on top of FFT.

nx/guides/advanced/complex_fft.livemd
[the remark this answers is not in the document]
ndrean: Yes, but what is important for the reader to understand is that the spatial or temporal distribution of the sample is related to the Fourier distribution precisely by the number of points of the samples. sample_time_i => fourier_frequency_i  BUT with limitations as you explain further down.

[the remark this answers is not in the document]
polvalente: Each point in the spectrum depends on all points in time, so this isn't really true as well. If you want to introduce a more formal definition of a signal, that's great, but the 1:1 mapping shouldn't be the sole reason

ndrean:
Ah! I recall I had a little home work on cyclotomic polynomials and FFT for polynomial products (a convolution in fact) but was a long time ago.

So if you already did it, I can only will bring in a kinda noob POV! Thanks for the link, I will run it.

polvalente:
> Ah! I recall I had a little home work on cyclotomic polynomials and FFT for polynomial products (a convolution in fact) but was a long time ago.
> 
> So if you already did it, I can only will bring in a kinda noob POV! Thanks for the link, I will run it.

The talk wasn't recorded, and putting it in a livebook format would be great!

To be clear, FFT can execute the convolution (with some restrictions) but it's important to keep in mind that each is its own primitive in Nx

ndrean:
In Scipy, I just had a look, there make some quick calculations to predict which method, direct or via FFT, is faster.

<https://github.com/scipy/scipy/blob/92d2a8592782ee19a1161d0bf3fc2241ba78bb63/scipy/signal/_signaltools.py#L1165>

and conv = IFFT( FFT x FFT(conjugate).

But for later!

polvalente (approved):

PR #1561, opened 2024-11-17
Feat tensor inspect pretty

fredguth:
This PR is not for merge, but for discussion. 
Nx.tensor `inspect` with `pretty` option renders like `lovely_tensors` python library.
e.g.
#Nx.Tensor<f32[2][128] n=256 x∈[0.006, 0.996] μ=0.517 σ=0.291>

I find it more useful than the current implementation.  Another option is to have a `custom_options[lovely:  true]` instead.

### DIFF
--- a/nx/lib/nx/tensor.ex
+++ b/nx/lib/nx/tensor.ex
@@ -220,7 +220,7 @@ defmodule Nx.Tensor do
       mean = tensor |> Nx.mean() |> Nx.to_number() |> :erlang.float()
       std = tensor |> Nx.standard_deviation() |> Nx.to_number() |> :erlang.float()
 
-      "n=#{n} " <>
+      " n=#{n} " <>
       "x∈[#{:io_lib.format("~.3f", [min])}, #{:io_lib.format("~.3f", [max])}] " <>
       "μ=#{:io_lib.format("~.3f", [mean])} σ=#{:io_lib.format("~.3f", [std])}"
     end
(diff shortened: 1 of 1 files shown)


nx/lib/nx/tensor.ex
```diff
@@ -210,48 +210,68 @@ defmodule Nx.Tensor do
...
 
-  defimpl Inspect do
+  defimpl Inspect, for: Nx.Tensor do
     import Inspect.Algebra
 
-    def inspect(
-          %{shape: shape, names: names, type: type, vectorized_axes: vectorized_axes} = tensor,
-          opts
-        ) do
+    defp format_stats(tensor) do
```
polvalente, line 216: Why are these specific stats chosen? Also, why not calculate them over each axis?

I think these are pretty meaningless for certain types of tensors. Identity, permutation, orthogonal tensors, for instance.

polvalente (changes requested):

nx/lib/nx/tensor.ex
```diff
@@ -210,48 +210,68 @@ defmodule Nx.Tensor do
...
+      n = Nx.size(tensor)
+      min = tensor |> Nx.reduce_min() |> Nx.to_number() |> :erlang.float()
```
polvalente, line 218: There's another conceptual problem here, im which Nx.Defn.Expr tensors will just break completely if this option is set.

If we decide to move forward with this, this mode should be implemented in the Nx.Backend inspect callback instead of if being placed here.

fredguth:
@polvalente Thanks for the review. 
The main idea is to transmit more information about the tensor in less text.   Humans tend not to grok raw data. 
The stats chose just give an overall sense of the domain and values and are there are many cases where they are meaningless. But 2 lines of the raw data are even less meaningful. 

I also don't know if its a good idea to use it as `:pretty` if pretty is being used for debug purposes and I did not even consider the Nx.Defn.Expr problem.  The other option that I  know of is using `custom_options`. Can I use custom_options without changing the Nx.tensor code?

I am new to Elixir, my experience is with PyTorch where I always "mokeypatch" the string representation of tensors using lovely_tensors. I find these stats very useful. If there is a way of changing the string representation without messing with Nx.tensor, I would prefer.

nx/lib/nx/tensor.ex
fredguth: This implementation is also not good if the tensor type is not float.

fredguth:
Thanks for the review. I will proceed in the way suggested.

PR #1564, opened 2024-12-07
docs: improve docs for Nx.conv/3

polvalente:


nx/lib/nx.ex
```diff
@@ -13000,6 +13004,29 @@ defmodule Nx do
   in the same way as with `:feature_group_size`, however, the input
   tensor will be split into groups along the batch dimension.
 
+
```
josevalim, line 13007:
```suggestion
```

Very important, I know. :D

nx/lib/nx.ex
```diff
@@ -12919,6 +12919,10 @@ defmodule Nx do
   of summing the element-wise products in the window across
   each input channel.
 
+  > #### Kernel Reflection {: .info}
```
josevalim, line 12922:
```suggestion
  > #### Kernel Reflection {: .info}
  >
```

josevalim (approved):

hunterboerner:
Assuming SciPy's convolve function is correct then these docs are wrong. To match SciPy's convolution you reverse the kernel, _don't apply conjugate_ , call `Nx.conv`, then reverse the output. This is being worked on in my fork of NxSignal: https://github.com/hunterboerner/nx_signal/tree/convolutions

PR #1566, opened 2024-12-12
use serving name as pg group name

SteffenDE:
The Nx documentation states that it is important to use the same Nx version when using distributed servings. This leads to a problem for people using blue green deployments and trying to upgrade Nx. Without some special handling, you'll often run into situations where both the old and new nodes run in the cluster at the same time. Also, if you previously tried to run different servings on different machines, this wasn't really possible, as Nx would choose a random machine from pg, so each machine in the cluster running Nx would always have to run all servings.

By separating the pg groups by serving name, we can run different servings on different nodes while keeping the cluster connected. This also allows us to upgrade Nx in a blue green deployment by encoding the Nx version into the serving name.

cc @josevalim

### DIFF
--- a/nx/lib/nx/serving.ex
+++ b/nx/lib/nx/serving.ex
@@ -1117,7 +1117,7 @@ defmodule Nx.Serving do
   end
 
   defp distributed_batched_run_with_retries!(name, input, retries) do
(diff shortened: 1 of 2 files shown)


josevalim (approved):

josevalim:

PR #1574, opened 2025-01-27
feat: improve precision for nx constants

polvalente:
If we rewrite the Nx defn AST during normalization to use the number values from :math or Complex for the arity-0 versions of i/e/pi, we can benefit from the previous pull requests on precision upcasting "for free".

This is implicit behavior, so I understand if we don't want this, but I believe it makes for a better UX.

### DIFF
--- a/nx/lib/nx/constants.ex
+++ b/nx/lib/nx/constants.ex
@@ -3,6 +3,10 @@ defmodule Nx.Constants do
   Common constants used in computations.
 
   This module can be used in `defn`.
+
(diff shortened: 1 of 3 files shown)


josevalim:
If the user defines their own constant, it won't have this special behaviour, right? So I think we should avoid this or have a mechanism that everyone can use, such as adding a `Nx.constant` or similar (maybe Nx.tensor already works like that?)

josevalim (approved):

PR #1575, opened 2025-01-28
feat: add better errors for invalid compiler configurations

polvalente:
Sometimes, users will try to use invalid combos such as using Torchx as a defn compiler.
This PR transforms the generic FunctionClauseError into a more actionable message.

### DIFF
--- a/nx/lib/nx/defn/compiler.ex
+++ b/nx/lib/nx/defn/compiler.ex
@@ -141,7 +141,7 @@ defmodule Nx.Defn.Compiler do
     case exception do
       %UndefinedFunctionError{module: compiler, function: ^name, arity: ^arity} ->
         raise ArgumentError,
-              "The expected compiler callback #{name}/#{arity} is missing. Please check that the module #{inspect(compiler)} is an Nx.Defn.Compiler."
+              "the expected compiler callback #{name}/#{arity} is missing. Please check that the module #{inspect(compiler)} is an Nx.Defn.Compiler."
 
       _ ->
         # This is not an error that should've been caught by this function, so we pass the exception along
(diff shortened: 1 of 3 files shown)


nx/lib/nx/defn/compiler.ex
```diff
@@ -120,12 +123,30 @@ defmodule Nx.Defn.Compiler do
...
     compiler.__jit__(fun, params, runtime_fun, args_list, opts)
+  rescue
+    e in [UndefinedFunctionError] ->
+      raise_missing_callback(e, :__jit__, 5, __STACKTRACE__)
+  end
+
+  defp raise_missing_callback(exception, name, arity, stacktrace) do
+    case exception do
+      %UndefinedFunctionError{module: compiler, function: ^name, arity: ^arity} ->
+        raise ArgumentError,
```
polvalente, line 143: @josevalim I used ArgumentError because this is directly caused by an invalid option being passed. If you have another suggestion for the exception type, I'm open to it!

nx/lib/nx/defn/compiler.ex
```diff
@@ -120,12 +123,30 @@ defmodule Nx.Defn.Compiler do
...
+              "The expected compiler callback #{name}/#{arity} is missing. Please check that the module #{inspect(compiler)} is an Nx.Defn.Compiler."
```
josevalim, line 144:
```suggestion
              "the expected compiler callback #{name}/#{arity} is missing. Please check that the module #{inspect(compiler)} is an Nx.Defn.Compiler."
```

nx/test/nx/defn/compiler_test.exs
```diff
@@ -0,0 +1,49 @@
+defmodule Nx.Defn.CompilerTest do
+  use ExUnit.Case, async: true
+
+  defmodule SomeInvalidServing do
+    def init(_, _, _) do
+      :ok
+    end
+  end
+
+  test "it raises an error if the __compile__ callback is missing" do
```
josevalim, line 10: I don't believe we use the `it` convention. :)

```suggestion
  test "raises an error if the __compile__ callback is missing" do
```

josevalim (approved):

nx/test/nx/defn/compiler_test.exs
polvalente: Probably autocomplete that got past by me -- I don't use it too

PR #1576, opened 2025-01-31
feat: create distributed function graph executor

polvalente:
This code is meant to be used to run a defn graph that has been split into multiple steps in a possibly distributed manner


nx/config/config.exs
```diff
@@ -5,3 +5,14 @@ import Config
...
+# config :logger,
+#   # handle_otp_reports: true,
+#   # handle_sasl_reports: true,
+#   backends: [:console]
+#   # sync_threshold: 50,
+#   # translate_otp_reports: true
+
+# config :logger, :console,
+#   format: "$time $metadata[$level] $message\n",
+#   metadata: [:node]
```
polvalente, line 18:
```suggestion
```

nx/lib/nx/sharding/partitioned_executor/function.ex
```diff
@@ -0,0 +1,111 @@
...
+  @moduledoc false
+
+  use GenServer
+
+  require Logger
+
+  # args is a list of {argument id, indices}
+  # results is a tuple with the results
+  # id is a unique identifier for the function
+  # function is a function of arity length(args) which returns a tuple of size length(results)
```
polvalente, line 11:
```suggestion
  # code is a function of arity 1 + length(extra_args) which returns a tuple of size length(results)
```

PR #1579, opened 2025-02-05
Bump ExDoc to v0.36

Benjamin-Philip:


josevalim:
We do a `mix deps.update` before releasing and that will fetch the latest, so no changes necessary for now. Thank you!

PR #1580, opened 2025-02-11
docs: autodiff docs

polvalente:
Adds a guide on automatic differentiation


nx/guides/advanced/automatic_differentiation.livemd
```diff
@@ -0,0 +1,182 @@
+# Automatic Differentation
+
+```elixir
+Mix.install([
+  {:nx, "~> 0.7"}
+])
+```
+
+## What is Function Differentiation?
```
josevalim, line 9: nitpick: let's make sure we are consistent with titles and if we use "What is function differentiation?" or "What is Function Differentiation?". FWIW, in the Elixir repo itself, we use the former, but we can follow another convention here.

nx/guides/advanced/automatic_differentiation.livemd
```diff
@@ -0,0 +1,182 @@
...
+
+Nx, through the `Nx.Defn.grad/2` and `Nx.Defn.value_and_grad/3` functions allows the user to differentiate functions that were defined through `defn`.
```
josevalim, line 11: For what is worth, we don't explain in this section what is function differentiation. We do talk about differentiation, gradient and derivative, but we don't explain any of the terms. It may be valuable to give a step back and provide a high-level overview. And then either reduce the amount of terms or explicitly explain them.

josevalim (commented):
Just some nits and ship it :)

nx/guides/advanced/automatic_differentiation.livemd
polvalente: I added 2 paragraphs explaining derivatives and gradients

PR #1581, opened 2025-02-19
Refactor EXLA NIFs to use Fine

jonatanklosko:
While working on [pythonx](https://github.com/livebook-dev/pythonx), I built a small C++ library to reduce the amount of boilerplate necessary for encoding/decoding terms between Elixir and C++, providing conveniences around resources, and more. I've just extracted it into [elixir-nx/fine](https://github.com/elixir-nx/fine).

This PR rewrites EXLA NIFs. I used this to further validate Fine, and given the amount of removed code, I believe it makes for a good case. The most important features is encoding/decoding based on NIF signature, and managing resources.

Also, while rewriting the resource management, I noticed a few possible memory leaks, which should now be addressed.


nx/lib/nx.ex
```diff
@@ -16823,11 +16823,11 @@ defmodule Nx do
 
       t = Nx.u8([10, 20, 30])
       Nx.to_pointer(t, mode: :local)
-      %Nx.Pointer{kind: :local, address: 1234, data_size: 3, handle: nil}
+      {:ok, %Nx.Pointer{kind: :local, address: 1234, data_size: 3, handle: nil}}
```
josevalim, line 16826:
```suggestion
      #=> {:ok, %Nx.Pointer{kind: :local, address: 1234, data_size: 3, handle: nil}}
```

nx/lib/nx.ex
```diff
@@ -16823,11 +16823,11 @@ defmodule Nx do
...
 
       t = Nx.s32([1, 2, 3])
       Nx.to_pointer(t, mode: :ipc)
-      %Nx.Pointer{kind: :ipc, address: nil, data_size: 32, handle: "some-ipc-handle"}
+      {:ok, %Nx.Pointer{kind: :ipc, address: nil, data_size: 32, handle: "some-ipc-handle"}}
```
josevalim, line 16830:
```suggestion
      #=> {:ok, %Nx.Pointer{kind: :ipc, address: nil, data_size: 32, handle: "some-ipc-handle"}}
```

josevalim (approved):

nx/lib/nx.ex
```diff
@@ -16823,11 +16823,11 @@ defmodule Nx do
 
       t = Nx.u8([10, 20, 30])
       Nx.to_pointer(t, mode: :local)
-      %Nx.Pointer{kind: :local, address: 1234, data_size: 3, handle: nil}
+      #=> {:ok, %Nx.Pointer{kind: :local, address: 1234, data_size: 3, handle: nil}}
```
jonatanklosko, line 16826: Btw. in retrospect we should probably return a pointer and raise on failure, because an error is likely a developer error, and the error EXLA returns is just a string, so it's pretty opaque. That would be breaking, so not sure. Maybe as part of v1?

cc @polvalente

jonatanklosko: Ok, I will send another PR soon!

exla/c_src/exla/exla_cuda.cc
```diff
@@ -5,24 +5,31 @@
...
-  std::vector<unsigned char> result(size);
-  memcpy(result.data(), &ipc_handle, size);
+  // Copy the memory handle to a buffer
+  std::string buffer;
+  buffer.resize(size);
+  memcpy(&(*(buffer.begin())), &ipc_handle, size);
 
-  return std::make_pair(result, status != cudaSuccess);
+  return buffer;
 }
```
polvalente, line 29: I don't think we should be using std::string here

jonatanklosko: `std::string` is actually just a sequence of `char`, the data is not said to have any encoding or similar, so it's basically Elixir binary. The buffer has 1 additional \0 byte at the end, so we can get `string.c_str()`, but if we use string as a buffer, this doesn't matter.

One thing though is that I did `&(*(buffer.begin()))` here, as it was elsewhere, but I later realised we should not do that (it works, but it is not the right thing to do). I will update it in the next PR.

Or is there any other reason not to use `std::string`?

jonatanklosko: The practical reason is that using a string, encoding is done automatically, otherwise we would need to convert the vector to binary term by hand (which is fine, but my point is that either of them is just buffer) :D

And for `host_ipc` the handle is an "actual" string, but in both cases we tread it as an opaque binary.

PR #1588, opened 2025-03-05
feat: allow explicitly disabling CUDA step

### LINKED ISSUE #1586: Improve EXLA's CUDA-conditional compilation
I recently faced an odd situation where I was running EXLA on a platform that isn't supported by pre-compiled XLA with CUDA, which caused the CPU libxla.so to be pulled, and this is expected.

However, `nvcc` and CUDA were available, and this caused EXLA to try to compile with CUDA support, only to fail at `exla_cuda.cc`  due to missing symbols. This was solved by removing the `cuda/bin` directory (and thus `nvcc`) from the path for just the `mix deps.compile` step.

We could fix this by trying to link against CUDA in EXLA directly, but I think solution should come in a few forms:

1. Provide some way for EXLA to detect whether the available libxla.so has CUDA support, so that we can default to the CPU-only behavior if needed.
2. Allow CUDA support to be explicitly turned off via an environment flag in EXLA's Makefile.

polvalente:
closes #1586


exla/mix.exs
```diff
@@ -33,7 +33,15 @@ defmodule EXLA.MixProject do
...
 
+        exla_cpu_only =
+          if System.get_env("EXLA_CPU_ONLY") in ["1", "true"] do
+            "true"
+          else
+            ""
+          end
+
         %{
+          "EXLA_CPU_ONLY" => exla_cpu_only,
```
jonatanklosko, line 44: If we leave it out of here, the check should still work?

jonatanklosko (approved):

exla/mix.exs
polvalente: I guess so. Originally I was checking for true under a different variable name, but now its only an ifdef

exla/Makefile
```diff
@@ -36,27 +36,29 @@ else
...
 
+NVCC := $(CXX)
+NVCCFLAGS = $(CFLAGS)
 LDFLAGS = -L$(XLA_EXTENSION_LIB) -lxla_extension -shared -fvisibility=hidden
 
 ifeq ($(CROSSCOMPILE),)
 	# Interrogate the system for local compilation
 	UNAME_S = $(shell uname -s)
 
+ifdef ($(EXLA_CPU_ONLY),)
```
jyc, line 47: Should this be ifndef? I see "EXLA_CPU_ONLY is set, skipping nvcc step" in an environment where I haven't set EXLA_CPU_ONLY

polvalente: Yup. Fixing it now.

polvalente: https://github.com/elixir-nx/nx/commit/5ecc82f9cbacdf87712640aa90d685c092918f2e

Fixed it. This is a remnant from the original solution for checking equality. The proper usage for ifdef/ifndef doesn't use the same syntax

jyc: Whoa! Thanks for the quick fix. Hope you have a good weekend!

PR #1594, opened 2025-03-17
test(exla): add more tests for LinAlg functions 

polvalente:
- **test: cholesky**
- **--version**
- **test: determinant**
- **test: matrix_power**
- **test: lu**
- **test: svd**
- **test: pinv**
- **chore: move triangular solve test to other category**

### DIFF
--- a/exla/test/exla/nx_linalg_doctest_test.exs
+++ b/exla/test/exla/nx_linalg_doctest_test.exs
@@ -1,5 +1,6 @@
 defmodule EXLA.MLIR.NxLinAlgDoctestTest do
   use EXLA.Case, async: true
+  import Nx, only: :sigils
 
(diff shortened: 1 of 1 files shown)


exla/test/exla/nx_linalg_doctest_test.exs
```diff
@@ -91,4 +88,318 @@ defmodule EXLA.MLIR.NxLinAlgDoctestTest do
       end
     end
   end
+
+  describe "cholesky" do
```
polvalente, line 92: The tests added here are generally extracted from the Nx linalg tests in nx itself. @TomasPegado is working on fixing triangular_solve and then we should also extract triangular_solve and solve tests to here as well.

PR #1607, opened 2025-05-08
Increase Torch to latest version 2.7

stewartj3:
Update Torch version to 2.7, and address namespace changes that occured in 2.6 which caused compilation failures. The linalg namespace appears to have been removed from the C++ api.


josevalim (approved):

polvalente (approved):
I'm assuming CUDA still works normally, but I can't test this until next week.

PR #1615, opened 2025-06-17
Fix hook order inside while loop

jonatanklosko:
We use tokens to order hooks, but currently when we enter `while` loop, we don't use the latest token, we always use the starting one. In certain cases this leads to outfeeds in unexpected order and results in an error.

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -312,14 +312,14 @@ defmodule EXLA.Defn do
        ) do
     [initial_arg, _arg, pred, body] = args
 
+    {initial, cache} = recur_composite(initial_arg, state, cache)
+
     initial =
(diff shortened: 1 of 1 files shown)


exla/lib/exla/defn.ex
```diff
@@ -312,15 +312,15 @@ defmodule EXLA.Defn do
        ) do
     [initial_arg, _arg, pred, body] = args
 
+    {initial, cache} = recur_composite(initial_arg, state, cache)
+
     initial =
       if token = get_token(cache) do
```
jonatanklosko, line 318: We need to recur on `initial_arg` first and use the cached token. Currently we get the token first, so it doesn't account for all the prior computation.

josevalim (approved):

PR #1619, opened 2025-07-18
fix: Nx.LinAlg.lu implementation

### LINKED ISSUE #1618: Unexpected determinant for 4X4 matrix
When upgrading to 0.10.0 we run into an expected 0 determinant for the following matrix

```
tensor = Nx.tensor([
  [-1.0, 1.0, -1.0, 0.0],
  [1.0, 0.0, 0.0, 0.0],
  [0.0, 0.0, 0.0, 1.0],
  [0.0, 1.0, 0.0, 0.0]
], type: :f64)

result = Nx.LinAlg.determinant(tensor)

IO.inspect(result)
```

With result

```
#Nx.Tensor<
  f64
  0.0
>
```

We had this trigger in our unit tests when upgrading to 0.10.0, but I'm having trouble at the moment reverting to a prior version to get a different value. Trying to figure this out at the moment.

But in the meantime -- numpy arrives at a 1.0 determinant for this matrix.

```
import numpy as np

if __name__ == "__main__":

    arr = np.array([
        [-1.0, 1.0, -1.0, 0.0],
        [1.0, 0.0, 0.0, 0.0],
        [0.0, 0.0, 0.0, 1.0],
        [0.0, 1.0, 0.0, 0.0]
    ])

    det = np.linalg.det(arr)

    print("Array:")
    print(arr)
    print("\nDeterminant:", det)
```

`Determinant: 1.0`

And I think 1.0 is correct, the matrix certainly doesn't seem ill-formed.

polvalente:
This PR refactors Nx.LinAlg.lu for correctness.

closes #1618


ja-b:
@polvalente thanks again for the help here! 🙏

polvalente:
> @polvalente thanks again for the help here! 🙏 

Sure! Thanks for reporting the bug.
I added this matrix as a regression test for both determinants and LU.

To be honest I'm not exactly sure what was causing this problem. I ended up rewriting the whole thing.

ja-b:
I'll be on the lookout for any other weirdness should it arise, and noticed the regression test beforehand. 😄

PR #1624, opened 2025-08-06
feat(Nx.Defn.Graph): allow splitting based on an accumulator

polvalente:
- Adds an optional accumulator argument to the split function
- Improves some of the argument reindexing


nx/lib/nx/defn/graph.ex
```diff
@@ -63,9 +71,28 @@ defmodule Nx.Defn.Graph do
...
+
+      stages = Nx.Defn.Graph.split(expr, 0, split_fn)
   """
   def split(expr, expr_split_fn) when is_function(expr_split_fn, 1) do
-    {chain, _, _} = __split__(expr, expr_split_fn)
+    {chain, _, _} = __split__(expr, nil, fn node, acc -> {expr_split_fn.(node), acc} end)
+    chain
+  end
+
+  def split(expr, initial_acc, expr_split_fn) when is_function(expr_split_fn, 2) do
```
josevalim, line 94: This is effectively a separate function. So it needs a separate `@doc "..."` bit. Which I think is good, we don't need to mix the examples in the doc above.

PR #1644, opened 2025-11-14
refactor: improve Nx.Defn.Evaluator debugging usability

Chapaman:
Changes `evaluator.ex` to produce `.exs` files that can be executed for comparison between nodes (see save_as documentation). 
Creates a `backend_comparison.livemd` file to document comparison between backends for the evaluator. 
Adds the `backend_comparison` and `complex_fft` to the `mix.exs` file

### DIFF
--- a/nx/lib/nx/defn/evaluator.ex
+++ b/nx/lib/nx/defn/evaluator.ex
@@ -539,8 +539,8 @@ defmodule Nx.Defn.Evaluator do
 
     inspect_opts =
       case inspect_limit do
-        nil -> []
-        limit -> [limit: limit]
+        nil -> [custom_options: [print_id: true]]
+        limit -> [custom_options: [print_id: true], limit: limit]
       end
 
     args_code =
--- a/nx/test/nx/defn/evaluator_test.exs
+++ b/nx/test/nx/defn/evaluator_test.exs
@@ -775,7 +775,7 @@ defmodule Nx.Defn.EvaluatorTest do
       opts = [compiler: Nx.Defn.Evaluator, debug_options: [inspect_limit: 5]]
       output = ExUnit.CaptureIO.capture_io(fn -> debug_test_fun(x, y, opts) end)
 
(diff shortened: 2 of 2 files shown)


nx/guides/advanced/backend_comparison.livemd
```diff
@@ -0,0 +1,267 @@
...
+IO.inspect(reconstructed, label: "Reconstructed tensor")
+IO.puts("\nProperties:")
+IO.puts("  Type: #{inspect(reconstructed.type)}")
+IO.puts("  Shape: #{inspect(reconstructed.shape)}")
+IO.puts("  Backend: #{inspect(reconstructed.data.__struct__)}")
+```
+
+## Running with Backend B
+
+Now let's run the same computation with the swapped operations. We leave `Nx` on its default backend, but temporarily enable the Mimic stubs so the evaluator will capture the modified behaviour.
```
polvalente:
```suggestion
Now let's run the same computation with the swapped operations. We leave `Nx` on its default backend, but temporarily enable the Mimic stubs so the evaluator will capture the modified behaviour.

In practice, this would be where we call a totally different backend to compare against the reference.
```

nx/lib/nx/defn/evaluator.ex
```diff
@@ -535,36 +535,105 @@ defmodule Nx.Defn.Evaluator do
...
+
+    # Result:
+    #{result_code}
+    """
+  end
+
+  defp serialize_tensor(%Nx.Tensor{data: %Expr{id: id}} = _tensor) when is_reference(id) do
+    # This is an unevaluated expression, not a concrete tensor
+    # Show the Node ID so users can find which file contains this tensor
+    id_str = :erlang.ref_to_list(id) |> List.to_string() |> String.replace(["#Ref<", ">"], "")
```
polvalente: Extract this line to a defp so we can use it when defining the filename on line 636 (but keep the replace with _ there too)

PR #1646, opened 2025-12-05
feat(sharding): add minimal EXLA support through Shardy

polvalente:
This PR adds support for Nx.Defn.shard_jit and related features to EXLA

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -247,7 +247,7 @@ defmodule EXLA.Defn do
          outputs,
          outfeed,
          run_options,
-         _is_sharded?
+         is_sharded?
        )
        when Outfeed.will_outfeed(outfeed) do
     if is_sharded? do


polvalente:
@josevalim We (me and @Chapaman) merged main into this branch and made it work with the new shard_jit things we added, expecting pre-sliced inputs as we previously discussed.

After we merge this we'll finish up the already-in-progress PR for all_gather

exla/lib/exla.ex
```diff
@@ -301,6 +301,30 @@ defmodule EXLA do
...
+      # Pass sharded inputs (one arglist per partition)
+      fun.([[Nx.tensor([1, 2, 3])], [Nx.tensor([4, 5, 6])]])
+      #=> #Nx.Tensor<
+      #     s32[6]
+      #     [2, 4, 6, 8, 10, 12]
+      #   >
+
+  ## Options
+
+    * `:input_shardings` - a list of lists of axis indices to shard the input tensors on.
```
josevalim: We say list of lists but the arugment is list of lists of lists, but i think this is confusing. I would use this:

    [%{tensor_dimension => list_of_mesh_dimensions}]

This way we add some variation, and we can allow the tensor dimension to be either an index or a name.

exla/lib/exla/defn.ex
```diff
@@ -35,6 +35,163 @@ defmodule EXLA.Defn do
...
+        # Build a mapping from dimension indices to mesh axes
+        dim_to_axes_map =
+          Enum.reduce(input_sharding, %{}, fn {tensor_dim, mesh_axes}, acc ->
+            # Normalize tensor dimension (could be integer, negative integer, or atom name)
+            dim_index = normalize_axis(tensor_dim, var.shape, var.names)
+
+            # Check for duplicate dimension specifications
+            if Map.has_key?(acc, dim_index) do
+              raise ArgumentError,
+                    "tensor dimension #{inspect(tensor_dim)} (index #{dim_index}) was specified multiple times"
```
josevalim: Can this still happen now that they are maps?

exla/lib/exla/defn.ex
```diff
@@ -35,6 +35,163 @@ defmodule EXLA.Defn do
...
+    {list_format, multipliers}
+  end
+
+  # Normalize a tensor dimension specification to an integer index.
+  # Follows the same strategy as Nx.Shape.normalize_axis/4.
+  # Supports:
+  # - Non-negative integers: 0, 1, 2, ...
+  # - Negative integers: -1 (last), -2 (second to last), ...
+  # - Atom names: :batch, :features, etc.
+  defp normalize_axis(axis, shape, _names) when is_integer(axis) do
```
josevalim: Can't we use Nx.Shape.normalize_axis or something similar?

exla/lib/exla/defn.ex
```diff
@@ -257,6 +452,30 @@ defmodule EXLA.Defn do
...
+                      "input sharding configuration provided but no device mesh was provided"
+              end
+
+              # Apply sharding annotations to function arguments if provided
+              if input_shardings != [] do
+                num_comp_args = length(comp_typespecs)
+
+                if length(input_shardings) != num_comp_args do
+                  raise ArgumentError,
+                        "expected #{num_comp_args} input sharding configuration (one per argument), got #{length(input_shardings)}"
```
josevalim: Didn't we already validate this?

exla/lib/exla/defn.ex
polvalente: For example a tensor with [:batch, :features] and the map has -2, 0 and :batch as keys, they all refer to the same axis

exla/lib/exla/defn.ex
polvalente: This is effectively the same logic, though

josevalim: Shall we expose it as `Nx.normalize_axis`? And then we build a tensor template and call it?

PR #1650, opened 2025-12-26
fix: added do_shard_jit_apply/4 + prepare_options/3 for __shard_jit__

Chapaman:
do_shard_jit_apply was not defined yet. Also, compiler function __shard_jit__ passes a mesh to prepare_options, which is now accepting it.

### DIFF
--- a/nx/lib/nx/defn/compiler.ex
+++ b/nx/lib/nx/defn/compiler.ex
@@ -181,10 +181,6 @@ defmodule Nx.Defn.Compiler do
     {compiler, &runtime_fun(&1, fun, compiler), opts}
   end
 
-  defp prepare_options(fun, _mesh, opts) do # This is a placeholder for sharding implementation
-    prepare_options(fun, opts)
-  end
-
   defp runtime_fun(args, fun, compiler) do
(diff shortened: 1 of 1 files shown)


nx/lib/nx/defn/compiler.ex
```diff
@@ -181,6 +181,10 @@ defmodule Nx.Defn.Compiler do
     {compiler, &runtime_fun(&1, fun, compiler), opts}
   end
 
+  defp prepare_options(fun, _mesh, opts) do # This is a placeholder for sharding implementation
+    prepare_options(fun, opts)
+  end
```
polvalente: This a placeholder for what exactly? I think we could just not pass the mesh and reuse the same `prepare_options` directly

polvalente (approved):

nx/lib/nx/defn/compiler.ex
Chapaman: > This a placeholder for what exactly? I think we could just not pass the mesh and reuse the same `prepare_options` directly

You're right! No need to pass mesh to prepare_options. maintaining it as arity /2 only

PR #1657, opened 2026-01-08
feat: Add    comprehensive FP8 type support (E4M3FN and E5M2)

nyo16:
Add FP8 E4M3FN type support

  Adds {:f8_e4m3fn, 8} type for the FP8 E4M3FN format from the OCP 8-bit Floating Point Specification (OFP8).

  E4M3FN format

  - 1 sign bit, 4 exponent bits, 3 mantissa bits (bias: 7)
  - No infinity ("FN" = Finite, No infinities)
  - Only S.1111.111 (0x7F/0xFF) is NaN
  - Range: -448.0 to 448.0

  Changes

  - nx/lib/nx/type.ex - Type spec, validation, and binary functions
  - nx/lib/nx/shared.ex - Read/write support for E4M3FN encoding
  - nx/test/nx/type_test.exs - Tests for type and special values
  - exla/ - MLIR and NIF support for F8E4M3FN

  Special value handling

  Since E4M3FN has no infinity representation, infinity_binary/1 and neg_infinity_binary/1 return NaN (0x7F), and max_binary/1 / min_binary/1 return finite max/min values.

  Values verified against OFP8 spec and Python ml_dtypes library.

  The existing {:f, 8} type (E5M2) remains unchanged.

### DIFF
--- a/exla/c_src/exla/exla_nif_util.h
+++ b/exla/c_src/exla/exla_nif_util.h
@@ -202,7 +202,9 @@ private:
       } else if (element_type.isBF16()) {
         type_name = exla::atoms::bf;
         type_size = 16;
-      } else if (element_type.isFloat8E4M3FN()) {
+      // Float8E4M3FNType doesn't have a convenience method like isBF16() on the
+      // base mlir::Type class, so we use mlir::isa<> for type checking instead.
+      } else if (mlir::isa<mlir::Float8E4M3FNType>(element_type)) {
         type_name = exla::atoms::f8_e4m3fn;
         type_size = 8;
       } else if (auto float_type =


nyo16:
Awesome feedback, thank you for your input @polvalente. Its great and i am learning the internals. I will work to simplify the PR and follow your insights.

josevalim:
@nyo16 note we are not going to introduce a new `f8_e5m2` type. We will instead document that f8 is e5m2. We will simply add a new `f8_e4m3fn`. My suggestion is actually to start from scratch, given most of the previous changes were no longer necessary. :)

polvalente:
I wonder if the original intention was to have :infinity map onto max finite for those unrepresentable values before.
I think these are only called when we explicitly encounter :infinity/:neg_infinity or their Nx.Constants counterparts.

exla/lib/exla/mlir/value.ex
```diff
@@ -1010,6 +1012,11 @@ defmodule EXLA.MLIR.Value do
     binary_part(<<x::float-big-16>>, 0, 1)
   end
 
+  defp f8E4M3FN_to_big(x) do
+    # E4M3FN encoding using Nx.Shared helper
+    Nx.Shared.write_finite_f8_e4m3fn(x)
```
josevalim: The `Nx.Shared` module is private to the Nx project. My suggestion is to add a new module called Nx.Floating, which can be public, and it has functions:

* load_bf16
* dump_bf16
* load_f8E5M2
* dump_f8E5M2
* and so on

nx/lib/nx/type.ex
```diff
@@ -107,13 +113,17 @@ defmodule Nx.Type do
...
+  def max_finite_binary({:f8_e4m3fn, 8}), do: <<0x7E::8-native>>
   def max_finite_binary({:f, 16}), do: <<0x7BFF::16-native>>
   def max_finite_binary({:f, 32}), do: <<0x7F7FFFFF::32-native>>
   def max_finite_binary({:f, 64}), do: <<0x7FEFFFFFFFFFFFFF::64-native>>
 
   @doc """
   Returns the maximum possible value for the given type.
   """
+  # E4M3FN has no infinity, return finite max
+  def max_binary({:f8_e4m3fn, 8} = type), do: max_finite_binary(type)
```
josevalim: Same as `min_binary`.

josevalim:
@nyo16 if it is any easier, we can break this into two pull requests, one for Nx and another for EXLA/Torchx.

nyo16:
> Or perhaps even better, what if we have a PR that simply adds the Nx.Floating model with `load_f8_e4m3fn` and `dump_f8_e4m3fn`? Then doing the rest is very straightforward and I can take care of it later.

I will try this first!

nx/lib/nx/shared.ex
```diff
@@ -264,6 +304,67 @@ defmodule Nx.Shared do
...
+
+      cond do
+        exp <= 0 ->
+          # Denormalized: value = mantissa/8 * 2^(-6)
+          mantissa = round(abs_val / :math.pow(2, -6) * 8)
+          <<sign::1, 0::4, min(7, mantissa)::3>>
+
+        exp >= 15 ->
+          # Overflow to max finite (not NaN, since E4M3FN saturates in our impl)
+          <<sign::1, 15::4, 6::3>>
```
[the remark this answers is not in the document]
nyo16: I've updated infinity_binary/1 and neg_infinity_binary/1 in Nx.Type to also saturate to max/min finite values (0x7E/0xFE) for consistency.

PR #1663, opened 2026-01-30
Basic all_gather implementation

Chapaman:
Implements Nx.Defn.Kernel.all_gather/2 to gather sharded tensor data across mesh partitions during distributed execution.
Changes
Nx
  Add all_gather/2 in defn/kernel.ex and defn/expr.ex with sharding semantics
  Add evaluator support for all_gather in defn/evaluator.ex
EXLA
  Lower all_gather to stablehlo.all_gather in defn.ex and mlir/value.ex
Test
  EXLA.Defn.ShardingTest: "generates correct MLIR with all_gather" checks MLIR generation and shard_jit output across a 2×2 mesh along axis 0 and 1

### DIFF
--- a/nx/test/nx/floating_test.exs
+++ b/nx/test/nx/floating_test.exs
@@ -156,7 +156,7 @@ defmodule Nx.FloatingTest do
       {0x7F, :nan},
       # Negative values (sign bit = 1)
       # Denormalized (exponent = 0): value = -mantissa/8 * 2^-6
-      if(+0.0 === -0.0, do: {0x80, 0.0}, else: {0x80, -0.0}),
+      if(+0.0 === -0.0, do: {0x00, -0.0}, else: {0x80, -0.0}),
       {0x81, -0.001953125},
       {0x82, -0.00390625},
       {0x83, -0.005859375},


nx/lib/nx/defn/expr.ex
```diff
@@ -1166,6 +1166,33 @@ defmodule Nx.Defn.Expr do
...
+
+    # Calculate output shape by multiplying the gather dimension by group_size
+    input_shape = tensor.shape
+    output_shape =
+      input_shape
+#      |> Tuple.to_list()
+#      |> List.update_at(all_gather_dim, &(&1 * group_size))
+#      |> List.to_tuple()
+
+    # Create output tensor with the new shape
```
polvalente: There are a few unused values here due to the stray comments that should all be removed. Also, just pass tensor as out directly

nx/lib/nx/defn/kernel.ex
```diff
@@ -1669,6 +1669,32 @@ defmodule Nx.Defn.Kernel do
...
+  Gathers tensors from all replicas along a specified dimension.
+
+  This operation concatenates tensors from multiple replicas/devices along
+  the specified dimension. Requires a backend that supports multi-device operations.
+
+  ## Parameters
+
+    * `tensor` - The input tensor to gather
+    * `all_gather_dim` - The dimension along which to gather
+    * `replica_groups` - 2D list defining how replicas are grouped (required)
```
polvalente: I'm not sure if this is the terminology we want to surface here. For now, let's make the function `all_gather(tensor, opts)` and defer the documentation of `opts` to the specific backend or compiler.

And in EXLA we should add a new section to the moduledoc of `EXLA` describing Sharding

polvalente (commented):
This is looking great! I think we need more tests in both Nx and EXLA

polvalente (approved):

PR #1670, opened 2026-02-16
Require OTP 27+, fix warnings on Elixir v1.20

josevalim:


josevalim:
Torchx has been broken for a while, but because we were not running the suite when nx changed, it was uncaught.

PR #1676, opened 2026-02-28
Fix BinaryBackend.select_and_scatter crash with :same padding (#1675)Fix/select and scatter negative padding

### LINKED ISSUE #1675: BinaryBackend.select_and_scatter/8 crashes with negative padding during max_pool gradient
# Description

Nx.Defn.value_and_grad crashes when computing gradients through a model containing max_pool with :same padding on BinaryBackend. The backward pass for max_pool calls select_and_scatter, which computes a negative padding value, causing :lists.duplicate/2 to fail with a FunctionClauseError.

  The same model works correctly with EXLA as the JIT compiler.

  Minimal reproduction

```
Mix.install([{:axon,` "~> 0.7"}, {:nx, "~> 0.10"}])

  model =
    Axon.input("input", shape: {nil, 16, 16, 1})
    |> Axon.conv(4, kernel_size: 7, strides: [2, 2], padding: :same, name: "conv")
    |> Axon.max_pool(kernel_size: {3, 3}, strides: [2, 2], padding: :same, name: "pool")
    |> Axon.flatten()
    |> Axon.dense(4, name: "fc")

  {init_fn, predict_fn} = Axon.build(model, mode: :inference)
  template = %{"input" => Nx.template({2, 16, 16, 1}, :f32)}
  model_state = init_fn.(template, Axon.ModelState.empty())

  # Forward pass works fine
  output = predict_fn.(model_state, %{"input" => Nx.broadcast(0.1, {2, 16, 16, 1})})
  IO.inspect(Nx.shape(output))  # => {2, 4}

  # Gradient crashes
  loss_fn = fn params ->
    state = %{model_state | data: params}
    output = predict_fn.(state, %{"input" => Nx.broadcast(0.1, {2, 16, 16, 1})})
    Nx.mean(output)
  end

  {_loss, _grads} = Nx.Defn.value_and_grad(model_state.data, loss_fn)
```
 #  Error

```
  ** (FunctionClauseError) no function clause matching in :lists.duplicate/2

      The following arguments were given to :lists.duplicate/2:

          # 1
          -36

          # 2
          <<0, 0, 0, 0>>

      (stdlib 6.2.2.2) lists.erl:510: :lists.duplicate/2
      (nx 0.10.0) lib/nx/binary_backend.ex:1669: Nx.BinaryBackend.select_and_scatter/8
      (nx 0.10.0) lib/nx/defn/evaluator.ex:461: Nx.Defn.Evaluator.eval_apply/4
```

#  Expected behavior

  value_and_grad should compute gradients through the max_pool layer without crashing.

#  Root cause

  BinaryBackend.select_and_scatter/8 computes negative padding values when the spatial dimensions after convolution are small relative to the pool kernel size. The -36 suggests the padding calculation (output_size - 1) * stride + kernel_size - input_size goes negative.

#  Workaround

  Use EXLA as the JIT compiler:
  Nx.Defn.jit(fn params -> Nx.Defn.value_and_grad(params, loss_fn) end, compiler:
  EXLA).(model_state.data)

# Look at 
lib/nx/binary_backend.ex:1669 in select_and_scatter/8.

#  Environment

  - Nx: 0.10.0
  - Elixir: 1.18.4
  - OTP: 27
  - OS: Linux (WSL2)

blasphemetheus:
## Summary

`select_and_scatter/8` computes `absolute_index` in the padded coordinate space but uses it against the unpadded `output_shape`, producing negative offsets that crash `List.duplicate/2`.

Fix: subtract low-padding from indices to map back to original coordinates, and filter out indices that land in the padding region.

## Changes

**`nx/lib/nx/binary_backend.ex`**
- Subtract per-dimension low-padding from `absolute_index` after computing it in padded space
- Filter `output_windows` to discard indices outside original tensor bounds

**`nx/test/nx/defn/grad_test.exs`**
- Add regression test: `grad(sum(window_max(x, {1,3,3,1}, padding: :same, strides: [1,2,2,1])))` on a `{1,4,4,1}` input — verifies output shape, finiteness, and gradient sum equals number of output elements

## Test Results

```
1170 tests, 0 failures, 1 skipped
1351 doctests, 0 failures
```

Closes #1675

### DIFF
--- a/nx/test/nx/defn/grad_test.exs
+++ b/nx/test/nx/defn/grad_test.exs
@@ -1435,7 +1435,9 @@ defmodule Nx.Defn.GradTest do
     defn grad_sum_sin_window_max_cos_same(t) do
       grad(
         t,
-        &Nx.sum(Nx.sin(Nx.window_max(Nx.cos(&1), {1, 3, 3, 1}, padding: :same, strides: [1, 2, 2, 1])))
+        &Nx.sum(
+          Nx.sin(Nx.window_max(Nx.cos(&1), {1, 3, 3, 1}, padding: :same, strides: [1, 2, 2, 1]))
+        )
       )
     end
 
--- a/nx/test/nx_test.exs
+++ b/nx/test/nx_test.exs
@@ -890,7 +890,9 @@ defmodule NxTest do
 
     test "computes a same padding window max 1D (#1675)" do
       t = Nx.iota({8}, type: {:f, 32})
-      assert Nx.window_max(t, {3}, padding: :same, strides: [2]) == Nx.tensor([1.0, 3.0, 5.0, 7.0])
+
+      assert Nx.window_max(t, {3}, padding: :same, strides: [2]) ==
+               Nx.tensor([1.0, 3.0, 5.0, 7.0])
     end
 
     test "computes a same padding window max 2D (#1675)" do


nx/lib/nx/binary_backend.ex
```diff
@@ -1625,11 +1629,18 @@ defmodule Nx.BinaryBackend do
...
+        # original (unpadded) coordinates by subtracting low-padding
+        padded_absolute_index =
           anchor
           |> Enum.zip(offset_from_anchor)
           |> Enum.map(fn {x, y} -> x + y end)
 
+        absolute_index =
+          padded_absolute_index
+          |> Enum.zip(low_pads)
+          |> Enum.map(fn {idx, lo} -> idx - lo end)
```
polvalente:
```suggestion
        absolute_index = Enum.zip_with(padded_absolute_index, low_pads,  &-/2)
```

PR #1677, opened 2026-03-02
Runtime call gpu

### LINKED ISSUE #1031: Allow on_host/2 to execute certain operations on host

Chapaman:
closes #1031


polvalente (commented):
This is basically in the right direction! I gotta check if there's more code we can share between the cuda and host paths.

polvalente (approved):

PR #1681, opened 2026-03-09
Fix/triangular solve conjugate grad

### LINKED ISSUE #1680: `triangular_solve` gradient missing `:conjugate` case for `transform_a`
## Description

The gradient implementation for `triangular_solve` in `grad.ex` only handles `:none` and `:transpose` in its `case opts[:transform_a]` blocks, but the forward pass also accepts `:conjugate`. When complex number support is fully available, computing gradients through `triangular_solve` with `transform_a: :conjugate` will crash with a `MatchError`.

The gradient docs at line 1046 of `grad.ex` explicitly mention all three transforms:

> The transformations are :none, :transpose and :conjugate, all of which can be applied beforehand to the a matrix.

## Location

`nx/lib/nx/defn/grad.ex` — two `case opts[:transform_a]` blocks:

1. Lines 1043-1048: transforms `a_input` before computing the gradient
2. Lines 1091-1095: transforms `da` after computing the gradient

Both are missing the `:conjugate` clause.

## Impact

Currently latent — the forward pass in `lin_alg.ex` raises `"complex numbers not supported yet"` for `:conjugate`, so the gradient code path can't be reached on BinaryBackend today. But when complex support lands, this will be a `MatchError` crash.

## Fix

Add `:conjugate -> Nx.conjugate(...)` to both case blocks. Fix with tests on this fork:

https://github.com/blasphemetheus/nx/tree/fix/triangular-solve-conjugate-grad

The branch includes:
- The two-line fix in `grad.ex`
- A BinaryBackend test verifying the error comes from the forward pass (not a `MatchError` in the gradient)
- EXLA tests with complex c64 tensors for verifying the gradient on backends that support `transform_a: :conjugate`

blasphemetheus:
# Support transform_a: :conjugate in triangular_solve

## Summary

Closes #1680.

Adds support for `transform_a: :conjugate` in `Nx.LinAlg.triangular_solve/3` across the Nx core API, BinaryBackend, and EXLA backend, including gradient support.

- **Nx core** (`nx/lib/nx/lin_alg.ex`): Removed the guard that rejected `:conjugate`, allowing it as a valid `transform_a` option alongside `:none` and `:transpose`. Updated doctest accordingly.
- **BinaryBackend** (`nx/lib/nx/binary_backend/matrix.ex`): Added conjugation of matrix A before solving when `transform_a: :conjugate` is set (conjugation without transpose). Includes a `conjugate_matrix/1` helper using `Complex.conjugate/1`.
- **EXLA backend** (`exla/lib/exla/defn.ex`): StableHLO's `TriangularSolve` supports `no_transpose`, `transpose`, and `adjoint` (conjugate transpose) but not conjugate-only. For complex types, we pre-conjugate A and then solve with `transform_a: :none`. For real types, conjugate is a no-op so we just use `:none`.
- **Gradient** (`nx/lib/nx/defn/grad.ex`): Added `:conjugate` clause to the `triangular_solve` gradient rule.

## Notes

- For real tensors, `:conjugate` is mathematically equivalent to `:none` (conjugation is a no-op on reals). Tests verify this equivalence.
- Complex tests use `c128` precision to ensure results are GPU-portable without needing loose tolerances.
- Torchx does not currently implement `triangular_solve` with conjugate support (it also lacks `left_side: false`). This can be addressed in a follow-up.

## Test plan

- [x] `nx/test/nx/lin_alg_test.exs` — forward pass with real tensors (lower/upper, left_side true/false)
- [x] `nx/test/nx/defn/grad_test.exs` — gradients wrt A and B with real tensors (lower/upper, left_side true/false)
- [x] `exla/test/exla/defn/expr_test.exs` — forward pass (real + complex c128), gradients (wrt A, wrt B, composed), lower/upper, left_side true/false
- [x] All existing `mix test` suites pass (nx and EXLA)

Co-Authored-By: Claude Opus 4.6

### DIFF
--- a/exla/test/exla/defn/expr_test.exs
+++ b/exla/test/exla/defn/expr_test.exs
@@ -3875,8 +3875,11 @@ defmodule EXLA.Defn.ExprTest do
         da,
         Nx.tensor(
           [
-            [Complex.new(-4.0, -4.6666665), Complex.new(-0.33333334, 0.33333334),
-             Complex.new(-2.0, 0)],
+            [
+              Complex.new(-4.0, -4.6666665),
+              Complex.new(-0.33333334, 0.33333334),
+              Complex.new(-2.0, 0)
+            ],
             [0, Complex.new(-0.11111112, -0.11111112), Complex.new(0, -0.6666667)],
             [0, 0, Complex.new(-2.6666667, -3.3333333)]
           ],
@@ -3939,8 +3942,11 @@ defmodule EXLA.Defn.ExprTest do
           [
             [Complex.new(-1.25, -2.75), 0, 0],
             [Complex.new(0.5, 0.5), Complex.new(0, 0.33333334), 0],
-            [Complex.new(-2.5, 0.5), Complex.new(-1.0, -0.6666667),
-             Complex.new(-5.6666665, -2.3333333)]
+            [
+              Complex.new(-2.5, 0.5),
+              Complex.new(-1.0, -0.6666667),
+              Complex.new(-5.6666665, -2.3333333)
+            ]
           ],
           type: :c64
         ),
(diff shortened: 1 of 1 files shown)


polvalente (approved):
Looking good! I think we just need those value assertions I mentioned and we can merge!

PR #1683, opened 2026-03-09
Fix CallbackServer process leak on repeated JIT compilation (get rid of CallbackServer)

### LINKED ISSUE #1682: EXLA.CallbackServer process leak on repeated JIT compilation
Every __compile__ call starts a CallbackServer process even when the compiled graph has no :runtime_call nodes. In long-running workloads with repeated JIT calls (~25K+), this exhausts the BEAM process limit. The server could be started lazily only when needed. 

(making issue just for PR to reference)

blasphemetheus:
# Fix CallbackServer process leak on repeated JIT compilation

## Summary

Closes #1682.

Every `__compile__` call unconditionally starts a `CallbackServer` process, even when the compiled graph contains no `:runtime_call` nodes. In long-running workloads with repeated JIT calls (~25K+), this exhausts the BEAM process limit.

The fix starts the `CallbackServer` lazily — only when the first `:runtime_call` node is encountered during compilation. The PID is cached in the process dictionary so subsequent `:runtime_call` nodes in the same compilation reuse the same server.

### Changes

- **`exla/lib/exla/defn.ex`**: Replace eager `CallbackServer` startup with lazy `ensure_callback_server/1`. Both `:host` and `:cuda` runtime_call handlers use the lazy pattern. On error, the lazily-started server is terminated; on success it self-terminates when the executable is GC'd.
- **`exla/test/exla/defn/runtime_call_test.exs`**: Add regression tests verifying that 1000 JIT calls without runtime_call and 100 calls with runtime_call do not leak processes.

## Test plan

- [x] Regression test confirms the leak: unfixed code grows process count by exactly 1000 after 1000 JIT calls; fixed code stays flat
- [x] All existing `runtime_call` tests pass
- [x] Both `:host` and `:cuda` runtime_call paths use lazy startup

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
------------------------------------
Edit: This became about getting rid of CallbackServer, replacing it with something, not just about lazy CallbackServer

### DIFF
--- a/exla/lib/exla/defn/outfeed.ex
+++ b/exla/lib/exla/defn/outfeed.ex
@@ -228,16 +228,23 @@ defmodule EXLA.Defn.Outfeed do
 
     caller = self()
 
-    result =
+    {:ok, pid} =
       Task.Supervisor.start_child(EXLA.Defn.TaskSupervisor, fn ->
         init(caller, client, device_id, hooks, compiled_hooks, callbacks, infeeds, group_leader)
       end)
+
+    ref = Process.monitor(pid)
 
     # Wait for the task to finish registering callback handlers before
     # returning, so that XLA execution can't fire a callback before
     # the dispatcher knows where to route it.
     receive do
-      :outfeed_ready -> result
+      :outfeed_ready ->
+        Process.demonitor(ref, [:flush])
+        {:ok, pid}
+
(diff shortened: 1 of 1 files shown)


josevalim:
In any case, we will need a new release once this is fixed!

PR #1685, opened 2026-03-11
Fix EXLA test tolerances for Blackwell GPU (SM 12.0a)

blasphemetheus:
- Loosen f32/c64 test tolerances across EXLA test suite for Blackwell GPU compatibility (RTX 5090, SM 12.0a)
  - Exclude additional doctests that rely on exact f32 string matching (atan2, conv, covariance, rsqrt, standard_deviation, variance, weighted_mean)
  - Loosen c64 triangular_solve conjugate gradient tolerances from 1e-5 to 1e-3
  - Loosen cholesky, eigh, LU, SVD, and custom call property test tolerances
  - Pin runtime_call and memory_tracking tests to :host client to avoid unrelated CUDA custom call issue  #1687
  - Remove SVD exact U/V assertions (singular vectors are not unique across GPU architectures)

  # Context

  Blackwell GPUs (SM 12.0a) produce slightly different FP32 results than older architectures due to FMA pipeline scheduling differences. The differences are small (1-200 ULPs) but exceed the previous tight tolerances. All loosened tolerances remain well within numerical correctness bounds.

  Test plan

  - nx tests pass (0 failures)
  - exla tests pass (1 pre-existing failure on main unrelated to this PR)

### DIFF
--- a/exla/test/exla/nx_linalg_doctest_test.exs
+++ b/exla/test/exla/nx_linalg_doctest_test.exs
@@ -283,9 +283,11 @@ defmodule EXLA.NxLinAlgDoctestTest do
 
       assert_all_close(t, u |> Nx.dot(s_matrix) |> Nx.dot(v), atol: 1.0e-2, rtol: 1.0e-2)
 
-      # U and V are orthonormal (exact values are not unique across GPU architectures)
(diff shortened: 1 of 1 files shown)


polvalente (approved):
Thanks for putting up with my reviews!

PR #1694, opened 2026-03-14
refactor: merge callback server into outfeed server

### LINKED ISSUE #1682: EXLA.CallbackServer process leak on repeated JIT compilation

polvalente:
closes #1682

### DIFF
--- a/exla/c_src/exla/exla.cc
+++ b/exla/c_src/exla/exla.cc
@@ -549,7 +549,7 @@ ExlaExecutable::RunResult run(ErlNifEnv *env,
   if (callback_server_pid.has_value()) {
     auto pid = callback_server_pid.value();
     auto msg_env = enif_alloc_env();
-    enif_send(msg_env, &pid, msg_env, exla::atoms::stop);
+    enif_send(msg_env, &pid, msg_env, fine::encode(msg_env, exla::atoms::stop));
     enif_free_env(msg_env);
   }
 


exla/lib/exla/defn.ex
```diff
@@ -245,24 +249,48 @@ defmodule EXLA.Defn do
...
     ref = Process.monitor(outfeed_pid)
 
-    receive do
-      {:DOWN, ^ref, _, _, _} ->
-        results = EXLA.Defn.Runner.read(runner)
+    run_options = Keyword.put(run_options, :callback_server_pid, outfeed_pid)
 
-        Enum.map(results, fn result ->
-          EXLA.Defn.Buffers.to_nx!(result, outputs, executable.mesh)
+    try do
```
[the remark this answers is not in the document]
josevalim: The short answer is that we need to still an internal outfeed message to teh end the loop. We can implement it using runtime call if we want.

exla/lib/exla/defn/outfeed.ex
```diff
@@ -263,4 +319,164 @@ defmodule EXLA.Defn.Outfeed do
...
+        %Nx.Tensor{} = template, [{bin, {type, shape_list}} | rest] ->
+          decoded =
+            bin
+            |> Nx.from_binary(type)
+            |> Nx.reshape(List.to_tuple(shape_list))
+
+          if Nx.compatible?(decoded, template) do
+            {decoded, rest}
+          else
+            throw({:error, {:shape_mismatch, decoded, template}})
```
josevalim: As per above, just have this return the correct exception directly.

```suggestion
            throw({:error, FooBar.exception(...)})
```

exla/lib/exla/defn.ex
```diff
@@ -245,16 +247,21 @@ defmodule EXLA.Defn do
...
       end)
 
+    infeeds = Map.new(infeeds)
+
+    {:ok, outfeed_pid} =
+      Outfeed.start_child(executable, outfeed, Process.group_leader(), infeeds)
+
+    ref = Process.monitor(outfeed_pid)
+
+    run_options = Keyword.put(run_options, :callback_server_pid, outfeed_pid)
```
josevalim: I believe these changes are not necessary now, so let's revert them and keep the same order of operations as before.

PR #1695, opened 2026-03-15
Test runtime_call inside while loops and optional computations

blasphemetheus:
Thread callback_pid_value through while loop regions and optional computation functions, which are IsolatedFromAbove in StableHLO. Also fix reset_token/merge_outfeed to preserve runtime_callbacks across cache boundaries so callbacks registered inside while bodies are not silently lost.

this is to fix a test hang on runtime_call inside while loops.

also bug has two layers:
  1. MLIR level — callback_pid_value from outer scope used inside IsolatedFromAbove while regions. This is the obvious one.
  2. Cache level — reset_token created a new map dropping runtime_callbacks_key(), and merge_outfeed didn't merge inner callbacks back to outer. So even with correct MLIR threading. callbacks registered inside while bodies were silently lost and the runtime passed zeroed PID bytes. This is the non-obvious one and worth highlighting since it could bite other cache-carried state in the future.
  
  will need rebasing if #1694 merges first for sure (touches some of the same areas)

### DIFF
--- a/exla/test/exla/defn/runtime_call_test.exs
+++ b/exla/test/exla/defn/runtime_call_test.exs
@@ -209,14 +209,112 @@ defmodule EXLA.Defn.RuntimeCallTest do
     assert_receive {:container_fun, ^ref}
   end
 
+  def add_one_callback(t, _opts), do: Nx.add(t, 1)
+  def double_callback(t, _opts), do: Nx.multiply(t, 2)
+  def negate_callback(t, _opts), do: Nx.negate(t)
+
   defn runtime_call_in_while(x) do
     while x, Nx.less(x, 10) do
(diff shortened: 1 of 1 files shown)


polvalente:
Let's wait until my other PR gets merged because the code structure might change a lot.

Thanks for the contribution

blasphemetheus:
ope on main after #1694 this bug is fixed. 
There are some tests I could include, but that'd just be a test-coverage/regression PR

blasphemetheus:
yeah, separate bug found on one of the tests, I'll just file that one as an issue, remove the test from this PR and reference this PR on that issue. That way I'm not pushing skipped tests but the issue gets tracked.

Issue #1672, opened 2026-02-18
Introduce Nx.io_callback as a replacement for token-based hooks and EXLA's outfeed


josevalim:
# Goal

Replace the current outfeed/token-based mechanism used by hooks (e.g. `print_value`, `hook/3`) with a new `io_callback` mechanism built on top of `stablehlo.custom_call`. This would eliminate the dependency on `stablehlo.outfeed` (which lives outside the computation graph) and instead use the same infrastructure that `runtime_call` already uses, but specialized for side-effect-only operations.

## Background: How hooks work today

NX hooks (`Nx.Defn.Kernel.hook/3`, `print_value/2`, `hook_token/4`) allow users to execute side effects during compiled computation. For example:

```elixir
defn my_fun(x) do
  x = Nx.cos(x)
  x = print_value(x, label: "after cos")  # side effect: prints tensor value
  Nx.sin(x)
end
```

Under the hood, hooks are **pure side effects** — they don't produce new values. To ensure they survive NX's tracing (which only retains expressions reachable from the output), the hook must be "attached" to an existing tensor via a token. The expression graph captures this as:

```
:attach_token(
  :token(%Nx.Defn.Token{hooks: [%{expr: <x>, name: :print_value, callback: fn ...}]}),
  <x>   # the original value passes through unchanged
)
```

The resulting tensor carries the same value as the input but now has a token dependency that keeps the side effect in the graph.

### Current compilation path (EXLA)

When EXLA compiles hooks, it uses the **outfeed mechanism**:

1. A `stablehlo.create_token` initializes a token at the start of the computation
2. For each hook, the compiler emits a `stablehlo.outfeed` with:
   - A u16 flag identifying which hook to call
   - The tensor data to send to the host
3. Tokens are threaded sequentially: each outfeed consumes the previous token and produces a new one, enforcing execution order
4. A final `outfeed(flag=0)` signals completion

On the host side, `EXLA.Defn.Outfeed` spawns a dedicated process that:
- Calls `EXLA.Client.from_outfeed` to receive the flag
- Looks up the hook callback in `compiled_hooks[flag]`
- Receives the tensor buffers and invokes the callback
- Loops until it receives flag=0

### Problems with outfeed

The outfeed mechanism has several drawbacks:

- **Lives outside the computation graph**: Outfeed operates as a separate channel outside the compiled computation, making it harder to reason about and optimize
[...]

PR #1696, opened 2026-03-15
Fix hook message ordering race in while loops

### LINKED ISSUE #1686: Hook messages within 'while' arrive out of order on OTP 28.3
test hooks executes hook within while in api_test.exs:230 fails intermittently — hook callbacks from successive while-loop iterations are delivered out of order. Expected {20.0, 3} (iteration 2) but received {60.0, 2} (iteration 3). Observed in CI on Elixir 1.18.4 / OTP 28.3.

blasphemetheus:
Execute hook callback before sending completion signal to outfeed loop, ensuring hook side-effects complete before the next iteration's hook process is spawned. Uses try/after to guarantee the parent is unblocked even if the hook raises.

Fixes #1686

## Tradeoff
this serializes hook execution. so the outfeed loop now waits for run.(...) to complete before proceeding. (before it could run concurrently with the next from_outfeed NIF call).
outfeed loop already blocked on `receive do ^ref` before looping and `from_outfeed` (mostly waiting for XLA to produce the next value.

if the concurrency is worth keeping, you could instead do a single long-lived hook executor process (ordering guaranteed via BEAM's single-sender message guarantee), but that seems like a bigger change

### DIFF
--- a/exla/lib/exla/defn/outfeed.ex
+++ b/exla/lib/exla/defn/outfeed.ex
@@ -260,7 +260,10 @@ defmodule EXLA.Defn.Outfeed do
         end
       end
 
-    send(parent, ref)
-    fun.(EXLA.Defn.Buffers.to_nx!(buffers, template))
+    try do
+      fun.(EXLA.Defn.Buffers.to_nx!(buffers, template))
+    after
+      send(parent, ref)
+    end
   end
 end
--- a/exla/test/exla/defn/api_test.exs
+++ b/exla/test/exla/defn/api_test.exs
@@ -243,6 +243,39 @@ defmodule EXLA.Defn.APITest do
       assert tuple == {Nx.tensor(120.0), Nx.tensor(1)}
     end
 
+    defn hook_countdown(x) do
+      while x, Nx.greater(x, 0) do
+        hook(Nx.subtract(x, 1), :countdown)
+      end
+    end
+
+    test "hook messages within while arrive in iteration order" do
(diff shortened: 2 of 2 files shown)


exla/lib/exla/defn/outfeed.ex
```diff
@@ -260,7 +260,10 @@ defmodule EXLA.Defn.Outfeed do
         end
       end
 
-    send(parent, ref)
-    fun.(EXLA.Defn.Buffers.to_nx!(buffers, template))
+    try do
+      fun.(EXLA.Defn.Buffers.to_nx!(buffers, template))
+    after
+      send(parent, ref)
+    end
```
josevalim: The `after` here is not guaranteed to be delivered (for example, in case of broken links). If we want to guarantee, then the best way is to replace `spawn(fn -> ... end)` above by `spawn_monitor(fn -> ... end)`. But the issue with this code is that we will now block future actions (such as infeeds and upcoming runtime callbacks) until the outfeed is done. The correct solution here is rather to have a queue of outfeeds (but then that introduces other complications).

On the other hand, we have a plan to rewrite outfeeds into runtime calls, so perhaps we revisit once that is done. The test is extremely useful though, thank you!

blasphemetheus: no problem! I'll make the PR a draft and can update later

blasphemetheus:
right, waiting on #1672 to resolve

PR #1697, opened 2026-03-15
Support vectorized gradients in Nx.Defn.grad

### LINKED ISSUE #1533: Nx.Defn.Grad returns sum of gradient over vectorized axis?

blasphemetheus:
fixes #1533

## Summary

Makes `Nx.Defn.grad` work correctly with vectorized tensors. The grad system now re-vectorizes expression tree nodes during the backward pass and adjusts per-op options (axes, padding, strides, etc.) to account for vectorized dimensions.

Key changes:

  - **Re-vectorization in backward pass**: `recur_to_grad` tracks vectorized axis names through `parents_tree` and re-vectorizes args/ans during gradient computation. `adjust_vectorized_args` handles ~20 ops.

  - **apply_vectorized boundaries**: Operations like `cumulative_sum`, `triangular_solve`, etc. use `apply_vectorized` internally, creating a devectorized region in the expression tree. The grad system detects these boundaries (in `parents_args(:optional)` and `update_grads(:optional)`) and stops propagating vectorized names / devectorizes gradients at the boundary.

  - **Cond fix**: Vectorized `cond` creates separate cond_then/cond_else nodes sharing the same to_grad variables. Prevents cross-contamination by cleaning to_grad entries before each clause's inner computation.

  - **Conv fix**: When conv collapses the vectorized dim into `batch_group_size`, devectorizes args/ans in `grad_conv` since the vectorized dim is already in the batch.

  - **Batched linalg helpers**: QR, Cholesky, and triangular_solve grads now use batch-aware matmul helpers for 3D+ tensors.

  - **Concrete tensor handling**: `parents_args` and `Expr.parameter` handle non-Expr tensors captured in grad closures.

  - **`assert_all_close` fix**: Test helper now devectorizes before comparing.

  ## Test plan

  - 346 tests pass (118 new), 0 failures
  - `vectorized_grad_test.exs`: 91 correctness tests comparing vectorized grad vs per-element grad across all supported ops
  - `grad_test.exs`: 27 edge case tests (vectorize/devectorize inside grad body, non-vectorized target, error messages, window_scatter, second-order grad, tuple grad)
  - Full Nx suite: 2,692 tests pass

### DIFF
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -68,6 +68,13 @@ defmodule Nx.Defn.Expr do
     parameter(tensor, context, pos)
   end
 
+  def parameter(%T{} = tensor, pos) do
+    # Concrete (non-Expr) tensor — convert to constant expression.
+    # This happens when captured tensors end up in optional args
+    # during grad callback re-evaluation.
+    to_expr(tensor) |> then(&parameter(&1, pos))
+  end
+
   @doc """
   Creates a tensor expression parameter at `pos` based on the given `tensor` and `context`.
   """
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -403,8 +403,11 @@ defmodule Nx.Defn.Grad do
     new_names = Enum.reject(vectorized_names, &(&1 in existing))
 
     cond do
-      new_names == [] -> node
-      node.shape == {} -> node
+      new_names == [] ->
+        node
+
+      node.shape == {} ->
+        node
 
       true ->
         Nx.vectorize(node, new_names)
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -135,10 +135,20 @@ defmodule Nx.Defn.Grad do
        ) do
     expr = apply(callback, call.data.args)
 
+    # When the optional wraps an apply_vectorized pattern (inner call has
+    # no vectorized axes but parent does), don't propagate vectorized names
+    # into the inner tree. The inner computation is in devectorized space.
+    inner_names =
+      if call.vectorized_axes == [] and parent_vectorized_names != [] do
+        []
+      else
+        parent_vectorized_names
+      end
+
     # Now traverse over the optional expression where args are the new parameters.
     # Once we access the parameter itself, we point the parameter to the arg.
     {{parents, nodes}, _} =
-      Composite.reduce(expr, {acc, parent_vectorized_names}, fn
+      Composite.reduce(expr, {acc, inner_names}, fn
         expr, {{parents, nodes}, expr_vectorized_names} ->
           arg_vectorized_names = compute_arg_vectorized_names(expr, expr_vectorized_names)
           parents = Map.update(parents, expr.data.id, [id], &[id | &1])
@@ -167,13 +177,19 @@ defmodule Nx.Defn.Grad do
 
   defp parents_args(op, t, parent_id, acc, parent_vectorized_names) do
     reduce_args(op, t, acc, fn arg, {parents, nodes} ->
-      if arg.data.op in @constants do
-        {parents, nodes}
-      else
-        arg_vectorized_names = compute_arg_vectorized_names(t, parent_vectorized_names)
-        parents = Map.update(parents, arg.data.id, [parent_id], &[parent_id | &1])
+      case arg.data do
+        %Nx.Defn.Expr{op: op} when op in @constants ->
+          {parents, nodes}
 
-        recur_parents_tree(arg, {parents, nodes}, arg_vectorized_names)
+        %Nx.Defn.Expr{id: arg_id} ->
+          arg_vectorized_names = compute_arg_vectorized_names(t, parent_vectorized_names)
+          parents = Map.update(parents, arg_id, [parent_id], &[parent_id | &1])
+
+          recur_parents_tree(arg, {parents, nodes}, arg_vectorized_names)
+
+        _concrete ->
+          # Concrete (non-Expr) tensor captured in closure — skip tree traversal
+          {parents, nodes}
       end
     end)
   end
@@ -209,16 +225,85 @@ defmodule Nx.Defn.Grad do
 
   defp to_grad(arg, to_grad_ids, parents, acc) do
     id = arg.data.id
-    acc = traverse_parents(__MODULE__, to_grad_ids, parents, acc)
-    acc = traverse_parents(id, to_grad_ids, parents, acc)
+
+    acc =
+      try do
+        acc = traverse_parents(__MODULE__, to_grad_ids, parents, acc)
+        traverse_parents(id, to_grad_ids, parents, acc)
+      rescue
+        e in ArgumentError ->
+          if arg.vectorized_axes != [] and
+               String.contains?(Exception.message(e), "cannot reshape") do
+            reraise ArgumentError,
+                    "cannot compute gradient: a vectorized axis name was reused with a " <>
+                      "different size inside the grad function. Redefining the meaning of " <>
+                      "a vectorized axis (e.g., devectorize then reshape then re-vectorize " <>
+                      "with the same name at a different size) is not supported. " <>
+                      "Original error: #{Exception.message(e)}",
+                    __STACKTRACE__
+          else
+            reraise e, __STACKTRACE__
+          end
+      end
+
     {nodes, grads} = acc
 
     res = sum_grad(Map.get(grads, id, []))
-    {Nx.broadcast(res, arg), {nodes, grads}}
+
+    res =
+      cond do
+        # Extra inner dims from devectorize baking batch into shape —
+        # extract each batch element's own gradient (the diagonal).
+        res.vectorized_axes == arg.vectorized_axes and
+            tuple_size(res.shape) > tuple_size(arg.shape) ->
+          batch_size = elem(res.shape, 0)
+
+          idx =
+            Nx.iota({batch_size}, type: :s32)
+            |> Nx.vectorize(arg.vectorized_axes)
+
+          result = Nx.take(res, idx, axis: 0)
+          Nx.broadcast(result, arg)
+
+        # Function added vectorization. Self-vectorization (same size) →
+        # reshape; external vectorization (different size) → broadcast.
+        res.vectorized_axes != [] and arg.vectorized_axes == [] ->
+          res_devec = Nx.devectorize(res, keep_names: false)
+
+          if Nx.size(res_devec) == Nx.size(arg) do
+            Nx.reshape(res_devec, arg)
+          else
+            Nx.broadcast(res, arg)
+          end
+
+        res.vectorized_axes == [] and arg.vectorized_axes != [] ->
+          Nx.broadcast(res, Nx.devectorize(arg, keep_names: false))
+          |> Nx.vectorize(arg.vectorized_axes)
+
+        true ->
+          Nx.broadcast(res, arg)
+      end
+
+    {res, {nodes, grads}}
   end
 
   defp sum_grad([]), do: Expr.tensor(0.0)
-  defp sum_grad(gs), do: Enum.reduce(gs, &Nx.add/2)
+
+  defp sum_grad(gs) do
+    # Mixed vectorized/non-vectorized terms from apply_vectorized —
+    # devectorize all before summing to avoid broadcast_vectors collision
+    has_vec = Enum.any?(gs, &(&1.vectorized_axes != []))
+    has_no_vec = Enum.any?(gs, &(&1.vectorized_axes == []))
+
+    gs =
+      if has_vec and has_no_vec do
+        Enum.map(gs, &Nx.devectorize(&1, keep_names: false))
+      else
+        gs
+      end
+
+    Enum.reduce(gs, &Nx.add/2)
+  end
 
   defp traverse_parents(id, to_grad_ids, parents, acc) do
     parents
@@ -236,6 +321,8 @@ defmodule Nx.Defn.Grad do
 
         {args, ans} =
           if vectorized_names != [] do
+            vec_offset = length(vectorized_names)
+
             args =
               Enum.map(args, fn
                 %T{} = arg ->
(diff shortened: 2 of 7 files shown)


blasphemetheus:
gather issue documented at #1698

blasphemetheus:
failure is a doctest with 1 different difference on a tensor, so should be clean. Any other tests that would be good to have?

josevalim (commented):
I wonder if there is a way to do holistically? Like we remove vectorization before grading and then reapply it afterwards, or similar. Or the opposite, we make sure we vectorize upfront so all operations inside a properly vectorized... unsure.

nx/lib/nx/defn/grad.ex
```diff
@@ -214,7 +214,22 @@ defmodule Nx.Defn.Grad do
...
+          Nx.broadcast(res, arg)
+
+        {vectorized_axes, []} ->
+          devec_arg = Nx.devectorize(arg, keep_names: true)
+          res = Nx.broadcast(res, devec_arg)
+          Nx.vectorize(res, vectorized_axes)
+
+        {_, _} ->
+          Nx.broadcast(res, arg)
+      end
```
polvalente: Is this really needed?

nx/lib/nx/defn/grad.ex
```diff
@@ -285,6 +311,170 @@ defmodule Nx.Defn.Grad do
...
+    [x, adjusted_dims, adjusted_opts]
+  end
+
+  # fft/ifft: [t, opts] — length refers to last axis, needs no axis adjustment
+  # but Nx.rank(t) is used in the grad clause (returns inner rank, which is correct)
+  defp adjust_vectorized_args(:fft, args, _offset), do: args
+  defp adjust_vectorized_args(:ifft, args, _offset), do: args
+
+  # conv: complex, skip for now
+  defp adjust_vectorized_args(:conv, args, _offset), do: args
```
polvalente: This should probably raise with unsupported

nx/lib/nx/defn/grad.ex
```diff
@@ -285,6 +311,170 @@ defmodule Nx.Defn.Grad do
...
+  # Default: no adjustment needed (elementwise ops, etc.)
+  defp adjust_vectorized_args(_op, args, _offset), do: args
+
+  # Helpers for adjust_vectorized_args
+
+  defp offset_axes(axes, offset) when is_list(axes) do
+    axes
+    |> Enum.map(&(&1 - offset))
+    |> Enum.filter(&(&1 >= 0))
+  end
```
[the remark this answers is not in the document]
polvalente: Actually, if we reach negative values, something is very wrong, I think

polvalente:
Mixed vectorization should work because computations work on that scenario. We even have things such as reshape_vectors and broacast_vectors for dealing with mixed vectors scenarios.

Also, not all vectorized tensors have the same vectorized axes

nx/lib/nx/defn/grad.ex
blasphemetheus: replaced with Nx.broadcast_vectors

nx/lib/nx/defn/grad.ex
blasphemetheus: oh absolutely, ArgumentError now

nx/lib/nx/defn/grad.ex
```diff
@@ -285,6 +311,170 @@ defmodule Nx.Defn.Grad do
...
+      |> adjust_keyword_drop(:window_dilations, offset)
+      |> adjust_keyword_padding(:padding, offset)
+
+    [x, adjusted_dims, adjusted_opts]
+  end
+
+  # fft/ifft: [t, opts] — length refers to last axis, needs no axis adjustment
+  # but Nx.rank(t) is used in the grad clause (returns inner rank, which is correct)
+  defp adjust_vectorized_args(:fft, args, _offset), do: args
+  defp adjust_vectorized_args(:ifft, args, _offset), do: args
```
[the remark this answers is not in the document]
blasphemetheus: aaand removed em :) following feedback on another PR (never gonna be used basically)

polvalente (commented):
Do we have a test where vectorize/devectorize/revectorize are used inside the grad function itself?

We're improving support for inputs with gradients, but what happens when the vectors are not observable from the output?

josevalim:
I don’t think we should worry about using vectorize inside grad. We can place it as a restriction for now.

nx/test/support/helpers.ex
```diff
@@ -23,8 +23,15 @@ defmodule Nx.Helpers do
     atol = opts[:atol] || 1.0e-4
     rtol = opts[:rtol] || 1.0e-4
 
-    unless Nx.all_close(lhs, rhs, atol: atol, rtol: rtol, equal_nan: opts[:equal_nan]) ==
-             Nx.tensor(1, type: {:u, 8}) do
+    close =
```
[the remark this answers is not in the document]
blasphemetheus: these changes are intended to match the approach used in `Nx.Testing.assert_all_close` (public API version)

blasphemetheus:
Reduced tests a bit, squashed into 3 commits. Passing on all tests, no tests left skipped for vectorized grad

nx/lib/nx/defn/expr.ex
```diff
@@ -68,6 +68,13 @@ defmodule Nx.Defn.Expr do
     parameter(tensor, context, pos)
   end
 
+  def parameter(%T{} = tensor, pos) do
+    # Concrete (non-Expr) tensor — convert to constant expression.
+    # This happens when captured tensors end up in optional args
+    # during grad callback re-evaluation.
+    to_expr(tensor) |> then(&parameter(&1, pos))
```
polvalente, line 75:
```suggestion
    tensor |> to_expr(tensor) |> parameter(pos)
```

Although @josevalim will probably be able to say if this is indeed a change we should be making

nx/lib/nx/lin_alg/cholesky.ex
```diff
@@ -76,8 +76,8 @@ defmodule Nx.LinAlg.Cholesky do
   end
 
   defn cholesky_grad(l, _input, g) do
-    num = g |> Nx.tril() |> Nx.dot([0], l, [0]) |> Nx.transpose()
-    den = l |> Nx.shape() |> Nx.eye() |> Nx.add(1)
+    num = g |> Nx.tril() |> matmul_contract_rows(l) |> batch_transpose()
```
polvalente, line 79:
```suggestion
    num = g |> Nx.tril() |> dot_contract_rows(l) |> batch_transpose()
```

we don't use matmul in the Nx codebase

polvalente: Also, if we're transposing the result:

what we have is `transpose(num) = transpose(tril(g)) dot l` which is equal to `num = l dot tril(g)`, so I think we can just do `num = Nx.dot(l, Nx.tril(g))`

nx/lib/nx/lin_alg/cholesky.ex
```diff
@@ -87,16 +87,43 @@ defmodule Nx.LinAlg.Cholesky do
...
   end
 
+  deftransformp matmul_contract_rows(a, b) do
+    rank = tuple_size(a.shape)
+
+    if rank <= 2 do
+      Nx.dot(a, [0], b, [0])
+    else
+      batch_axes = Enum.to_list(0..(rank - 3))
+      Nx.dot(a, [-2], batch_axes, b, [-2], batch_axes)
```
polvalente, line 100: using `Nx.dot(a, [-2], b, [-2])` should be equivalent in all cases, I think.

polvalente: But, see my comment about rewriting it altogether

nx/lib/nx/lin_alg/cholesky.ex
```diff
@@ -87,16 +87,43 @@ defmodule Nx.LinAlg.Cholesky do
...
+  deftransformp eye_like(l) do
+    rank = tuple_size(l.shape)
+    n = elem(l.shape, rank - 1)
+
+    if rank <= 2 do
+      Nx.eye(n)
+    else
+      batch_dims = l.shape |> Tuple.to_list() |> Enum.take(rank - 2)
+      Nx.eye(n) |> Nx.broadcast(List.to_tuple(batch_dims ++ [n, n]))
+    end
```
polvalente, line 124:
```suggestion
    shape = put_elem(l.shape, rank - 2, n)
    Nx.eye(shape)
```

polvalente: and since this code is small we can put it back instead of having a defp

nx/lib/nx/lin_alg/qr.ex
```diff
@@ -151,13 +151,26 @@ defmodule Nx.LinAlg.QR do
     # Equation (3)
     r_inv = Nx.LinAlg.invert(r)
 
-    m = Nx.dot(r, Nx.LinAlg.adjoint(dr)) |> Nx.subtract(Nx.dot(Nx.LinAlg.adjoint(dq), q))
+    m =
+      matmul(r, Nx.LinAlg.adjoint(dr))
```
polvalente, line 155: same consideration about matmul

nx/lib/nx/lin_alg/qr.ex
```diff
@@ -151,13 +151,26 @@ defmodule Nx.LinAlg.QR do
...
   end
+
+  deftransformp matmul(a, b) do
+    rank = tuple_size(a.shape)
+
+    if rank <= 2 do
+      Nx.dot(a, b)
+    else
+      batch_axes = Enum.to_list(0..(rank - 3))
+      Nx.dot(a, [-1], batch_axes, b, [-2], batch_axes)
```
polvalente, line 173: This code assumes both sides have the same rank

polvalente: Let's just use it in-place because it relies on this implied condition

nx/test/support/helpers.ex
```diff
@@ -23,8 +23,15 @@ defmodule Nx.Helpers do
...
+      lhs
+      |> Nx.all_close(rhs, atol: atol, rtol: rtol, equal_nan: opts[:equal_nan])
+      |> Nx.devectorize(keep_names: false)
+      |> Nx.backend_transfer(Nx.BinaryBackend)
+      |> Nx.to_flat_list()
+      |> Enum.all?(&(&1 == 1))
+
+    unless close do
```
polvalente, line 34: We should probably be fixing Nx.Testing and using it here instead

nx/test/nx/defn/vectorized_grad_test.exs
```diff
@@ -0,0 +1,808 @@
+defmodule Nx.Defn.VectorizedGradTest do
```
polvalente, line 1: Not sure if this should be a separate file

nx/lib/nx/defn/grad.ex
```diff
@@ -167,13 +177,19 @@ defmodule Nx.Defn.Grad do
...
 
-        recur_parents_tree(arg, {parents, nodes}, arg_vectorized_names)
+        %Nx.Defn.Expr{id: arg_id} ->
+          arg_vectorized_names = compute_arg_vectorized_names(t, parent_vectorized_names)
+          parents = Map.update(parents, arg_id, [parent_id], &[parent_id | &1])
+
+          recur_parents_tree(arg, {parents, nodes}, arg_vectorized_names)
+
+        _concrete ->
+          # Concrete (non-Expr) tensor captured in closure — skip tree traversal
```
polvalente, line 191: @josevalim should this be happening here?

nx/lib/nx/defn/grad.ex
```diff
@@ -209,16 +225,85 @@ defmodule Nx.Defn.Grad do
...
+            reraise ArgumentError,
+                    "cannot compute gradient: a vectorized axis name was reused with a " <>
+                      "different size inside the grad function. Redefining the meaning of " <>
+                      "a vectorized axis (e.g., devectorize then reshape then re-vectorize " <>
+                      "with the same name at a different size) is not supported. " <>
+                      "Original error: #{Exception.message(e)}",
+                    __STACKTRACE__
+          else
+            reraise e, __STACKTRACE__
+          end
```
polvalente, line 246: This shouldn't be captured here. At best, we should capture closer to where the error happens.

nx/lib/nx/defn/grad.ex
```diff
@@ -209,16 +225,85 @@ defmodule Nx.Defn.Grad do
...
+      cond do
+        # Extra inner dims from devectorize baking batch into shape —
+        # extract each batch element's own gradient (the diagonal).
+        res.vectorized_axes == arg.vectorized_axes and
+            tuple_size(res.shape) > tuple_size(arg.shape) ->
+          batch_size = elem(res.shape, 0)
+
+          idx =
+            Nx.iota({batch_size}, type: :s32)
+            |> Nx.vectorize(arg.vectorized_axes)
```
polvalente, line 263: I think this is overfitting a very specific case. Can you walk me through the logic?
This code only works if length(arg.vectorized_axes) is 0 or 1

nx/lib/nx/defn/grad.ex
```diff
@@ -209,16 +225,85 @@ defmodule Nx.Defn.Grad do
...
+    # Mixed vectorized/non-vectorized terms from apply_vectorized —
+    # devectorize all before summing to avoid broadcast_vectors collision
+    has_vec = Enum.any?(gs, &(&1.vectorized_axes != []))
+    has_no_vec = Enum.any?(gs, &(&1.vectorized_axes == []))
+
+    gs =
+      if has_vec and has_no_vec do
+        Enum.map(gs, &Nx.devectorize(&1, keep_names: false))
+      else
+        gs
```
polvalente, line 302: what's the case where we don't want devectorize to happen?

polvalente:
@blasphemetheus after some discussion with @josevalim, we think this is the wrong approach overall.

We think that the solution should be akin to a thin wrapper on `Nx.Defn.Grad.transform` instead, where we deal with the vectors in the input and output steps, without leaking them downwards, kinda like what Nx does with passing tensors to backends.

This approach can include a requirement for functions not revectorizing stuff internally, I think.

blasphemetheus:
thanks for the review! Hopefully the tests are useful here.

PR #1703, opened 2026-03-16
Use precision: :highest in EXLA test suite

blasphemetheus:
on cpu, no tensor cores, no cuBLAS. MLIR attribute stablehlo.precision = HIGHEST will be emitted with this change, but CPU backend ignores it since it always uses full f32

----
XLA defaults to TF32 tensor cores for f32 GEMM on GPUs with tensor cores (Ampere+). TF32 truncates the f32 mantissa from 23 to 10 bits, giving ~1e-3 precision instead of ~1e-7. This is the correct default for ML training performance, but causes linalg property tests to fail when they verify decompositions via matmul reconstruction (e.g. assert dot(L, L^T) ≈ A).

The decomposition factors themselves are accurate to 1-2 ULP — only the reconstruction matmul amplifies TF32 rounding. Setting precision to :highest forces full f32 precision for dot operations, matching the accuracy that the test tolerances expect.

This fixes 7 test failures on tensor-core GPUs (eigh, cholesky, LU, triangular_solve gradient) without changing any tolerances.

CPU tests are unaffected — precision: :highest is a no-op on CPU since there are no tensor cores.

### DIFF
--- a/exla/test/exla/backend_test.exs
+++ b/exla/test/exla/backend_test.exs
@@ -5,6 +5,7 @@ defmodule EXLA.BackendTest do
 
   setup do
     Nx.default_backend(EXLA.Backend)
+    Nx.Defn.default_options(compiler: EXLA, precision: :default)
     :ok
   end
 
@@ -21,7 +22,8 @@ defmodule EXLA.BackendTest do
     asinh: 1,
     logsumexp: 2,
     exp: 1,
-    expm1: 1
+    expm1: 1,
+    rsqrt: 1
   ]
 
   doctest Nx,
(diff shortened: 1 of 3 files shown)


blasphemetheus:
alright that fixes 6, the other one was the cholesky property c64 test, which was fixed by `--xla_gpu_deterministic_ops=true` and a from-source XLA build with a native sm_120 cubin, but not by precision: :highest. It's a non-deterministic reduction order issue, not a TF32 precision issue. so this won't help.

polvalente (approved):

blasphemetheus:
oh i think the ci failure is just a rounding difference, `0.5773502691896257` vs        
  `...258`. exact string match doctest on an element-wise op, unrelated to precision

PR #1709, opened 2026-03-20
Basic Nx.block implementation

### LINKED ISSUE #946: Make optional backend mechanism extensible with protocols

Chapaman:
## Extensible computation blocks (`Nx.block/4`): structs, required `backend.block/4`, no `backend_args`

`Nx.block/4` is the Nx API for **named blocks**: a `%Nx.Block.*{}` holds static options, `args` is a **list** of tensor inputs (and optional trailing keyword fragments where the API needs them), `output` is the **result template** (`Nx.Tensor` or tuple of tensors), and `fun` is the **default** implementation `fn struct, ...tensors -> ... end`.

**Dispatch:** `Nx.block/4` resolves a common backend from tensor leaves in `args`, validates `fun` arity (`1 + length(args)`), and calls **only**:

`backend.block(struct, output, args, fun)`

There is **no** second hop through `Nx.Shared.optional/4`, **no** `function_exported?(backend, op, …)` before `block`, and **no** `Nx.Block.backend_args/2` — backends receive the **same** `(struct, output, args, fun)` contract and interpret structs / tensors themselves.

Earlier call sites that used **`Nx.Shared.optional/4`** (or similar) for things like **`logical_not`** and **`phase`** now use **`block`** with **`Nx.Block.LogicalNot`**, **`Nx.Block.Phase`**, etc.

### `Nx.block/4` (`nx/lib/nx.ex`)

- Validates default `fun` arity.
- **`backend = Nx.Shared.list_impl!(args)`** then **`backend.block(struct, output, args, fun)`** (single path).

### `Nx.Backend` (`nx/lib/nx/backend.ex`)

- **`@callback block(struct, output :: tensor | tuple, args :: [term], fun) :: tensor | tuple`** is **required**.
- Default portable behaviour: **`apply(fun, [struct | args])`**.
- Former **optional** per-op surface is folded into **`block/4`** or dropped from the behaviour where ops are only reached via blocks; **`@optional_callbacks`** / duplicate **`@impl`**s are cleaned up accordingly (see Torchx).

### `Nx.Block` (`nx/lib/nx/block.ex`)

- **No** separate `Nx.Block.name/1` or `:name` field — identity is the **struct module** (`Nx.Block.QR`, `Nx.Block.LU`, …). Defn / JIT **cache keys** and **EXLA** codegen use **`struct.__struct__`** (and related args), not a dedicated name API.
- **`backend_args/2` removed** — no separate argument reshaping layer; options live on structs, tensors in `args`.
- Structs for **`logical_not`**, **`phase`**, **`all_close`**, cumulatives, **`take`**, **`take_along_axis`**, **`top_k`**, **`fft2` / `ifft2`**, LinAlg (**`QR`**, **`Eigh`**, **`SVD`**, **`LU`**, **`Cholesky`**, **`Solve`**, **`Determinant`**, …) with **`@derive {Nx.Container, …}`** where option fields must traverse **`defn`** boundaries — **`Nx.Defn.Compiler.to_lazy_params/2`** walks arguments with **`Nx.LazyContainer.traverse/3`**, which (via **`Nx.Container`**) treats those fields consistently as part of the lazy parameter pipeline.

### Call sites and LinAlg

- **`Nx.logical_not`**, **`phase`**, **`all_close`**, cumulatives, **`take`**, **`take_along_axis`**, **`top_k`**, **`fft2` / `ifft2`**, and LinAlg entry points use **`Nx.block/4`** with the matching **`%Nx.Block.*{}`** and tensor **`args`**.
- LinAlg helpers use **`struct(Nx.Block.*, opts)`** and list **`args`** as before; **`defn`** code consumes **`%Nx.Block.*{}`** and struct fields instead of ad-hoc maps where applicable.

### Defn / backends

- **`Nx.Defn.Expr`**: **`block/3`** (public) / **`block/4`** (`@impl Nx.Backend`) build **`op: :block`** expressions.
- **`Nx.Defn.Evaluator`**: evaluates **`:block`** by building a default **`fun`** whose arity matches the tensor prefix of **`args`**, then **`backend.block(struct, out, param_prefix, fun)`** (no **`backend_args`**). **Compute-cache** keys for **`:block`** use **`struct.__struct__`** together with the cached tensor prefix (aligned with EXLA).
- **EXLA** (`exla/lib/exla/defn.ex`): **`:block`** compilation / cache keys use **`struct.__struct__`** and call args in one place; nested MLIR subfunctions are named from the struct module (**`block_subfunction_description/1`** — last segment of **`struct.__struct__`**, underscored), not from a **`Block.name`** field.
- **Torchx** (`torchx/lib/torchx/backend.ex`): **`block/4`** dispatches with **`case {block_name, args}`** (e.g. **`{Nx.Block.Eigh, [t]}`**, **`{Nx.Block.Solve, [a, b]}`**) for native paths; on **MPS**, unsupported LinAlg blocks are detected by **`block_name`** (struct module) and delegate to **`apply(fun, [struct | args])`**; otherwise **`apply(fun, [struct | args])`** for unknown or unhandled shapes. Standalone **`fft2`/`ifft2`** behaviour impls removed where redundant; stray **`@impl`** removed for ops no longer on **`Nx.Backend`**.

### Tests & tooling

- **`nx/test/nx/optional_test.exs`**: backends exercise **`block/4`**; assertions updated where defaults run on **`BinaryBackend`**.
- **Torchx**: **`fft2` / `ifft2`** doctests excluded in **`torchx/nx_doctest_test.exs`** (**`inspect`** / signed-zero vs **`BinaryBackend`**); **`torchx/nx_block_test.exs`** adds numerical checks vs **`BinaryBackend`**.
- **`Nx.all_close/3`**: tolerance uses a safe **`abs(b)`** for integer **`b`** (e.g. via **`f64`**) so signed-min overflow does not break **`assert_all_close`** (relevant once **`all_close`** goes through **`%Nx.Block.AllClose{}`**).


torchx/lib/torchx/backend.ex
```diff
@@ -46,10 +46,16 @@ defmodule Torchx.Backend do
     Keyword.validate!(opts, [:device])
   end
 
-  ## Optional callback for MPS compatibility
+  ## Block callback for MPS compatibility
 
   @impl true
-  def optional(function_name, args, default_impl) do
+  def block(struct, args, fun) do
```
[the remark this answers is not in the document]
polvalente: Then use block_name to match instead of the function_name conversion

nx/lib/nx.ex
```diff
@@ -6876,12 +6876,44 @@ defmodule Nx do
...
+      function_exported?(backend, :block, 3) ->
+        backend.block(struct, args, fun)
+
+      function_exported?(backend, name, length(backend_args) + 1) ->
+        apply(backend, name, [output | backend_args])
+
+      true ->
+        apply(fun, [struct | args])
+    end
+  end
```
Chapaman: @josevalim we have three main approaches here

1. current: keep optional callbacks with the same contract. this requires the Nx.Block.backend_args trick or some protocol implementation
2. change the optional callbacks contracts to receive their structs as the first args and no opts (removes the backend_args trick)
3. @polvalente 's preferred approach: remove optional callbacks altogether and backends have to deal with def block directly, which won't be optional

josevalim: Option 3. I think we always call `block/3`, we just make it easier to dispatch to the default implementation.

polvalente: Yeah, let's move forward with that then. This will let us get rid of the `backend_args` separation in Nx.Block since now backends will adapt to the new (struct, ...args) contract

exla/lib/exla/defn.ex
```diff
@@ -645,7 +647,8 @@ defmodule EXLA.Defn do
          %{client: %EXLA.Client{platform: :host}, builder: %Function{}} = state,
          cache
        )
-       when evec_type_kind != :c and eval_type_kind != :c do
+       when op == :block and evec_type_kind != :c and eval_type_kind != :c do
+    tensor = hd(in_args)
```
josevalim: Let's pattern match on the args as it was done before!

exla/lib/exla/defn.ex
```diff
@@ -672,16 +675,17 @@ defmodule EXLA.Defn do
...
-             args: [%{data: %{op: :take, args: [tensor, indices, opts]}}, expr, _callback]
+             args: [%Nx.Block.Take{axis: axis}, in_args, expr, _callback]
            }
          },
          state,
          cache
-       ) do
-    axis = opts[:axis]
+       )
+       when op == :block do
```
josevalim: Let's move it to pattern matching!

exla/lib/exla/defn.ex
```diff
@@ -645,7 +647,8 @@ defmodule EXLA.Defn do
          %{client: %EXLA.Client{platform: :host}, builder: %Function{}} = state,
          cache
        )
-       when evec_type_kind != :c and eval_type_kind != :c do
+       when op == :block and evec_type_kind != :c and eval_type_kind != :c do
```
josevalim: Let's move the op == :block back into a pattern matching!

exla/lib/exla/defn.ex
```diff
@@ -714,56 +718,74 @@ defmodule EXLA.Defn do
...
-         :optional,
-         %T{data: %Expr{args: [%{data: %{op: :top_k, args: [tensor, opts]}}, expr, _callback]}},
+         op,
+         %T{data: %Expr{args: [%Nx.Block.TopK{k: k}, in_args, expr, _callback]}},
          state,
          cache
-       ) do
+       )
+       when op == :block do
+    tensor = hd(in_args)
```
josevalim: Same as above, pattern matching!

exla/lib/exla/defn.ex
```diff
@@ -714,56 +718,74 @@ defmodule EXLA.Defn do
...
-         :optional,
-         %T{data: %Expr{args: [%{data: %{op: :fft2, args: [tensor, opts]}}, expr, _callback]}},
+         op,
+         %T{data: %Expr{args: [%Nx.Block.FFT2{} = fft2_struct, in_args, expr, _callback]}},
          state,
          cache
-       ) do
+       )
+       when op == :block do
+    tensor = hd(in_args)
```
josevalim: Same as above, pattern matching!

exla/lib/exla/defn.ex
```diff
@@ -714,56 +718,74 @@ defmodule EXLA.Defn do
...
-         :optional,
-         %T{data: %Expr{args: [%{data: %{op: :ifft2, args: [tensor, opts]}}, expr, _callback]}},
+         op,
+         %T{data: %Expr{args: [%Nx.Block.IFFT2{} = ifft2_struct, in_args, expr, _callback]}},
          state,
          cache
-       ) do
+       )
+       when op == :block do
+    tensor = hd(in_args)
```
josevalim: Same as above, pattern matching!

exla/lib/exla/defn.ex
```diff
@@ -714,56 +718,74 @@ defmodule EXLA.Defn do
...
+    [struct, in_args, expr, _callback] = args
+    op = Nx.Block.name(struct)
 
-    {args, opts} = Enum.split_while(in_args, &(not is_list(&1)))
+    {_call_prefix, rest} = Enum.split_while(in_args, &(not is_list(&1)))
 
-    {call_args, cache} = Enum.map_reduce(args, cache, &recur_operator(&1, state, &2))
-    key = computation_key(op, call_args ++ opts)
+    {call_args, cache} = Enum.map_reduce(in_args, cache, &recur_operator(&1, state, &2))
+    key = computation_key(op, [struct | call_args ++ rest])
```
josevalim: Because you are mapping over `in_args`, it feels like `opts` will be included twice? Do we even have `opts` now? Aren't those all struct fields now?

nx/lib/nx/defn/evaluator.ex
```diff
@@ -426,6 +420,53 @@ defmodule Nx.Defn.Evaluator do
...
+
+  defp block_default_fun(expr, state, expr_cache, 7),
+    do: fn struct, a, b, c, d, e, f, g ->
+      block_apply_default(expr, state, expr_cache, struct, [a, b, c, d, e, f, g])
+    end
+
+  defp block_default_fun(expr, state, expr_cache, 8),
+    do: fn struct, a, b, c, d, e, f, g, h ->
+      block_apply_default(expr, state, expr_cache, struct, [a, b, c, d, e, f, g, h])
+    end
```
josevalim: We already have logic for this in `Nx.Defn.Compiler.fun`

josevalim (commented):
I did one pass. The current code is a bit mixed in style. My understanding is that `in_args` would not have `opts` anymore, because the `opts` would be field in the structure.

Futhermore, the structs are all values, never containers, so we shouldn't need to implement the Nx.Container protocol for them.

nx/lib/nx/block.ex
```diff
@@ -0,0 +1,117 @@
...
+
+defmodule Nx.Block.IFFT2 do
+  @derive {Nx.Container, containers: [], keep: [:eps, :lengths, :axes]}
+  defstruct eps: nil, lengths: nil, axes: nil
+end
+
+defmodule Nx.Block do
+  @moduledoc false
+
+  def name(%{__struct__: module}) do
```
polvalente: Let's get rid of this function! We're using it for the cache key names in EXLA and for function routing in Torchx. Both can use the struct name from: `%name{} = struct`

nx/lib/nx/lin_alg.ex
```diff
@@ -1744,8 +1748,9 @@ defmodule Nx.LinAlg do
        %{tensor | type: output_type, shape: l_shape, names: names},
        %{tensor | type: output_type, shape: u_shape, names: names}}
 
-    :lu
-    |> Nx.Shared.optional([tensor, opts], output, &Nx.LinAlg.LU.lu/2)
+    Nx.block(struct(Nx.Block.LU, opts), [tensor], output, fn %Nx.Block.LU{} = opts, t ->
+      Nx.LinAlg.LU.lu(opts, t)
```
polvalente: LU.lu doesn't use opts, let's change it to lu/1?

nx/lib/nx/block.ex
```diff
@@ -87,31 +87,3 @@ defmodule Nx.Block.IFFT2 do
   @derive {Nx.Container, containers: [], keep: [:eps, :lengths, :axes]}
```
Chapaman: @josevalim 
On `@derive {Nx.Container, ...}` on the `Nx.Block.*` option structs: from what @polvalente and I can see, we still need these structs to participate in `Nx.Container` (and therefore `Nx.LazyContainer`) for the defn pipeline to treat them like normal arguments.

Right now `Nx.Defn.Compiler.to_lazy_params/2` walks each argument with `Nx.LazyContainer.traverse/3` so option structs that carry tensor-related fields get turned into parameters/templates correctly. Without the `@derive`, those block option structs wouldn’t go through that path the same way, and we weren’t able to find another approach that kept the current behavior without an overhaul.

That said, this is our reading of the code path — if you see a cleaner way to structure this (or a spot we missed), we’d be very glad to follow your lead.

josevalim: @Chapaman when I discussed it with @polvalente, the idea was exactly that the tensors would be passed separately (as a container, as done by this PR) and the struct would only have "keep" values, this way we could skip traversing the struct (because by definition they would not have anything interesting).

IIRC, we had a technical reason for doing such, but it should be a consequence that they don't need to be containers.

polvalente: We tried removing the container implementation for the structs yesterday and quite a few tests failed in ways that l couldn't find a simple fix that would boil down to "parse this block argument differently".

Do you want to take a stab at this?

josevalim: Ok, when this is ready, let me know, and I can review and try to change it. I assume the challenge is that we want them as cache args but not traverse them. I can take a look. All of them simply declare all fields as keep, so it seems we are in the right direction anyway.

exla/lib/exla/defn.ex
```diff
@@ -714,56 +715,73 @@ defmodule EXLA.Defn do
...
          state,
          cache
        ) do
     {tensor, cache} = recur_operator(tensor, state, cache) |> unwrap_single_tensor!()
 
+    opts =
+      [lengths: fft2_struct.lengths, axes: fft2_struct.axes]
+      |> then(fn kw ->
+        if fft2_struct.eps != nil, do: Keyword.put(kw, :eps, fft2_struct.eps), else: kw
+      end)
```
polvalente: I think we can rewrite this for clarity

exla/lib/exla/defn.ex
```diff
@@ -714,56 +715,73 @@ defmodule EXLA.Defn do
...
          state,
          cache
        ) do
     {tensor, cache} = recur_operator(tensor, state, cache) |> unwrap_single_tensor!()
 
+    opts =
+      [lengths: ifft2_struct.lengths, axes: ifft2_struct.axes]
+      |> then(fn kw ->
+        if ifft2_struct.eps != nil, do: Keyword.put(kw, :eps, ifft2_struct.eps), else: kw
+      end)
```
polvalente: I think we can rewrite this for clarity

exla/lib/exla/defn.ex
```diff
@@ -714,56 +715,73 @@ defmodule EXLA.Defn do
...
-    %{data: %{args: in_args, op: op}} = call
+  defp cached_recur_operator(:block, %T{data: %Expr{args: args}}, state, cache) do
+    [struct, in_args, expr, _callback] = args
 
-    {args, opts} = Enum.split_while(in_args, &(not is_list(&1)))
-
-    {call_args, cache} = Enum.map_reduce(args, cache, &recur_operator(&1, state, &2))
-    key = computation_key(op, call_args ++ opts)
+    {call_args, cache} = Enum.map_reduce(in_args, cache, &recur_operator(&1, state, &2))
+    key = computation_key(struct.__struct__, [struct | call_args])
```
polvalente: José can correct me, but I think accessing __struct__ is generally an antipattern, with `%name{} = struct` being the preferred way to get the module name

polvalente: Also, maybe we can use {:block, name}, but I think just the struct name is fine.

exla/lib/exla/defn.ex
```diff
@@ -1818,8 +1836,15 @@ defmodule EXLA.Defn do
     {region, merge_outfeed(cache, comp_cache)}
   end
 
-  defp optional_computation(name, args, expr, %{builder: %Function{}} = state, cache) do
-    %Function{module: module, name: name} = subbuilder(state.builder, name)
+  defp block_subfunction_description(%_{} = struct) do
```
polvalente: We should just use the full name here because we also want to support user structs. What are the implications for removing this transformation?

nx/lib/nx/block.ex
```diff
@@ -0,0 +1,89 @@
...
+  defstruct []
+end
+
+defmodule Nx.Block.Cholesky do
+  defstruct []
+end
+
+defmodule Nx.Block.Solve do
+  defstruct []
+end
```
polvalente, line 15: These are missing the container implementation. So either these functions are doing something correctly that the other failing ones aren't, or there are tests we can add to catch this.

@josevalim I think this is a good hint on whether containers are really needed

nx/lib/nx/defn/evaluator.ex
```diff
@@ -133,14 +133,13 @@ defmodule Nx.Defn.Evaluator do
...
+  defp compute_cache(:block, %{data: %Expr{args: args}}, state, cache) do
+    [struct, in_args, expr, _callback] = args
 
-    {call_prefix, call_suffix} = Enum.split_while(call_args, &(not is_list(&1)))
+    {call_prefix, call_suffix} = Enum.split_while(in_args, &(not is_list(&1)))
     {call_prefix, cache} = Enum.map_reduce(call_prefix, cache, &compute_cache(&1, state, &2))
-    call_args = call_prefix ++ call_suffix
-    key = computation_key(call_name, call_args)
+    in_args = call_prefix ++ call_suffix
+    key = computation_key(struct.__struct__, call_prefix)
```
polvalente: Same consideration on `__struct__` access

nx/lib/nx/defn/expr.ex
```diff
@@ -803,6 +812,11 @@ defmodule Nx.Defn.Expr do
 
   @behaviour Nx.Backend
 
+  @impl true
+  def block(struct, _output, in_args, fun) do
+    expr_block(struct, in_args, fun)
+  end
+
```
polvalente, line 814: Is this clause really needed?

nx/lib/nx/lin_alg.ex
```diff
@@ -149,8 +149,9 @@ defmodule Nx.LinAlg do
 
     out = %{tensor | type: output_type, shape: output_shape, names: output_names}
 
-    :cholesky
-    |> Nx.Shared.optional([tensor], out, &Nx.LinAlg.Cholesky.cholesky/1)
+    Nx.block(%Nx.Block.Cholesky{}, [tensor], out, fn %Nx.Block.Cholesky{}, t ->
```
polvalente, line 152: I think we shouldn't need to pass `out` as an argument

torchx/lib/torchx/backend.ex
```diff
@@ -542,7 +565,6 @@ defmodule Torchx.Backend do
     |> to_nx(out)
   end
 
-  @impl true
```
polvalente, line 545: All of these should become defp and be called by block

torchx/lib/torchx/backend.ex
```diff
@@ -629,7 +651,6 @@ defmodule Torchx.Backend do
     to_nx(result, out)
   end
 
-  @impl true
   def all_close(%T{} = out, %T{} = a, %T{} = b, opts) do
```
polvalente: I think all_close is missing the delegate in def block. Please do a pass ensuring all functions that lost @impl true now get a clause there or are removed

torchx/lib/torchx/backend.ex
```diff
@@ -988,6 +1005,10 @@ defmodule Torchx.Backend do
     end
   end
 
+  def logical_not(out, tensor) do
```
polvalente: Why was this extracted?

nx/lib/nx/defn/expr.ex
Chapaman: > Is this clause really needed?

Nx.block always ends up calling backend's block/4 from what I could see, because we always supply an output template

polvalente: I think we should do away with supplying the template, leaving only block/3

exla/lib/exla/defn.ex
Chapaman: Changed it to use the full name. It will help with avoiding any confusion when using

exla/lib/exla/defn.ex
```diff
@@ -1818,8 +1843,14 @@ defmodule EXLA.Defn do
     {region, merge_outfeed(cache, comp_cache)}
   end
 
-  defp optional_computation(name, args, expr, %{builder: %Function{}} = state, cache) do
-    %Function{module: module, name: name} = subbuilder(state.builder, name)
+  defp block_subfunction_description(%module{} = _) do
+    module
+    |> Module.split()
+    |> Enum.map_join("_", &Macro.underscore(to_string(&1)))
```
josevalim: My suggestion would be to simply use:

```suggestion
    module
    |> Atom.to_string()
```

If we cannot have dots in the name, then:

```suggestion
    module
    |> Atom.to_string()
    |> String.replace(".", "_")
```

nx/lib/nx/defn/expr.ex
```diff
@@ -1765,6 +1775,11 @@ defmodule Nx.Defn.Expr do
   defp traverse_args(:while, [initial, _arg, _condition, _body], state),
     do: traverse_args([initial], state)
 
+  defp traverse_args(:block, [struct, in_args, _body, _callback], state) do
+    {in_args_io, state} = Enum.map_reduce(in_args, state, &recur_inspect/2)
+    {[inspect(struct) | in_args_io], state}
```
josevalim, line 1775: Why are we returning the inspected struct? 🤔

nx/lib/nx/defn/tree.ex
```diff
@@ -254,4 +254,18 @@ defmodule Nx.Defn.Tree do
...
     end)
   end
+
+  @doc false
+  def map_block_args(list, acc, fun) when is_list(list) do
+    Enum.map_reduce(list, acc, fn
+      %T{} = arg, acc ->
+        fun.(arg, acc)
+
+      arg, acc when is_list(arg) ->
```
josevalim: Do we really need list arguments? I thought all list arguments are converted into the struct fields? It feels like we should remove this one.

nx/lib/nx/lin_alg.ex
```diff
@@ -1154,8 +1155,9 @@ defmodule Nx.LinAlg do
            names: List.duplicate(nil, tuple_size(r_shape))
        }}
 
-    :qr
-    |> Nx.Shared.optional([tensor, opts], output, &Nx.LinAlg.QR.qr/2)
+    Nx.block(struct(Nx.Block.QR, opts), [tensor], output, fn %Nx.Block.QR{} = opts, t ->
```
josevalim: Use `struct!` here and elsewhere in this file so we catch typos.

```suggestion
    Nx.block(struct!(Nx.Block.QR, opts), [tensor], output, fn %Nx.Block.QR{} = opts, t ->
```

nx/lib/nx/block.ex
josevalim: Here is a patch that solves the container issue for SVD:

```diff
diff --git a/nx/lib/nx/lin_alg.ex b/nx/lib/nx/lin_alg.ex
index b34c154f..a6de3667 100644
--- a/nx/lib/nx/lin_alg.ex
+++ b/nx/lib/nx/lin_alg.ex
@@ -1504,8 +1504,9 @@ defmodule Nx.LinAlg do
   """
   def svd(tensor, opts \\ []) do
     opts = keyword!(opts, max_iter: 100, full_matrices?: true)
-    %T{vectorized_axes: vectorized_axes} = tensor = Nx.to_tensor(tensor)
+    _ = opts[:max_iter] || raise ArgumentError, "missing option :max_iter"

+    %T{vectorized_axes: vectorized_axes} = tensor = Nx.to_tensor(tensor)
     %T{type: type, shape: shape} = tensor = Nx.devectorize(tensor)

     Nx.Shared.raise_complex_not_implemented_yet(type, "LinAlg.svd", 2)
@@ -1518,8 +1519,8 @@ defmodule Nx.LinAlg do
        %{tensor | names: List.duplicate(nil, rank - 1), type: output_type, shape: s_shape},
        %{tensor | names: List.duplicate(nil, rank), type: output_type, shape: v_shape}}

-    Nx.block(struct(Nx.Block.SVD, opts), [tensor], output, fn %Nx.Block.SVD{} = opts, t ->
-      Nx.LinAlg.SVD.svd(opts, t)
+    Nx.block(struct!(Nx.Block.SVD, opts), [tensor], output, fn %Nx.Block.SVD{}, t ->
+      Nx.LinAlg.SVD.svd(t, opts)
     end)
     |> Nx.vectorize(vectorized_axes)
   end
diff --git a/nx/lib/nx/lin_alg/svd.ex b/nx/lib/nx/lin_alg/svd.ex
index a1597bd7..66b704c7 100644
--- a/nx/lib/nx/lin_alg/svd.ex
+++ b/nx/lib/nx/lin_alg/svd.ex
@@ -21,15 +21,8 @@ defmodule Nx.LinAlg.SVD do
   import Nx.Defn
   @eps 1.1920929e-07

-  defn svd(%Nx.Block.SVD{} = o, input_tensor) do
-    opts =
-      keyword!([max_iter: o.max_iter, full_matrices?: o.full_matrices?],
-        max_iter: 100,
-        full_matrices?: true
-      )
-
-    validate_opts(opts)
-
+  defn svd(input_tensor, opts) do
     {target_shape, u_shape, s_shape, vt_shape} = calculate_shapes(input_tensor)

     tensor = Nx.revectorize(input_tensor, [vector: :auto], target_shape: target_shape)
@@ -53,10 +46,6 @@ defmodule Nx.LinAlg.SVD do
     end)
   end

-  deftransformp validate_opts(opts \\ []) do
-    opts[:max_iter] || raise ArgumentError, "missing option :max_iter"
-  end
-
   deftransformp calculate_shapes(t) do
     shape = Nx.shape(t)
     rank = tuple_size(shape)
```

In a nutshell, the issue is that you are passing the struct as arguments to `defn` function. The fix is rather simple, pass the already validated options as arguments, instead of the struct. The struct is useful for `exla` and `torchx` but, for the default implementation, we can just just use the given options as is.

So in a nutshell:

1. Do not pass the block structs to `defn`
2. Move option validate to before the `block` is created
3. Remove default arguments from internal functions (e.g. Nx.Linalg.SVD.svd doesn't need opts as default args)

PS: a few functions in the `svd` module are `defn` but I think they could be `defnp`.

nx/lib/nx/lin_alg/block_eigh.ex
```diff
@@ -9,8 +9,8 @@ defmodule Nx.LinAlg.BlockEigh do
 
   import Nx.Defn
 
-  defn eigh(matrix, opts \\ []) do
-    opts = keyword!(opts, eps: 1.0e-6, max_iter: 100)
+  defn eigh(%Nx.Block.Eigh{} = o, matrix) do
+    opts = keyword!([eps: o.eps, max_iter: o.max_iter], eps: 1.0e-4, max_iter: 1000)
```
josevalim: Basically, if you undo these changes, you will solve the struct/container issue.

nx/lib/nx/defn/expr.ex
Chapaman: Dropped it. block inspect now only shows tensor arguments
```
b = block a     tuple2
```
instead of
```
b = block %Nx.Block.QR{eps: 1.0e-10, mode: :reduced}, a   tuple2
```

nx/lib/nx/defn/tree.ex
Chapaman: Removed it
`map_block_args` now flat-maps tensors and passes everything else through.

nx/lib/nx/defn/expr.ex
polvalente: We should return the inspected struct. It's important to inspect it both so we know which block it is as well as the options passed (as we had it for optional)

polvalente (approved):

nx/lib/nx/defn/expr.ex
josevalim: Agreed. I missed originally this was in the context of inspect.

PR #1715, opened 2026-03-20
Warn when closure-captured tensors cause recompilation storms

jeffreyksmithjr:
## Summary

When a non-scalar tensor with 10 or more elements is used as a constant inside a defn expression, this change emits a `Logger.warning` during graph construction. The warning explains that constants are embedded into the computation graph and will force recompilation if their values change between calls.

This lives entirely in Nx core (`Nx.Defn.Expr` and `Nx.Defn.Compiler`). No changes to EXLA.

The original version of this PR used an ETS-based recompilation counter in EXLA. Based on review feedback, I moved the check upstream to where constants actually enter the graph. This catches all sources of tensor constants (closure captures, deftransform, anything that flows through `to_expr`), not just the closure capture case.

## How it works

In `Nx.Defn.Expr.to_expr/1`, when a `BinaryBackend` tensor with non-scalar shape and 10+ elements is converted to a `:tensor` expression node, the warning fires. A process dictionary flag limits it to once per trace. The flag is saved and restored in `Nx.Defn.Compiler.runtime_fun/3` so that nested jit calls do not interfere with the outer trace's state.

The 10-element threshold avoids warning on small intentional constants like masks or indices. The warning message notes that intentionally constant tensors can be safely ignored.

## Test plan

- [x] Warning fires when a large tensor is used as a constant in a closure
- [x] No warning for scalar constants
- [x] No warning for small tensor constants (below threshold)
- [x] No warning when tensors are passed as function arguments
- [x] Only one warning per trace, even with multiple large constants
- [x] Full `mix test` passes in both `nx` and `exla`

Closes #1714

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -486,9 +486,37 @@ defmodule EXLA.Defn do
         :telemetry.execute([:exla, :compilation], measurements, %{key: key})
       end
 
+      if evaled, do: check_recompilation(key, args_key)
+
       outfeed = Outfeed.with_user_hooks(outfeed, hooks)
       {executable, {used_inputs, outputs, outfeed, inputs_and_typespecs}}
     end)
+  end
+
+  @recompilation_threshold 10
+
+  defp check_recompilation(key, args_key) do
+    {:module, mod} = :erlang.fun_info(key, :module)
+    {:new_index, idx} = :erlang.fun_info(key, :new_index)
+    counter_key = {:recompilation, {mod, idx}, args_key}
+
+    count =
+      :ets.update_counter(
+        EXLA.Defn.LockedCache,
+        counter_key,
+        {2, 1},
+        {counter_key, 0}
+      )
+
+    if count == @recompilation_threshold do
+      Logger.warning(
+        "EXLA has compiled the same function #{count} times with the same input " <>
+          "shapes. This typically means tensor values are being captured inside a " <>
+          "closure passed to defn, jit, or value_and_grad. Each distinct captured " <>
+          "value forces a full recompilation. Pass changing tensors as explicit " <>
+          "function arguments instead."
+      )
+    end
   end
 
   defp us_to_ms(time), do: Float.round(time / 1000, 1)
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -486,7 +486,7 @@ defmodule EXLA.Defn do
         :telemetry.execute([:exla, :compilation], measurements, %{key: key})
       end
 
-      if evaled, do: check_recompilation(key, args_key)
+      if evaled && cache, do: check_recompilation(key, args_key)
 
       outfeed = Outfeed.with_user_hooks(outfeed, hooks)
       {executable, {used_inputs, outputs, outfeed, inputs_and_typespecs}}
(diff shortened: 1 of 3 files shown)


exla/lib/exla/defn.ex
```diff
@@ -483,7 +483,15 @@ defmodule EXLA.Defn do
           total_time: eval_time + comp_time
         }
 
-        :telemetry.execute([:exla, :compilation], measurements, %{key: key})
+        :telemetry.execute([:exla, :compilation], measurements, %{
+          key: key,
+          function_identity: fun_identity(key),
```
josevalim: `key` is always a function, so we don't need to pass this information.

exla/lib/exla/defn.ex
```diff
@@ -493,6 +501,70 @@ defmodule EXLA.Defn do
...
+    end
+  end
+
+  defp fun_identity(fun) when is_function(fun) do
+    {:new_uniq, uniq} = :erlang.fun_info(fun, :new_uniq)
+    uniq
+  end
+
+  defp fun_identity(key) do
+    :erlang.phash2(key)
```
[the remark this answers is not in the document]
jeffreyksmithjr: Removed.

exla/lib/exla/defn.ex
```diff
@@ -483,7 +483,15 @@ defmodule EXLA.Defn do
           total_time: eval_time + comp_time
         }
 
-        :telemetry.execute([:exla, :compilation], measurements, %{key: key})
+        :telemetry.execute([:exla, :compilation], measurements, %{
+          key: key,
+          function_identity: fun_identity(key),
+          compilation_status: if(evaled, do: :compiled, else: :cache_hit)
```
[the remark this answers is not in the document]
jeffreyksmithjr: I simplified this out, but could restore it or do it elsewhere, if helpful.

jeffreyksmithjr:
Maybe this approach is better? It at least seems closer to the point of the user's needed action. The choice of 10 elements is arbitrary but would seem to be likely to help people making the same sort of error I made in my code.

Definitely less added infra to achieve this approach.

exla/lib/exla/defn.ex
```diff
@@ -486,11 +486,39 @@ defmodule EXLA.Defn do
...
+      :ets.update_counter(
+        EXLA.Defn.LockedCache,
+        counter_key,
+        {2, 1},
+        {counter_key, 0}
+      )
+
+    if count == @recompilation_threshold do
+      Logger.warning(
+        "EXLA has compiled the same function #{count} times with the same input " <>
```
josevalim: We should print the function (`inspect(key)`)!

exla/lib/exla/defn.ex
```diff
@@ -486,11 +486,39 @@ defmodule EXLA.Defn do
         :telemetry.execute([:exla, :compilation], measurements, %{key: key})
       end
 
+      if evaled, do: check_recompilation(key, args_key)
```
josevalim:
```suggestion
      if evaled && cache, do: check_recompilation(key, args_key)
```

exla/lib/exla/defn.ex
```diff
@@ -486,11 +486,39 @@ defmodule EXLA.Defn do
...
+  @recompilation_threshold 10
+
+  defp check_recompilation(key, args_key) do
+    {:module, mod} = :erlang.fun_info(key, :module)
+    {:new_index, idx} = :erlang.fun_info(key, :new_index)
+    counter_key = {:recompilation, {mod, idx}, args_key}
+
+    count =
+      :ets.update_counter(
+        EXLA.Defn.LockedCache,
```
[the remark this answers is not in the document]
josevalim: Sounds good but let's have a function in like `EXLA.Defn.LockedCache.count(key)`, whicn you would call as `EXLA.Defn.LockedCache.count({mod, idx, args_key})`. Then the count function itself wraps it in a `{:counter, key}` tuple. This way we don't leak ETS names throughout the codebase.

exla/lib/exla/defn.ex
jeffreyksmithjr: Done.

exla/lib/exla/defn.ex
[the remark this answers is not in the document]
jeffreyksmithjr: Done.

exla/lib/exla/defn.ex
jeffreyksmithjr: Done.

jeffreyksmithjr:
Addressed the latest round of feedback.

josevalim (approved):
@polvalente we can merge if you don't have any objections!

polvalente (approved):

PR #1717, opened 2026-03-26
Support vectorized gradients via boundary devectorization

blasphemetheus:
Minimal approach: instead of per-op vectorization adjustments in the backward pass, handle vectorization at the boundary only:

1. Devectorize the gradient seed (constant 1.0 uses devec output shape)
2. Devectorize expression args in recur_to_grad (backward pass stays in devectorized space)
3. Re-vectorize the final gradient in to_grad to match input's vectorized_axes

229/232 existing tests pass. 3 failures are the mixed-vectorization case (non-vectorized target in vectorized context) where unbroadcast sums over batch dims — needs further work.

### DIFF
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -26,14 +26,13 @@ defmodule Nx.Defn.Grad do
 
     {parents, nodes} = parents_tree(transformed_expr, ids)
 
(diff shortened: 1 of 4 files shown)


blasphemetheus:
sorry meant to open that to my fork so i can iterate with ci

PR #1719, opened 2026-03-31
Add more padding variations: cycle, replicate and mirror

laszlokorte:
Until now only constante padding (`Nx.pad/3`) and reflection padding `Nx.reflect/2` are supported by Nx.

This pull request adds `Nx.cycle/2` for cyclic padding, `Nx.replicate/2` for using the outmost value along each axis as pad value and `Nx.mirror/2` which behaves similar to `Nx.reflect/2` but does repeat the edge values.

I did already propose especially the cyclic padding in https://github.com/elixir-nx/nx/issues/1708

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -4241,15 +4241,13 @@ defmodule Nx do
   defp left_cycle_index_period(1), do: Nx.tensor([0])
 
   defp left_cycle_index_period(n) do
-    n |> Nx.subtract(Nx.iota({n})) |> Nx.subtract(1) |> Nx.remainder(n)
+    Nx.subtract(n - 1, Nx.iota({n})) |> Nx.remainder(n)
   end
 
   # n == 0 is handled at the call side
   defp right_cycle_index_period(1), do: Nx.tensor([0])
 
-  defp right_cycle_index_period(n) do
-    Nx.iota({n})
-  end
+  defp right_cycle_index_period(n), do: Nx.iota({n})
 
   # n == 0 is handled at the call side
   defp left_mirror_index_period(1), do: Nx.tensor([0])
@@ -4278,7 +4276,7 @@ defmodule Nx do
   defp left_reflect_index_period(n) do
     # Generates the indices for pre-reflecting on the axis
     base = Nx.iota({n - 1})
-    left = base |> Nx.add(1)
+    left = Nx.add(base, 1)
     right = Nx.subtract(n - 2, base)
     Nx.concatenate([left, right])
   end
@@ -4290,23 +4288,19 @@ defmodule Nx do
     # Generates the indices for post-reflecting on the axis
     base = Nx.iota({n - 1})
     left = Nx.subtract(n - 2, base)
-    right = base |> Nx.add(1)
+    right = Nx.add(base, 1)
     Nx.concatenate([left, right])
   end
 
   # n == 0 is handled at the call side
   defp left_replicate_index_period(1), do: Nx.tensor([0])
 
-  defp left_replicate_index_period(n) do
-    Nx.broadcast(0, {n})
-  end
+  defp left_replicate_index_period(n), do: Nx.broadcast(0, {n})
 
   # n == 0 is handled at the call side
   defp right_replicate_index_period(1), do: Nx.tensor([0])
 
-  defp right_replicate_index_period(n) do
-    Nx.broadcast(n - 1, {n})
-  end
+  defp right_replicate_index_period(n), do: Nx.broadcast(n - 1, {n})
 
   defp padding_with_index(tensor, opts) do
     opts = keyword!(opts, [:padding_config, :left_index_period, :right_index_period])
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -16988,7 +16988,7 @@ defmodule Nx do
       which specify the length (0 or greater) of the reflection before and
       after the tensor along a each axis.
 
-  See also: `pad/3`
+  See also: `pad/3`, `cycle/2`, `replicate/2`, `mirror/2`
 
   ## Examples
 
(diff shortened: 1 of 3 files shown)


nx/lib/nx.ex
```diff
@@ -17013,8 +17013,205 @@ defmodule Nx do
...
+    right = Nx.subtract(n - 1, Nx.iota({n}))
+    Nx.concatenate([left, right])
+  end
+
+  defp right_mirror_index_period(1), do: Nx.tensor([0])
+
+  defp right_mirror_index_period(n) do
+    # Generates the indices for post-mirroring on the axis
+    left = Nx.subtract(n - 1, Nx.iota({n}))
+    right = Nx.iota({n})
```
```suggestion
    right = Nx.iota({n})
    left = Nx.subtract(n - 1, right)
```

we might need to do something a bit more complicated here, in that iota being a creation function might need to receive the backend opts extracted from the input argument

nx/lib/nx.ex
```diff
@@ -17013,8 +17013,205 @@ defmodule Nx do
...
+      right_index_period: &right_mirror_index_period/1
+    )
+  end
+
+  defp left_mirror_index_period(1), do: Nx.tensor([0])
+
+  defp left_mirror_index_period(n) do
+    # Generates the indices for pre-mirroring on the axis
+    left = Nx.iota({n})
+    right = Nx.subtract(n - 1, Nx.iota({n}))
```
```suggestion
    left = Nx.iota({n})
    right = Nx.subtract(n - 1, left)
```

nx/lib/nx.ex
```diff
@@ -17013,8 +17013,205 @@ defmodule Nx do
...
+        s32[x: 4][y: 7]
+        [
+          [1, 0, 0, 1, 2, 2, 1],
+          [1, 0, 0, 1, 2, 2, 1],
+          [4, 3, 3, 4, 5, 5, 4],
+          [4, 3, 3, 4, 5, 5, 4]
+        ]
+      >
+  """
+  @doc type: :shape
```
polvalente: @josevalim WDYT we add `type: :padding` instead and move some of the previous funs there as well?

nx/lib/nx.ex
```diff
@@ -17013,8 +17013,205 @@ defmodule Nx do
...
+    right = Nx.subtract(n - 2, Nx.iota({n - 1}))
+    Nx.concatenate([left, right])
+  end
+
+  defp right_reflect_index_period(1), do: Nx.tensor([0])
+
+  defp right_reflect_index_period(n) do
+    # Generates the indices for post-reflecting on the axis
+    left = Nx.subtract(n - 2, Nx.iota({n - 1}))
+    right = Nx.iota({n - 1}) |> Nx.add(1)
```
```suggestion
    # Generates the indices for post-reflecting on the axis
    base = Nx.iota({n - 1})
    left = Nx.subtract(n - 2, base)
    right = Nx.add(base, 1)
```

nx/lib/nx.ex
```diff
@@ -17013,8 +17013,205 @@ defmodule Nx do
...
+      right_index_period: &right_reflect_index_period/1
+    )
+  end
+
+  defp left_reflect_index_period(1), do: Nx.tensor([0])
+
+  defp left_reflect_index_period(n) do
+    # Generates the indices for pre-reflecting on the axis
+    left = Nx.iota({n - 1}) |> Nx.add(1)
+    right = Nx.subtract(n - 2, Nx.iota({n - 1}))
```
```suggestion
    base = Nx.iota({n - 1})
    left = Nx.add(base, 1)
    right = Nx.subtract(n - 2, base)
```

nx/lib/nx.ex
```diff
@@ -17013,8 +17013,205 @@ defmodule Nx do
...
     opts = keyword!(opts, [:padding_config])
 
+    padding_with_index(tensor,
+      padding_config: opts[:padding_config],
+      left_index_period: &left_reflect_index_period/1,
+      right_index_period: &right_reflect_index_period/1
+    )
+  end
+
+  defp left_reflect_index_period(1), do: Nx.tensor([0])
```
polvalente: ~should we be matching on 0 as well?~

Let's add a comment above each of these saying that we match on 1 specifically because 0 is handled at the call site

nx/lib/nx.ex
```diff
@@ -16988,7 +16988,7 @@ defmodule Nx do
       which specify the length (0 or greater) of the reflection before and
       after the tensor along a each axis.
 
-  See also: `pad/3`
+  See also: `pad/3`, `cycle/2`, `replicate/2`, `mirror/2`
```
polvalente: Shall we add :cycle, :replicate, :reflect and :mirror as padding modes to Nx.pad?

nx/lib/nx.ex
polvalente: @josevalim do you think this is something we should be concerned about?

polvalente (commented):
Very nice PR!

Let's also add some tests (maybe in nx_test.exs?) that stress test some of the edge cases, such as negative padding config (Nx.pad does slicing in the case of negative padding, maybe these need to do the same? Do we have Numpy references for these?)

josevalim:
Sorry, I originally thought we were adding only a single function, but after seeing we need multiple and looking at the implementation (thank you!), I think we should go ahead and pass the `mode` as second argument to `pad`, as done for numpy: https://numpy.org/doc/stable/reference/generated/numpy.pad.html (using same names too). Especially because the padding_config is still a required argument. In this scenario, we would deprecate `Nx.reflect/2`.

For example: `Nx.pad(tensor, :reflect, padding_config)`, `Nx.pad(tensor, :symmetric, padding_config)`. Plus `numpy` has a few other modes we may want to add, so this will avoid adding 7-8 different functions.

laszlokorte:
I agree that unifying with  `pad/3` would be much cleaner.

nx/lib/nx.ex
```diff
@@ -17013,8 +17013,205 @@ defmodule Nx do
...
+  defp left_replicate_index_period(1), do: Nx.tensor([0])
+
+  defp left_replicate_index_period(n) do
+    0 |> Nx.broadcast({n})
+  end
+
+  defp right_replicate_index_period(1), do: Nx.tensor([0])
+
+  defp right_replicate_index_period(n) do
+    (n - 1) |> Nx.broadcast({n}) |> Nx.remainder(n)
```
josevalim: I don't think we need the remainder on this one?

laszlokorte: you are right!

laszlokorte:
I have update the code to unify the 4 padding styles into the `pad/3` function. currently the padding_config for `pad/3` expected `{left, right, interior}` for the constant padding but the `reflect/2` function - that I based the other padding styles on - expects `{left, right}`, not supporting interior padding. 

I am not sure what makes most sense API-wise. Just document that the padding_config format differs depending on padding type or come up with some sensible value for interior values (just constant `0`?) for the other padding types.

nx/lib/nx.ex
```diff
@@ -17010,98 +17275,15 @@ defmodule Nx do
       >
   """
   @doc type: :shape
+  @deprecated "Use pad/3 instead"
```
laszlokorte: currently the deprecation causes the doctests for `reflect/2` trigger a warning. I am not sure how to disable this.

josevalim: That's fine for now!

josevalim:
Ah, if they have different padding_config then it may make sense to give the function a different name. Overall, I think this approach is better, we just need to figure out the name.

laszlokorte:
I agree that making the padding_config structure vary by padding type is confusing but a unifying approach could be:
For constant padding allow {left, right, interior} (as before) as well as {left, right} (new) defaulting {left, right, 0}
and for the other padding types also allow both {left, right, interior \\ 0} and {left, right} but just using 0 as constant interior value.

Alternative names could be `pad_outside` or `pad_around` but at a first glance I do not like either.

polvalente:
@laszlokorte @josevalim How about we go with pad_outer for the new function name?

josevalim:
`pad_outer` sounds like a winner to me!

josevalim (approved):

PR #1722, opened 2026-04-01
fix: ensure return body doesn't contain closures in while

polvalente:
closes #1716

### DIFF
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -298,6 +298,50 @@ defmodule Nx.Defn.Expr do
             "condition for while cannot be vectorized, got: #{inspect(condition)}"
     end
 
+    # Extract the parent scope's context from `initial`. The body must not contain
+    # tensors with this context: that would mean a tensor was captured from outside
+    # the while and placed directly in the return tuple, bypassing the usual
+    # merge_context! checks that Nx operations perform. Tensors from nested inner
+    # whiles (which carry a different {:while, ref} context) are allowed.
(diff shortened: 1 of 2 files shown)


nx/lib/nx/defn/expr.ex
```diff
@@ -298,6 +298,50 @@ defmodule Nx.Defn.Expr do
             "condition for while cannot be vectorized, got: #{inspect(condition)}"
     end
 
+    # Extract the parent scope's context from `initial`. The body must not contain
+    # tensors with this context: that would mean a tensor was captured from outside
+    # the while and placed directly in the return tuple, bypassing the usual
+    # merge_context! checks that Nx operations perform. Tensors from nested inner
+    # whiles (which carry a different {:while, ref} context) are allowed.
+    parent_context = Composite.reduce(initial, nil, fn tensor, acc ->
+      merge_context!(tensor, acc)
```
josevalim: This means initial itself cannot have mixed context but I think that's fine, right?

josevalim (approved):

nx/lib/nx/defn/expr.ex
polvalente: I think so. Although if you have an example on which initial would have different ctx we could check this in tests

josevalim: Yeah, I can't think of any example. I think `cond` no longer assigns different contexts, right? So the only example would be a while inside a while with the inner while referring to a variable inside of the while and one from outside, which is clearly broken too.

polvalente: Yeah, that's what I was thinking. Since no previous tests broke, I think we can move forward with this one

PR #1732, opened 2026-04-14
feat(exla): configurable shm permissions for IPC buffers, default 0400

blasphemetheus:
Addresses @polvalente's review feedback on the original security-fix PR: make the shared memory permissions a configurable option with a functional-by-default `0o400` default, per Pol's explicit suggestion.

## Problem

`exla/c_src/exla/ipc.cc` previously called `shm_open` with mode `0o666`, making EXLA's IPC shared memory segments world-readable (and potentially world-writable depending on the caller's umask). On multi-tenant systems — shared GPU workstations, cluster nodes, CI runners with multiple users — any other local user could read tensor data from a running EXLA process: model weights, training data, inference inputs, intermediate activations.

The segments are only ever accessed by the same user's EXLA processes (the IPC channel is between the BEAM and the local NIF), so there is no legitimate default use case for cross-user access.

## Fix

```elixir
# Default — owner-read-only (was world-readable 0o666).
# Matches EXLA's functional-by-default spirit: the tensor is an
# immutable value, so the shared segment is read-only to attachers.
Nx.to_pointer(tensor, mode: :ipc)

# Explicit alternative — e.g. writable by the owner
Nx.to_pointer(tensor, mode: :ipc, shm_permissions: 0o600)
```

The creator still gets a read/write `O_RDWR` fd (Linux skips the access check on file creation), so `ftruncate` and the memcpy from device memory succeed even with a `0o400` mode. Only subsequent opens by other processes are read-only by default.

`:shm_permissions` is validated at the Elixir surface: must be an integer in the range `0..0o7777` (POSIX file mode bits, including setuid/setgid/sticky). The NIF takes it as `int64_t` and `get_ipc_handle` casts to `mode_t` before handing it to `shm_open(3)`.

## Changes

- **`exla/c_src/exla/ipc.h` / `ipc.cc`** — `get_ipc_handle` takes a `mode_t mode` parameter; previously hardcoded to `0o666`.
- **`exla/c_src/exla/exla.cc`** — `get_buffer_device_pointer` NIF takes a new `shm_permissions` `int64_t` argument and passes it through.
- **`exla/lib/exla/nif.ex`** — NIF declaration updated to `/4`.
- **`exla/lib/exla/backend.ex`** — `to_pointer/2` accepts `shm_permissions: 0o400` in its `Keyword.validate!`, validates the range, and passes it to the NIF.
- **`nx/lib/nx.ex`** — docstring for `Nx.to_pointer/2` mentions `:shm_permissions` as a relayed option with examples.
- **`exla/test/exla/device_memory_sharing_test.exs`** — five new tests:
  - `defaults to 0o400 (owner-read-only, functional-by-default)` — `File.stat!`s the `/dev/shm` segment and asserts the lower 12 bits are `0o400`.
  - `accepts an explicit permission value` — same, with `0o600`.
  - `rejects non-integer shm_permissions` / `rejects negative` / `rejects shm_permissions above 0o7777` — argument validation.
  - The `/dev/shm`-based tests are gated on `File.dir?("/dev/shm")` so they no-op on platforms (macOS, some BSDs) that don't expose shm there, and clean up the segment via `on_exit` to avoid polluting the CI runner's `/dev/shm`. The argument-validation tests are universal.

## Threat model note

`0o400` prevents **cross-user** read access on multi-tenant systems. It does not protect against **same-UID compromise** — a malicious sibling process running as the same user can still open the segment read-only and map it. True isolation would require mount namespaces or `memfd_create`, which are out of scope for this PR.

## Compatibility

- Existing callers of `Nx.to_pointer(t, mode: :ipc)` now get `0o400` instead of `0o666`, which is the security improvement. The creator's writes still work normally.
- Any caller who explicitly depended on world-readable shm (undocumented, not a supported use case) can opt back in by passing `shm_permissions: 0o666`.
- The NIF signature change is internal: `EXLA.NIF.get_buffer_device_pointer` is a private API with no external callers (verified via grep across `/home/blewf/git/`).

## Test plan

- [x] `mix test test/exla/device_memory_sharing_test.exs` — 8 tests, 0 failures (3 pre-existing + 5 new).
- [x] `File.stat!("/dev/shm/exla:ipc:...")` confirms the `shm_open` mode matches the option passed.
- [x] `mix compile --warnings-as-errors` — clean.

### DIFF
--- a/exla/c_src/exla/exla.cc
+++ b/exla/c_src/exla/exla.cc
@@ -322,7 +322,7 @@ get_buffer_device_pointer(ErlNifEnv *env, fine::ResourcePtr<ExlaClient> client,
 
   if (pointer_kind == "host_ipc") {
     auto handle_name =
-        "exla:ipc:" + std::to_string(device_size) + ":" + std::to_string(ptr);
+        "exla:ipc:" + std::to_string(getpid()) + ":" + std::to_string(ptr);
     auto fd = get_ipc_handle(handle_name.c_str(), device_size,
                              static_cast<mode_t>(shm_permissions));
 
@@ -330,7 +330,7 @@ get_buffer_device_pointer(ErlNifEnv *env, fine::ResourcePtr<ExlaClient> client,
       throw std::runtime_error("unable to get IPC handle");
     }
 
-    auto ipc_ptr = open_ipc_handle(fd, device_size, 1);
+    auto ipc_ptr = open_ipc_handle(fd, device_size, /*writable=*/1);
     if (ipc_ptr == nullptr) {
       throw std::runtime_error("unable to open IPC handle");
     }
--- a/exla/c_src/exla/exla.cc
+++ b/exla/c_src/exla/exla.cc
@@ -4,6 +4,7 @@
 #include <stdio.h>
 #include <string>
 #include <tuple>
+#include <unistd.h>
 #include <unordered_map>
 
 #include "absl/log/globals.h"
@@ -306,7 +307,6 @@ FINE_NIF(mlir_compile, ERL_NIF_DIRTY_JOB_CPU_BOUND);
 // ExlaBuffer Functions
 
 std::variant<std::tuple<fine::Atom, uint64_t, uint64_t>,
-             std::tuple<fine::Atom, std::string, uint64_t, uint64_t>,
              std::tuple<fine::Atom, std::string, uint64_t>>
 get_buffer_device_pointer(ErlNifEnv *env, fine::ResourcePtr<ExlaClient> client,
                           fine::Term buffer_term, fine::Atom pointer_kind,
@@ -330,15 +330,14 @@ get_buffer_device_pointer(ErlNifEnv *env, fine::ResourcePtr<ExlaClient> client,
       throw std::runtime_error("unable to get IPC handle");
     }
 
-    auto ipc_ptr = open_ipc_handle(fd, device_size);
+    auto ipc_ptr = open_ipc_handle(fd, device_size, 1);
     if (ipc_ptr == nullptr) {
       throw std::runtime_error("unable to open IPC handle");
     }
 
     memcpy(ipc_ptr, reinterpret_cast<void *>(ptr), device_size);
 
-    return std::make_tuple(pointer_kind, handle_name, static_cast<uint64_t>(fd),
-                           device_size);
+    return std::make_tuple(pointer_kind, handle_name, device_size);
   }
 
   if (pointer_kind == "cuda_ipc") {
@@ -371,17 +370,20 @@ fine::ResourcePtr<ExlaBuffer> create_buffer_from_device_pointer(
     }
     ptr = maybe_pointer.value();
   } else if (pointer_kind == "host_ipc") {
-    auto tuple =
-        fine::decode<std::tuple<uint64_t, std::string>>(env, pointer_data);
-    auto fd = std::get<0>(tuple);
-    auto memname = std::get<1>(tuple);
+    auto memname = fine::decode<std::string>(env, pointer_data);
     auto device_size = xla::ShapeUtil::ByteSizeOf(shape);
-    ptr = open_ipc_handle(fd, device_size);
+    int writable = 0;
+    auto fd = open_existing_ipc_handle(memname.c_str(), &writable);
+    if (fd == -1) {
+      throw std::runtime_error("unable to get fd for IPC handle");
+    }
+    ptr = open_ipc_handle(fd, device_size, writable);
     if (ptr == nullptr) {
+      close(fd);
       throw std::runtime_error("unable to get pointer for IPC handle");
     }
-    on_delete_callback = [fd, memname, ptr, device_size]() {
-      close_ipc_handle(fd, ptr, memname.c_str(), device_size);
+    on_delete_callback = [fd, ptr, device_size]() {
+      close_imported_ipc_handle(fd, ptr, device_size);
     };
   } else if (pointer_kind == "local") {
     auto ptr_int = fine::decode<int64_t>(env, pointer_data);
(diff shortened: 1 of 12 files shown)


exla/c_src/exla/ipc.cc
```diff
@@ -8,7 +8,7 @@
 
 // Function to create or open a shared memory object and set its size
 int get_ipc_handle(const char* memname, size_t memsize) {
-  int fd = shm_open(memname, O_CREAT | O_RDWR, 0666);
+  int fd = shm_open(memname, O_CREAT | O_RDWR, 0600);
```
polvalente: Let's make this an option instead. Default to 0b0600 but allow the caller to pass 0o0400 or whatever else suits the application.

exla/lib/exla/backend.ex
```diff
@@ -84,7 +84,13 @@ defmodule EXLA.Backend do
 
   @impl true
   def to_pointer(%T{data: %B{buffer: buffer}}, opts \\ []) do
-    opts = Keyword.validate!(opts, mode: :local)
+    opts = Keyword.validate!(opts, mode: :local, shm_permissions: 0o600)
```
polvalente:
```suggestion
    opts = Keyword.validate!(opts, mode: :local, shm_permissions: 0o400)
```

Since EXLA aims to be functional, let's default to 400

exla/lib/exla/backend.ex
```diff
@@ -84,7 +84,13 @@ defmodule EXLA.Backend do
...
+
+    unless is_integer(opts[:shm_permissions]) and opts[:shm_permissions] >= 0 do
+      raise ArgumentError,
+            ":shm_permissions must be a non-negative integer (typically an octal " <>
+              "literal like 0o600), got: #{inspect(opts[:shm_permissions])}"
```
polvalente:
```suggestion
              "literal like 0o400), got: #{inspect(opts[:shm_permissions])}"
```

polvalente:
@blasphemetheus I pushed a few changes, mainly to allow testing IPC stuff, but also some improvements to the memory mapping and permissions. Can you give this a review before we merge?

polvalente (approved):

blasphemetheus:
awesome, yeah I'll take a look and give a review

exla/lib/exla/backend.ex
```diff
@@ -125,8 +125,8 @@ defmodule EXLA.Backend do
         # Pointer is an integer here
         %Nx.Pointer{kind: :local, address: ptr, data_size: size}
 
-      {:host_ipc, handle_name, fd, size} ->
-        %Nx.Pointer{kind: :ipc, handle: handle_name, address: fd, data_size: size}
+      {:host_ipc, handle_name, size} ->
```
blasphemetheus, line 128: api change, i think this is an improvement

blasphemetheus: so now receiver calls `open_existing_ipc_hangle(memname, ...)` to get the fd (file description) instead of having fd  passed across the NIF boundary into elixir like before, which was worse.

exla/c_src/exla/exla.cc
```diff
@@ -679,6 +681,23 @@ fine::Ok<> start_log_sink(ErlNifEnv *env, ErlNifPid logger_pid) {
 
 FINE_NIF(start_log_sink, 0);
 
+// Test-only NIFs — excluded from production builds.
+#ifndef EXLA_PROD
```
blasphemetheus, line 701: ok, excluded from prod cause there's no bounds check, no type check, no safety. raw memory write exposed to the beam. 

Guard makes it not compiled into production releases.

exla/lib/exla/nif.ex
```diff
@@ -97,5 +97,11 @@ defmodule EXLA.NIF do
   def encode_local_pid(_pid), do: err!()
   def decode_local_pid(_pid_bin), do: err!()
 
+  if Mix.env() != :prod do
```
blasphemetheus, line 100: defense in depth

exla/test/test_helper.exs
```diff
@@ -38,8 +48,33 @@ cuda_required =
     [:cuda_required]
   end
 
+distributed_exclude =
+  cond do
```
blasphemetheus, line 52: when to skip distributed tagged tests (new tag). and when to run them

exla/c_src/exla/exla.cc
```diff
@@ -330,15 +330,14 @@ get_buffer_device_pointer(ErlNifEnv *env, fine::ResourcePtr<ExlaClient> client,
       throw std::runtime_error("unable to get IPC handle");
     }
 
-    auto ipc_ptr = open_ipc_handle(fd, device_size);
+    auto ipc_ptr = open_ipc_handle(fd, device_size, 1);
```
blasphemetheus: could use a named constant/enum to read better instead of a random number. always writable is the meaning i think

exla/test/test_helper.exs
```diff
@@ -38,8 +48,33 @@ cuda_required =
...
+    :distributed in Keyword.get(ExUnit.configuration(), :exclude, []) ->
+      [:distributed]
+
+    Code.ensure_loaded?(:peer) and try_starting_epmd?.() and
+        match?({:ok, _}, Node.start(:"primary@127.0.0.1", :longnames)) ->
+      {:ok, _pid, node2} = :peer.start(%{name: :"secondary@127.0.0.1"})
```
blasphemetheus, line 58: could write a try rescue that gives [:distributed] if something goes wrong if 
- stale node name (so previous test run aborted without cleanup leaves secondary@127.0.0.1 registered with epmd, might result in MatchError, test suite down
- Port collision (something's using the distribution port)
- if the peer isn't started by nx or exla for some reason idk

blasphemetheus (commented):
Looks good! might add one test for a new path. could add a try rescue in the distributed positive case (so when you're trying to run distributed tests) to guard against weirdness. There is a change to public function (address: fd removed), so could note that somewhere.

blasphemetheus:
test didn't really add anything not already covered, reverted that

exla/lib/exla/backend.ex
polvalente: Yeah, the fd was passed in case we tried to re-open the pointer from the same instance, but this still works with just the handle

PR #1737, opened 2026-04-19
docs: fix undefined c:block/4 callback reference

preciz:
The @doc for Nx.block/4 incorrectly referenced c:block/4, which is not defined in the Nx module. This updates it to point to the correct callback c:Nx.Backend.block/4.


polvalente (approved):

polvalente:
Thanks!

PR #1739, opened 2026-04-20
Refactor EXLA block lowering through EXLA.CustomCall protocol

Chapaman:
## Summary

- Introduced **`EXLA.CustomCall`** with a single callback **`call/4`** (no separate `apply?`, `function_name`, or `config`). It returns **`:skip`** to compile the block body, or **`{:ok, %EXLA.CustomCall.Spec{}}`** with **`call_target_name`**, optional **`backend_config`** (StableHLO backend config), and **`operand_element_types`** (**`:infer`** by default, or an explicit list of per-input Nx types so **`EXLA.Defn`** inserts **`convert`**s when traced dtypes differ from what the native FFI expects). Uses **`@fallback_to_any true`**; built-in C-backed lowerings for **`Nx.Block.LinAlg.QR`** and **`Nx.Block.LinAlg.Eigh`** live in **`defimpl EXLA.CustomCall, for: Any`** (with **`defp`** target helpers). Blocks that are not XLA custom calls (e.g. **`Take`**, **`TopK`**, FFT family) remain **lowered inline** in **`EXLA.Defn`** — see the protocol moduledoc.

- **`EXLA.Defn`**: generic **`cached_recur_operator(:block, …)`** path — recurse tensor args, then **`EXLA.CustomCall.call/4`** → if **`:skip`**, **`default_block_implementation/5`**; if **`{:ok, spec}`**, apply optional operand casts, then **`Value.custom_call`**.

- **Helpers**: **`to_type`**, **`op_type`**, **`op_shape`**, **`expr_to_typespec`**, **`axes_for_rank`** are **`defp`** again (not public API). **`fft` / `fft2`** remain **`def`** helpers for the inline FFT **`:block`** clauses (builder comes from **`%Value{}.function`** where applicable).

- **Note**: QR eligibility still uses the **kind** from the **Q** slot of the output template (e.g. **`{:f, _}`** vs **`{:c, _}`**), not **`q.type != :c`** (types are tuples).


exla/lib/exla/custom_call.ex
```diff
@@ -0,0 +1,193 @@
...
+  # --- apply?/4 ---
+
+  def apply?(
+        %Nx.Block.LinAlg.QR{},
+        {%{type: {q_type_kind, _}}, _r},
+        _args,
+        client
+      ) do
+    q_type_kind != :c and client.platform == :host
+  end
```
Chapaman: We split this into **`apply?/4`** and **`call/4`** because a single callback would mix “should EXLA take the native path for *this* compile?” with “emit MLIR.” Keeping two functions separates **eligibility** from **lowering**.

- **`apply?`** answers whether the native path applies for this **block tag** *in context*: e.g. for QR we still have `%Nx.Block.LinAlg.QR{}`, but we only native-lower when the **output type** is real (not complex) **and** the **client** is `:host`. Another block could gate on arity, struct fields, mesh, etc.—without duplicating the full `call` body.
- **`call`** only runs when `apply?` is true and contains the actual **`Value.*` / Defn lowering**.

`EXLA.Defn` mirrors that split: recurse operands → `apply?` → either `call` or the generic block fallback (`default_block_implementation`).

josevalim: You could have a single callback and instead return `:skip` in `call` when it cannot be lowered.

exla/lib/exla/custom_call.ex
```diff
@@ -0,0 +1,193 @@
...
+  @doc """
+  Lowers the block natively.
+
+  Must return the list of `EXLA.MLIR.Value`s (or a single value) that
+  represents the block result, matching the shape of `out`.
+  """
+  def call(struct, out, args, client)
+end
+
+defimpl EXLA.CustomCall, for: Any do
```
josevalim: Should those be different implementations rather than having it all in `Any`?

josevalim:
@polvalente @Chapaman very nice! However, I think we should couple this more closely to actual custom calls (in C), because that's how it will work in practice. In other words, we should only move to the protocol blocks which are implemented as custom C calls. This will make it easier to see what they have in common (and we won't need to get distracted with things like take, which is quite different).

exla/lib/exla/custom_call.ex
polvalente: This way the user can provide their own overrides!

josevalim: Ah, I see! Let's add a comment explaining why then! :D

exla/c_src/exla/exla.cc
```diff
@@ -535,6 +537,20 @@ fine::Ok<> load_pjrt_plugin(ErlNifEnv *env, std::string device_type,
...
+  void *handle = dlopen(path.c_str(), RTLD_NOW | RTLD_GLOBAL);
+  if (handle == nullptr) {
+    const char *err = dlerror();
+    throw std::invalid_argument(err ? err : "dlopen failed");
+  }
+  (void)handle;
+  return fine::Ok();
+}
+
+FINE_NIF(dlopen_test_plugin, 0);
```
polvalente: Let's expose this as load_dylib so that it's not used just for tests

exla/c_src/exla_test_plugin/qr_alias_registration.cc
```diff
@@ -0,0 +1,16 @@
+// Test-only shared library: registers an alias FFI name that reuses the
+// existing qr_cpu_custom_call_f32 handler symbol from libxla_extension.so.
+//
+// Built when BUILD_EXLA_TEST_PLUGIN=1 (Mix test). Load with RTLD_GLOBAL via
+// EXLA.NIF.dlopen_test_plugin/1 before compiling or running graphs that emit
+// the alias call_target_name.
```
polvalente: We can wrap this file in `ifndef EXLA_PROD` to exclude its contents from in MIX_ENV=prod.
Let's also remove the plugin suffix. I suggest exla_test/custom_calls.cc as the file path in c_src

exla/Makefile
```diff
@@ -86,6 +86,21 @@ else
...
+# Optional test plugin: registers qr_cpu_custom_call_f32_exla_alias -> same
+# handler as qr_cpu_custom_call_f32 (Mix sets BUILD_EXLA_TEST_PLUGIN=1 in :test).
+TEST_PLUGIN_CC = c_src/exla_test_plugin/qr_alias_registration.cc
+TEST_PLUGIN_SO = $(PRIV_DIR)/test/exla_qr_alias_plugin.so
+
+ifneq ($(BUILD_EXLA_TEST_PLUGIN),)
+ifneq ($(BUILD_EXLA_TEST_PLUGIN),0)
+$(EXLA_SO): $(TEST_PLUGIN_SO)
+endif
+endif
```
polvalente:
```suggestion
```

exla/mix.exs
```diff
@@ -33,7 +33,8 @@ defmodule EXLA.MixProject do
           "FINE_INCLUDE_DIR" => Fine.include_dir(),
           "MIX_BUILD_EMBEDDED" => "#{Mix.Project.config()[:build_embedded]}",
           "CWD_RELATIVE_TO_PRIV_PATH" => cwd_relative_to_priv,
-          "EXLA_VERSION" => "#{@version}"
+          "EXLA_VERSION" => "#{@version}",
+          "BUILD_EXLA_TEST_PLUGIN" => if(Mix.env() == :test, do: "1", else: "0")
```
polvalente:
```suggestion
          "EXLA_VERSION" => "#{@version}"
```

exla/Makefile
```diff
@@ -86,6 +86,21 @@ else
...
+
+$(TEST_PLUGIN_SO): $(TEST_PLUGIN_CC) | $(XLA_EXTENSION_DIR)
+	@ mkdir -p $(dir $@)
+	$(CXX) $(CFLAGS) -shared $(TEST_PLUGIN_CC) -o $@ $(LDFLAGS)
+
 $(EXLA_SO): $(EXLA_CACHE_SO)
```
polvalente:
```suggestion
$(TEST_PLUGIN_SO): $(TEST_PLUGIN_CC) | $(XLA_EXTENSION_DIR)
	@ mkdir -p $(PRIV_DIR)
	$(CXX) $(CFLAGS) -shared $(TEST_PLUGIN_CC) -o $@ $(LDFLAGS)

EXLA_SO_DEPS = $(EXLA_CACHE_SO)
ifeq($(MIX_ENV),test)
EXLA_SO_DEPS += $(TEST_PLUGIN_SO)
end

$(EXLA_SO): $(EXLA_SO_DEPS)
```

exla/lib/exla/mlir/value.ex
```diff
@@ -785,6 +785,28 @@ defmodule EXLA.MLIR.Value do
...
   end
 
+  @doc false
+  def qr_with_call_target(
+        %Value{function: func} = value,
+        q_typespec,
+        r_typespec,
+        call_target_name
+      )
+      when is_binary(call_target_name) do
```
polvalente: What if the protocol instead had 2 callbacks:

```elixir
@spec function_name(struct, output_container, input_templates_list, client) :: String.t() | :skip
@spec config(struct, output_container, input_templates_list, client) :: map() | nil
```

function_name would be the string name registered by invoking the .so (what we have here as call_target_name)

input and output typespecs we can infer ourselves

and backend_config would be obtained from the config callback. We'd have to validate all map values are encodable as mlir::DictionaryAttr (but we can encode them ourselves).

polvalente: There's a separate discussion on whether we expose other things such as has_side_effect and output_operand_aliases, but we can add these afterwards.

exla/lib/exla/custom_call.ex
```diff
@@ -94,6 +94,58 @@ defprotocol EXLA.CustomCall do
...
+      {{:s, 8}, {:f, 32}} -> "eigh_cpu_custom_call_s8"
+      {{:s, 16}, {:f, 32}} -> "eigh_cpu_custom_call_s16"
+      {{:s, 32}, {:f, 32}} -> "eigh_cpu_custom_call_s32"
+      {{:s, 64}, {:f, 32}} -> "eigh_cpu_custom_call_s64"
+      {{:u, 8}, {:f, 32}} -> "eigh_cpu_custom_call_u8"
+      {{:u, 16}, {:f, 32}} -> "eigh_cpu_custom_call_u16"
+      {{:u, 32}, {:f, 32}} -> "eigh_cpu_custom_call_u32"
+      {{:u, 64}, {:f, 32}} -> "eigh_cpu_custom_call_u64"
+      _ -> :skip
+    end
```
polvalente: I think the name is invariant to the output type. We can just use the input type

polvalente: Also, why are these not defps in the protocol impl?

Chapaman: > Also, why are these not defps in the protocol impl?

I thought I needed to make it a `def` and not `defp` because I'm using it in both `EXLA.CustomCall` and `EXLA.MLIR.Value` should I change it?

polvalente: Are Value.qr and Value.eigh actually used? I think if they are, they should just be aliases to Value.custom_call (which will end up calling the protocol).

If this is indeed possible, than these functions will just be used in the protocol implementation.

Chapaman: 😅😅 it was not being used lol
removed it :~

exla/lib/exla/custom_call.ex
```diff
@@ -0,0 +1,179 @@
...
+
+  @doc """
+  Host CPU `stablehlo.custom_call` target for `Nx.LinAlg.qr/2`, or `:skip`.
+
+  `operand_type` is the input matrix element type.
+  """
+  def qr_cpu_target(operand_type) do
+    case operand_type do
+      {:f, 32} -> "qr_cpu_custom_call_f32"
+      {:f, 64} -> "qr_cpu_custom_call_f64"
```
polvalente: Let's get rid of this module and only define the target resolution inside the call sites

exla/c_src/exla/exla.cc
```diff
@@ -715,4 +736,145 @@ FINE_NIF(write_to_pointer, 0);
 
 } // namespace exla
 
+// Host QR custom calls: integer operands with f32 Q/R (see Nx.Type.to_floating/1
```
josevalim: This should not be here, right? Is this for test only?

exla/lib/exla/custom_call.ex
```diff
@@ -0,0 +1,157 @@
...
+      {:u, 16} -> "eigh_cpu_custom_call_u16"
+      {:u, 32} -> "eigh_cpu_custom_call_u32"
+      {:u, 64} -> "eigh_cpu_custom_call_u64"
+      _ -> :skip
+    end
+  end
+
+  def function_name(_, _, _, _), do: :skip
+
+  def config(_, _, _, _), do: nil
```
josevalim: We don't need a separate callback, make it return `{:ok, name, config}`. We also need to better document what is the config.

exla/lib/exla/defn.ex
```diff
@@ -2251,19 +2193,23 @@ defmodule EXLA.Defn do
...
+  @doc false
+  def op_type(%Value{} = op), do: Value.get_typespec(op).type
 
-  defp op_shape(%Value{} = op), do: Value.get_typespec(op).shape
+  @doc false
+  def op_shape(%Value{} = op), do: Value.get_typespec(op).shape
 
-  defp to_type(%Value{} = op, type) do
+  @doc false
+  def to_type(%Value{} = op, type) do
```
josevalim: Why are we converting all of these to public?

exla/c_src/exla/exla.cc
polvalente: In refactoring to use the protocol, we ended up losing the ability to do typecasts for the operands like we did before.

We either would have to add these or we'd have to add an extra callback to the protocol to make it work in this simpler version which doesn't leak EXLA.MLIR.

In the final implementation these will be absorbed into elixir-nx/xla and we'll just use them.

exla/lib/exla/custom_call.ex
polvalente: What about if we add output operand aliases too? Would you still prefer a single callback?

exla/lib/exla/defn.ex
polvalente: I think this is leftover from a previous iteration

exla/lib/exla/custom_call.ex
josevalim: We return everything from a single callback, so we make sure the format supports extensions.

exla/c_src/exla/exla.cc
josevalim: @polvalente we could also allow the custom call to return the types of the inputs and we cast them if they different. For most cases, the types of the inputs are the ones given as argument.

polvalente: I guess this is a good middle ground. Let's do this and we'll have a proper struct for the return format with the input types, callback name, attributes and whatnot

exla/lib/exla/defn.ex
```diff
@@ -998,6 +910,65 @@ defmodule EXLA.Defn do
     {to_operator(op, args, expr, state), cache}
   end
 
+  defp cast_custom_call_operands(call_args, :infer), do: call_args
```
polvalente: How about we also use :skip? :infer I think would have this implicit meaning of "convert to some type if needed" instead of just "pass as is"

polvalente: or even just use nil

exla/lib/exla/custom_call/spec.ex
```diff
@@ -0,0 +1,33 @@
+defmodule EXLA.CustomCall.Spec do
+  @moduledoc """
+  Result of `EXLA.CustomCall.call/4` when lowering a tagged `Nx.block/4` to
+  `stablehlo.custom_call`.
+
+    * **`call_target_name`** — XLA FFI handler name (`call_target_name` on the op).
+
+    * **`backend_config`** — Optional StableHLO dictionary attribute (`nil` omits it).
+      Same constraints as `EXLA.MLIR.Value.custom_call/4`.
```
josevalim: Shouldn't we call it attributes instead? Why backend_config?

`EXLA.MLIR.Value.custom_call/4` is private API, no? Should we be referring to it here?

exla/lib/exla/mlir/value.ex
```diff
@@ -1088,6 +1047,44 @@ defmodule EXLA.MLIR.Value do
...
+    else
+      raise ArgumentError,
+            "custom_call backend_config key must match [A-Za-z_][A-Za-z0-9_]*, got: #{inspect(key)}"
+    end
+  end
+
+  defp attr_dict_key(key) do
+    raise ArgumentError,
+          "custom_call backend_config key must be an atom or string, got: #{inspect(key)}"
+  end
```
josevalim: It is still unclear to me why we have the backend_config abstraction. Should we just say they are attributes which must be a pair of strings that we pass as is? The only people who will set this will be already knee deep inside StableHLO. I am not sure adding an abstraction on top will benefit anyone.

exla/lib/exla/custom_call.ex
```diff
@@ -0,0 +1,150 @@
...
+
+  ## `call/4` arguments
+
+    * `struct` — the **tag** passed as the first argument to `Nx.block/4`
+      (your own `defstruct` or an existing tag such as `%Nx.Block.LinAlg.QR{}`).
+
+    * `out` — the **output template** tuple passed to `Nx.block/4` (expression
+      metadata for shapes and types, not runtime tensors).
+
+    * `args` — list of **input templates**, in the same order as `inputs` in
```
josevalim: Nitpick: maybe `args` before `out` in the argument list?

exla/lib/exla/custom_call/spec.ex
```diff
@@ -0,0 +1,33 @@
...
+
+    * **`operand_element_types`** — How operand SSA values are presented to the handler:
+
+      * **`:infer`** (default) — use each lowered operand’s element type as produced
+        from the block inputs. No extra converts.
```
josevalim: It is not really infer, right? Perhaps it should be `:default`?

exla/lib/exla/custom_call/spec.ex
```diff
@@ -0,0 +1,33 @@
...
+        types (for example after promotion rules).
+  """
+
+  @enforce_keys [:call_target_name]
+
+  defstruct [:call_target_name, backend_config: nil, operand_element_types: :infer]
+
+  @type t :: %__MODULE__{
+          call_target_name: String.t(),
+          backend_config: map() | nil,
```
josevalim: Why not default to an empty map? 🤔

exla/lib/exla/defn.ex
josevalim: I commented above: `:default`

exla/lib/exla/defn.ex
```diff
@@ -1662,7 +1634,8 @@ defmodule EXLA.Defn do
     end
   end
 
-  defp fft2(exla_op, [%Value{} = tensor, opts], %{type: type} = ans, state) do
+  @doc false
```
josevalim: Do we need to make all of these public? Please double check.

Chapaman: > Do we need to make all of these public? Please double check.

you're right! changed to defp

exla/lib/exla/custom_call/spec.ex
Chapaman: > It is not really infer, right? Perhaps it should be `:default`?

yes, renamed to `:default`

exla/lib/exla/custom_call/spec.ex
Chapaman: > Shouldn't we call it attributes instead? Why backend_config?
> 
> `EXLA.MLIR.Value.custom_call/4` is private API, no? Should we be referring to it here?

`backend_config` was the stablehlo's name, added a mention on the docs but now public field is `attributes`
`EXLA.CustomCall.Spec` docs no longer point at `EXLA.MLIR.Value.custom_call/4`

exla/lib/exla/mlir/value.ex
Chapaman: > It is still unclear to me why we have the backend_config abstraction. Should we just say they are attributes which must be a pair of strings that we pass as is? The only people who will set this will be already knee deep inside StableHLO. I am not sure adding an abstraction on top will benefit anyone.

callers now pass `attributes` as `[{binary_name, binary_attr}]`

exla/lib/exla/custom_call/spec.ex
Chapaman: > Why not default to an empty map? 🤔

agreed - now it defaults to an empty collection (list of `{name, attr}` string pairs)
not exactly a map, but this way both sides stay literal MLIR strings

exla/lib/exla/custom_call/spec.ex
```diff
@@ -0,0 +1,37 @@
+defmodule EXLA.CustomCall.Spec do
+  @moduledoc """
+  Result of `EXLA.CustomCall.call/4` when lowering a tagged `Nx.block/4` to
+  `stablehlo.custom_call`.
+
+    * **`call_target_name`** — XLA FFI handler name (`call_target_name` on the op).
+
+    * **`attributes`** — Optional `{name, attr}` pairs, default `[]`, merged into
+      the `backend_config` dictionary on `stablehlo.custom_call` (StableHLO’s name
+      for that attribute). Each `name` must be a **binary** MLIR identifier; each
```
polvalente:
```suggestion
      for that attribute). Each `name` must be a **BEAM binary** MLIR identifier; each
```

exla/lib/exla/custom_call/spec.ex
```diff
@@ -0,0 +1,37 @@
...
+      `attr` must be a **binary** with valid MLIR attribute syntax for the RHS after
```
polvalente:
```suggestion
      `attr` must be a **BEAM binary** with valid MLIR attribute syntax for the RHS after
```

polvalente: I think it would be nice to support some elixir representations in the future like we do in EXLA.MLIR.Value, but we can leave this for a future iteration

polvalente (approved):

exla/lib/exla/mlir/value.ex
```diff
@@ -1088,6 +1047,32 @@ defmodule EXLA.MLIR.Value do
...
+
+  defp validate_backend_config_key!(key) when is_binary(key) do
+    if Regex.match?(~r/^[A-Za-z_][A-Za-z0-9_]*$/, key) do
+      key
+    else
+      raise ArgumentError,
+            "custom_call backend_config key must match [A-Za-z_][A-Za-z0-9_]*, got: #{inspect(key)}"
+    end
+  end
```
josevalim, line 1074: Remove the regex validation. Users should be responsible to pass what they need anyway. It is a low-level API after all.

exla/lib/exla/custom_call.ex
```diff
@@ -0,0 +1,156 @@
+defprotocol EXLA.CustomCall do
+  @moduledoc """
+  Extension point for lowering selected `Nx.block/4` tags to **XLA custom calls**
+  (`stablehlo.custom_call` in MLIR), the same style as helpers on
+  `EXLA.MLIR.Value` such as `qr/3` and `eigh/3`.
```
josevalim, line 5: This module is private, isn't it? We should not refer to it in public docs.

exla/lib/exla/custom_call.ex
```diff
@@ -0,0 +1,156 @@
...
+
+  Other blocks (for example gather-based `take` or FFT) are lowered inline in
+  `EXLA.Defn` and do not use this protocol.
+
+  ## When `EXLA.Defn` calls it
+
+  During compilation with `compiler: EXLA`, when the builder is an MLIR
+  `EXLA.MLIR.Function`, each `Nx.block(tag, inputs, outputs, fn ... end)` is
```
josevalim, line 13: Internal details leaking through here as well. We should document it from the point of view of users, not of how it is implemented.

josevalim (commented):
Beautiful! I have added only very minor comments! 💯

PR #1752, opened 2026-05-21
Basic io_callback implementation

### LINKED ISSUE #1672: Introduce Nx.io_callback as a replacement for token-based hooks and EXLA's outfeed

Chapaman:
closes #1672

First iteration of `Nx.io_callback/2`: side-effect host callbacks with **passthrough** inputs/outputs, lowered to `stablehlo.custom_call` (`exla_io_callback`). Hooks and `print_value` now use `io_callback` instead of the outfeed/token hook path.

- **Nx:** new `:io_callback` expr op (`{:fn, fun}` / `{:hook, name, callback}`), public `Nx.io_callback/2`, evaluator, tree/grad
- **Hooks:** `hook/3`, `print_value`, and `attach_token` lowered via `io_callback`; `hook_token` + `Nx.Defn.Token` kept only for the shared-token / `attach_token` pattern
- **EXLA:** `Value.io_callback/3`, compiler lowering (host + CUDA), Outfeed `{:exla_io_callback, ...}` handler, C++ FFI (`io_callback.cc` / `io_callback_cuda.cc`)
- **Hook outfeed removed:** EXLA no longer lowers `:token` / `:attach_token` for hooks (raises if legacy nodes appear); `stablehlo.create_token` only when **infeeds** need it
- **Zero-copy:** `output_operand_aliases` per leaf (PID + tensors: `output[0]` → `operand[0]`, `output[i+1]` → `operand[i+1]`; operand `0` = callback server PID)
- **Graph retention:** reassign the result (`x = Nx.io_callback(x, ...)`); Nx DCE runs before XLA sees `has_side_effect`
- **Ordered execution:** PID tensor threading between `io_callback`s — each custom call returns the PID (operand 0 → result 0) and the compiler chains it into the next call, so XLA cannot reorder independent side effects; `if`/`cond` branches reset to a dominating PID (no StableHLO tokens, no evaluator changes)

## Not in this PR (yet)

- Removing `Nx.Defn.Token`, `create_token`, `hook_token`, and the `attach_token` API


exla/c_src/exla/custom_calls/io_callback.cc
```diff
@@ -0,0 +1,75 @@
...
+      callback_server_pid_arg = std::move(tensor);
+    } else {
+      inputs.push_back(std::move(tensor));
+    }
+  }
+
+  // Result buffers are aliased to input buffers via output_operand_aliases —
+  // we must NOT touch them here. Pass no output buffers to the bridge.
+  exla::callback_bridge::Result result =
+      exla::callback_bridge::InvokeIoCallback(
```
polvalente, line 55:
```suggestion
      exla::callback_bridge::InvokeIOCallback(
```

nitpick on naming
I generally keep all letters in an acronym capital (HTTP vs Http, GRPC vs Grpc, etc)

exla/lib/exla/defn/outfeed.ex
```diff
@@ -370,6 +477,74 @@ defmodule EXLA.Defn.Outfeed do
...
+                    end
+                end
+
+              {:error, _} = error ->
+                error
+            end
+
+          :error ->
+            Logger.error(
+              "EXLA.Outfeed received io_callback id #{inspect(callback_id)} that is not registered"
```
polvalente, line 515: Can we try to unnest this into a with?

exla/lib/exla/mlir/value.ex
```diff
@@ -823,6 +823,63 @@ defmodule EXLA.MLIR.Value do
...
+    leaf_count = length(typespecs)
+
+    {callback_id_words, callback_id_size} = term_to_int64_list(callback_id)
+
+    # Each leaf at operand[i+1] (skipping PID at 0) is aliased to result[i].
+    # The position in the list determines which result index is aliased.
+    aliases =
+      Enum.map(0..(leaf_count - 1)//1, fn i ->
+        attr_output_operand_alias(i + 1)
+      end)
```
polvalente:
```suggestion
    {callback_id_words, callback_id_size} = term_to_int64_list(callback_id)

    # Each leaf at operand[i+1] (skipping PID at 0) is aliased to result[i].
    # The position in the list determines which result index is aliased.
    aliases =
      Enum.with_index(typespecs, fn _typespec, index ->
        attr_output_operand_alias(i + 1)
      end)
```

exla/lib/exla/mlir/value.ex
```diff
@@ -823,6 +823,63 @@ defmodule EXLA.MLIR.Value do
...
+    ]
+
+    op(func, "stablehlo.custom_call", operands, result_types, attributes: attributes)
+  end
+
+  # Emits a single #stablehlo.output_operand_alias<...> attribute.
+  # `operand_index` is the 0-based index into the custom_call's operand list.
+  # Note: operand 0 is the PID, so the first leaf is at operand_index 1.
+  # The output index is implicit: aliases[i] refers to result i.
+  # Both tuple-indices lists are empty because we operate on flat tensors.
```
polvalente:
```suggestion
  # Both tuple-indices lists are empty because we operate on flat tensor lists.
```

exla/lib/exla/defn.ex
```diff
@@ -847,6 +848,51 @@ defmodule EXLA.Defn do
...
+           state,
+         cache
+       )
+       when platform in [:host, :cuda] do
+    tensor_exprs = Composite.flatten_list([tensor_expr])
+
+    {arg_values, cache} =
+      Enum.map_reduce(tensor_exprs, cache, fn arg, cache ->
+        recur_operator(arg, state, cache) |> unwrap_single_tensor!()
+      end)
```
polvalente: I think this can be done through a single Composite.reduce that accumulates both reverse_arg_values and cache

nx/lib/nx/defn/expr.ex
```diff
@@ -447,22 +451,31 @@ defmodule Nx.Defn.Expr do
...
+  defp io_callback_depend(acc, hooked) do
+    Composite.traverse(acc, fn leaf -> io_callback_depend_pair(leaf, hooked) end)
+  end
+
+  defp io_callback_depend_pair(%T{data: %Expr{id: id}}, %T{data: %Expr{id: id}} = hooked),
+    do: hooked
+
+  defp io_callback_depend_pair(acc, hooked) do
+    zero = Nx.tensor(0, type: acc.type)
+    Nx.add(acc, Nx.multiply(zero, Nx.sum(hooked)))
```
polvalente, line 478: I'm not sure if this is the best solution. Maybe we can add a metadata node or introduce a new expression node that can carry the nodes + attached callback?

@josevalim do you have suggestions?

nx/lib/nx.ex
```diff
@@ -2220,6 +2220,81 @@ defmodule Nx do
...
+  input is returned unchanged.
+  """
+  @doc type: :backend
+  def io_callback(tensor_or_container, callback)
+      when is_function(callback, 1) or is_atom(callback) do
+    tensors = Nx.Defn.Composite.flatten_list([tensor_or_container])
+    backend = Nx.Shared.list_impl!(tensors)
+
+    if backend == Nx.Defn.Expr do
+      backend.io_callback(tensor_or_container, callback)
```
polvalente, line 2288:
```suggestion
      Nx.Defn.Expr.io_callback(tensor_or_container, callback)
```

nx/test/nx/defn/expr_test.exs
```diff
@@ -425,24 +425,14 @@ defmodule Nx.Defn.ExprTest do
...
-               f = attach_token e, d       f32
-               g = attach_token e, c       f32
-               h = subtract f, g           f32
-             >\
-             """
+      assert rendered =~ "io_callback"
+      assert rendered =~ "subtract"
+      refute rendered =~ "token"
+      refute rendered =~ "attach_token"
     end
```
polvalente, line 457: Let's try to keep the same type of verbose assertions

polvalente (commented):
This generally looks good. Super exciting that we're progressing on this

nx/lib/nx.ex
```diff
@@ -2220,6 +2220,85 @@ defmodule Nx do
...
+      x = Nx.io_callback(x, &MyMod.log/1)   # correct — x is reassigned
+      Nx.io_callback(x, &MyMod.log/1)        # wrong — callback will be eliminated
+
+  Ordering between sequential `io_callback` calls is guaranteed through data
+  dependencies: since each call returns its input, chaining callbacks produces
+  an explicit dependency edge. On EXLA, independent callbacks are additionally
+  serialized by threading the callback-server PID through each `io_callback`
+  custom call (operand 0 → result 0), so XLA cannot reorder side effects even
+  when their tensor inputs are unrelated. No StableHLO token machinery is
+  required at the Nx expression level.
```
josevalim, line 2244: We should not discuss StableHLO on user facing APIs, as it is an implementation detail.

I am also not sure the EXLA note is relevant here. We say they are serialised but that means they do not run concurrently, it is not ultimately about order. I think we should rather say: "Ordering between IO callbacks are not guarantee, unless they are chained. For example, if you have this expression, `io_callback(a) + io_callback(b)`, there is no guarantee about order. But this one has: `io_callback(a + io_callback(b))`.

nx/lib/nx.ex
```diff
@@ -2220,6 +2220,85 @@ defmodule Nx do
...
+  >
+  > `io_callback/2` will generally operate on tensors allocated on a given
+  > physical device. When calling other Nx computations from within the callback,
+  > those computations cannot use the same device or a deadlock will result.
+
+  > #### Backend transfers {: .warning}
+  >
+  > When executing inside `Nx.Defn.Evaluator`, do not transfer tensors with
+  > `Nx.backend_transfer/2` inside the callback because the values may still be
+  > used in the rest of the computation. Use `Nx.backend_copy/2` instead.
```
josevalim, line 2256: So maybe we should emit it as a binary reference pointing to device? This way you are forced to copy it (and it solves both the lock and backend transfer issue above).

nx/lib/nx/defn/kernel.ex
```diff
@@ -230,9 +230,7 @@ defmodule Nx.Defn.Kernel do
 
   """
   def print_value(expr, fun, opts) when Kernel.and(is_function(fun, 1), is_list(opts)) do
-    token = create_token()
-    {token, _} = hook_token(token, fun.(expr), &IO.inspect(&1, opts))
-    attach_token(token, expr)
+    Nx.io_callback(expr, fn t -> IO.inspect(fun.(t), opts) end)
```
josevalim, line 233: I am wondering if we should call it `io_call` for consisatency with `runtime_call`?

nx/lib/nx/defn/expr.ex
```diff
@@ -1449,6 +1462,86 @@ defmodule Nx.Defn.Expr do
...
+    end
+  end
+
+  defp normalize_io_callback_spec(fun) when is_function(fun, 1), do: {:fn, fun}
+
+  defp normalize_io_callback_spec(name) when is_atom(name), do: {:hook, name, nil}
+
+  defp normalize_io_callback_spec({:hook, name, callback})
+       when is_atom(name) and (is_function(callback, 1) or is_nil(callback)),
+       do: {:hook, name, callback}
```
josevalim, line 1543: Why are we normalizing it here? Shouldn't we normalize in the callers of Expr.io_callback, since the callers already have the exact shape?

nx/lib/nx/defn/expr.ex
```diff
@@ -1449,6 +1462,86 @@ defmodule Nx.Defn.Expr do
...
+  Helper for defining an `:io_callback` expression node.
+
+  The callback is a side effect: the node's output is the input passed through
+  unchanged. A unique `ref` is embedded in the args to prevent CSE from
+  collapsing two distinct `io_callback` calls that share the same inputs and
+  callback function.
+
+  The user **must** reassign the result. Nx-level DCE prunes unreachable nodes
+  before any compiler sees `has_side_effect`, so discarding the return value
+  silently drops the callback from the graph.
```
josevalim, line 1475: This seems to be talking about implementation details of EXLA?

nx/lib/nx/defn/expr.ex
```diff
@@ -1788,6 +1886,9 @@ defmodule Nx.Defn.Expr do
     Enum.map_reduce(args, state, &recur_inspect/2)
   end
 
+  defp io_callback_spec_inspect({:hook, name, _}), do: inspect({:hook, name, nil})
+  defp io_callback_spec_inspect({:fn, _fun}), do: "&fn"
```
josevalim, line 1890: We can inspect the whole fun, no?

exla/lib/exla/defn/outfeed.ex
```diff
@@ -478,15 +655,39 @@ defmodule EXLA.Defn.Outfeed do
...
         %Nx.Tensor{} = template, [{bin, {type, shape_list}} | rest] ->
           decoded =
             bin
             |> Nx.from_binary(type)
             |> Nx.reshape(List.to_tuple(shape_list))
+            |> maybe_revectorize(template, mode)
+
+          compatible? =
+            case mode do
+              :io_callback -> true
```
josevalim, line 686: Why don't we need to check for compatibility here?

exla/lib/exla/defn/outfeed.ex
```diff
@@ -370,6 +482,71 @@ defmodule EXLA.Defn.Outfeed do
...
+          {:error, _} = error ->
+            error
+        end
+      rescue
+        exception ->
+          send(self(), :stop)
+          {:error, {:exception, Exception.message(exception)}}
+      catch
+        kind, reason ->
+          send(self(), :stop)
```
josevalim, line 529: We are sending stop, but will this effectively stop the execution on the callback too? Can we have tests for this scenario?

exla/lib/exla/defn/outfeed.ex
```diff
@@ -370,6 +482,71 @@ defmodule EXLA.Defn.Outfeed do
...
+          {:error, {kind, format_runtime_callback_reason(reason)}}
```
josevalim, line 530: This should likely be: `Exception.format(kind, reason, __STACKTRACE__)`.  We need to update runtime_call accordingly.

exla/lib/exla/defn.ex
```diff
@@ -1736,17 +1778,30 @@ defmodule EXLA.Defn do
...
+    # leak into sibling branches (SSA dominance). Linear chaining outside
+    # control flow still updates the cache directly via update_callback_pid/2.
+    merged = %{new_outfeed | token: outfeed.token, callback_pid: outfeed.callback_pid}
+    %{cache | __MODULE__ => merged}
+  end
 
   defp reset_token(%{__MODULE__ => outfeed}, token),
     do: %{__MODULE__ => Outfeed.with_token(outfeed, token)}
 
+  defp reset_callback_pid(%{__MODULE__ => outfeed}, callback_pid),
```
josevalim, line 1792: I don't think we need this functionality. This is necessary for tokens because we had to use different tokens on each branch of a conditional. But a callback_pid is a static value across all conditionals. Basically, we need a simple helper, which is "get_or_set_callback_pid" which returns the callback_pid (or creates one if it does not exist). We could even use the process dictionary if we want to, and forget this cache craziness.

PR #1755, opened 2026-05-25
Introduce buffer donation construct

seanmor5:
It's been awhile :)

I am working on some voice-to-voice models, and they're pretty tight on memory requirements. After some research, I realized we are actually being quite wasteful with a lot of our implementations because we don't support [Buffer Donation](https://docs.jax.dev/en/latest/buffer_donation.html). With a lot of algorithms, we have signatures that look a lot like reductions, e.g. an Axon training loop is:

```elixir
train(data :: tensor(), params :: map(binary(), tensor()) :: map(binary(), tensor())
```

Where the output is an updated parameters map. On every training iteration, we are actually duplicating the training state (e.g. we'll have copy of `Tn` and `T(n+1)` and as many other copies while we wait for garbage collection). To avoid OOMs we explicitly garbage collect in the Axon training loop, and while this works for models where 2 copies of the state fit on device, it does not work as model size grows.

The solution is to introduce a buffer donation concept. The idea is that XLA/PjRt will be smart and avoid allocating a new buffer if we're explicit and say "this one is okay to overwrite." This also significantly reduces memory pressure for text generation algorithms that rely on KV-cache!

This is a draft for now, because this is a vibed implementation and I just want to propose the change and settle on an API. This copies the JAX API, but I actually prefer something I think that's more explicit, e.g. maybe annotating inputs before passing them to JIT functions:

```elixir
# mark a buffer as okay to donate
EXLA.donate(buffer)
# call the function
fun(buffer)
```

The API/relationship here is a bit harder to workout because this is not really a natural concept in Elixir.

I can confirm that the changes work for my use case at least, and I am able to load and run models that had previously OOMed without buffer donation.

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -322,6 +322,14 @@ defmodule EXLA.Defn do
     {hooks, options} = Keyword.pop(options, :hooks, %{})
     {debug?, options} = Keyword.pop(options, :debug, false)
     {lazy_transfers, options} = Keyword.pop(options, :lazy_transfers, :opt_in)
+    {donate_argnums, options} = Keyword.pop(options, :donate_argnums, [])
+    donate_argnums = validate_donate_argnums!(donate_argnums, length(vars))
+
+    # Put the normalized list back so it lands in disk_key and comp_key.
+    options =
+      if donate_argnums == [],
+        do: options,
+        else: Keyword.put(options, :donate_argnums, donate_argnums)
 
     {client_name, options} = Keyword.pop_lazy(options, :client, &EXLA.Client.default_name/0)
     client = EXLA.Client.fetch!(client_name)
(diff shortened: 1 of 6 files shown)


exla/lib/exla/defn.ex
```diff
@@ -322,6 +322,14 @@ defmodule EXLA.Defn do
     {hooks, options} = Keyword.pop(options, :hooks, %{})
     {debug?, options} = Keyword.pop(options, :debug, false)
     {lazy_transfers, options} = Keyword.pop(options, :lazy_transfers, :opt_in)
+    {donate_argnums, options} = Keyword.pop(options, :donate_argnums, [])
```
polvalente, line 325: I think the only think that is a bit confusing for users here is that we're exposing the flat arg structure as the argument, while users operate on containers.

Maybe we should push this to Nx for 2 reasons:

1. Other compilers can agree on the same interface (very useful for Torchx deciding whether to clone a tensor, same with EMLX)
2. Nx exposes the non-flat container inputs, so the interface would be closer to what the user passes. For instance, a user could alias just part of a container. They can't do that with map containers because we can't rely on ordering.

polvalente: Other than this, the PR looks great

PR #1761, opened 2026-06-10
Basic io_call implementation

### LINKED ISSUE #1672: Introduce Nx.io_callback as a replacement for token-based hooks and EXLA's outfeed

Chapaman:
this PR has been split into #1767 #1766

Closes #1672.

Clean restart from `main` (not a continuation of the now defunct #1752). Side-effect callbacks in `defn` via **`Nx.io_call/3–4`**, which returns **`{token, result}`** and threads explicit ordering tokens. **`hook/3`**, **`hook_token/4`**, and **`print_value/2`** are thin proxies over `io_call`.

## Nx

- **`:io_call`** expr op — args `[token, data, callback_spec, template, ref]`; returns `{token, data}` via tuple root + `:elem`
- **`:create_token`** expr op — `create_token/0` builds a token tensor (`type: :token`), not `%Nx.Defn.Token{}`
- **`Nx.io_call/3–4`** — public API; callback spec normalized to `{:fn, fun}` / `{:hook, name, cb}` internally at the call site (not exposed as public clauses)
- **`Nx.create_token/0`** — public entry point for token creation (replaces `Nx.Defn.Kernel.create_token/0` in user docs)
- **`hook/3`, `hook_token/4`, `print_value/2`** — route through `io_call` + `attach_token/2`
- **Evaluator, tree, grad, inspect** — support for `:io_call` / `:create_token`; named `io_call` inspect renders as `io_call add: token, data`
- **Undefined named hooks** — raise by default; opt in to skip with `ignore_undefined_io_calls: true` (Evaluator + EXLA)
- **`:hooks` option kept** — JIT mechanism to override named `io_call` callbacks (and potentially runtime calls later)
- **`Tree.has_hooks?/2`** — checks each `:io_call` node independently against the hooks map
- **Removed** — `%Nx.Defn.Token{}` struct, `:token` batch outfeed expr paths, `add_hook/4`

## EXLA

- **`exla_io_call`** custom call (Host + CUDA) via outfeed bridge
- **MLIR** — `create_token`, `io_call` with token in/out + `after_all` for ordering
- **Static `callback_pid`** per computation (no branch-reset / PID chaining between calls)
- **Zero-copy** — `output_operand_aliases` on leaf tensors; token output is not aliased
- **`attach_token` lowering kept** — token edge stays in the graph
- **Outfeed** — `Exception.format/3` on errors, compatibility checks on materialization, halt on any `io_call` error

- **Partial hook streaming** — automatically sets `ignore_undefined_io_calls: true` so unregistered hooks (e.g. `:to_be_ignored`) are skipped when only a subset of hooks is streamed


nx/lib/nx.ex
```diff
@@ -2220,6 +2220,83 @@ defmodule Nx do
...
+  def io_call(%Nx.Tensor{} = token, tensor_or_container, callback)
+      when is_function(callback, 1) do
+    io_call(token, tensor_or_container, {:fn, callback})
+  end
+
+  def io_call(%Nx.Tensor{} = token, tensor_or_container, {:hook, _, _} = callback_spec),
+    do: io_call_impl(token, tensor_or_container, callback_spec)
+
+  def io_call(%Nx.Tensor{} = token, tensor_or_container, {:fn, _} = callback_spec),
+    do: io_call_impl(token, tensor_or_container, callback_spec)
```
polvalente: Move these 2 clauses inside io_call_impl, as they are leaking an internal representation to the API

nx/lib/nx.ex
```diff
@@ -2220,6 +2220,83 @@ defmodule Nx do
...
+    end
+  end
+
+  defp run_io_call_eager!({:fn, fun}, container) when is_function(fun, 1), do: fun.(container)
+
+  defp run_io_call_eager!({:hook, _name, callback}, container)
+       when is_function(callback, 1),
+       do: callback.(container)
+
+  defp run_io_call_eager!({:hook, _name, nil}, _container), do: :ok
```
```suggestion
```

And maybe add the hook name to the exception below

nx/test/nx/defn/expr_test.exs
```diff
@@ -433,14 +433,20 @@ defmodule Nx.Defn.ExprTest do
...
-               g = attach_token e, c       f32
-               h = subtract f, g           f32
+               parameter b:0                           f32
+               parameter c:1                           f32
+               a = create_token                        token
+               d = add b, c                            f32
+               e = io_call a, d, {:hook, :add, nil}    tuple2
+               f = elem e, 0                           token
+               g = multiply b, c                       f32
+               h = io_call f, g, {:hook, :mult, nil}   tuple2
```
polvalente:
```suggestion
               e = io_call add: a, d    tuple2
               f = elem e, 0                           token
               g = multiply b, c                       f32
               h = io_call mult: f, g   tuple2
```

I think this style of inspect would be easier to read. And for non-named calls we just use `io_call arg1, arg2, ...` without the leading name

nx/lib/nx/defn/tree.ex
```diff
@@ -17,8 +17,14 @@ defmodule Nx.Defn.Tree do
...
   end
 
-  defp detect_hook(%T{data: %Expr{op: :token, args: [token]}} = t, cache, hooks) do
-    if Enum.any?(token.hooks, &(hooks[&1.name] || &1.callback)) do
+  defp detect_hook(%T{data: %Expr{op: :io_call, args: [_, _, spec, _, _]}} = t, cache, hooks) do
+    side_effect? =
+      case spec do
+        {:hook, name, callback} -> hooks[name] || callback
+        {:fn, _} -> true
+      end
```
polvalente: We need to look deeper into whether this is breaking named hooks, given that each io_call effectively only knows about the hook referenced in itself

nx/lib/nx/defn/evaluator.ex
```diff
@@ -357,6 +330,42 @@ defmodule Nx.Defn.Evaluator do
...
+         [token, tensor_expr, callback_spec, out_template, _ref],
+         _ans,
+         state,
+         caches
+       ) do
+    {_token, caches} = eval(token, state, caches)
+    {tensor_value, caches} = composite_eval(tensor_expr, state, caches)
+
+    case resolve_io_call(callback_spec, state.hooks) do
+      nil -> :ok
```
polvalente: I think :ok here is not the correct handling.
We should have an atom signifying "this hook has already ran" and `nil` should mean an invalid graph, where a named hook was referenced before we collected it somehow.

polvalente: This is actually keeping the current behavior. Maybe change it and make an opt-in "ignore undefined io_call" thing?

We also need to deprecate the :hooks option and make it :io_calls in Nx.Defn.Evaluator and likely EXLA as well

exla/lib/exla/defn.ex
```diff
@@ -234,7 +234,8 @@ defmodule EXLA.Defn do
          run_options,
          is_sharded?
        )
-       when Outfeed.will_outfeed(outfeed) or Outfeed.has_runtime_calls(outfeed) do
+       when Outfeed.will_outfeed(outfeed) or Outfeed.has_runtime_calls(outfeed) or
```
polvalente: I think we don't need will_outfeed anymore, we're replacing it with io_calls

exla/lib/exla/defn.ex
```diff
@@ -769,6 +770,60 @@ defmodule EXLA.Defn do
...
+       ) do
+    raise """
+    Nx.io_call/3 is currently only supported for EXLA CPU (platform: :host) and CUDA (platform: :cuda),
+    but the active EXLA client is configured for platform #{inspect(platform)}.
+    Please run on the :host or :cuda client or wait for future segmentation-based support.
+    """
+  end
+
+  defp cached_recur_operator(:create_token, _expr, %{builder: builder}, cache) do
+    {Value.create_token(builder), cache}
```
polvalente: I guess we could keep create_token like this, but effectively even an s1 tensor would work as we just need _any_ value, not a token per se

exla/test/exla/defn/api_test.exs
```diff
@@ -289,6 +289,37 @@ defmodule EXLA.Defn.APITest do
...
+      send_value = fn value -> send(parent, Nx.to_number(value)) end
+
+      assert_equal(
+        EXLA.jit(&side_effect_hooks/2, hooks: %{a: send_value, b: send_value}).(1, 2),
+        Nx.tensor(3)
+      )
+
+      assert_received 2
+      assert_received 1
+    end
```
polvalente, line 337: We should have io_call tests as well, not just hook tests (even though hooks use io_call now)

nx/lib/nx.ex
```diff
@@ -2220,6 +2220,83 @@ defmodule Nx do
...
+  ## Examples
+
+      iex> defmodule IoCallExample do
+      ...>   def log(t), do: t
+      ...> end
+      iex> token = Nx.Defn.Expr.create_token()
+      iex> x = Nx.tensor([1, 2, 3])
+      iex> {updated_token, x} = Nx.io_call(token, x, &IoCallExample.log/1)
+      iex> match?(%Nx.Tensor{type: :token}, updated_token)
+      true
```
```suggestion
      iex> {%Nx.Tensor{type: :token}, x} = Nx.io_call(token, x, &IoCallExample.log/1)
```

nx/lib/nx.ex
```diff
@@ -2220,6 +2220,83 @@ defmodule Nx do
...
+      iex> Nx.to_flat_list(x)
+      [1, 2, 3]
```
polvalente: maybe assert on the inspected tensor directly, not a flat list of it

nx/lib/nx.ex
```diff
@@ -2220,6 +2220,83 @@ defmodule Nx do
...
+
+  Each call takes the current token and returns `{token, data}` where `data` is the
+  input passed through unchanged:
+
+      token = Nx.Defn.Kernel.create_token()
+      {token, x} = Nx.io_call(token, x, &MyMod.log/1)
+
+  Named hooks can be overridden via `Nx.Defn.jit/2`:
+
+      Nx.io_call(token, x, :my_hook)
```
polvalente: We wanna remove hook naming in favor of io_call (i.e. hooks -> io_calls)

nx/lib/nx.ex
```diff
@@ -2220,6 +2220,83 @@ defmodule Nx do
...
+
+  Unlike `runtime_call/4`, `io_call/3` does not compute a new value in the callback.
+  Its purpose is to execute side effects (logging, sending messages, etc.) while
+  preserving explicit ordering through tokens.
+
+  Each call takes the current token and returns `{token, data}` where `data` is the
+  input passed through unchanged:
+
+      token = Nx.Defn.Kernel.create_token()
+      {token, x} = Nx.io_call(token, x, &MyMod.log/1)
```
polvalente: Maybe we wanna bring Nx.create_token out of Nx.Defn.Kernel.create_token since we're exposing io_call in Nx instead of Nx.Defn.Kernel

nx/lib/nx/defn/evaluator.ex
josevalim: Maybe we keep this one as hooks and we explain it as a mechanism to hook into IO calls (and potentially runtime ones in the future)?

exla/lib/exla/defn.ex
Chapaman: > I think we don't need will_outfeed anymore, we're replacing it with io_calls

`will_outfeed` is used on the infeed path, should we change it as well?

exla/lib/exla/mlir/value.ex
```diff
@@ -696,6 +696,50 @@ defmodule EXLA.MLIR.Value do
     op(func, "stablehlo.create_token", [], result_types) |> one!()
   end
 
+  def after_all([%Value{function: func} | _] = tokens) do
+    result_types = [type_token()]
+    op(func, "stablehlo.after_all", tokens, result_types) |> one!()
+  end
```
polvalente, line 702: So this is supposed to ensure all tokens execute by the end of the computation?

polvalente (approved):
Looks great! Let's get @josevalim to take a look as well, but this is well on track for us to remove infeed and outfeed altogether!

exla/lib/exla/mlir/value.ex
polvalente: I think we might be able to remove this for now. We're threading tokens through host_callback directly

nx/lib/nx/defn/evaluator.ex
```diff
@@ -8,6 +8,15 @@ defmodule Nx.Defn.Evaluator do
...
   The following options are specific to this compiler:
 
+    * `:io_calls` - a map of callbacks to override named `io_call` side effects.
+      This allows overriding named `io_call`s at JIT time without recompiling the
+      graph.
+
+    * `:ignore_undefined_io_calls` - when `true`, named `io_call`s without a
+      default callback or a matching entry in `:io_calls` are skipped during
+      evaluation. Defaults to `false`, in which case an undefined named `io_call`
+      raises.
```
josevalim, line 18: Why do we have this feature? What is the use case? :)

nx/lib/nx/defn/expr.ex
```diff
@@ -441,30 +445,68 @@ defmodule Nx.Defn.Expr do
   def id(), do: make_ref()
 
   @doc false
-  def add_hook(token, expr, name, function) do
-    expr = to_container_expr(expr)
-    token = Nx.Defn.Token.add_hook(token, expr, name, function)
-    {token, expr}
+  def create_token do
+    expr(%T{shape: {}, type: :token, names: []}, nil, :create_token, [])
```
josevalim: If we still need a token, I would prefer to keep reusing Nx.Defn.Token as a separate structure rather than mess-up tensors. This means that now we need to handle a possible type: :token in a bunch of different places.

nx/lib/nx/defn/grad.ex
```diff
@@ -1051,6 +1069,30 @@ defmodule Nx.Defn.Grad do
...
+        is_tuple(g) and tuple_size(g) == leaf_count + 1 -> g |> Tuple.to_list() |> tl()
+        true -> List.wrap(g)
+      end
+
+    {pairs, []} =
+      Composite.reduce(tensor_expr, {[], gs}, fn child, {pairs, [grad | gs]} ->
+        {[{child, grad} | pairs], gs}
+      end)
+
+    Enum.reverse(pairs)
```
josevalim, line 1082: This looks a bit complicated. We did not need to do this dance with `add_hook`, and I assume it is because we used `add_hook` used `to_container_expr` to normalize the input expression.

nx/lib/nx/defn/evaluator.ex
polvalente: Retrocompatibility with the existing hook behavior

nx/lib/nx/defn/expr.ex
polvalente: Let's try to eliminate it in favor of using a pseudo token s1 tensor we keep incrementing

nx/lib/nx/defn/evaluator.ex
josevalim: I don't think we need this then. If you do `io_call(:foo)` without specifying a callback, then it means you want to make it pluggable and be ignored. If you require it to be given, then you can validate it upfront. So I think we should ignore by default and remove this. The previous one was designed like this on purpose. :)

nx/lib/nx/defn/kernel.ex
```diff
@@ -1361,176 +1360,70 @@ defmodule Nx.Defn.Kernel do
...
-  end
-
-  @doc """
-  Attaches a token to an expression. See `hook/3`.
-  """
-  def attach_token(%Nx.Defn.Token{} = token, expr) do
-    Nx.Defn.Expr.attach_token(token, expr)
-  end
+  def io_call(expr, name, function) when Kernel.and(is_atom(name), is_function(function, 1)),
+    do: Nx.io_call(expr, name, function)
```
josevalim, line 1426: We need to keep create_token, attach_token, hook_token e hook as deprecated functions. They are either no-op or simply delegate to io_call.

nx/lib/nx/defn/kernel.ex
```diff
@@ -1425,6 +1425,53 @@ defmodule Nx.Defn.Kernel do
...
+    do: {create_token(), io_call(expr, name)}
+
+  @deprecated "Tokens are no longer needed. Use io_call/2 instead."
+  def io_call_token(_token, expr, function) when is_function(function, 1),
+    do: {create_token(), io_call(expr, function)}
+
+  @deprecated "Tokens are no longer needed. Use io_call/3 instead."
+  def io_call_token(_token, expr, name, function)
+      when Kernel.and(is_atom(name), is_function(function, 1)),
+      do: {create_token(), io_call(expr, name, function)}
```
josevalim, line 1465: We don't need these.

exla/lib/exla/defn/outfeed.ex
```diff
@@ -500,6 +609,8 @@ defmodule EXLA.Defn.Outfeed do
       [] -> {:ok, container}
       _ -> {:error, {:invalid_args_spec, :extra_values}}
     end
+  catch
+    :throw, {:error, _} = error -> error
```
josevalim, line 613: Do we need this one?

nx/lib/nx/defn/evaluator.ex
```diff
@@ -40,15 +49,17 @@ defmodule Nx.Defn.Evaluator do
 
   @impl true
   def __compile__(_key, vars, fun, opts) do
-    hooks = Keyword.get(opts, :hooks, %{})
+    io_calls = Keyword.get(opts, :io_calls, %{})
```
josevalim, line 52: We have to keep reading the hooks key for backwards compatibility for now.

exla/lib/exla/defn/outfeed.ex
```diff
@@ -227,45 +225,88 @@ defmodule EXLA.Defn.Outfeed do
...
   end
 
-  defp init(client, device_id, hooks, compiled_hooks, infeeds, rt_callbacks, group_leader) do
+  defp init(
+         client,
+         device_id,
+         infeed_flags,
+         infeeds,
+         callbacks,
+         ignore_undefined_io_calls,
```
josevalim, line 258: Let's remove `ignore_undefined_io_calls` and then revert this back to a single line. Also some of the multi-line calls below.

nx/lib/nx/serving.ex
```diff
@@ -636,13 +636,13 @@ defmodule Nx.Serving do
   most common batches.
   """
   def streaming(%Nx.Serving{} = serving, opts \\ []) do
-    hooks = Keyword.get(opts, :hooks, [])
```
josevalim, line 639: We need to keep reading :hooks for backwards compatibility.

nx/lib/nx.ex
```diff
@@ -2220,6 +2220,71 @@ defmodule Nx do
     list
   end
 
+  @doc """
```
josevalim, line 2223: Why do we need to define this function here? Having it inside `defn` is not enough?

polvalente: I think the flexibility of being able to define these inside a `fn` or inside `deftransform` is preferrable. I'd rather have Nx.Defn.Kernel have as few exclusive features as possible, if any at all.

Chapaman:
Split into two PRs per maintainer feedback:

- #1766 — EXLA hooks via host callback (no Nx API changes)
- #1767 — rename hooks → `io_call`, remove tokens (merge after #1766)

This PR can be closed in favor of those two.

PR #1764, opened 2026-06-18
fix: mitigate false positives on recompilation check

### LINKED ISSUE #1758: Warning: EXLA has compiled Function 10 times with the same input shapes.
I am using Nx inside Livebooks for educational purposes (rather than running production code). Recently I observe many warnings of this kind:

> 13:21:31.619 [warning] EXLA has compiled #Function<93.124553489/1 in EXLA.Backend.slice/5> 10 times with the same input shapes. This typically means tensor values are being captured inside a closure passed to defn, jit, or value_and_grad. Each distinct captured value forces a full recompilation. Pass changing tensors as explicit function arguments instead.

As far as I understand the warning was introduced by #1715 and is meant to be helpful but in my case I am not sure what exactly causes it. It is especially difficult to debug since the warning is triggered only sometimes and even disappears after running the same Livebook cell again. This makes it hard to understand if any of my code changes caused or resolved the warning.

Here is an example of a Livebook in which almost all cells trigger the warning sometimes:

https://github.com/laszlokorte/elixir-nx-mnist-livebook/blob/main/handwriting.livemd

My guess is that its somehow triggered by my plotting library:
https://github.com/laszlokorte/kino_zoetrope/blob/main/lib/kino_zoetrope/tensor_stack.ex#L70-L82

But I do not feel the warning message pointing me into the correct direction to fix it, especially since I am not calling `defn`, `jit`, or `value_and_grad` on any of my code paths directly (not knowing about Nx internals).

polvalente:
closes #1758

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -494,33 +494,35 @@ defmodule EXLA.Defn do
     end)
   end
 
-  @recompilation_threshold 10
-
   defp check_recompilation(key, args_key, outputs) do
-    {:module, mod} = :erlang.fun_info(key, :module)
-    {:new_index, idx} = :erlang.fun_info(key, :new_index)
-    {:env, env} = :erlang.fun_info(key, :env)
+    threshold = Application.get_env(:exla, :check_recompilation, :infinity)
 
-    out_key =
-      [outputs]
-      |> Nx.Defn.Composite.flatten_list()
-      |> Enum.map(&{&1.type, &1.shape})
+    if is_integer(threshold) do
+      {:module, mod} = :erlang.fun_info(key, :module)
+      {:new_index, idx} = :erlang.fun_info(key, :new_index)
+      {:env, env} = :erlang.fun_info(key, :env)
 
-    # Normalize the env by replacing %Nx.Tensor{} values with {type, shape}
-    # so the counter key is stable for same-shaped tensors but distinguishes
-    # closures that capture different non-tensor metadata (e.g. different block ops).
-    normalized_env = normalize_env(env)
+      out_key =
+        [outputs]
+        |> Nx.Defn.Composite.flatten_list()
+        |> Enum.map(&{&1.type, &1.shape})
 
-    count = EXLA.Defn.LockedCache.count({mod, idx, args_key, out_key, normalized_env})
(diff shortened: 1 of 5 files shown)


exla/lib/exla/defn.ex
```diff
@@ -496,10 +496,22 @@ defmodule EXLA.Defn do
 
   @recompilation_threshold 10
 
-  defp check_recompilation(key, args_key) do
+  defp check_recompilation(key, args_key, outputs) do
```
josevalim, line 497: I think we should revert the feature because traversing the environment makes it much more expensive. Or  alternatively we don't do it by default and there is a flag we can use in dev.

josevalim: To give more context, we now traverse the whole environment, copy it to the ETS table (which can balloon), and then compare with any existing entries there.

polvalente: How about we hide it under a runtime config?

josevalim: What do you mean? Anyway, what I proposed was `config :exla, check_recompilation: threshold` and the default is  infinity (which means no checking).

polvalente: Exactly that!

exla/lib/exla/defn.ex
```diff
@@ -487,31 +487,58 @@ defmodule EXLA.Defn do
...
+  defp normalize_env_value(%Nx.Tensor{type: type, shape: shape}), do: {type, shape}
+  defp normalize_env_value(v) when is_list(v), do: Enum.map(v, &normalize_env_value/1)
+
+  defp normalize_env_value(v) when is_tuple(v),
+    do: v |> Tuple.to_list() |> Enum.map(&normalize_env_value/1) |> List.to_tuple()
+
+  defp normalize_env_value(v) when is_map(v) and not is_struct(v),
+    do: Map.new(v, fn {k, val} -> {normalize_env_value(k), normalize_env_value(val)} end)
+
+  defp normalize_env_value(v), do: v
```
josevalim, line 543: Be careful because you can have improper lists  in the environment and the code above will fail to traverse it on `Enum.map`.

josevalim (approved):
One comment and ship it!

PR #1766, opened 2026-06-23
Implement EXLA hooks via host callback without outfeed

Chapaman:
## Summary

Splits the EXLA side of #1761 into a standalone change (per maintainer feedback from @josevalim and @polvalente).

- Replaces the outfeed u16-flag hook protocol with XLA `host_callback` / `exla_runtime_callback` for hook side effects and runtime calls
- Keeps the existing Nx API unchanged (`hook/3`, `:hooks`, `%Nx.Defn.Token{}`) so this can land independently
- Unifies host and CUDA runtime callback paths; outfeed is retained only for lazy-transfer infeed
- Adds ordering and side-effect regression tests in `exla/test/exla/defn/api_test.exs`

Part 1 of 2. Follow-up: rename hooks → `io_call` and remove tokens from Nx (stacked PR).

Supersedes the EXLA portions of #1761.

## Test plan

- [x] `cd exla && mix test test/exla/defn/api_test.exs` (26 tests, 0 failures)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/exla/c_src/exla/custom_calls/runtime_callback.cc
+++ b/exla/c_src/exla/custom_calls/runtime_callback.cc
@@ -1,80 +1,86 @@
 #include "runtime_callback_bridge.h"
 
 #include <cstring>
-#include <vector>
 #include <string>
+#include <vector>
 
 #include "xla/ffi/api/ffi.h"
 #include "xla/ffi/ffi_api.h"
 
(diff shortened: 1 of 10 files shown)


josevalim (approved):
This is fantastic, awesome job!!!!

polvalente (approved):

PR #1767, opened 2026-06-23
Remove token plumbing from Nx hooks

Chapaman:
## Summary
Split internal token plumbing removal from the `io_call` rename so we can verify functionality is preserved before deciding on naming.
- Removes internal `:token` / `:attach_token` graph nodes and token plumbing from evaluator, compiler, tree, and EXLA lowering
- Introduces `:hook` expr nodes for side effects (data passes through unchanged, ordered by program order)
- **Keeps** `hook/2-3`, `:hooks`, and the EXLA host-callback mechanism from #1766
- **Keeps** `%Nx.Defn.Token{}`, `create_token/0`, `hook_token/3-4`, and `attach_token/2` as deprecated backwards-compat shims (they lower to `{:token_hook, ...}` internally, not graph token nodes)
- Fixes EXLA hook lowering for container and `while` inputs via `recur_hook_pass/4`
Rename `hook` → `io_call` deferred to a follow-up PR.


josevalim:
@Chapaman let's break this one in two as well, one that removes the tokens, so we can test it preserves the functionality, and then we decide on the rename!!! :heart:

exla/lib/exla/defn.ex
```diff
@@ -817,21 +817,60 @@ defmodule EXLA.Defn do
...
+    arg_template = Nx.to_template(tensor_expr)
+
+    cache = add_callback(cache, {id, callback_spec, nil, arg_template})
+
+    unless callback_pid_value do
+      raise "internal bug: hook callback pid operand is missing"
+    end
+
+    leaf_typespecs = Enum.map(arg_values, &Value.get_typespec/1)
+    num_aliased = length(leaf_typespecs)
```
polvalente: Should we optimize and calculate this in the Composite.reduce call above?

nx/lib/nx/defn/expr.ex
```diff
@@ -441,28 +441,50 @@ defmodule Nx.Defn.Expr do
...
-    Composite.traverse(expr, fn tensor ->
-      expr = to_expr(tensor)
-      expr(expr, expr.data.context, :attach_token, [token, expr])
-    end)
+    [%T{data: %Expr{context: context}} | _] =
+      Composite.flatten_list([tensor_expr])
+
+    ref = make_ref()
+    user_template = Nx.to_template(tensor_or_container)
+    leaf_count = user_template |> List.wrap() |> Composite.flatten_list() |> length()
```
polvalente:
```suggestion
    {_, leaf_count} = user_template |> List.wrap() |> Composite.reduce(0, fn _, acc -> acc + 1 end)
```

nx/lib/nx/defn/expr.ex
```diff
@@ -1745,20 +1767,30 @@ defmodule Nx.Defn.Expr do
...
+    {data, state} = recur_inspect(data, state)
+    var_name = var_name(state)
+
+    expr =
+      case callback_spec do
+        {:named, name, _} ->
+          IO.iodata_to_binary([var_name, " = hook ", Atom.to_string(name), ": ", data])
+
+        {:fn, fun} ->
+          IO.iodata_to_binary([var_name, " = hook ", data, ", ", inspect(fun)])
```
polvalente: do we want to render as io_call instead of hook in the PR?

nx/lib/nx/defn/evaluator.ex
```diff
@@ -40,7 +44,11 @@ defmodule Nx.Defn.Evaluator do
 
   @impl true
   def __compile__(_key, vars, fun, opts) do
-    hooks = Keyword.get(opts, :hooks, %{})
+    hooks =
+      opts
+      |> Keyword.get(:hooks, %{})
+      |> then(&Map.merge(Keyword.get(opts, :hooks, %{}), &1))
```
josevalim: This is probably a leftover.

nx/lib/nx/defn/expr.ex
josevalim: no, let's keep it as a hook. no renames at all during this pr.

nx/lib/nx/defn/kernel.ex
```diff
@@ -1417,120 +1413,17 @@ defmodule Nx.Defn.Kernel do
...
-    Nx.Defn.Token.new()
-  end
-
-  @doc """
-  Attaches a token to an expression. See `hook/3`.
-  """
-  def attach_token(%Nx.Defn.Token{} = token, expr) do
-    Nx.Defn.Expr.attach_token(token, expr)
-  end
+    do: Nx.hook(expr, name, function)
```
josevalim: Let's keep this only on kernel for now, let's not add Nx.hook if we plan to rename it in the future.

nx/lib/nx/defn/token.ex
josevalim, line 1: We should probably keep this file and the `defstruct`, for backwards compatibility, and remove it in a future release.

josevalim (commented):
This is looking great. I dropped some comments but we are almost there.

1. Ideally we would not change tests in this PR!
2. Let's not add Nx.hook, since we plan to introduce `io_call`. Let's keep the main API in `Nx.Defn.Kernel` still
3. I am thinking we should even launch a new Nx version with this change, and then introduce `io_call` in a future. That should help us migrate with ease

nx/lib/nx/defn/grad.ex
```diff
@@ -1047,8 +1058,28 @@ defmodule Nx.Defn.Grad do
...
+        is_tuple(g) and tuple_size(g) == leaf_count -> Tuple.to_list(g)
+        true -> List.wrap(g)
+      end
+
+    {pairs, []} =
+      Composite.reduce(tensor_expr, {[], gs}, fn child, {pairs, [grad | gs]} ->
+        {[{child, grad} | pairs], gs}
+      end)
+
+    Enum.reverse(pairs)
```
josevalim: This one still feels like a regression to me. The reason we didn't have to do this with tokens is because before we would emit a tuple as the result of the hook and use element access to fetch parts of the result. It would be nice if we could keep the same semantics as before. It seems we already fixed this in `expr.ex`, maybe we don't need this anymore!

josevalim:
It seems the rename was accidentally pushed to this branch!

exla/lib/exla/defn.ex
```diff
@@ -817,21 +817,95 @@ defmodule EXLA.Defn do
...
+         %{client: %EXLA.Client{platform: platform}, callback_pid_value: callback_pid_value} =
+           state,
+         cache
+       )
+       when platform in [:host, :cuda] do
+    {reverse_arg_values, reverse_typespecs, cache} =
+      Composite.reduce(tensor_expr, {[], [], cache}, fn %T{} = expr, {acc, typespecs, cache} ->
+        {value, cache} = recur_operator(expr, state, cache) |> unwrap_single_tensor!()
+        {[value | acc], [Value.get_typespec(value) | typespecs], cache}
       end)
```
polvalente:
```suggestion
    {reverse_arg_values, reverse_typespecs, cache, num_aliased} =
      Composite.reduce(tensor_expr, {[], [], cache, 0}, fn %T{} = expr, {acc, typespecs, cache, num_aliased} ->
        {value, cache} = recur_operator(expr, state, cache) |> unwrap_single_tensor!()
        {[value | acc], [Value.get_typespec(value) | typespecs], cache, num_aliased + 1}
end)
```

exla/lib/exla/defn.ex
polvalente: I left a suggestion above

nx/lib/nx/defn/kernel.ex
```diff
@@ -1343,7 +1342,7 @@ defmodule Nx.Defn.Kernel do
         end
       end
 
-  Note a hook can only access the variables passed as arguments
+  Note an hook can only access the variables passed as arguments
```
polvalente:
```suggestion
  Note a hook can only access the variables passed as arguments
```

nx/lib/nx/defn/kernel.ex
```diff
@@ -1448,59 +1433,27 @@ defmodule Nx.Defn.Kernel do
...
     do: unguarded_hook(expr, name, function)
 
   defp unguarded_hook(expr, name, function) do
-    {token, result} = Nx.Defn.Expr.add_hook(create_token(), expr, name, function)
-    attach_token(token, result)
+    spec =
+      case function do
+        nil -> {:named, name, nil}
+        fun -> {:named, name, fun}
+      end
```
polvalente: Isn't this just 
```suggestion
    spec = {:named, name, nil}
```

polvalente (approved):

polvalente:
> This is looking great. I dropped some comments but we are almost there.
> 
> 1. Ideally we would not change tests in this PR!

Only tests that really changed were inspect tests, which I think we're fine to do.
EXLA didn't change tests at all, and there is one extra assertion that makes sense to add even in the previous implementation. ✅ for me
> 2. Let's not add Nx.hook, since we plan to introduce `io_call`. Let's keep the main API in `Nx.Defn.Kernel` still
✅ 
> 3. I am thinking we should even launch a new Nx version with this change, and then introduce `io_call` in a future. That should help us migrate with ease
I think we can hold off the io_call and put in a few more bugfixes, to launch 0.13 by EOW!

josevalim (commented):
Beautiful!!!!!!!

PR #1770, opened 2026-06-29
Standardize docs across nx, exla, torchx

### LINKED ISSUE #1759: Introduce documentation standard for backends
Issue #1754 along with some recent code review highlighted the need for an easy way for users to find, in each backend, documentation on implementation details for each Nx function.

A markdown documentation structure that mirrors the Nx modules and function structure is the desired approach. As a fallback alternative, we can introduce "ghost" modules for documenting in hexdocs that don't get compiled with the code.

Chapaman:
Closes #1759 

- Add Nx.Docs to define the convention for backend-specific documentation modules (Kernel.SpecialForms-style ghost modules that mirror the Nx API).
- Add EXLA.Docs.Nx, EXLA.Docs.Nx.LinAlg, Torchx.Docs.Nx, and Torchx.Docs.Nx.LinAlg documenting how each backend lowers or executes block-backed and backend-specific operations.
- Document all current Nx.block/4 entry points: top-level Nx (take, FFT, cumulative ops, transfers, runtime_call, etc.) and Nx.LinAlg (cholesky, QR, SVD, …).
- Add ## Backend notes sections on 29 Nx / Nx.LinAlg functions linking to the EXLA and Torchx docs modules.
- Wire discoverability through Nx, Nx.LinAlg, Nx.Backend, Nx.block/4, EXLA, and Torchx moduledocs, plus Hex “Backend documentation” groups in each package.

### DIFF
--- a/exla/guides/backend_documentation/index.md
+++ b/exla/guides/backend_documentation/index.md
@@ -13,12 +13,3 @@ the [backend documentation convention](https://hexdocs.pm/nx/backend_documentati
 
 Implementation code lives in `EXLA.Defn`, `EXLA.CustomCall`, and related modules.
 
-## Example
-
-```elixir
-iex> Nx.take(Nx.tensor([10, 20, 30]), Nx.tensor([0, 2]))
-#Nx.Tensor<
-  s32[2]
-  [10, 30]
->
-```
--- a/exla/guides/backend_documentation/index.md
+++ b/exla/guides/backend_documentation/index.md
@@ -1,15 +1,12 @@
 # Backend documentation
 
-EXLA-specific documentation for Nx backend behaviour.
+EXLA-specific documentation for Nx backend behavior.
 
-These guides document divergent behaviour, backend-specific options, and
+These guides document divergent behavior, backend-specific options, and
 limitations for Nx operations. They mirror the structure of the Nx API — see
 the [backend documentation convention](https://hexdocs.pm/nx/backend_documentation-convention.html).
 
 ## Guides
 
-  * [Nx](backend_documentation-nx.html) — top-level `Nx` blocks, transfers, and `defn` integration
-  * [Nx.LinAlg](backend_documentation-nx_lin_alg.html) — linear algebra blocks
-
-Implementation code lives in `EXLA.Defn`, `EXLA.CustomCall`, and related modules.
-
+  * [Nx](backend_documentation-nx.html) — top-level `Nx` operations, transfers, and `defn` integration
+  * [Nx.LinAlg](backend_documentation-nx_lin_alg.html) — linear algebra
--- a/exla/guides/backend_documentation/index.md
+++ b/exla/guides/backend_documentation/index.md
@@ -1,0 +1,24 @@
+# Backend documentation
+
+EXLA-specific documentation for Nx backend behaviour.
+
+These guides describe how EXLA lowers and executes Nx operations. They mirror
+the structure of the Nx API (see the [backend documentation convention](https://hexdocs.pm/nx/backend_documentation-convention.html))
+but are **not callable** — use `Nx` and `compiler: EXLA` in your code.
+
+## Guides
+
+  * [Nx](backend_documentation-nx.html) — top-level `Nx` blocks, transfers, and `defn` integration
+  * [Nx.LinAlg](backend_documentation-nx_lin_alg.html) — linear algebra blocks and related lowerings
+
+Implementation code lives in `EXLA.Defn`, `EXLA.CustomCall`, and related modules.
+
+## Example
+
+```elixir
+iex> Nx.take(Nx.tensor([10, 20, 30]), Nx.tensor([0, 2]))
+#Nx.Tensor<
+  s32[2]
+  [10, 30]
+>
+```
--- a/exla/guides/backend_documentation/index.md
+++ b/exla/guides/backend_documentation/index.md
@@ -1,12 +1,7 @@
 # Backend documentation
 
-EXLA-specific documentation for Nx backend behavior.
+Things that work differently under EXLA than under the portable Nx API. See the
+[backend documentation convention](https://hexdocs.pm/nx/backend_documentation-convention.html).
 
-These guides document divergent behavior, backend-specific options, and
-limitations for Nx operations. They mirror the structure of the Nx API — see
-the [backend documentation convention](https://hexdocs.pm/nx/backend_documentation-convention.html).
-
-## Guides
-
-  * [Nx](backend_documentation-nx.html) — top-level `Nx` operations, transfers, and `defn` integration
-  * [Nx.LinAlg](backend_documentation-nx_lin_alg.html) — linear algebra
+  * [Nx](backend_documentation-nx.html)
+  * [Nx.LinAlg](backend_documentation-nx_lin_alg.html)
--- a/exla/guides/backend_documentation/index.md
+++ b/exla/guides/backend_documentation/index.md
@@ -2,14 +2,14 @@
 
 EXLA-specific documentation for Nx backend behaviour.
 
-These guides describe how EXLA lowers and executes Nx operations. They mirror
-the structure of the Nx API (see the [backend documentation convention](https://hexdocs.pm/nx/backend_documentation-convention.html))
-but are **not callable** — use `Nx` and `compiler: EXLA` in your code.
+These guides document divergent behaviour, backend-specific options, and
+limitations for Nx operations. They mirror the structure of the Nx API — see
+the [backend documentation convention](https://hexdocs.pm/nx/backend_documentation-convention.html).
 
 ## Guides
 
   * [Nx](backend_documentation-nx.html) — top-level `Nx` blocks, transfers, and `defn` integration
-  * [Nx.LinAlg](backend_documentation-nx_lin_alg.html) — linear algebra blocks and related lowerings
+  * [Nx.LinAlg](backend_documentation-nx_lin_alg.html) — linear algebra blocks
 
 Implementation code lives in `EXLA.Defn`, `EXLA.CustomCall`, and related modules.
 
--- a/exla/guides/backend_documentation/nx.md
+++ b/exla/guides/backend_documentation/nx.md
@@ -1,0 +1,264 @@
+# Nx (EXLA)
+
+```elixir
+iex> Nx.take(Nx.tensor([10, 20, 30]), Nx.tensor([0, 2]))
+#Nx.Tensor<
+  s32[2]
+  [10, 30]
+>
+```
+
+EXLA implementation notes for top-level `Nx` operations.
+
+Each function below documents how EXLA handles the corresponding `Nx` API when
+`compiler: EXLA` is used (or when `EXLA.Backend` stores tensors).
+
+Linear algebra is documented separately in [Nx.LinAlg](backend_documentation-nx_lin_alg.html).
+
+Many `Nx` functions delegate to `Nx.block/4`. During EXLA compilation,
+specialized lowerings exist for some block tags; otherwise EXLA compiles the
+default callback as an XLA subcomputation.
+
+## take/3
+
+Tensor indexing (`%Nx.Block.Take{}`).
+
+### Lowering
+
+Lowered to StableHLO `gather` in `EXLA.Defn` (not a custom call).
+
+### Options
+
+* `:axis` — honoured; indices are gathered along this axis
+
+### Platforms
+
+All EXLA clients.
+
+## take_along_axis/3
+
+Take along axis (`%Nx.Block.TakeAlongAxis{}`).
+
+### Lowering
+
+No dedicated gather lowering — EXLA compiles the default Nx callback as an XLA
+subcomputation.
+
+### Options
+
+* `:axis` — honoured by the default callback
+
+## top_k/2
+
+Top-k values and indices (`%Nx.Block.TopK{}`).
+
+### Lowering
+
+Lowered to StableHLO `top_k` via `EXLA.MLIR.Value.top_k/3`.
+
+### Options
+
+* `:k` — number of elements to return per slice
+
+## fft2/2
+
+Two-dimensional FFT (`%Nx.Block.FFT2{}`).
+
+### Lowering
+
+Lowered to StableHLO FFT ops via `EXLA.MLIR.Value.fft/4` with mode `:fft`.
+
+### Options
+
+* `:lengths`, `:axes`, `:eps` — forwarded to the lowering; `:eps` may be used
+for cleanup in the default callback when compilation falls back
+
+## ifft2/2
+
+Two-dimensional inverse FFT (`%Nx.Block.IFFT2{}`).
+
+### Lowering
+
+Lowered to StableHLO FFT ops with mode `:ifft`.
+
+## rfft/2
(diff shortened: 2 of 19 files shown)


josevalim:
Apologies but I am not a fan of this approach. We are introducing a bunch of modules, which have a runtime cost, to document APIs that do not really exist and people should not invoke. Otherwise this will appear as suggestions on LSP, ExDoc autocompletes, etc.

We should figure out how to structure this using regular pages/guides as much as possible.

polvalente:
> Apologies but I am not a fan of this approach. We are introducing a bunch of modules, which have a runtime cost, to document APIs that do not really exist and people should not invoke. Otherwise this will appear as suggestions on LSP, ExDoc autocompletes, etc.
> 
> We should figure out how to structure this using regular pages/guides as much as possible.

I think we can keep the same general idea where we have a standardized nested structure that corresponds to the Nx modules (e.g. EXLA > Backend Documentation >  Nx > LinAlg) which is written as Markdown, Livemd or cheatsheets, as  relevant. The most important part of this PR for me is having an easy way to look things up structurally.

I suspect livebooks will provide the closest experience to regular modules, and cheatsheets can be complimentary documentation to that. We might also want some sort of CI check to ensure the documentation gets updated if we add or remove functions to the original Nx modules, but we can leave this as a follow-up.

josevalim:
I think you can pass files to doctest to make sure their content is correct. But that means using .cheatmd or .md files. Any should be fine!

exla/guides/backend_documentation/index.md
```diff
@@ -0,0 +1,24 @@
+# Backend documentation
+
+EXLA-specific documentation for Nx backend behaviour.
+
+These guides describe how EXLA lowers and executes Nx operations. They mirror
+the structure of the Nx API (see the [backend documentation convention](https://hexdocs.pm/nx/backend_documentation-convention.html))
+but are **not callable** — use `Nx` and `compiler: EXLA` in your code.
```
josevalim: This is wrong now, as it is only markdown files.

exla/guides/backend_documentation/nx.md
```diff
@@ -0,0 +1,264 @@
...
+
+Lowered to StableHLO `gather` in `EXLA.Defn` (not a custom call).
+
+### Options
+
+* `:axis` — honoured; indices are gathered along this axis
+
+### Platforms
+
+All EXLA clients.
```
josevalim: Question: what documenting this particular behaviour is giving us? What it becomes in StableHLO is not necessarily relevant. If it is, we will probably benefit more from having a debug_stablehlo option that prints it.

exla/mix.exs
```diff
@@ -87,9 +88,24 @@ defmodule EXLA.MixProject do
...
+          filename: "backend_documentation",
+          title: "Backend documentation"
+        ],
+        "guides/backend_documentation/nx.md": [
+          filename: "backend_documentation-nx",
+          title: "Nx (EXLA)"
+        ],
+        "guides/backend_documentation/nx_lin_alg.md": [
+          filename: "backend_documentation-nx_lin_alg",
+          title: "Nx.LinAlg (EXLA)"
```
josevalim: This is already the title, do we need to give it again? Same above.

exla/guides/backend_documentation/nx.md
polvalente: We do have this in EXLA.to_mlir_module

polvalente: Now to think out loud about lowering patterns, this function is implemented as an Nx.Block, so. users might want to override the imolementation.

Though at that point they might as well look at the source code.

Not to decide anything, but maybe. what we want here is to scope documenting divergent behavior and backend specific options only?

josevalim: Divergent behavior or specific trade-offs. Yes. Most of what EXLA does is already abstracted several layers away, saying we use gather won’t add much. But saying function X does not support A option or has a special flag can help. The question is what users of this library may want to know. If you are implementing custom blocks, then the source is already the best source of truth anyway.

exla/guides/backend_documentation/index.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Example
+
+```elixir
+iex> Nx.take(Nx.tensor([10, 20, 30]), Nx.tensor([0, 2]))
+#Nx.Tensor<
+  s32[2]
+  [10, 30]
+>
+```
```
polvalente: I don't see why we need this example here

exla/guides/backend_documentation/nx.md
```diff
@@ -0,0 +1,100 @@
+# Nx (EXLA)
+
+```elixir
+iex> Nx.take(Nx.tensor([10, 20, 30]), Nx.tensor([0, 2]))
+#Nx.Tensor<
+  s32[2]
+  [10, 30]
+>
+```
+
```
polvalente:
```suggestion
```

exla/guides/backend_documentation/index.md
polvalente:
```suggestion
```

exla/guides/backend_documentation/nx_lin_alg.md
```diff
@@ -0,0 +1,76 @@
+# Nx.LinAlg (EXLA)
+
+```elixir
+iex> a = Nx.tensor([[4.0, 2.0], [2.0, 3.0]])
+iex> l = Nx.LinAlg.cholesky(a)
+iex> Nx.shape(l)
+{2, 2}
+```
+
```
polvalente:
```suggestion
```

nx/lib/nx/lin_alg.ex
```diff
@@ -136,6 +136,10 @@ defmodule Nx.LinAlg do
           ]
         ]
       >
+
+  ## Backend notes
+
+  See [Torchx](https://hexdocs.pm/torchx/backend_documentation-nx_lin_alg.html#cholesky-1).
```
polvalente: I don't think we should reference backends from Nx itself

nx/lib/nx.ex
```diff
@@ -39,6 +39,14 @@ defmodule Nx do
 
     * `Nx.Constants` declares many constants commonly used in numerical code
 
+  Backend-specific notes for extensible operations (`Nx.block/4`, transfers, and
+  related APIs) are documented in the [backend documentation convention](backend_documentation-convention.html)
+  guide and in each backend's **Backend documentation** pages on HexDocs when a
+  backend has divergent behaviour or backend-specific options (for example
+  [EXLA](https://hexdocs.pm/exla/backend_documentation.html) and
+  [Torchx](https://hexdocs.pm/torchx/backend_documentation.html)).
```
```suggestion
  Backend-specific notes for extensible operations (`Nx.block/4`, transfers, and
  related APIs) are documented in the [backend documentation convention](backend_documentation-convention.html)
  guide and in each backend's **Backend documentation** pages on HexDocs when a
  backend has divergent behaviour or backend-specific option.
```

torchx/guides/backend_documentation/index.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Example
+
+```elixir
+iex> Nx.take(Nx.tensor([10, 20, 30]), Nx.tensor([0, 2]))
+#Nx.Tensor<
+  s32[2]
+  [10, 30]
+>
+```
```
polvalente:
```suggestion
```

torchx/guides/backend_documentation/nx.md
```diff
@@ -0,0 +1,82 @@
+# Nx (Torchx)
+
+```elixir
+iex> Nx.take(Nx.tensor([10, 20, 30]), Nx.tensor([0, 2]))
+#Nx.Tensor<
+  s32[2]
+  [10, 30]
+>
+```
```
polvalente:
```suggestion
```

torchx/guides/backend_documentation/nx_lin_alg.md
```diff
@@ -0,0 +1,143 @@
+# Nx.LinAlg (Torchx)
+
+```elixir
+iex> a = Nx.tensor([[4.0, 2.0], [2.0, 3.0]])
+iex> l = Nx.LinAlg.cholesky(a)
+iex> Nx.shape(l)
+{2, 2}
+```
```
polvalente:
```suggestion
```

polvalente (commented):
Overall this looks great, I just think we need to clean up some of the things I pointed out

exla/guides/backend_documentation/nx.md
```diff
@@ -0,0 +1,92 @@
...
+### Options
+
+* `:permissions` — octal file mode for host IPC shared memory (default
+`0o400`)
+
+### Limitations
+
+Not supported for ROCm, TPU, or non-device buffers.
+
```
polvalente: This is _exactly_ the kind of documentation that this PR let's shine through!

exla/guides/backend_documentation/nx_lin_alg.md
```diff
@@ -0,0 +1,69 @@
...
+### Options
+
+* `:mode` — honoured by whichever implementation is selected (`:reduced` or
+`:complete`)
+* `:eps` — used by the default callback only; ignored by the native custom call
+
+### Numerical notes
+
+Native and default implementations may differ slightly in orthogonality of `q`
+and triangular structure of `r` within floating-point tolerance. Reconstruction
```
polvalente:
```suggestion
and triangular structure of `r` within floating-point tolerance. 
For example, for an eigenpair `(lambda, v)`, `(-lambda, -v)` is
an equally valid eigenpair that would lead to radically different 
`q` and `r` matrices. Reconstruction of
```

exla/guides/backend_documentation/nx_lin_alg.md
```diff
@@ -0,0 +1,69 @@
...
+### Supported platforms
+
+* **Host** — native CPU custom call for supported **real** dtypes (`f32`,
+`f64`, and integer types cast to `f32`); default callback for complex inputs and
+unsupported types
+* **CUDA / ROCm / TPU** — default callback only
+
+### Options
+
+* `:max_iter` and `:eps` — honoured by the default callback; ignored by the
```
polvalente:
```suggestion
* `:max_iter` and `:eps` — honored by the default callback; ignored by the
```

exla/guides/backend_documentation/nx_lin_alg.md
```diff
@@ -0,0 +1,69 @@
...
+using reconstruction `v · diag(λ) · vᵀ` rather than direct eigenvector equality
+across backends.
+
+## svd/2
+
+Singular value decomposition (`%Nx.Block.LinAlg.SVD{}`).
+
+### Options
+
+* `:max_iter` — honoured (default iterative algorithm in Nx)
```
polvalente:
```suggestion
* `:max_iter` — honored (default iterative algorithm in Nx)
```

exla/guides/backend_documentation/nx_lin_alg.md
```diff
@@ -0,0 +1,69 @@
...
+* `:full_matrices?` — honoured
```
polvalente:
```suggestion
* `:full_matrices?` — honored
```

exla/lib/exla.ex
```diff
@@ -82,6 +82,13 @@ defmodule EXLA do
   your block tag struct; see `EXLA.CustomCall` for the `call/4` contract,
   including returning `:skip` to fall back to the block's default Elixir callback.
 
+  Backend-specific behaviour, options, and limitations for Nx functions are
```
polvalente, line 85:
```suggestion
  Backend-specific behavior, options, and limitations for Nx functions are
```

exla/mix.exs
```diff
@@ -70,7 +70,8 @@ defmodule EXLA.MixProject do
 
   defp deps do
     [
-      {:nx, path: "../nx"},
+      {:nx, "~> 0.12.0"},
+      # {:nx, path: "../nx"},
```
polvalente: I think this is a stray change

nx/lib/nx.ex
```diff
@@ -2271,6 +2277,11 @@ defmodule Nx do
   Inside `defn`, this builds an expression node that is supported by compilers.
   Outside of `defn` or in backends without special support, it executes `fun`
   directly and validates that the result matches the provided template.
+
+  ## Backend notes
+
+  See [EXLA](https://hexdocs.pm/exla/backend_documentation-nx.html#runtime_call-4) and
+  [Torchx](https://hexdocs.pm/torchx/backend_documentation-nx.html#runtime_call-4).
```
polvalente: Let's remove all of these mentions to EXLA and Torchx

torchx/guides/backend_documentation/nx_lin_alg.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Options
+
+* `:mode` — `:reduced` or `:complete`, passed to LibTorch
+* `:eps` — ignored (used only by the default Nx callback on MPS fallback paths)
+
+### Numerical notes
+
+`q` and `r` are not guaranteed to match other backends bit-for-bit. Verify
+via reconstruction `q · r ≈ input`.
```
polvalente: Let's copy the docs from EXLA to here

exla/guides/backend_documentation/nx.md
```diff
@@ -0,0 +1,92 @@
...
+Linear algebra is documented separately in [Nx.LinAlg](backend_documentation-nx_lin_alg.html).
+
+## runtime_call/4
+
+Runtime Elixir callback from `defn` (`runtime_call` expression).
+
+### Platforms
+
+* **Host / CUDA** — supported via a device-side callback bridged through
+`EXLA.MLIR.Value.runtime_call/3`
```
josevalim: This is still leaking implementation details :)

josevalim (changes requested):
This is much better but I think the style here is dry and hard to read. It reads like someone traced the source code and added references and it does not feel suitable for library users.

For example, here is what I would write for `Nx.runtime_call/4`:

```
## `Nx.runtime_call/4`

This function allows running Elixir code within `defn` code. Outside of a `defn`, it executes the callback directly, it does not compile the code.

Currently supported only for host and CUDA backends. Raises for ROCm/TPU devices.

> #### Avoid device deadlocks {: .warning}
> 
> The tensors given to the callback are allocated in the binary backend. You must be careful to not use the same device the `defn` itself is running on, as that will lead to a deadlock.
```

So I think we need to considerably improve the quality of the documentation if we want to ship this. Remove reference to internal functions, and focus on what users of the library want to extract from this.

josevalim (approved):

josevalim:
It looks good to me. One last suggestion is that perhaps we don't need this to be three different files. Maybe just one file called "Backend documentation" that expands both Nx and Nx.LinAlg remarks. Three files are likely overkill now when each is less than 100 LOC.

polvalente:
> It looks good to me. One last suggestion is that perhaps we don't need this to be three different files. Maybe just one file called "Backend documentation" that expands both Nx and Nx.LinAlg remarks. Three files are likely overkill now when each is less than 100 LOC.

I think keeping separate files is better because the goal here is to define a standard that other backend maintainers are encouraged to follow and the size of these might vary for each backend.

polvalente (approved):

PR #1774, opened 2026-07-06
Rename hooks to io_call

### LINKED ISSUE #1672: Introduce Nx.io_callback as a replacement for token-based hooks and EXLA's outfeed

Chapaman:
## Summary:

- Introduces Nx.io_call/2-3 and :io_calls as the primary side-effect API (originally featured in #1761
- Keeps deprecated hook/2-3, :hooks, and token shims for backward compatibility
- Part 3 of 3 after #1766 and #1767

closes #1672


nx/lib/nx.ex
```diff
@@ -3899,20 +3968,6 @@ defmodule Nx do
...
-      #Nx.Tensor<
-        s32[11]
-        [0, 0, 10, 0, 20, 0, 30, 0, 0, 0]
-      >
-      # └ low ┘  └ interior ┘   └ high ┘
-      ```
-
   Padding widths can be negative. If they are negative,
   the tensor is clipped on either end according to the
   padding width. Interior padding widths cannot be negative.
```
polvalente, line 3918: This should be kept

Chapaman: Kept — the pad docs example (including the interior-padding illustration and the negative-width note) is present again on this branch after the earlier conflict resolution.

polvalente (approved):

exla/lib/exla/defn.ex
```diff
@@ -320,6 +320,7 @@ defmodule EXLA.Defn do
        ) do
     {cache, options} = Keyword.pop(options, :cache, true)
     {hooks, options} = Keyword.pop(options, :hooks, %{})
```
josevalim, line 322: We need to add a deprecation warning here. Although I would make keep this one as hook in case we want to use it to replace runtime calls or something else in the future too?

nx/lib/nx/defn.ex
```diff
@@ -488,11 +491,18 @@ defmodule Nx.Defn do
   defp prepare_options(opts) do
     opts = Keyword.merge(default_options(), opts)
 
-    if not is_map(Keyword.get(opts, :hooks, %{})) do
-      raise ArgumentError, ":hooks option must be a map"
+    io_calls =
+      opts
+      |> Keyword.get(:io_calls, %{})
+      |> then(&Map.merge(Keyword.get(opts, :hooks, %{}), &1))
```
josevalim, line 497: Here we are merging, in EXLA we are using as default. We need to make them consistent and make hooks warn.

josevalim (commented):
I have dropped some comments. I am not sure about renaming `:hooks` as the option to compilers/backends. That's definitely going to have a larger blast radius and maybe keeping it generic would be more useful in the future?

I also think we should remove the deprecated code before we go with the renaming, because there are some changes in this pull request that are unrelated to the renaming itself!

PR #1775, opened 2026-07-08
Fix vectorized grad/2 for ops that cross the devectorization boundary

### LINKED ISSUE #1729: Vectorized gradient fails for apply_vectorized ops (Cholesky, triangular_solve, cond)

Chapaman:
## Summary
Fixes vectorized `Nx.Defn.grad/2` for ops that internally devectorize and re-vectorize their inputs (`cholesky`, `triangular_solve`, `select`, `cond`), as reported in #1729.
The boundary-wrapper grad from #1533 runs the backward pass in devectorized space. Failures were caused by shape/name collisions at the vectorized ↔ devectorized seam (e.g. duplicate `:batch` axis names), not by missing vectorization support in the ops themselves. This PR strengthens that boundary instead of rewriting forward `apply_vectorized` wrappers.
- **`grad(:triangular_solve)`** — re-vectorize batch axes before backward math and devectorize before returning
- **`cholesky_grad`** — use explicit batch axes in dot products (same pattern as `qr_grad`)
- **`grad(:cond)`** — normalize branch gradients, clear axis-name collisions, and replace (not append) stored grads
- **`to_grad/5`** — use `revectorize` when crossing the boundary; preserve input axis names
- **`expr.ex`** — fix vectorized `cond` zero template to use `revectorize`
- **Docs/error** — clearer message when `grad/2` is called with non-tensor-aware operators like `-x` in a raw anonymous function
Adds vectorized grad tests for `cholesky`, `triangular_solve`, `select`, `cond`, and the `-x` edge case.
Closes #1729.

### DIFF
--- a/nx/lib/nx/binary_backend.ex
+++ b/nx/lib/nx/binary_backend.ex
@@ -527,11 +527,13 @@ defmodule Nx.BinaryBackend do
         right_batch_item_bits = right_batch_item_length * right_size
 
         <<_::bitstring-size(^left_offset_bits),
-          left_batch_item_binary::bitstring-size(^left_batch_item_bits), _::bitstring>> =
+          left_batch_item_binary::bitstring-size(^left_batch_item_bits),
+          _::bitstring>> =
           left_binary
 
         <<_::bitstring-size(^right_offset_bits),
-          right_batch_item_binary::bitstring-size(^right_batch_item_bits), _::bitstring>> =
+          right_batch_item_binary::bitstring-size(^right_batch_item_bits),
+          _::bitstring>> =
           right_binary
 
         bin_dot(
@@ -1779,7 +1781,8 @@ defmodule Nx.BinaryBackend do
             before_slice_size = current - previous
 
             <<before_offset::bitstring-size(^before_slice_size),
-              current_bitstring::bitstring-size(^target_chunk), to_traverse::bitstring>> =
+              current_bitstring::bitstring-size(^target_chunk),
+              to_traverse::bitstring>> =
               to_traverse
 
             updated_elements =
--- a/nx/lib/nx/binary_backend.ex
+++ b/nx/lib/nx/binary_backend.ex
@@ -527,13 +527,11 @@ defmodule Nx.BinaryBackend do
         right_batch_item_bits = right_batch_item_length * right_size
 
         <<_::bitstring-size(^left_offset_bits),
-          left_batch_item_binary::bitstring-size(^left_batch_item_bits),
-          _::bitstring>> =
+          left_batch_item_binary::bitstring-size(^left_batch_item_bits), _::bitstring>> =
           left_binary
 
         <<_::bitstring-size(^right_offset_bits),
-          right_batch_item_binary::bitstring-size(^right_batch_item_bits),
-          _::bitstring>> =
+          right_batch_item_binary::bitstring-size(^right_batch_item_bits), _::bitstring>> =
           right_binary
 
         bin_dot(
@@ -1781,8 +1779,7 @@ defmodule Nx.BinaryBackend do
             before_slice_size = current - previous
 
             <<before_offset::bitstring-size(^before_slice_size),
-              current_bitstring::bitstring-size(^target_chunk),
-              to_traverse::bitstring>> =
+              current_bitstring::bitstring-size(^target_chunk), to_traverse::bitstring>> =
               to_traverse
 
             updated_elements =
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -626,10 +626,12 @@ defmodule Nx.Defn.Grad do
         0.0
       )
 
-    # w.r.t operand
+    # w.r.t operand. Elements exactly at the boundaries pass the gradient
+    # through to the operand (matching clip_by_value/clamp in other
+    # frameworks); the limits only receive gradient when strictly exceeded.
     w_operand =
       Nx.select(
-        Nx.bitwise_and(Nx.greater(operand, min), Nx.less(operand, max)),
+        Nx.bitwise_and(Nx.greater_equal(operand, min), Nx.less_equal(operand, max)),
         g,
         0.0
       )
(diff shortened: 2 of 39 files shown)


nx/lib/nx/defn/grad.ex
```diff
@@ -1074,7 +1199,14 @@ defmodule Nx.Defn.Grad do
   defp grad(:fft, args, ans, g, _batch_count), do: grad_fft(:fft, args, ans, g)
   defp grad(:ifft, args, ans, g, _batch_count), do: grad_fft(:ifft, args, ans, g)
 
-  defp grad(:triangular_solve, [a_input, b, opts], x_input, g, _batch_count) do
+  defp grad(:triangular_solve, [a_input, b_input, opts], x_input, g, batch_count) do
+    vectorized_axes = batch_vectorized_axes(b_input, batch_count)
+
+    a = revectorize_batch_axes(a_input, vectorized_axes)
+    _b = revectorize_batch_axes(b_input, vectorized_axes)
```
```suggestion
```

nx/lib/nx/defn/grad.ex
```diff
@@ -1504,6 +1639,28 @@ defmodule Nx.Defn.Grad do
...
+  defp revectorize_batch_axes(%T{shape: shape} = tensor, vectorized_axes) do
+    batch_count = length(vectorized_axes)
+    {batch_sizes, rest_shape} = shape |> Tuple.to_list() |> Enum.split(batch_count)
+    expected_sizes = Enum.map(vectorized_axes, &elem(&1, 1))
+
+    if batch_sizes == expected_sizes do
+      Nx.revectorize(tensor, vectorized_axes, target_shape: List.to_tuple(rest_shape))
+    else
+      tensor
+    end
```
polvalente: This smells like a silent failure. In which case would a tensor come in with sizes which aren't the expected ones?

In the very least I think something where `batch_sizes` are broadcastable to `expected_sizes` might occur and we're silently passing the tensor through. (e.g. {2, 1, 3} vs {2, 10, 3})

nx/lib/nx/binary_backend.ex
```diff
@@ -527,11 +527,13 @@ defmodule Nx.BinaryBackend do
...
         <<_::bitstring-size(^left_offset_bits),
-          left_batch_item_binary::bitstring-size(^left_batch_item_bits), _::bitstring>> =
+          left_batch_item_binary::bitstring-size(^left_batch_item_bits),
+          _::bitstring>> =
           left_binary
 
         <<_::bitstring-size(^right_offset_bits),
-          right_batch_item_binary::bitstring-size(^right_batch_item_bits), _::bitstring>> =
+          right_batch_item_binary::bitstring-size(^right_batch_item_bits),
+          _::bitstring>> =
```
polvalente: I think this might be stale formatting wrt elixir 1.20

nx/lib/nx/defn/grad.ex
```diff
@@ -250,25 +278,98 @@ defmodule Nx.Defn.Grad do
...
+  defp normalize_vectorized_shape(%T{vectorized_axes: []} = t), do: t
+
+  defp normalize_vectorized_shape(%T{shape: shape, names: names, vectorized_axes: vec_axes} = t) do
+    vec_count = length(vec_axes)
+
+    if tuple_size(shape) >= vec_count and
+         Enum.take(Tuple.to_list(shape), vec_count) == Enum.map(vec_axes, &elem(&1, 1)) do
+      %{
+        t
+        | shape: shape |> Tuple.to_list() |> Enum.drop(vec_count) |> List.to_tuple(),
```
polvalente: instead of Enum.take + Enum.drop, we could use Enum.split to avoid double traversal

nx/lib/nx/defn/grad.ex
```diff
@@ -1522,6 +1656,54 @@ defmodule Nx.Defn.Grad do
     end
   end
 
+  defp batch_vectorized_axes(_tensor, 0), do: []
+
+  defp batch_vectorized_axes(%T{shape: shape, names: names}, batch_count) do
+    {batch_sizes, _} = shape |> Tuple.to_list() |> Enum.split(batch_count)
+    {batch_names, _} = Enum.split(names, batch_count)
```
polvalente: Enum.take instead of split

nx/lib/nx/defn/grad.ex
```diff
@@ -1522,6 +1656,54 @@ defmodule Nx.Defn.Grad do
...
+
+  defp revectorize_batch_axes(%T{shape: shape} = tensor, vectorized_axes) do
+    batch_count = length(vectorized_axes)
+    expected_sizes = Enum.map(vectorized_axes, &elem(&1, 1))
+    shape_l = Tuple.to_list(shape)
+
+    # Captured non-vectorized operands (e.g. `a` when only `b` is vectorized) do not
+    # carry a batch prefix — leave them alone. If a leading prefix is present and
+    # broadcast-compatible but not equal, fail loudly instead of silently skipping.
+    if length(shape_l) < batch_count do
```
polvalente: tuple_size(shape) instead of lenght(shape_l) is more efficient

polvalente (approved):

PR #1777, opened 2026-07-15
Add :pjrt_plugin platform for loading external PJRT plugins

chapai:
## What

Adds a `:pjrt_plugin` client platform to EXLA that loads an external PJRT C-API plugin (a shared library) at runtime and builds a client from it. The plugin is selected per-client via two new options:

- `:device_type` — the device type string the plugin registers under (e.g. `"tt"`)
- `:plugin_path` — path to the plugin shared library (e.g. `pjrt_plugin_tt.so`)

```elixir
config :exla, :clients,
  tt: [platform: :pjrt_plugin , device_type: "tt", plugin_path: "/path/to/pjrt_plugin_tt.so"]
```
Implementation is a single new branch in EXLA.Client.build_client/2 that calls the existing EXLA.NIF.load_pjrt_plugin/2 and EXLA.NIF.get_c_api_client/1 NIFs, plus docs in the EXLA "Clients" section.

## Why

Today each accelerator needs a dedicated platform branch (:cuda, :rocm, :tpu). PJRT already defines a stable C-API for out-of-tree plugins, and the loader NIFs are already in the tree — this just exposes them through client config. It lets EXLA target any hardware shipping a PJRT plugin without a bespoke backend. My motivating case is running Nx on Tenstorrent hardware via tt-xla's pjrt_plugin_tt.so, but nothing here is TT-specific.

## Scope

- No C++ / no NIF changes. Elixir-only
- Missing :device_type or :plugin_path raises a clear ArgumentError.
- Any environment a plugin itself requires (e.g. a vendor runtime root dir) must be set in the OS environment before the VM starts; EXLA does not manage it. This is documented.

## Testing

- Existing EXLA host suite is unchanged and passes (957 doctests, 445 tests; the one Nx.rsqrt/1 f64 last-ULP doctest diff is pre-existing and reproduces on a clean tree — a precompiled xla_extension CPU machine-feature mismatch, unrelated to this change).
- End-to-end validated on real hardware: with tt-xla's pjrt_plugin_tt.so, Nx.add/2 and a full linear-regression training run execute on TT hardware and PCC-match :host.

### DIFF
--- a/exla/lib/exla.ex
+++ b/exla/lib/exla.ex
@@ -126,7 +126,15 @@ defmodule EXLA do
   Each client configuration accepts the following options:
 
     * `:platform` - the platform the client runs on. It can be
-      `:host` (CPU), `:cuda`, `:rocm`, or `:tpu`. Defaults to `:host`.
+      `:host` (CPU), `:cuda`, `:rocm`, `:tpu`, or `:c_api` (an
+      externally loaded PJRT plugin, see below). Defaults to `:host`.
+
+    * `:device_type` - required for the `:c_api` platform. The device
+      type string the PJRT plugin registers under (for example `"tt"`).
+
+    * `:plugin_path` - required for the `:c_api` platform. The path to
+      the PJRT plugin shared library to load (for example the tt-xla
+      `pjrt_plugin_tt.so`).
 
     * `:default_device_id` - the default device ID to run on.
       For example, if you have two GPUs, you can choose a different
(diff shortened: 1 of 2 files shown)


josevalim (approved):
Looks great to me! Perhaps we only want to change the platform name to `:pjrt_plugin` but let's see what @polvalente thinks!

chapai:
Thanks for the review @josevalim! I was struggling to pick the name an `:pjrt_plugin` sounds way better to me. 
I will follow with the change after I check the failed tests. Most likely tomorrow.

polvalente (approved):
I agree with José regarding the name! Awesome contribution

chapai:
@josevalim @polvalente I checked the failed test from the previous run ([one](https://github.com/elixir-nx/nx/actions/runs/29388222405/job/87300766112), [two](https://github.com/elixir-nx/nx/actions/runs/29388222405/job/87300766133)) and it looks like both are ` doctest Nx.rsqrt/1 (435) (EXLA.BackendTest)` (unrelated to my changes). 

Should I fix them as a separate PR or it is fine to skip it?

Edit: nvm, the fresh run passed

josevalim:

Issue #1746, opened 2026-04-21
Nx.LinAlg.determinant gradient broken for batched 4x4+ inputs


blasphemetheus:
`Nx.LinAlg.determinant` gradient raises on batched 4×4 and larger inputs (3×3 happens to work on the diagonal-case I tested). Shape mismatch in the grad's internal `Nx.dot` — same class as the other batched-grad bugs.

```elixir
x = Nx.tensor([[
      [5.0, 1.0, 2.0, 3.0],
      [4.0, 10.0, 6.0, 7.0],
      [8.0, 9.0, 15.0, 11.0],
      [12.0, 13.0, 14.0, 20.0]
    ]])                                 # {1, 4, 4}

Nx.Defn.grad(x, fn a -> Nx.sum(Nx.LinAlg.determinant(a)) end)
# ** (ArgumentError) dot/zip expects shapes to be compatible,
#    dimension 3 of left-side (4) does not equal dimension 2 of right-side (1)
```

Same class as #1740/#1741/#1742/#1743. Found by the batched-grad fuzz probe.

polvalente:
@blasphemetheus for future reference, you should open a single issue if many functions are failing due to using the same coding strategy.

The fixes will likely be in the same PR, and I think having too much granularity in the issue tracker adds more noise than clarity

blasphemetheus:
yeah, sorry about that

Issue #1744, opened 2026-04-21
Nx.LinAlg.invert gradient broken for batched inputs


blasphemetheus:
`Nx.LinAlg.invert` gradient raises on batched inputs. The `custom_grad` at `nx/lib/nx/lin_alg.ex:866` uses `Nx.dot` without batch axes, producing a higher-rank intermediate that can't broadcast.

```elixir
x = Nx.tensor([
      [[4.0, 1.0], [1.0, 5.0]],
      [[9.0, 1.0], [1.0, 10.0]]
    ])                                  # {2, 2, 2}

Nx.Defn.grad(x, fn a -> Nx.sum(Nx.LinAlg.invert(a)) end)
# ** (ArgumentError) cannot broadcast tensor of dimensions
#    {2, 3, 2, 2, 3} to {2, 3, 3}
```

Same class as #1741/#1742/#1743 (triangular_solve / LU / SVD batched-grad bugs). Fix pattern mirrors QR/Cholesky in #1731: use `batch_axes/1` + `Nx.dot(a, [-1], ba, g, [-2], ba)` style in the custom_grad formula.

Found while extending the batched-grad fuzz probe across remaining hand-derived-grad linalg ops.

Issue #1742, opened 2026-04-21
Nx.LinAlg.lu gradient broken for batched inputs


blasphemetheus:
`Nx.LinAlg.lu` gradient raises on batched inputs. `lu_grad` uses unqualified `Nx.dot` calls that contract the wrong axes when the input has a leading batch dimension.

```elixir
x = Nx.tensor([
      [[4.0, 3.0], [6.0, 3.0]],
      [[2.0, 1.0], [5.0, 7.0]]
    ])                                  # {2, 2, 2}

Nx.Defn.grad(x, fn t ->
  {_p, l, u} = Nx.LinAlg.lu(t)
  Nx.add(Nx.sum(l), Nx.sum(u))
end)
# ** (ArgumentError) cannot broadcast tensor of dimensions
#    {2, 2, 2, 2, 2, 2, 2} to {2, 2, 2}
```

Same class as #1741 (triangular_solve), and same class as the QR/Cholesky batched-grad bugs fixed on the branch behind #1731. Likely fixable with the same `batch_axes/1` helper pattern — `Nx.dot(a, [-1], ba, b, [-2], ba)` instead of `Nx.dot(a, b)`.

Issue #1741, opened 2026-04-21
Nx.LinAlg.triangular_solve gradient broken for batched inputs


blasphemetheus:
`Nx.LinAlg.triangular_solve` gradient raises on batched (3D+) inputs — failure mode is a shape-mismatch in the `Nx.dot` call inside the custom grad formula. Related to but distinct from #1729: this one uses a plain batched tensor with no `Nx.vectorize` wrapping.

```elixir
a = Nx.tensor([
      [[4.0, 0.0], [2.0, 5.0]],
      [[9.0, 0.0], [3.0, 10.0]]
    ])                                # {2, 2, 2} lower-triangular
b = Nx.broadcast(1.0, {2, 2})

Nx.Defn.grad(a, fn x ->
  Nx.sum(Nx.LinAlg.triangular_solve(x, b))
end)
# ** (ArgumentError) dot/zip expects shapes to be compatible,
#    dimension 2 of left-side (2) does not equal dimension 0 of right-side (2)
```

Same class of bug as the LU/SVD batched-grad issues — the `custom_grad` formulas were written assuming 2D inputs and haven't been made batch-aware. Likely fixable with the same `batch_axes/1` helper pattern PR #1731 applies to `qr_grad` and `cholesky_grad`.

Found by a batched-grad fuzz probe while auditing linalg gradients.

PR #1778, opened 2026-07-15
fix: batched gradients for lu, invert, triangular_solve and determinant

RicardoSantos-99:
The gradient formulas for these linalg ops used Nx.dot/2, which assumes 2D inputs. On batched (3D+) tensors the contraction axes were wrong and the backward pass either raised or produced bad shapes. Forward passes were fine.

Added a batch_axes helper (same approach already used by qr_grad) and switched the dot calls to contract only the last two axes, keeping the leading batch axes. triangular_solve also had two Nx.transpose/1 calls that reversed every axis; they now transpose only the last two, and the vector RHS reshape is rank based so it covers both a plain {n} vector and a batched {batch, n} one (the shape used by [#1741](https://github.com/RicardoSantos-99/dominex/issues/1741)'s own reproducer).

determinant goes through lu, so its batched grad is fixed too. Tests compare batched grads against the stack of per-slice 2D grads, plus finite differences for triangular_solve.

Same batched-grad class and fix pattern as #1731 (qr, cholesky).

Closes [#1741](https://github.com/RicardoSantos-99/dominex/issues/1741)
Closes [#1742](https://github.com/RicardoSantos-99/dominex/issues/1742)
Closes [#1744](https://github.com/RicardoSantos-99/dominex/issues/1744)
Closes [#1746](https://github.com/RicardoSantos-99/dominex/issues/1746)

### DIFF
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -1112,7 +1112,7 @@ defmodule Nx.Defn.Grad do
       end
 
     a_inv_hermitian = Nx.LinAlg.invert(Nx.LinAlg.adjoint(a))
-    ba = linalg_batch_axes(a_inv_hermitian)
+    batch_axes = linalg_batch_axes(a_inv_hermitian)
 
     x =
       case {Nx.shape(x_input), opts[:left_side]} do
@@ -1140,18 +1140,31 @@ defmodule Nx.Defn.Grad do
         # which means that:
         # A_bar = inv(A^H).X_bar.X^H
         # B_bar = inv(A^H).X_bar
-        gx = Nx.dot(g, [-1], ba, Nx.LinAlg.adjoint(x), [-2], ba)
-        da = a_inv_hermitian |> Nx.dot([-1], ba, gx, [-2], ba) |> Nx.negate()
-        db = Nx.dot(a_inv_hermitian, [-1], ba, g, [-2], ba)
+        da =
+          a_inv_hermitian
+          |> Nx.dot(
+            [-1],
+            batch_axes,
+            Nx.dot(g, [-1], batch_axes, Nx.LinAlg.adjoint(x), [-2], batch_axes),
+            [-2],
+            batch_axes
+          )
+          |> Nx.negate()
+
+        db = Nx.dot(a_inv_hermitian, [-1], batch_axes, g, [-2], batch_axes)
         {da, db}
       else
         # X.A = B -> X = B.inv(A)
         # taking a similar approach to the branch above, we get
         # A_bar = -X^H.X_bar.inv(A^H)
         # B_bar = X_bar.inv(A^H)
-        xg = Nx.dot(Nx.LinAlg.adjoint(x), [-1], ba, g, [-2], ba)
-        da = xg |> Nx.dot([-1], ba, a_inv_hermitian, [-2], ba) |> Nx.negate()
-        db = Nx.dot(g, [-1], ba, a_inv_hermitian, [-2], ba)
+        da =
+          Nx.LinAlg.adjoint(x)
+          |> Nx.dot([-1], batch_axes, g, [-2], batch_axes)
+          |> Nx.dot([-1], batch_axes, a_inv_hermitian, [-2], batch_axes)
+          |> Nx.negate()
+
+        db = Nx.dot(g, [-1], batch_axes, a_inv_hermitian, [-2], batch_axes)
         {da, db}
       end
 
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -1107,11 +1107,12 @@ defmodule Nx.Defn.Grad do
     a =
       case opts[:transform_a] do
         :none -> a_input
-        :transpose -> Nx.transpose(a_input)
+        :transpose -> transpose_matrix(a_input)
         :conjugate -> Nx.conjugate(a_input)
       end
 
     a_inv_hermitian = Nx.LinAlg.invert(Nx.LinAlg.adjoint(a))
+    ba = linalg_batch_axes(a_inv_hermitian)
 
     x =
       case {Nx.shape(x_input), opts[:left_side]} do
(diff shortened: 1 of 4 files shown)


nx/lib/nx/lin_alg/lu.ex
```diff
@@ -159,16 +159,27 @@ defmodule Nx.LinAlg.LU do
     u_h = Nx.LinAlg.adjoint(u)
     l_h = Nx.LinAlg.adjoint(l)
     p_t = Nx.LinAlg.adjoint(p)
+    ba = batch_axes(u)
```
polvalente: let's call this variable `batch_axes`. The other variables have bad names because they are direct translations of math formulas, but we generally prefer readable names.

```suggestion
    batch_axes = batch_axes(u)
```

nx/lib/nx/defn/grad.ex
```diff
@@ -1107,11 +1107,12 @@ defmodule Nx.Defn.Grad do
     a =
       case opts[:transform_a] do
         :none -> a_input
-        :transpose -> Nx.transpose(a_input)
+        :transpose -> transpose_matrix(a_input)
         :conjugate -> Nx.conjugate(a_input)
       end
 
     a_inv_hermitian = Nx.LinAlg.invert(Nx.LinAlg.adjoint(a))
+    ba = linalg_batch_axes(a_inv_hermitian)
```
```suggestion
    batch_axes = linalg_batch_axes(a_inv_hermitian)
```

nx/lib/nx/defn/grad.ex
```diff
@@ -1139,23 +1140,25 @@ defmodule Nx.Defn.Grad do
...
         {da, db}
       else
         # X.A = B -> X = B.inv(A)
         # taking a similar approach to the branch above, we get
         # A_bar = -X^H.X_bar.inv(A^H)
         # B_bar = X_bar.inv(A^H)
-        da = x |> Nx.LinAlg.adjoint() |> Nx.dot(g) |> Nx.dot(a_inv_hermitian) |> Nx.negate()
-        db = Nx.dot(g, a_inv_hermitian)
+        xg = Nx.dot(Nx.LinAlg.adjoint(x), [-1], ba, g, [-2], ba)
+        da = xg |> Nx.dot([-1], ba, a_inv_hermitian, [-2], ba) |> Nx.negate()
```
```suggestion
        da = Nx.LinAlg.adjoint(x) |> Nx.dot([-1], ba, g, [-2], ba) |> Nx.dot([-1], ba, a_inv_hermitian, [-2], ba) |> Nx.negate()
```

xg is not used anywhere else, so this avoids adding another hard to read variable

nx/lib/nx/defn/grad.ex
```diff
@@ -1139,23 +1140,25 @@ defmodule Nx.Defn.Grad do
         # which means that:
         # A_bar = inv(A^H).X_bar.X^H
         # B_bar = inv(A^H).X_bar
-        da = a_inv_hermitian |> Nx.dot(g |> Nx.dot(Nx.LinAlg.adjoint(x))) |> Nx.negate()
-        db = Nx.dot(a_inv_hermitian, g)
+        gx = Nx.dot(g, [-1], ba, Nx.LinAlg.adjoint(x), [-2], ba)
+        da = a_inv_hermitian |> Nx.dot([-1], ba, gx, [-2], ba) |> Nx.negate()
```
```suggestion
        da = a_inv_hermitian |> Nx.dot([-1], ba, Nx.dot(g, [-1], ba, Nx.LinAlg.adjoint(x), [-2], ba), [-2], ba) |> Nx.negate()
```

`gx` could be conflated with the name pattern for `gradient of x`, which is why I'd rather avoid adding it

nx/lib/nx/lin_alg.ex
```diff
@@ -866,10 +866,22 @@ defmodule Nx.LinAlg do
...
       ans_h = adjoint(ans)
-      [ans_h |> Nx.negate() |> Nx.dot(g) |> Nx.dot(ans_h)]
+      ba = batch_axes(ans_h)
+
+      [
+        ans_h
+        |> Nx.negate()
+        |> Nx.dot([-1], ba, g, [-2], ba)
+        |> Nx.dot([-1], ba, ans_h, [-2], ba)
+      ]
```
polvalente:
```suggestion
      batch_axes = batch_axes(ans_h)

      [
        ans_h
        |> Nx.negate()
        |> Nx.dot([-1], batch_axes, g, [-2], batch_axes)
        |> Nx.dot([-1], batch_axes, ans_h, [-2], batch_axes)
      ]
```

polvalente (approved):
Thanks for the PR! I'll deal with the suggested changes :)

RicardoSantos-99:
> Thanks for the PR! I'll deal with the suggested changes :)

Thanks for the review! I'm going to pull triangular_solve out of this PR before merging, for two reasons:

There's already a draft PR ([#1775](https://github.com/RicardoSantos-99/dominex/issues/1775)) reworking grad(:triangular_solve) for the vectorized case.
While double-checking against #1741's exact reproducer, I found our fix only handles a batched matrix RHS, not a batched vector RHS (e.g. b shaped {batch, n}), which is the actual shape in that issue. So it's not a complete fix yet.
I'll keep this PR scoped to lu, invert and determinant ([#1742](https://github.com/RicardoSantos-99/dominex/issues/1742), [#1744](https://github.com/RicardoSantos-99/dominex/issues/1744), [#1746](https://github.com/RicardoSantos-99/dominex/issues/1746)), and follow up on triangular_solve separately once #1775 lands or the vector case is handled properly.

polvalente:
@RicardoSantos-99 I think it's ok if you keep triangular_solve in this one. That PR reworks something tangential to this which is vectorized axes, not batch axes directly :)

I'll wait for the next commits before merging, then

RicardoSantos-99:
@polvalente I fixed the vector RHS case, triangular_solve is back in with a batched vector b now handled correctly (verified against #1741's exact reproducer plus finite differences and the full test suite). Should be ready for another look.

nx/lib/nx/defn/grad.ex
```diff
@@ -1107,25 +1107,25 @@ defmodule Nx.Defn.Grad do
...
 
-    x =
-      case {Nx.shape(x_input), opts[:left_side]} do
-        {{n}, true} -> Nx.reshape(x_input, {n, 1})
-        {{n}, false} -> Nx.reshape(x_input, {1, n})
-        _ -> x_input
+    # When b is a (possibly batched) vector, its rank is one less than a's.
+    # Add a unit axis so the backward math below can treat it as a matrix;
+    # the axis is removed from db at the end.
+    vector_axis =
```
polvalente: vector_axis might not be the ideal name due to vectorization naming colision.
How about we call this `b_rank_correction_axis`

nx/test/nx/defn/grad_test.exs
```diff
@@ -2560,6 +2570,49 @@ defmodule Nx.Defn.GradTest do
...
     end
+
+    test "computes grad for batched tensor" do
+      t =
+        Nx.tensor([
+          [[1.0, 2.0, 3.0], [0.0, 4.0, 5.0], [0.0, 0.0, -6.0]],
+          [[2.0, 0.0, 1.0], [1.0, 3.0, 0.0], [0.0, 1.0, 4.0]]
+        ])
+
+      assert_all_close(invert_grad(t), Nx.stack([invert_grad(t[0]), invert_grad(t[1])]))
```
polvalente: I expect this to pass `assert_equal`

polvalente: Same for similar assertions in this file

polvalente (commented):
Looks, great! Just left 2 small comments and we can merge

nx/lib/nx/defn/grad.ex
RicardoSantos-99: done

nx/test/nx/defn/grad_test.exs
RicardoSantos-99: done

polvalente (approved):

PR #1779, opened 2026-07-15
Fix Nx.clip gradient applying upstream grad twice (g²)

blasphemetheus:
Hey there — I hit NaNs deep into a training run on my project (Elixir/Nx
under the hood), and added a `Nx.clip` guard to the loss to stabilize it.
The guard itself silently broke training: loss climbed from init, gradients
were wrong from step 0. Tracked it down to the `:clip` grad rule and wrote
pin tests that isolate it. They fail on current main, so here's a fix.

## The bug

`clip` should pass the upstream gradient `g` through unchanged for
in-range elements. But in `Nx.Defn.Grad`, the `:clip` rule builds its
selection masks with `g` already inside them:

```elixir
w_operand = Nx.select(..., g, 0.0)
w_min     = Nx.select(..., Nx.broadcast(g, operand), 0.0)
w_max     = Nx.select(..., Nx.broadcast(g, operand), 0.0)
```

…and then multiplies by `g` a second time when assembling the result. So
`clip` backprops **g² instead of g**. The forward pass is fine — only the
backward is wrong. Squaring destroys sign (every gradient component comes
out non-negative) and shrinks magnitude (under `mean`, gradients are
1/n² instead of 1/n).

The wrong values are identical on BinaryBackend and EXLA, so this is the
grad rule itself, not a compiler.

## Why the existing tests pass

All three existing clip grad tests (`test/nx/defn/grad_test.exs`, "clip"
describe block) differentiate `Nx.sum(Nx.clip(...))`. Under sum the
upstream gradient is `g = 1` everywhere, and `1² == 1` — the bug is
mathematically invisible. It only shows once `clip` sits inside a
composite expression where `g` varies (a `mean`, a loss, anything
downstream of the clip).

## The fix

Delete the redundant multiply — the masks already carry `g`:

```elixir
[
  {operand, w_operand},
  {min, Nx.sum(w_min)},
  {max, Nx.sum(w_max)}
]
```

## Tests

Six new tests in the existing `describe "clip"` block of
`test/nx/defn/grad_test.exs`, ordered from bare `mean(clip(x))` up to the
stable-form BCE shape that bit me in the wild, each against analytically
derived expected gradients:

- On unpatched main: 5 of the 6 fail (the 6th is a control — the
  `Nx.min(Nx.max(x, lo), hi)` equivalent, whose gradients were already
  correct).
- With the fix: the full grad suite passes 280/280.

## Notes

- I audited the other grad rules in `grad.ex` for the same shape (mask
  built with `g` inside a `select`, then re-multiplied by `g`):
  `:select`, `:product`, binary min/max, abs all apply `g` exactly once.
  `:clip` is the only instance.
- Workaround for released versions, if anyone else hits this:
  `Nx.min(Nx.max(x, lo), hi)` composes to the same forward values with
  correct gradients.


nx/lib/nx/defn/grad.ex
polvalente, line 516: I wonder if this should be greater_equal and less_equal

polvalente, line 503: Should this be just Nx.greater_equal(min, operand)?

nx/test/nx/fuzz_clip_grad_test.exs
```diff
@@ -0,0 +1,124 @@
+defmodule Nx.FuzzClipGradTest do
```
polvalente: Let's not add a specific module for this and remove the heavy AI annotations in docs/comments

nx/test/nx/defn/grad_test.exs
```diff
@@ -3318,6 +3318,106 @@ defmodule Nx.Defn.GradTest do
     test "computes gradient wrt to upper lim" do
       assert grad_sum_clip_wrt_to_upper_lim(Nx.iota({5}), 2.5) == Nx.tensor(2.0)
     end
+
+    # The tests below differentiate through mean-based composite expressions,
+    # where the gradient arriving at the clip node is not 1. Sum-based tests
+    # alone cannot detect gradient-scaling bugs in the clip rule: with g == 1,
+    # g² == g (the rule used to backprop g² instead of g). Expectations are
+    # analytic, and the min∘max variant pins that both clamp spellings
+    # produce identical gradients.
```
polvalente: I don't think we should document what the rule used to do as that was a bug

nx/lib/nx/defn/grad.ex
blasphemetheus: Huh, yes. And I think if we use greater_equal on w_operand and w_min there could be a tie value that magnifies to 2g instead of g. I'll try and write a test to confirm that behavior

blasphemetheus:
Lmk if I should remove more annotations, it's a bit less wordy for no reason now, any other thoughts, can turn off draft status whenever

polvalente:
> Lmk if I should remove more annotations, it's a bit less wordy for no reason now, any other thoughts, can turn off draft status whenever

General shape looks great! If it's ready I'll do a final review pass

polvalente (approved):

polvalente (approved):

PR #1780, opened 2026-07-18
chore: fully deprecate tokens + use :hooks as the option for named io_calls

polvalente:
Continuation of the move from token hooks into io_calls


josevalim (approved):
!!!!!!

PR #1781, opened 2026-07-19
perf(cholesky): replace zero_out_dynamic_slice with mask-based approach, ~2x speedup

Ljzn:
### Problem

The inner loop of `cholesky_matrix/2` calls `dot_with_dynamic_slice/6` O(n²/2) times to compute prefix dot products of rows `L[i][0..j)` and `L[j][0..j)`.

Since `j` is a `defn` while-loop variable (a dynamic tensor), `Nx.slice/4` cannot be used directly — its `lengths` parameter requires static integers (`Nx.Shape.slice/4` line 1263).

The previous workaround, `zero_out_dynamic_slice/3`, used `Nx.argsort` to move the desired range `[start, end)` to the front of the vector and zero out the rest. This is unnecessarily expensive — O(n log n) per call due to `argsort`, and creates 7 intermediate tensors per invocation.

### Solution

Replace the argsort-based approach with independent mask + select for each operand. Since `Nx.dot` already ignores zero-valued elements, there is no need to physically rearrange the vector — zeroing out the out-of-range elements is sufficient:

```elixir
defnp dot_with_dynamic_slice(left, left_start, left_end, right, right_start, right_end) do
  n = Nx.axis_size(left, 0)
  idx = Nx.iota({n})
  left_mask = Nx.logical_and(idx >= left_start, idx < left_end)
  right_mask = Nx.logical_and(idx >= right_start, idx < right_end)

  Nx.dot(Nx.select(left_mask, left, 0), Nx.select(right_mask, right, 0))
end
```

This is O(n) per call (down from O(n log n)), uses 4 intermediate tensors (down from 13), and is functionally identical. Unlike the argsort approach (which aligns by position-within-range), this approach aligns by original index — but this produces the same result whenever `left_start == right_start`, which is the case at both current call sites.

The dead function `zero_out_dynamic_slice/3` is removed entirely.

### Benchmark

Full `Nx.LinAlg.cholesky/2` on random 32-bit float positive definite matrices, averaged over 3 runs per size:

| 50 | 298.41 | 177.60 | 1.68x |
| 100 | 2051.62 | 1081.13 | 1.90x |
| 200 | 21610.71 | 11165.04 | 1.94x |

Reconstruction error `L * L' ≈ A` is identical between both versions (~1.7e-7 for n=50, ~1.9e-7 for n=100, ~2.7e-7 for n=200).

<details>

```elixir
defmodule Bench do
  def run do
    sizes = [50, 100, 200]
    iters = 3
    IO.puts("n\\ttime_ms\\terror")
    for n <- sizes do
      key = Nx.Random.key(42)
      a = Nx.dot(Nx.LinAlg.adjoint(a), a) |> Nx.add(Nx.eye(n) |> Nx.multiply(n))
      times = Enum.map(1..iters, fn _ -> {t, _} = :timer.tc(fn -> Nx.LinAlg.cholesky(a) end); t end)
      l = Nx.LinAlg.cholesky(a)
      err = Nx.subtract(a, Nx.dot(l, Nx.LinAlg.adjoint(l))) |> Nx.abs() |> Nx.mean() |> Nx.to_number()
      IO.puts("#{n}\\t#{Float.round(avg_ms, 2)}\\t#{err}")
    end
  end
end
Bench.run()
```

</details>

### Correctness

- All 2702 existing tests pass (`mix test`).
- Both current call sites use `left_start == right_start` (both `0`), so the result is identical — both implementations compute `∑ left[i] × right[i]` over the range.
- The `L * L'" ≈ A` reconstruction error is unchanged.
- The independent-mask form is more robust: if someone later passes different ranges for left and right, the function correctly computes the per-element product over the intersection of both ranges, rather than silently misaligning them as the old argsort approach would.

### Notes

`zero_out_dynamic_slice` was only used within this module and is safely removed.

### DIFF
--- a/nx/lib/nx/lin_alg/cholesky.ex
+++ b/nx/lib/nx/lin_alg/cholesky.ex
@@ -59,20 +59,12 @@ defmodule Nx.LinAlg.Cholesky do
   defnp approximate_zeros(matrix, eps), do: Nx.select(Nx.abs(matrix) <= eps, 0, matrix)
 
   defnp dot_with_dynamic_slice(left, left_start, left_end, right, right_start, right_end) do
-    lhs = zero_out_dynamic_slice(left, left_start, left_end)
-    rhs = zero_out_dynamic_slice(right, right_start, right_end)
+    n = Nx.axis_size(left, 0)
+    idx = Nx.iota({n})
+    left_mask = Nx.logical_and(idx >= left_start, idx < left_end)
+    right_mask = Nx.logical_and(idx >= right_start, idx < right_end)
 
-    Nx.dot(lhs, rhs)
-  end
-
-  defnp zero_out_dynamic_slice(t, start_idx, end_idx) do
-    # assumes t has rank 1
-    zero_out_selector = Nx.logical_or(Nx.iota(t.shape) < start_idx, Nx.iota(t.shape) >= end_idx)
-
-    Nx.take(
-      Nx.select(zero_out_selector, 0, t),
-      Nx.argsort(zero_out_selector, direction: :asc, stable: true)
-    )
+    Nx.dot(Nx.select(left_mask, left, 0), Nx.select(right_mask, right, 0))
   end
 
   defn cholesky_grad(l, _input, g) do


polvalente:
Closing this in favor of #1782 because I couldn't push the fix to your branch

(to be clear, you will receive proper commit attribution on merge)

polvalente:
@Ljzn just merged into main!

PR #1782, opened 2026-07-19
perf(cholesky): replace zero_out_dynamic_slice with specific dot solution

polvalente:
Supersedes #1781 to incorporate an iterative improvement on the proposed refactor


polvalente:
Squash-merged locally as bdea65ed with author zhanglinjie <zhanglinjie@pxn.one> (gh --author-email cannot set another user's email).

PR #1784, opened 2026-07-20
perf(qr): replace explicit Householder matrix with implicit rank-1 update, ~4-9x speedup

Ljzn:
### Problem

The QR decomposition in `Nx.LinAlg.QR` constructs the full Householder reflector matrix `H`
explicitly at each iteration of the while loop:

```elixir
selector = Nx.iota({Nx.size(x)}) |> Nx.greater_equal(i) |> then(&Nx.outer(&1, &1))
eye = Nx.eye(Nx.size(x))
Nx.select(selector, eye - scale * Nx.outer(v, v), eye)
```

This builds an `m×m` matrix (7 intermediate tensors per call), then applies it with
two `Nx.dot(H, r)` and `Nx.dot(q, H)` operations — each O(n³) per iteration.
For an `n×n` matrix with `n` iterations, the total complexity is **O(n⁴)**.

### Solution

Following the approach used by LAPACK's `DLARF` / `ZLARF` routines, replace the explicit
Householder matrix construction with an implicit rank-1 update. Since

```
H·r = (I - β·v·vᴴ)·r = r - β·v·(vᴴ·r)
q·H = q·(I - β·v·vᴴ) = q - β·(q·v)·vᴴ
```

each iteration becomes O(n²) instead of O(n³), reducing the overall QR complexity
from **O(n⁴) to O(n³)**:

```elixir
{v, scale} = householder_reflector(x, i, eps)

vh_r = Nx.dot(Nx.LinAlg.adjoint(v), r)
r = r - scale * Nx.dot(Nx.new_axis(v, 1), Nx.new_axis(vh_r, 0))

q_v = Nx.dot(q, [1], v, [0])
q = q - scale * Nx.outer(q_v, v)
```

Note the asymmetry between the two updates: for `H·r`, Nx.outer(a, b) computes
`a[i]·conj(b[j])` (conjugate transpose), so we must avoid the built-in conjugation
by explicitly computing `v·(vᴴ·r)` via `Nx.dot` on reshaped vectors. For `q·H`,
the outer product `(q·v)·vᴴ` naturally matches Nx.outer's conjugation, so we can
use it directly.

### Handling Degenerate Columns (NaN Guard)

The original code's explicit H construction contained a `Nx.select(selector, ..., eye)`
that isolated NaN from degenerate columns (all-zero columns after row zeroing in
wide-matrix padding) to only affect the padded submatrix, which was later sliced away.

In the implicit form, there is no selector to contain NaN. Following LAPACK's
convention (`IF( TAU.NE.ZERO )` in both `DLARF` and `ZLARF`), we return `scale = 0`
when the column norm is effectively zero, making the update a no-op:

```elixir
# Complex branch (the real branch already had this guard via norm_selector)
{:c, _} ->
  ...
  {n_u, n_u_sq} = norm(u)
  norm_selector = Nx.less(Nx.real(n_u_sq), 1.0e-30)
  {u / Nx.select(norm_selector, 1, n_u), Nx.select(norm_selector, 0, 2)}
```

This ensures `v = 0` and `scale = 0` for degenerate columns, so `r - 0·... = r`
and `q - 0·... = q` without NaN.

### Benchmark

Full `Nx.LinAlg.qr/2` on random 32-bit float square matrices (BinaryBackend),
averaged over 3 runs per size:

| 30 | 260 | 69 | **3.8x** |
| 50 | 1772 | 296 | **6.0x** |
| 80 | 11021 | 1238 | **8.9x** |

The speedup increases with matrix size, consistent with the O(n⁴) → O(n³)
improvement.

Reconstruction error `Q·R ≈ A` is identical between both versions (~1.3-2.2e-7).

<details>

```elixir
defmodule Bench do
  def run do
    sizes = [30, 50, 80]
    iters = 3
    IO.puts("n\\ttime_ms\\terror")
    for n <- sizes do
      key = Nx.Random.key(42)
        {t, _} = :timer.tc(fn -> Nx.LinAlg.qr(a) end); t
      end)
      {q, r} = Nx.LinAlg.qr(a)
      err = Nx.subtract(a, Nx.dot(q, r)) |> Nx.abs() |> Nx.mean() |> Nx.to_number()
      IO.puts("#{n}\\t#{Float.round(avg_ms, 2)}\\t#{err}")
    end
  end
end
Bench.run()
```

</details>

### Correctness

- All 2707 existing tests pass (`mix test`).
- The rank-1 update formula is mathematically equivalent to the explicit H matrix
  (`H·r = r - β·v·(vᴴ·r)` is an identity; see LAPACK Working Note 14).
- SVD, EIGH, and PINV (which internally use QR via QDWH) continue to work correctly.
- SVD zero-singular-value columns (which are not uniquely determined) no longer
  cause spurious test failures — the test now only checks columns corresponding
  to non-zero singular values.

### Test Updates

Three numeric-precision tests were updated to accommodate the different floating-point
rounding between the two mathematically equivalent algorithms:

1. **QR doctest** (`lin_alg.ex`): `r[2][2]` changed from `1.4142137` to `1.4142135`
   (both are equally valid `sqrt(2.0)` approximations within f32 precision).
2. **SVD U matrix test** (`lin_alg_test.exs`): Only the first `rank` columns of U
   are checked; columns corresponding to zero singular values span the null space
   and are not uniquely determined.
3. **Defn evaluator QR test** (`evaluator_test.exs`): Changed from exact `==`
   comparison to `assert_all_close(atol: 1e-6)` to allow valid f32 precision
   differences.

---


polvalente:
Thanks for another great improvement! I tested this locally with EXLA and the current defn implementation is basically on-par with the custom call EXLA provides, which is a great sign for other backends which might rely on our custom impl (or EXLA on cuda)

polvalente (approved):

PR #1786, opened 2026-07-20
perf(lu): replace invert with triangular_solve in gradient, ~4x speedup

Ljzn:
### Problem

The LU decomposition gradient (`lu_grad/2` in `lu.ex`) computes the inverse of the
triangular factors `L` and `U` using `Nx.LinAlg.invert/1`:

```elixir
lt_inv = Nx.LinAlg.invert(l_h)   # l_h = adjoint(L), upper triangular
ut_inv = Nx.LinAlg.invert(u_h)   # u_h = adjoint(U), lower triangular
```

`Nx.LinAlg.invert/1` is a general-purpose routine that computes:
1. The determinant (O(n³))
2. A general solve via LU decomposition (O(n³))
3. Row scaling for numerical stability

For triangular matrices, all three steps are unnecessary — the inverse can be
computed directly via forward/backward substitution.

### Solution

Replace `invert` with `triangular_solve(A, I)`, which solves `A·X = I` using
a single triangular solve (O(n²) per column, O(n³) total). This is equivalent
to LAPACK's `DTRTRI`/`ZTRTRI` routines, which are the standard approach for
inverting triangular matrices:

```elixir
eye = Nx.eye(Nx.axis_size(l_h, -1), type: Nx.type(l_h))
eye = Nx.broadcast(eye, Nx.shape(l_h))
lt_inv = Nx.LinAlg.triangular_solve(l_h, eye, lower: false)
ut_inv = Nx.LinAlg.triangular_solve(u_h, eye, lower: true)
```

The `lower: false` / `lower: true` options match the triangular structure:
`l_h = adjoint(L)` is upper triangular (adjoint of a lower triangular matrix),
and `u_h = adjoint(U)` is lower triangular.

### Benchmark

Gradient of `sum(L) + sum(U)` w.r.t. input, measured on random 32-bit float
square matrices (BinaryBackend), averaged over 3 runs:

| 32 | 591 | 122 | **4.8x** |
| 64 | 3683 | 969 | **3.8x** |
| 128 | 32528 | 8587 | **3.8x** |

Since this optimization affects the gradient path, it benefits any operation
that differentiates through LU, including `Nx.LinAlg.solve`, `Nx.LinAlg.determinant`,
and `Nx.LinAlg.invert` themselves.

<details>

```elixir
defmodule Bench do
  import Nx.Defn

  defn lu_sum(t), do: Nx.LinAlg.lu(t) |> then(fn {_, l, u} -> Nx.sum(l) + Nx.sum(u) end)

  def run do
    sizes = [32, 64, 128]
    iters = 3
    IO.puts("n\\ttime_ms")
    for n <- sizes do
      key = Nx.Random.key(42)
        {t, _} = :timer.tc(fn -> grad(a, &lu_sum/1) end)
        t
      end)
      IO.puts("#{n}\\t#{Float.round(avg_ms, 2)}")
    end
  end
end

Bench.run()
```

</details>

### Correctness

- All 2712 existing tests pass (`mix test`).
- `triangular_solve(T, I)` and `invert(T)` compute the same mathematical result
  (`inv(T)`) for any invertible triangular matrix T.
- The test suite includes gradient checks for LU, solve, and determinant with
  batched tensors.

---

### DIFF
--- a/nx/lib/nx/lin_alg/lu.ex
+++ b/nx/lib/nx/lin_alg/lu.ex
@@ -164,8 +164,9 @@ defmodule Nx.LinAlg.LU do
     lh_dl = Nx.dot(l_h, [-1], batch_axes, dl, [-2], batch_axes)
     du_uh = Nx.dot(du, [-1], batch_axes, u_h, [-2], batch_axes)
 
-    lt_inv = Nx.LinAlg.invert(l_h)
-    ut_inv = Nx.LinAlg.invert(u_h)
+    eye = Nx.eye(Nx.shape(l_h), type: Nx.type(l_h))
+    lt_inv = Nx.LinAlg.triangular_solve(l_h, eye, lower: false)
+    ut_inv = Nx.LinAlg.triangular_solve(u_h, eye, lower: true)
 
     df = lh_dl |> Nx.tril(k: -1) |> Nx.add(Nx.triu(du_uh))
 


nx/lib/nx/lin_alg/lu.ex
```diff
@@ -164,8 +164,10 @@ defmodule Nx.LinAlg.LU do
     lh_dl = Nx.dot(l_h, [-1], batch_axes, dl, [-2], batch_axes)
     du_uh = Nx.dot(du, [-1], batch_axes, u_h, [-2], batch_axes)
 
-    lt_inv = Nx.LinAlg.invert(l_h)
-    ut_inv = Nx.LinAlg.invert(u_h)
+    eye = Nx.eye(Nx.axis_size(l_h, -1), type: Nx.type(l_h))
+    eye = Nx.broadcast(eye, Nx.shape(l_h))
+    lt_inv = Nx.LinAlg.triangular_solve(l_h, eye, lower: false)
+    ut_inv = Nx.LinAlg.triangular_solve(u_h, eye, lower: true)
```
polvalente: are L and U always guaranteed to be of the same shape?

Can't we avoid the broadcast by passing the shape of l_h directly to the Nx.eye function?

Ljzn: > are L and U always guaranteed to be of the same shape?

Yes — `Nx.LinAlg.lu/1` only accepts square matrices, so L and U are both n×n,
and their adjoints (l_h, u_h) share the same shape as well.

polvalente (approved):

PR #1787, opened 2026-07-20
perf(binary_backend): replace element-wise O(n) scan with zip-reduce, ~3-4x speedup

Ljzn:
### Problem

`element_wise_bin_op/4` in `Nx.BinaryBackend` handles all tensor–tensor element-wise
operations (add, multiply, divide, pow, etc. — 22 operators). Its inner loop scans
left and right binaries by computing `i * size` and constructing a sub-binary from
the start on each iteration:

```elixir
for i <- 0..(count - 1), into: <<>> do
  left_consumed = i * left_size
  <<_::size(^left_consumed)-bitstring, match!(x, 0), _::bitstring>> = left_data
  # ... same for right ...
end
```

Each iteration creates two sub-binaries (left + right) from scratch — O(2n) binary
allocations. For a 1M-element tensor, this means 2M sub-binary creations and
reference-count updates.

### Solution

Replace the range-based loop with a `for` comprehension generator over the left
binary (BEAM compiles this to a tight cursor-advancing loop with zero allocations)
while zipping the right binary through `reduce`:

```elixir
match_types [left_type, right_type, type] do
  {data, _} =
    for <<match!(x, 0) <- left_data>>, reduce: {<<>>, right_data} do
      {acc, <<match!(y, 1), right_rest::bitstring>>} ->
        {<<acc::bitstring, write!(fun.(type, read!(x, 0), read!(y, 1)), 2)>>, right_rest}
    end

  data
end
```

This eliminates:
- `i * left_size` / `i * right_size` integer multiplications per element
- Left sub-binary creation (comprehension generator uses a cursor, not sub-binaries)
- Right sub-binary creation is reduced to one incremental extraction per element
  (the `right_rest` in the accumulator)

No intermediate lists or additional memory is allocated.

### Benchmark

`Nx.add(a, b)` on random 32-bit float vectors, averaged over 10 runs per size:

| 1,000 | 0.486 | 0.179 | **2.7x** |
| 100,000 | 68.1 | 16.4 | **4.2x** |
| 1,000,000 | 932 | 212 | **4.4x** |

The speedup grows with tensor size as integer multiplication and sub-binary
allocation overhead accumulates linearly in the original loop.

<details>

```elixir
defmodule Bench do
  def run do
    sizes = [1000, 100000, 1000000]
    iters = 10
    IO.puts("n\\ttime_ms")
    for n <- sizes do
      a = Nx.iota({n}, type: {:f, 32})
      b = Nx.iota({n}, type: {:f, 32})
        {t, _} = :timer.tc(fn -> Nx.add(a, b) end); t
      end)
      IO.puts("#{n}\\t#{Float.round(avg_ms, 3)}")
    end
  end
end
Bench.run()
```

</details>

### Correctness

- All 2712 existing tests pass (`mix test`).
- The change only affects `element_wise_bin_op/4`, which is the tensor–tensor
  path (not scalar–tensor, which already uses efficient `binary_to_binary`).
- The `match_types` macro expansion is unchanged; only the loop construct around
  it was replaced.

---

### DIFF
--- a/nx/lib/nx/binary_backend.ex
+++ b/nx/lib/nx/binary_backend.ex
@@ -689,26 +689,24 @@ defmodule Nx.BinaryBackend do
   end
 
   defp element_wise_bin_op(%{shape: shape, type: type} = out, left, right, fun) do
-    %T{type: {_, left_size} = left_type} = left
-    %T{type: {_, right_size} = right_type} = right
+    %T{type: left_type} = left
+    %T{type: right_type} = right
 
-    count = Nx.size(shape)
     left_data = broadcast_data(left, shape)
     right_data = broadcast_data(right, shape)
 
     data =
       match_types [left_type, right_type, type] do
-        for i <- 0..(count - 1), into: <<>> do
-          left_consumed = i * left_size
-          <<_::size(^left_consumed)-bitstring, match!(x, 0), _::bitstring>> = left_data
-          x = read!(x, 0)
+        {data, _} =
+          for <<match!(x, 0) <- left_data>>, reduce: {<<>>, right_data} do
+            {acc, <<match!(y, 1), right_rest::bitstring>>} ->
+              x_val = read!(x, 0)
+              y_val = read!(y, 1)
+              result = <<write!(fun.(type, x_val, y_val), 2)>>
+              {<<acc::bitstring, result::binary>>, right_rest}
+          end
 
-          right_consumed = i * right_size
-          <<_::size(^right_consumed)-bitstring, match!(y, 1), _::bitstring>> = right_data
-          y = read!(y, 1)
-
-          <<write!(fun.(type, x, y), 2)>>
-        end
+        data
       end
 
     from_binary(out, data)


nx/lib/nx/binary_backend.ex
```diff
@@ -689,26 +689,21 @@ defmodule Nx.BinaryBackend do
...
-          <<_::size(^left_consumed)-bitstring, match!(x, 0), _::bitstring>> = left_data
-          x = read!(x, 0)
-
-          right_consumed = i * right_size
-          <<_::size(^right_consumed)-bitstring, match!(y, 1), _::bitstring>> = right_data
-          y = read!(y, 1)
+        {data, _} =
+          for <<match!(x, 0) <- left_data>>, reduce: {<<>>, right_data} do
+            {acc, <<match!(y, 1), right_rest::bitstring>>} ->
+              {<<acc::bitstring, write!(fun.(type, read!(x, 0), read!(y, 1)), 2)>>, right_rest}
```
polvalente: let's split this line a little bit for readability?

Ljzn: Good point! How about this?
```ex
{data, _} =
  for <<match!(x, 0) <- left_data>>, reduce: {<<>>, right_data} do
    {acc, <<match!(y, 1), right_rest::bitstring>>} ->
      result = <<write!(fun.(type, read!(x, 0), read!(y, 1)), 2)>>
      {<<acc::bitstring, result::binary>>, right_rest}
  end
```

polvalente: I think it would be better to also extract the reads back out into variables

polvalente (approved):

PR #1788, opened 2026-07-20
Raw data pointer and dep path

m0rt3nlund:
Added function to get raw device data pointer to be able to pass it to other libraries
Also added ability to specify the location of the dependencies


polvalente (changes requested):
Let's implement Nx.to_pointer and from_pointer properly instead

polvalente:
closing in favor of #1796

PR #1795, opened 2026-07-20
Lift buffer donation to Nx.Defn

Chapaman:
Builds on @seanmor5's EXLA buffer-donation work in #1755 (merged that branch onto current `main`).


josevalim:
@Chapaman @polvalente if we are planning to go ahead with this, we should discuss a better API for it, because the argnum approach is not very intuitive for complex data structures once we have maps, structs, etc. Perhaps some sort of `defn` annotation?

polvalente:
This is what I was thinking. Defn annotation would be great for defn graph too. The one thing Handoff is missing to deploy Axon accross a network is args annotated as parameters.

The correct abstraction could solve both problems. But I'm prioritizing this as the latter is solvable by other means.

josevalim:
I believe we should at least spend a bit of time discussing some options. We have been working hard to stabilize the API for v1.0. Rushing and adding an API that we are aware to be inferior just before v1.0 goes counter to that effort. Some options I have in mind:

1. Introduce `EXLA.donatable` or similar to mark the tensor as donatable before calling the backend. We then store a field in the EXLA.Backend struct. This function can use Nx.Container to traverse the elements. We could make this part of Nx itself as well (then it is just a new backend callback).

2. Introduce a function to Nx.Defn.Kernel that marks some AST nodes as donatable and traverse them. This seems to be hard.

nx/lib/nx/defn/compiler.ex
```diff
@@ -868,51 +868,64 @@ defmodule Nx.Defn.Compiler do
...
+              template, fun, {acc_templates, acc_funs, i, donated} ->
+                donated =
+                  if template.donatable do
+                    MapSet.put(donated, i)
+                  else
+                    donated
+                  end
+
+                # Strip so donatable does not leak through template copies / expr params.
+                template = %{template | donatable: false}
```
polvalente: I think we're stripping information from compilers too early, and this ends up just becoming available again at runtime when the lazy params get unwrapped. I believe certain compilers (perhaps a future OpenCL one, or even the EMLX compiler) might be able to leverage this information at compile-time. For Torchx runtime is fine.

Keeping this would also allow us to not accumulate `donated`, and therefore not have to add new info to the `:options`, which can break some caching strategies.

nx/lib/nx/defn/compiler.ex
```diff
@@ -868,28 +868,29 @@ defmodule Nx.Defn.Compiler do
...
     {fun, params, templates, [first_flatten | rest_flattens]}
   end
 
   @doc false
-  def to_lazy_params(fun, args) do
+  def to_lazy_params(fun, args, _opts \\ []) do
     {params, cache, {templates, funs, _}} =
-      Enum.reduce(args, {[], [], {[], [], 0}}, fn
+      args
+      |> Enum.reduce({[], [], {[], [], 0}}, fn
```
polvalente: we can revert this file completely!

polvalente (approved):

nx/lib/nx/defn/compiler.ex
josevalim: The changes to this file do not seem necessary?

nx/lib/nx/defn.ex
```diff
@@ -316,11 +316,21 @@ defmodule Nx.Defn do
 
     * `:hooks` - a map of callbacks to override named io_calls. See `Nx.Defn.Kernel.io_call/3`.
 
+  Buffer donation is baked from the templates at compile time: pass
+  `Nx.donate/1` on the templates (or tensors used as templates) that should
+  be donated. Marking live arguments with `Nx.donate/1` only when invoking
+  the compiled function does not enable donation if the template was not
+  donatable; omitting `donate/1` on invoke also does not disable donation
+  already baked from templates. See `Nx.donate/1`.
```
josevalim: Can we please not have public docs generated by agents? :)

nx/lib/nx/defn.ex
```diff
@@ -883,7 +900,10 @@ defmodule Nx.Defn do
 
   defp do_shard_jit_apply(fun, mesh, args_list, opts) do
     opts = prepare_options(opts)
-    {fun, params, _templates, args_list} = Nx.Defn.Compiler.to_lazy_params_sharded(fun, args_list)
+
+    {fun, params, _templates, args_list} =
+      Nx.Defn.Compiler.to_lazy_params_sharded(fun, args_list, opts)
+
```
josevalim: Many of the changes to this file do not seem to be necessary either?

nx/lib/nx/tensor.ex
```diff
@@ -235,16 +243,23 @@ defmodule Nx.Tensor do
           ])
         end
 
+      donatable =
+        if tensor.donatable do
+          concat([line(), "donatable: true"])
```
josevalim: Instead of a newline, maybe we show just ` [donatable]` after the shape?

nx/lib/nx.ex
```diff
@@ -1028,6 +1028,69 @@ defmodule Nx do
...
+      iex> t.donatable
+      true
+
+      iex> %{a: a, b: b} = Nx.donate(%{a: Nx.tensor(1), b: Nx.tensor(2)})
+      iex> {a.donatable, b.donatable}
+      {true, true}
+
+  """
+  @doc type: :conversion
+  def donate(tensor_or_container) do
```
josevalim: Not sure I like the name `donate` in active tense because we are really not donating at this moment.

nx/lib/nx/tensor.ex
```diff
@@ -27,11 +29,17 @@ defmodule Nx.Tensor do
   @type axes :: [axis]
   @type name :: atom
 
-  @type t :: %Nx.Tensor{data: data, type: type, shape: shape, names: [name]}
-  @type t(data) :: %Nx.Tensor{data: data, type: type, shape: shape, names: [name]}
+  @type t :: %Nx.Tensor{data: data, type: type, shape: shape, names: [name], donatable: boolean}
```
josevalim: Should this be `dontable?`?

```suggestion
  @type t :: %Nx.Tensor{data: data, type: type, shape: shape, names: [name], donatable?: boolean}
```

nx/lib/nx.ex
```diff
@@ -1028,6 +1028,69 @@ defmodule Nx do
...
+  end
+
+  @doc """
+  Returns `true` if `tensor` is marked as donatable.
+
+  See `donate/1`.
+  """
+  @doc type: :conversion
+  def donatable?(%T{donatable: true}), do: true
+  def donatable?(%T{}), do: false
```
josevalim:
```suggestion
  def donatable?(t) when is_tensor(t), do: false
```

Please also add a test for it testing integers or complex numbers as arguments.

nx/lib/nx.ex
```diff
@@ -8469,7 +8548,13 @@ defmodule Nx do
...
 
     impl!(target, indices, updates)
-    |> apply(op, [%{target | type: type}, target, indices, updates, [axes: axes]])
+    |> apply(op, [
+      %{target | type: type},
+      target,
+      indices,
+      updates,
+      [axes: axes]
+    ])
```
josevalim: This change and a bunch of changes below and above seem to be unrelated!

Chapaman:
@josevalim addressed your review comments:

- Reverted `nx/lib/nx/defn/compiler.ex` and the unnecessary `Nx.Defn` opts plumbing / agent-generated docs
- Reverted the unrelated formatter churn in `nx.ex`
- Renamed the marker to `Nx.donatable/1` (your earlier suggestion of `*.donatable`) and the field to `donatable?`
- `donatable?/1` now uses `when is_tensor(t)` and has tests for integers / complex
- Inspect now shows ` [donatable]` after the shape

Happy to rename `donatable/1` further if you prefer something like `mark_donatable/1`.

polvalente (approved):

exla/lib/exla/defn.ex
```diff
@@ -562,6 +585,57 @@ defmodule EXLA.Defn do
...
+        user_args_with_leaf_index
+        |> Enum.with_index(fn {leaf_idx, ts}, k -> {leaf_idx, {k, ts}} end)
+        |> Map.new()
+
+      out_with_index = Enum.with_index(out_typespecs)
+
+      {pairs, _used_outs} =
+        donated_leaf_set
+        |> Enum.sort()
+        |> Enum.reduce({[], MapSet.new()}, fn leaf_idx, {pairs, used_outs} ->
```
josevalim, line 605: You can use a map_reduce then instead of a reduce+reverse.

exla/lib/exla.ex
```diff
@@ -303,6 +303,12 @@ defmodule EXLA do
       this if the input tensors are allocated on host and the computation is
       running on GPU/TPU with a limited amount of memory**
 
+  ## Buffer donation
+
+  Mark inputs with `Nx.donatable/1` (or pass donatable templates to `compile/3`).
+  EXLA reads `donatable?` on argument tensors and aliases donated inputs onto
+  matching outputs. Not supported with sharded execution. See `Nx.donatable/1`.
```
josevalim, line 310: Can we phrase this better? It is a bit unclear? I'm generally not a fan of AI written docs. The lack of subject doesn't make it clear either: "Mark inputs with `Nx.donatable/1`" the user or the function marks them?

josevalim (approved):
Two nitpicks and ship it!!!!!!!!!!

exla/lib/exla.ex
Chapaman: Altered the text to specify that the user should mark them and also made a bit more clear on what happens to the marked tensor

exla/lib/exla/defn.ex
```diff
@@ -427,8 +445,13 @@ defmodule EXLA.Defn do
                 end)
               end
 
+              for {arg_index, output_index} <- alias_pairs do
+                Function.set_arg_aliasing(builder, arg_index, output_index)
+              end
+
```
polvalente, line 451: There is a slight bug in behavior, where if we just compile a function with donatable templates, we don't validate the input tensors afterwards and end up donating any non-donatable buffer:

```elixir
args = [Nx.iota({3, 3}, backend: EXLA.Backend)]
# note that args is not donatable, but template_args below is
template_args = args |> Enum.map(fn t -> Nx.to_template(t) |> Nx.donatable() end)

f = Nx.Defn.compile(fn x -> Nx.add(x, 1) end, template_args, compiler: EXLA)

result = apply(f, args) 

Enum.at(args, 0)
```

polvalente: we should also send a follow-up PR that validates that donation works with sharding or raises and is documented accordingly

nx/lib/nx/tensor.ex
```diff
@@ -237,14 +245,21 @@ defmodule Nx.Tensor do
...
 
+      donatable =
+        if tensor.donatable? do
+          " [donatable]"
+        else
+          empty()
+        end
+
       # Devectorizing helps validate results
       data = tensor.data.__struct__.inspect(Nx.devectorize(tensor), opts)
```
polvalente, line 256:
```
#Inspect.Error<
  got RuntimeError with message:

      """
      ToLiteral() called on deleted or donated buffer: INVALID_ARGUMENT: Buffer has been deleted or donated.
      """
```

We might want EXLA.Backend.inspect to check the flag to return a proper error if the tensor has been deleted instead of surfacing the exception during inspection.

polvalente: I think something along the lines of

```elixir
#Nx.Tensor<
  s32[3][3] [donatable]

  Deallocated due to buffer donation
>
```

polvalente (approved):

PR #1800, opened 2026-08-06
docs: grammar in put_slice

greetingsfellowhumans:


nx/lib/nx.ex
```diff
@@ -14418,8 +14418,8 @@ defmodule Nx do
 
   ## Vectorized tensors
 
-  The both tensor to be sliced and the slices can be vectorized,
-  but indices must be non-vectorized.
+  Both the tensor and the slices can be vectorized, but the indices
+  must remain non-vectorized.
```
```suggestion
  must be non-vectorized.
```

Remain I think implies the possibility of transformation instead of the characteristic of the tensors at the time of the call

polvalente (approved):

PR #1804, opened 2026-08-11
fix: return the modified tensor expression from the cache computation for :hooks

polvalente:


polvalente:
this branch was created from v0.13.0 so that we can release on 0.13.1

Issue #1806, opened 2026-08-12
Gradients through Nx.LinAlg decompositions fail unless every output is consumed (svd regression on main)


cash-mckeeman:
### Summary

Gradients through `Nx.LinAlg` decompositions only work when **every** output of the decomposition is consumed. Consuming a subset raises.

On `main` this currently includes `svd`, which is a regression introduced after 0.13.0 — see the bisect below.

### Reproduction

```elixir
a = Nx.tensor([[4.0, 1.0, 0.5], [1.0, 3.0, 0.2], [0.5, 0.2, 2.0]])

# consuming one output — raises
Nx.Defn.grad(fn x -> {_q, r} = Nx.LinAlg.qr(x); Nx.sum(r) end).(a)
Nx.Defn.grad(fn x -> {_p, l, _u} = Nx.LinAlg.lu(x); Nx.sum(l) end).(a)
Nx.Defn.grad(fn x -> {_u, s, _vt} = Nx.LinAlg.svd(x); Nx.sum(s) end).(a)

# consuming all outputs — fine
Nx.Defn.grad(fn x -> {q, r} = Nx.LinAlg.qr(x); Nx.sum(q) + Nx.sum(r) end).(a)
```

```
** (ArgumentError) given axis (-2) invalid for shape with rank 0
```

`eigh` fails the same way on released versions with `cannot reshape, current shape {} is not compatible with new shape {1, 3, 3}`.

### Current state

| grad, one output consumed | Hex 0.13.1 | `main` (08a1ab2) |
|---|---|---|
| `eigh` (w or v) | FAIL | OK |
| `svd` (s only) | OK | **FAIL** |
| `qr` (q or r) | FAIL | FAIL |
| `lu` (l only) | FAIL | FAIL |
| `cholesky` | OK | OK |

### Bisect

```
bdea65e (parent)                         svd s-only OK    eigh w-only FAIL
2eabb4f  fix: batched LinAlg grads        svd s-only FAIL  eigh w-only OK
         for svd and eigh (#1783)
```

#1783 rewrote `svd_grad` to be batch-aware, replacing concrete-shape operations with `Nx.axis_size(t, -2)` and batched `Nx.dot`. It also gave `eigh` an analytic VJP plus a `broadcast_grad(g, template)` helper that reshapes a malformed cotangent before use.

The net effect was to fix `eigh`'s partial-output gradient and break `svd`'s in the same commit.

### Cause

[...]

PR #1807, opened 2026-08-12
fix(defn): accept constant tensor arguments in Nx.block

### LINKED ISSUE #1805: Nx.block raises when given a constant tensor argument (breaks Nx.LinAlg.solve/2 under jit)
### Summary

`Nx.block/4` raises when any of its arguments is a plain constant tensor rather than a traced expression. `Nx.LinAlg.solve/2` is the only multi-argument block op today, so it is where this surfaces: solving against a closed-over right-hand side fails inside `jit`.

This is **not** gradient-specific — it reproduces under plain `Nx.Defn.jit/1`.

### Reproduction

```elixir
a = Nx.tensor([[4.0, 1.0, 0.5], [1.0, 3.0, 0.2], [0.5, 0.2, 2.0]])
b = Nx.tensor([1.0, 2.0, 3.0])

Nx.LinAlg.solve(a, b)                                     # OK
Nx.Defn.jit(fn x, y -> Nx.LinAlg.solve(x, y) end).(a, b)  # OK
Nx.Defn.jit(fn x -> Nx.LinAlg.solve(x, b) end).(a)        # raises
Nx.Defn.grad(fn x -> Nx.sum(Nx.LinAlg.solve(x, b)) end).(a)  # raises
```

```
** (FunctionClauseError) no function clause matching in Nx.Defn.Expr.parameter/2

    The following arguments were given to Nx.Defn.Expr.parameter/2:

        # 1
        #Nx.Tensor<
          f32[3]
          [1.0, 2.0, 3.0]
        >

        # 2
        1

    (nx 0.13.1) lib/nx/defn/expr.ex:72: Nx.Defn.Expr.parameter/2
    (elixir 1.20.2) lib/enum.ex:5079: Enum.with_index_list/3
    (nx 0.13.1) lib/nx/defn/expr.ex:433: Nx.Defn.Expr.expr_block/3
    (nx 0.13.1) lib/nx/lin_alg.ex:717: Nx.LinAlg.solve/2
```

Deriving the right-hand side from the traced input instead (`Nx.LinAlg.solve(x, Nx.sum(x, axes: [0]))`) works, which is consistent with the diagnosis below.

### Cause

`expr_block/3` builds one parameter per argument:

```elixir
params = Enum.with_index(args, &parameter/2)
```

`Expr.parameter/2` matches only `%T{data: %Expr{context: context}}` — it derives the tracing context from the argument itself, so a constant tensor has no matching clause.

Worth noting for whoever picks this up: making `parameter/2` accept the constant is not sufficient on its own. The block node still stores the raw tensor in `in_args`, and the failure then moves to `Nx.Defn.Tree.scope_ids_each/3`, which has no clause for a non-`Expr` tensor either. Both the parameter context and the stored `in_args` need to hold expressions.

### Affected versions

Reproduces on `main` (08a1ab2) and on released 0.13.0 and 0.13.1.

### Test coverage

There are currently no tests exercising `Nx.block/4` anywhere in the repo, and every existing `solve` grad test in `test/nx/defn/grad_test.exs` passes the right-hand side as a `defn` parameter, so the constant path is never reached.

I have a fix and regression tests ready and am happy to open a PR if you'd like it.

cash-mckeeman:
Closes #1805.

`Nx.block/4` raised whenever one of its arguments was a plain constant tensor rather than a traced expression. `Nx.LinAlg.solve/2` is the only multi-argument block op today, so it is where this surfaced: solving against a closed-over right-hand side failed.

This is not gradient-specific — it reproduces under plain `Nx.Defn.jit/1`:

```elixir
a = Nx.tensor([[4.0, 1.0, 0.5], [1.0, 3.0, 0.2], [0.5, 0.2, 2.0]])
b = Nx.tensor([1.0, 2.0, 3.0])

Nx.Defn.jit(fn x, y -> Nx.LinAlg.solve(x, y) end).(a, b)  # OK
Nx.Defn.jit(fn x -> Nx.LinAlg.solve(x, b) end).(a)        # ** (FunctionClauseError)
```

Present in released 0.13.0 and 0.13.1 as well as `main`.

### Cause

`expr_block/3` built one parameter per argument with `parameter/2`, which matches only `%T{data: %Expr{context: context}}` — it derives the tracing context from the argument itself, so a constant tensor had no matching clause.

### Fix

Two parts, both needed:

1. Derive a fallback context from the first traced argument (`block_context/1`), and give it *only* to arguments that carry no context of their own. Arguments that were already expressions keep their own context exactly as before, so the change is confined to the constants that previously had no clause at all. If nothing is traced, the fallback is `:root`.
2. Convert the arguments to expressions and store those in the block node's `in_args`.

Part 2 is not cosmetic. With only part 1, `parameter/2` stops raising but the block node still holds a raw tensor, and the failure moves to `Nx.Defn.Tree.scope_ids_each/3`, which likewise has no clause for a non-`Expr` tensor.

The per-argument contexts are captured before the `to_expr/1` conversion, since converting erases the distinction between a traced argument and a constant.

The `:root` fallback in `block_context/1` is defensive rather than reachable: `Nx.block/4` dispatches through `Nx.Shared.list_impl!(args)`, so an all-constant argument list resolves to `BinaryBackend` and never reaches `expr_block/3`. Happy to drop it if you would rather not carry unreachable branches.

### Tests

Two, both verified failing on `main` before the fix:

- `test/nx/defn_test.exs` — a new `describe "block"`, jitting a solve against a constant right-hand side. No test previously called `Nx.block/4` directly, though the linalg grad suite drives the block path throughout.
- `test/nx/defn/grad_test.exs` — in the existing `describe "solve"`, asserting the constant-rhs gradient equals the traced-rhs gradient computed by the existing `solve_grad_wrt_a/2`. Same mathematics, different tracing path, so any divergence is the defect.

### Verification

- `nx` suite: 2725 passed, 0 failures (2723 before, plus these two).
- `exla` suite: 1416 passed, 0 failures, 49 excluded.
- `mix compile --force`: no new warnings.

Independent of #1806 / its PR — different files, both branched from `main`.

### DIFF
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -427,48 +427,6 @@ defmodule Nx.Defn.Expr do
 
     out
   end
-
-  # A block's arguments may mix traced expressions with plain constant tensors
-  # (for example `Nx.LinAlg.solve(x, b)` where `b` is closed over). Constants
-  # carry no context of their own, so take the context from the first traced
-  # argument as the fallback for those constants.
-  defp block_context(args) do
-    Enum.find_value(args, :root, fn
-      %T{data: %Expr{context: context}} -> context
-      _ -> nil
-    end)
-  end
-
-  # Preserve each already-traced argument's own context exactly as before the
-  # fix; only arguments with no context of their own (constants) fall back to
-  # the block's derived context. Must run before `to_expr/1` converts the
-  # constants into expressions, since that erases the distinction.
-  defp arg_context(%T{data: %Expr{context: context}}, _fallback), do: context
-  defp arg_context(_arg, fallback), do: fallback
-
-  defp expr_block(struct, in_args, fun) do
-    {args, opts} = Enum.split_while(in_args, &(not is_list(&1)))
-    context = block_context(args)
-    contexts = Enum.map(args, &arg_context(&1, context))
-    args = Enum.map(args, &to_expr/1)
-    in_args = args ++ opts
-
-    params =
-      args
-      |> Enum.zip(contexts)
-      |> Enum.with_index(fn {arg, arg_context}, pos -> parameter(arg, arg_context, pos) end)
-
-    case apply(fun, [struct | params ++ opts]) do
-      %{data: %{context: context}} = res ->
-        expr(res, context, :block, [struct, in_args, res, fun])
-
-      t when is_tuple(t) ->
-        context = elem(t, 0).data.context
-        out = tuple_out(tuple_size(t))
-        tuple(expr(out, context, :block, [struct, in_args, t, fun]), Tuple.to_list(t))
-    end
-  end
-
   ## Nx.Defn AST callbacks
 
   @doc false
@@ -865,7 +823,21 @@ defmodule Nx.Defn.Expr do
 
   @impl true
   def block(struct, _output \\ nil, in_args, fun) do
-    expr_block(struct, in_args, fun)
+    {args, opts} = Enum.split_while(in_args, &(not is_list(&1)))
+    {args, context} = to_exprs(args)
+    context = context || :root
+    in_args = args ++ opts
+    params = Enum.with_index(args, fn arg, pos -> parameter(arg, context, pos) end)
+
+    case apply(fun, [struct | params ++ opts]) do
+      %{data: %{context: context}} = res ->
+        expr(res, context, :block, [struct, in_args, res, fun])
+
+      t when is_tuple(t) ->
+        context = elem(t, 0).data.context
+        out = tuple_out(tuple_size(t))
+        tuple(expr(out, context, :block, [struct, in_args, t, fun]), Tuple.to_list(t))
+    end
   end
 
   @impl true
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -427,6 +427,7 @@ defmodule Nx.Defn.Expr do
 
     out
   end
+
   ## Nx.Defn AST callbacks
 
   @doc false
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -428,9 +428,35 @@ defmodule Nx.Defn.Expr do
     out
   end
 
+  # A block's arguments may mix traced expressions with plain constant tensors
+  # (for example `Nx.LinAlg.solve(x, b)` where `b` is closed over). Constants
+  # carry no context of their own, so take the context from the first traced
+  # argument as the fallback for those constants.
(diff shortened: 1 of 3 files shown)


nx/lib/nx/defn/expr.ex
```diff
@@ -428,9 +428,35 @@ defmodule Nx.Defn.Expr do
...
+  # A block's arguments may mix traced expressions with plain constant tensors
+  # (for example `Nx.LinAlg.solve(x, b)` where `b` is closed over). Constants
+  # carry no context of their own, so take the context from the first traced
+  # argument as the fallback for those constants.
+  defp block_context(args) do
+    Enum.find_value(args, :root, fn
+      %T{data: %Expr{context: context}} -> context
+      _ -> nil
+    end)
+  end
```
polvalente: I think this might have a slight bug, though I'm not sure if it's actually reproducible: If we have a block inside a while using a closure from outside the while, the order in which they appear might result in a hidden invalid context:


```elixir
defn foo_lhs(a) do
  x = a + 1
  while {x, i = 0}, i < 10 do
    x_next = Nx.LinAlg.solve(a, x)
    {x, i + 1}
  end
end

defn foo_rhs(a) do
  x = a + 1
  while {x, i = 0}, i < 10 do
    x_next = Nx.LinAlg.solve(x, a)
    {x_next, i + 1}
  end
end
```

Neither of these should work, and I have a feeling that the second one will end up passing through.

nx/test/nx/defn_test.exs
```diff
@@ -931,6 +932,19 @@ defmodule Nx.DefnTest do
     end
   end
 
+  describe "block" do
+    @rhs Nx.tensor([4.0, 3.0, 2.0])
+    @mat Nx.tensor([[4.0, 1.0, 0.5], [1.0, 3.0, 0.2], [0.5, 0.2, 2.0]])
+
+    @tag compiler: Evaluator
```
polvalente: I don't think it makes sense to have these tensors as module attributes

nx/test/nx/defn/grad_test.exs
```diff
@@ -4803,6 +4803,25 @@ defmodule Nx.Defn.GradTest do
       )
     end
 
+    @constant_rhs Nx.tensor([4.0, 3.0, 2.0])
```
polvalente: We shouldn't have this as a module attribute. If you really need to reuse the value, either pass it down as a defn option or have it in the describe setup as a shared value.

nx/lib/nx/defn/expr.ex
polvalente: Both fail correctly in the branch, which is good. I am pushing a simplified implementation along with these as tests.

PR #1808, opened 2026-08-12
fix(grad): functions with tuple outputs should broadcast grads onto unused outputs

### LINKED ISSUE #1806: Gradients through Nx.LinAlg decompositions fail unless every output is consumed (svd regression on main)

### LINKED ISSUE #1809: custom_grad with tuple output: unconsumed outputs get a rank-0 cotangent
### Summary

A `custom_grad` whose primal is a tuple gets a rank-0 zero cotangent for any output the caller does not consume. If the gradient body then does anything shape-sensitive with that cotangent, it raises.

This is the same root cause as #1806, but reached through a different entry point — a bare `custom_grad`, with no `Nx.block` involved — so a fix at the block boundary does not cover it.

### Reproduction

```elixir
defmodule Adj do
  import Nx.Defn

  defn pair(x) do
    r = {Nx.multiply(x, 2.0), Nx.multiply(x, 3.0)}

    custom_grad(r, [x], fn {ga, gb} ->
      n = Nx.axis_size(ga, -2)
      [Nx.add(Nx.multiply(ga, n * 1.0), gb)]
    end)
  end
end

a = Nx.tensor([[4.0, 1.0, 0.5], [1.0, 3.0, 0.2], [0.5, 0.2, 2.0]])

Nx.Defn.grad(fn x -> {p, q} = Adj.pair(x); Nx.sum(p) + Nx.sum(q) end).(a)  # OK
Nx.Defn.grad(fn x -> {p, _} = Adj.pair(x); Nx.sum(p) end).(a)              # OK
Nx.Defn.grad(fn x -> {_, q} = Adj.pair(x); Nx.sum(q) end).(a)              # raises
```

```
** (ArgumentError) given axis (-2) invalid for shape with rank 0
```

The asymmetry is the tell. Consuming only the second output leaves `ga` unconsumed, and `ga` is the one this body inspects with `Nx.axis_size/2`. Consuming only the first leaves `gb` unconsumed, and `gb` is only ever used elementwise, where broadcasting against a rank-0 zero happens to work.

So the failure depends on whether the *unconsumed* output's cotangent is used shape-sensitively — which makes it easy to miss.

### Cause

`sum_grad([])` (`lib/nx/defn/grad.ex:366`) returns a rank-0 `Expr.tensor(0.0)` for any node that accumulated no gradient, and `recur_to_grad`'s tuple branch maps it over positions with no per-output template.

### Relationship to #1806 / #1808

#1806 is the same defect reached through `Nx.block`, and the fix I proposed in #1808 restores the invariant at the block boundary. I deliberately did not extend it here, because this is a separate entry point that deserves its own change — and because for a `:metadata`/`custom_grad` node, `ans` is a `tuple_out/1` template carrying no per-element shapes, so the shapes needed to build a correctly-shaped zero are only present in `args`. That is what makes the block clause a natural place for it and `sum_grad/1` a poor one.

Verified: this reproduction fails identically on `main` (`08a1ab27`) and with both #1807 and #1808 applied.

Worth noting `Nx.Block.LinAlg.Eigh` uses exactly this pattern — a tuple-output `custom_grad` — and needed its own `broadcast_grad/2` helper for precisely this reason. It is shielded today only because that `custom_grad` sits inside a block.

Happy to put up a PR if you have a preferred shape for the fix.

cash-mckeeman:
closes #1806
closes #1809

Gradients through `Nx.LinAlg` decompositions only worked when every output was consumed:

```elixir
a = Nx.tensor([[4.0, 1.0, 0.5], [1.0, 3.0, 0.2], [0.5, 0.2, 2.0]])

Nx.Defn.grad(fn x -> {_q, r} = Nx.LinAlg.qr(x); Nx.sum(r) end).(a)
# ** (ArgumentError) given axis (-1) invalid for shape with rank 0

Nx.Defn.grad(fn x -> {q, r} = Nx.LinAlg.qr(x); Nx.sum(q) + Nx.sum(r) end).(a)
# fine
```

| grad, one output consumed | 0.13.1 | `main` before this PR |
|---|---|---|
| `eigh` | FAIL | OK |
| `svd` | OK | **FAIL** |
| `qr` | FAIL | FAIL |
| `lu` | FAIL | FAIL |

`svd` is a regression on `main`, bisected to 2eabb4f (#1783):

```
bdea65e (parent)   svd s-only OK    eigh w-only FAIL
2eabb4f  #1783      svd s-only FAIL  eigh w-only OK
```

#1783 made `svd_grad` batch-aware — replacing concrete shapes with `Nx.axis_size(t, -2)` and batched `Nx.dot` — and separately gave `eigh` an analytic VJP plus a `broadcast_grad/2` guard. The guard made `eigh` tolerate the malformed cotangent; `svd` got the batch-aware code without it. So the commit fixed one op's partial-output gradient and broke another's.

### Cause

`sum_grad([])` returns a rank-0 `Expr.tensor(0.0)` for any node that accumulated no gradient. For an unconsumed tuple output of a block, that scalar reaches the grad body, which then indexes axes on it.

### Fix

Restore the invariant once, at the block boundary, rather than in each grad body:

> A block body receives one cotangent per output, each shaped like that output.

`update_grads(:block, …)` already zips cotangents against the block's outputs, and each output expression carries its own shape and type, so the shaping happens there. `qr`, `lu` and `svd` need no per-op guards, and `eigh`'s `broadcast_grad` becomes dead code rather than load-bearing.

This mirrors what `update_grads(:while, …)` already does a few clauses up — `gs = Enum.zip_with(gs, flatten_initial, &Nx.broadcast/2)` — so the block clause is now normalizing its cotangents the same way the `while` clause always has.

**Why not fix `sum_grad/1` itself?** Because the shapes are not available there. The obvious deeper fix is to make the tuple branch of `recur_to_grad` hand `sum_grad/1` a per-position template, but for a `:metadata`/`custom_grad` node `ans` is a `tuple_out/1` template that carries no per-element shapes — the only place the output shapes exist is `args`, which is exactly what the block clause has and `sum_grad/1` does not. Hence the boundary.

That does leave a sibling hole, which I would rather name than hide: the same rank-0 cotangent reaches a bare tuple-output `custom_grad` outside any block, and fails identically. `Nx.Block.LinAlg.Eigh` uses that pattern and is only shielded today because its `custom_grad` sits inside a block. I have not touched it here — it is a distinct entry point and deserves its own change. Happy to file it separately, or fold it in if you would prefer one PR.

I confirmed that: deleting the two `broadcast_grad` calls from `eigh_grad/3` leaves the grad suite green and the compiler reports `broadcast_grad/2 unused`. I have **not** removed it in this PR — that is your call, and it would also mean dropping the now-unused `defnp`.

I considered guarding each grad body instead, mirroring `broadcast_grad`. #1783 is the argument against it: per-op guards are exactly what got missed, and every future block op reintroduces the bug.

One thing worth your opinion: the helper's final clause passes a cotangent through unchanged when its shape differs from the output's and it is not rank 0. That branch should be unreachable — `sum_grad([])` is the only producer of a rank-0 cotangent, and every other path preserves the accumulated shape, which already matches the primal output. I left it as a permissive no-op rather than raising, but if you would rather it raise so a future violation surfaces loudly instead of silently, say so and I will change it.

### Tests

Seven, in the existing `describe` blocks. Five fail on `main` before the fix; the two `eigh` ones already pass and are guard-rails against this change regressing what `broadcast_grad` currently covers.

The main oracle is linearity of the cotangent — partial-output gradients must sum to the all-outputs gradient, which the existing tests already cover. That is precisely the invariant the defect breaks, and it covers every output position. Where a closed form exists it is asserted directly: `∇sum(σ) = UVᵀ` and `∇sum(λ) = I`. Plus a batching check.

Every existing `qr`/`lu`/`svd` grad test destructures all outputs, which is why this shipped — `eigh` had the suite's only partial-output test, added by #1783 itself.

### Verification

- `nx` suite: 2730 passed, 0 failures (2723 before, plus these seven).
- `exla` suite: 1416 passed, 0 failures, 49 excluded.
- `mix compile --force`: no new warnings.

Independent of #1807 — different files, both branched from `main`.

### DIFF
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -421,6 +421,7 @@ defmodule Nx.Defn.Grad do
 
     {grads, []} =
       Composite.reduce(expr, {grads, gs}, fn child, {grads, [g | gs]} ->
+        g = match_block_cotangent(g, child)
         {Map.update(grads, child.data.id, [g], &[g | &1]), gs}
       end)
 
@@ -618,6 +619,18 @@ defmodule Nx.Defn.Grad do
       g = clear_axis_names(g)
       Map.update(grads, child.data.id, [g], &[g | &1])
     end)
+  end
+
+  # A block body receives one cotangent per output, each shaped like that
+  # output. An output nothing depends on accumulates no gradient, so
+  # `sum_grad/1` hands back the rank-0 zero, which block bodies cannot index
+  # or batch over. Restore the invariant here, once, rather than in each body.
+  defp match_block_cotangent(g, out) do
+    case {Nx.shape(g), Nx.shape(out)} do
+      {shape, shape} -> g
+      {{}, _} -> Nx.broadcast(Nx.as_type(g, Nx.type(out)), out)
+      {_, _} -> g
+    end
   end
 
   defp select_composite(pred, left, right) do
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -401,6 +401,7 @@ defmodule Nx.Defn.Grad do
 
           _ ->
             g = gs |> Tuple.to_list() |> Enum.map(&sum_grad/1)
+            g = broadcast_tuple_grads(op, args, g)
             {nodes, update_grads(op, args, ans, g, to_grad_ids, grads)}
         end
 
@@ -421,7 +422,6 @@ defmodule Nx.Defn.Grad do
 
     {grads, []} =
       Composite.reduce(expr, {grads, gs}, fn child, {grads, [g | gs]} ->
-        g = match_block_cotangent(g, child)
         {Map.update(grads, child.data.id, [g], &[g | &1]), gs}
       end)
 
@@ -621,17 +621,25 @@ defmodule Nx.Defn.Grad do
     end)
   end
 
-  # A block body receives one cotangent per output, each shaped like that
-  # output. An output nothing depends on accumulates no gradient, so
-  # `sum_grad/1` hands back the rank-0 zero, which block bodies cannot index
-  # or batch over. Restore the invariant here, once, rather than in each body.
-  defp match_block_cotangent(g, out) do
-    case {Nx.shape(g), Nx.shape(out)} do
-      {shape, shape} -> g
-      {{}, _} -> Nx.broadcast(Nx.as_type(g, Nx.type(out)), out)
-      {_, _} -> g
+  # Unused tuple outputs get a rank-0 zero from `sum_grad([])`. Broadcast
+  # each slot onto the primal leaf, as `update_grads(:while, ...)` does.
+  defp broadcast_tuple_grads(op, args, gs) do
+    case tuple_primal(op, args) do
+      nil ->
+        gs
+
+      primal ->
+        Enum.zip_with(gs, Composite.flatten_list([primal]), fn g, out ->
+          Nx.broadcast(g, Nx.devectorize(out, keep_names: false))
+        end)
     end
   end
+
+  defp tuple_primal(:block, [_, _, expr | _]), do: expr
+  defp tuple_primal(:metadata, [expr | _]), do: expr
+  defp tuple_primal(:cond, [_, last]), do: last
+  defp tuple_primal(:io_call, [tensor_expr | _]), do: tensor_expr
+  defp tuple_primal(_, _), do: nil
 
   defp select_composite(pred, left, right) do
     {selected, []} =
--- a/nx/test/nx/defn/grad_test.exs
+++ b/nx/test/nx/defn/grad_test.exs
@@ -2334,6 +2334,25 @@ defmodule Nx.Defn.GradTest do
         Nx.stack([eigh_evals_grad(t[0]), eigh_evals_grad(t[1])])
       )
     end
+
+    defn eigh_evecs_grad(t) do
+      grad(t, fn x ->
+        {_evals, evecs} = Nx.LinAlg.eigh(x)
+        Nx.sum(evecs)
+      end)
+    end
+
+    test "computes the grad when only one output is consumed" do
+      t = Nx.tensor([[4.0, 1.0, 0.5], [1.0, 3.0, 0.2], [0.5, 0.2, 2.0]])
+
+      assert_all_close(Nx.add(eigh_evals_grad(t), eigh_evecs_grad(t)), eigh_sum_grad(t))
+    end
+
+    test "grad of the eigenvalue sum is the identity" do
+      t = Nx.tensor([[4.0, 1.0, 0.5], [1.0, 3.0, 0.2], [0.5, 0.2, 2.0]])
+
+      assert_all_close(eigh_evals_grad(t), Nx.eye(3, type: :f32))
+    end
   end
 
   describe "cholesky" do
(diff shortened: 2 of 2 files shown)


nx/lib/nx/defn/grad.ex
```diff
@@ -620,6 +621,18 @@ defmodule Nx.Defn.Grad do
...
+  # output. An output nothing depends on accumulates no gradient, so
+  # `sum_grad/1` hands back the rank-0 zero, which block bodies cannot index
+  # or batch over. Restore the invariant here, once, rather than in each body.
+  defp match_block_cotangent(g, out) do
+    case {Nx.shape(g), Nx.shape(out)} do
+      {shape, shape} -> g
+      {{}, _} -> Nx.broadcast(Nx.as_type(g, Nx.type(out)), out)
+      {_, _} -> g
+    end
+  end
```
polvalente: We dont call them cotangents in this module, please rename to grads for consistency.
Also, matching solely on the shape might end up introducing bugs regarding vectorization.

polvalente: Additionally, I have a feeling that this might be the wrong fix or fixing in the wrong place.
Why does a block break here, but before block this worked fine for QR and LU?

polvalente: While should also be prone to the same types of bugs, so the solution should be similar to what is done for while

polvalente: > eigh has the only partial-output test in the suite, which is why it is also the only op https://github.com/elixir-nx/nx/pull/1783 got right.

The issue mentions a PR that doesn't really touch this file at all, so I don't think that exercising the case explains why it works for that function but not for others.

polvalente (changes requested):
As evidenced by the other issue, the fix in this PR is too specific to blocks (and TBH, very brittle). This seems to be an issue related to partially used tuples in general.

polvalente (approved):
I changed the fix into a holistic approach that fixes all similar cases, and kept the tests you added

PR #1815, opened 2026-08-16
Fix multi-tensor op dispatch to consider all tensor arguments

blasphemetheus:
## Problem

`Nx.put_slice`, `Nx.clip`, `Nx.gather`, `Nx.reduce`, and `Nx.window_reduce` dispatch on the first tensor's backend only (`impl!(tensor)`). Whenever a concrete tensor is combined with an `Expr` argument — a `Nx.Defn.jit` closure over a captured tensor, or a grad re-trace of a captured value — `Nx.BinaryBackend` gets picked and crashes:

```elixir
t = Nx.tensor([1.0, 2.0, 3.0, 4.0, 5.0])
Nx.Defn.jit(fn p -> Nx.put_slice(t, [1], p) end).(Nx.tensor([10.0, 20.0]))
# ** (FunctionClauseError) no function clause matching in Nx.BinaryBackend.to_binary/1
```

Found by fuzzing grad/jit with captured closure tensors; the same crash fires for all five ops in both jit and grad contexts.

## Fix

Dispatch through the existing `impl!/2,3` (whose `pick_struct` already prefers any backend over `Nx.BinaryBackend`), considering the tensor arguments that can independently be Exprs:

| Op | Dispatch |
|---|---|
| `put_slice` | `impl!(tensor, slice)` |
| `clip` | `impl!(tensor, min, max)` |
| `gather` | `impl!(tensor, indices)` |
| `reduce`, `window_reduce` | `impl!(tensor, acc)` |

One line per site in `nx/lib/nx.ex`.

A side effect worth noting: for `reduce`/`window_reduce` under grad, the dispatch crash was *masking* the intended "cannot compute gradient for Nx.reduce/4" ArgumentError — with correct dispatch, the documented error surfaces.

## Tests

- grad-side regressions (put_slice/clip values, reduce/window_reduce documented-error assertions) appended to the captured-closure area of `test/nx/defn/grad_test.exs`
- jit-closure regressions (put_slice/clip/gather values) in `test/nx/defn_test.exs` under `compiler: Evaluator`
- full nx suite green: 1384 doctests + 1375 tests

### DIFF
--- a/nx/lib/nx.ex
+++ b/nx/lib/nx.ex
@@ -12480,7 +12480,7 @@ defmodule Nx do
         tensor
       else
         out = %{tensor | type: type, shape: shape, names: names}
-        impl!(tensor).reduce(out, tensor, acc, [axes: axes, keep_axes: keep_axes], fun)
+        impl!(tensor, acc).reduce(out, tensor, acc, [axes: axes, keep_axes: keep_axes], fun)
       end
 
     vectorize(output, vectorized_axes)
@@ -12617,7 +12617,7 @@ defmodule Nx do
 
       out = %{tensor | shape: output_shape}
       opts = [padding: padding_config, strides: strides, window_dilations: dilations]
-      impl!(tensor).window_reduce(out, tensor, acc, window_dimensions, opts, fun)
+      impl!(tensor, acc).window_reduce(out, tensor, acc, window_dimensions, opts, fun)
     end)
   end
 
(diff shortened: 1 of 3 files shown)


polvalente (approved):
@blasphemetheus Looks good. Any reason for keeping the PR as draft?

blasphemetheus:
Nope no reason!

blasphemetheus:
Just what I put it as before review, now I'm just on me phone so I don't know if I can mark as ready til I get home

PR #1819, opened 2026-08-17
Raise ArgumentError for multiple :auto dimensions in reshape

blasphemetheus:
`Nx.reshape` with more than one `:auto` dimension leaked a bare `ArithmeticError`: after resolving the first `:auto`, `Tuple.product` multiplied by the second `:auto` atom.

```elixir
Nx.reshape(Nx.iota({12}), {:auto, :auto})
# before: ** (ArithmeticError) bad argument in arithmetic expression
# after:  ** (ArgumentError) only a single :auto dimension is allowed in reshape, got: {:auto, :auto}
```

### DIFF
--- a/nx/lib/nx/shape.ex
+++ b/nx/lib/nx/shape.ex
@@ -162,6 +162,11 @@ defmodule Nx.Shape do
   def reshape(old_shape, new_shape) do
     old_size = Tuple.product(old_shape)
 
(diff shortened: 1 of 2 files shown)


blasphemetheus:
this one is more opinionated, but covers a uncovered and unintentional error that would be confusing to run into in the wild

polvalente (approved):
Maybe worth a pass in `Nx.revectorize` `:target_shape`

PR #1822, opened 2026-08-19
Add runtime_raise inside defn

Chapaman:
Nx.Defn.Kernel.runtime_raise inside if/cond/while runs at runtime when that branch is taken. raise stays a trace-time check for shapes, ranks, and case.

Failed EXLA callbacks used to crash the Runner GenServer. They now exit the Outfeed with CallbackError and raise in the original caller, so they can be rescued or tested with assert_raise.

### DIFF
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -246,9 +246,16 @@ defmodule EXLA.Defn do
       end)
 
     infeeds = Map.new(infeeds)
+    error_ref = make_ref()
 
     {:ok, outfeed_pid} =
-      Outfeed.start_child(executable, outfeed, Process.group_leader(), infeeds)
+      Outfeed.start_child(
+        executable,
+        outfeed,
+        Process.group_leader(),
+        infeeds,
+        {self(), error_ref}
+      )
 
     ref = Process.monitor(outfeed_pid)
 
@@ -256,18 +263,39 @@ defmodule EXLA.Defn do
 
     {:ok, runner} =
       EXLA.Defn.Runner.start_link(lock, fn ->
-        EXLA.Executable.run(executable, [Enum.reverse(buffers)], run_options)
+        try do
+          EXLA.Executable.run(executable, [Enum.reverse(buffers)], run_options)
+        after
+          send(outfeed_pid, :stop)
+        end
       end)
 
     _ = EXLA.Defn.Lock.transfer(lock, fn -> send(runner, lock) end, outfeed_pid)
 
     receive do
       {:DOWN, ^ref, _, _, _} ->
-        results = EXLA.Defn.Runner.read(runner)
+        runner_result = EXLA.Defn.Runner.read(runner)
 
-        Enum.map(results, fn result ->
-          EXLA.Defn.Buffers.to_nx!(result, outputs, executable.mesh)
-        end)
+        callback_error =
+          receive do
+            {:exla_callback_error, ^error_ref, kind, reason, stacktrace} ->
+              {kind, reason, stacktrace}
+          after
+            0 -> nil
+          end
+
+        case {callback_error, runner_result} do
+          {{kind, reason, stacktrace}, _runner_result} ->
+            :erlang.raise(kind, reason, stacktrace)
+
+          {nil, {:error, kind, reason, stacktrace}} ->
+            :erlang.raise(kind, reason, stacktrace)
+
+          {nil, {:ok, results}} ->
+            Enum.map(results, fn result ->
+              EXLA.Defn.Buffers.to_nx!(result, outputs, executable.mesh)
+            end)
+        end
     end
   end
 
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -248,12 +248,7 @@ defmodule EXLA.Defn do
     infeeds = Map.new(infeeds)
 
     {:ok, outfeed_pid} =
-      Outfeed.start_child(
-        executable,
-        outfeed,
-        Process.group_leader(),
-        infeeds
-      )
+      Outfeed.start_child(executable, outfeed, Process.group_leader(), infeeds)
 
     ref = Process.monitor(outfeed_pid)
 
@@ -261,11 +256,7 @@ defmodule EXLA.Defn do
 
     {:ok, runner} =
       EXLA.Defn.Runner.start_link(lock, fn ->
-        try do
-          EXLA.Executable.run(executable, [Enum.reverse(buffers)], run_options)
-        after
-          send(outfeed_pid, :stop)
-        end
+        EXLA.Executable.run(executable, [Enum.reverse(buffers)], run_options)
       end)
 
     _ = EXLA.Defn.Lock.transfer(lock, fn -> send(runner, lock) end, outfeed_pid)
@@ -278,8 +269,8 @@ defmodule EXLA.Defn do
           {{:shutdown, %EXLA.Defn.CallbackError{} = error}, _runner_result} ->
             :erlang.raise(error.kind, error.reason, error.stacktrace)
 
-          {_down_reason, {:error, kind, reason, stacktrace}} ->
-            :erlang.raise(kind, reason, stacktrace)
+          {_down_reason, {:error, message}} ->
+            raise RuntimeError, message: message
 
           {_down_reason, {:ok, results}} ->
             Enum.map(results, fn result ->
@@ -310,13 +301,15 @@ defmodule EXLA.Defn do
           end)
         end)
 
-      EXLA.Executable.run(executable, input_lists, run_options)
-    else
-      [_ | _] = results ->
-        # For sharded execution, we get a list of results (one per partition)
-        Enum.map(results, fn result ->
-          EXLA.Defn.Buffers.to_nx!(result, outputs, executable.mesh)
-        end)
+      case EXLA.Executable.run(executable, input_lists, run_options) do
+        {:ok, results} ->
+          Enum.map(results, fn result ->
+            EXLA.Defn.Buffers.to_nx!(result, outputs, executable.mesh)
+          end)
+
+        {:error, message} ->
+          raise RuntimeError, message: message
+      end
     after
       EXLA.Defn.Lock.unlock(lock)
     end
--- a/exla/lib/exla/defn.ex
+++ b/exla/lib/exla/defn.ex
@@ -246,15 +246,13 @@ defmodule EXLA.Defn do
       end)
 
     infeeds = Map.new(infeeds)
-    error_ref = make_ref()
 
     {:ok, outfeed_pid} =
       Outfeed.start_child(
         executable,
         outfeed,
         Process.group_leader(),
-        infeeds,
-        {self(), error_ref}
+        infeeds
       )
 
     ref = Process.monitor(outfeed_pid)
@@ -273,25 +271,17 @@ defmodule EXLA.Defn do
     _ = EXLA.Defn.Lock.transfer(lock, fn -> send(runner, lock) end, outfeed_pid)
 
     receive do
-      {:DOWN, ^ref, _, _, _} ->
+      {:DOWN, ^ref, _, _, down_reason} ->
         runner_result = EXLA.Defn.Runner.read(runner)
 
-        callback_error =
-          receive do
-            {:exla_callback_error, ^error_ref, kind, reason, stacktrace} ->
-              {kind, reason, stacktrace}
-          after
-            0 -> nil
-          end
+        case {down_reason, runner_result} do
+          {{:shutdown, %EXLA.Defn.CallbackError{} = error}, _runner_result} ->
+            :erlang.raise(error.kind, error.reason, error.stacktrace)
 
-        case {callback_error, runner_result} do
-          {{kind, reason, stacktrace}, _runner_result} ->
+          {_down_reason, {:error, kind, reason, stacktrace}} ->
             :erlang.raise(kind, reason, stacktrace)
 
-          {nil, {:error, kind, reason, stacktrace}} ->
-            :erlang.raise(kind, reason, stacktrace)
-
-          {nil, {:ok, results}} ->
+          {_down_reason, {:ok, results}} ->
             Enum.map(results, fn result ->
               EXLA.Defn.Buffers.to_nx!(result, outputs, executable.mesh)
             end)
--- a/exla/lib/exla/defn/callback_error.ex
+++ b/exla/lib/exla/defn/callback_error.ex
@@ -3,7 +3,7 @@ defmodule EXLA.Defn.CallbackError do
   defexception [:kind, :reason, :stacktrace]
 
   @impl true
-  def message(%{kind: kind, reason: reason}) do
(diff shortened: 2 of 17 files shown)


nx/lib/nx/defn/kernel.ex
```diff
@@ -1438,6 +1439,77 @@ defmodule Nx.Defn.Kernel do
...
+
+      Nx.Defn.Kernel.if(unquote(predicate),
+        do:
+          Nx.Defn.Kernel.io_call(value, fn _ ->
+            Elixir.Kernel.raise(unquote(exception_or_message))
+          end),
+        else: value
+      )
+    end
+  end
```
polvalente: I think we should try to make `raise` itself become a special node that works in runtime when called inside defn. I'm just not sure how to make `raise` work inside `case`. `raise_if` is a nice out if we can't make it happen, though.

exla/lib/exla/defn.ex
```diff
@@ -246,28 +246,56 @@ defmodule EXLA.Defn do
...
 
     ref = Process.monitor(outfeed_pid)
 
     run_options = Keyword.put(run_options, :callback_server_pid, outfeed_pid)
 
     {:ok, runner} =
       EXLA.Defn.Runner.start_link(lock, fn ->
-        EXLA.Executable.run(executable, [Enum.reverse(buffers)], run_options)
+        try do
+          EXLA.Executable.run(executable, [Enum.reverse(buffers)], run_options)
```
polvalente: I think we're missing a change in the c++ code to not raise when unwrapping the result. This way if the callback returns `{:error, ...}` we can just surface that tuple without raising

exla/lib/exla/defn.ex
```diff
@@ -246,28 +246,56 @@ defmodule EXLA.Defn do
...
     _ = EXLA.Defn.Lock.transfer(lock, fn -> send(runner, lock) end, outfeed_pid)
 
     receive do
       {:DOWN, ^ref, _, _, _} ->
-        results = EXLA.Defn.Runner.read(runner)
+        runner_result = EXLA.Defn.Runner.read(runner)
+
+        callback_error =
+          receive do
+            {:exla_callback_error, ^error_ref, kind, reason, stacktrace} ->
```
josevalim: It would be nice to not have to deal with additional messages. One option is to wrap the raised exception in `EXLA.Defn.CallbackError{kind: kind, reason: reason, stacktrace: stacktrace}` and then in here, you can check the DOWN reason (the fifth element of the tuple) and unpack it accordingly. Then we don't need the error ref, we don't need to pass self, etc.

nx/lib/nx/defn/kernel.ex
Chapaman: Went with throw-during-trace, catch in cond/while, then the existing io_call raise shape. No new Expr op: those are tensors and raise has no type/shape. case stays trace-time (it's Elixir case). raise_if is now if + raise.

exla/lib/exla/defn.ex
Chapaman: Done. Outfeed exits with {:shutdown, %EXLA.Defn.CallbackError{...}} and maybe_outfeed unpacks the DOWN reason. {:shutdown, _} so Task.Supervisor doesn't log it. No error ref, extra messages, or passing self.

exla/lib/exla/defn.ex
Chapaman: run no longer unwrap-throws. Failed status comes back as {:error, message} from the NIF. Executable.run still raises that so the public contract is unchanged. Callback exceptions themselves come from the Outfeed DOWN.

nx/lib/nx/defn/kernel.ex
```diff
@@ -1511,22 +1577,35 @@ defmodule Nx.Defn.Kernel do
...
+  output, or the raising branch may be eliminated.
+
+  `raise_if/3` and `raise_if/4` are pipeable wrappers around the same
+  `if` plus `raise` pattern.
   """
   defmacro raise(exception, arguments) do
     quote do
-      Elixir.Kernel.raise(unquote(exception), unquote(arguments))
+      Nx.Defn.Expr.trace_raise(unquote(exception), unquote(arguments))
     end
```
polvalente: I think if we denormalize this node to regular `Elixir.Kernel.raise` when inside `:case`, we can implement `Nx.Defn.Expr.raise` such that it is always runtime only.

https://github.com/Chapaman/nx/blob/0fe88be090643087b1ce03a836b8289bc4900ba7/nx/lib/nx/defn/compiler.ex#L485

polvalente: If this works, we can then remove raise-if and the throw-catch trick

Chapaman: Case already ends up as Kernel.raise: only the matching clause runs, the throw escapes, and the compiler converts it. Rewriting in normalize({:case}) is the same outcome.

Always-runtime Expr.raise doesn't cover `if Nx.rank(t) != 2, do: raise` then `{r, c} = shape`. Rank is only known while tracing, and a discarded runtime node lets tracing continue into the match (that's the LinAlg CI failure). 4392b554 Kernel.raises when the remaining if/cond clause is already known true.

I couldn't drop the throw-catch: if raise only returns a node, `raise; value` falls through. Same question for raise_if as the pipeable wrapper. Do you have something in mind for those?

josevalim: @polvalente ah, the issue is that we already allow `raise` while building the graph, and that actually raises, so we cannot convert the same raise to raise at runtime. We will likely need `runtime_raise` or `io_raise` or something instead. :( Thoughts?

polvalente: I like runtime_raise!

Chapaman: Went with runtime_raise. raise is back to Kernel.raise while building the graph. raise_if is if plus runtime_raise.

polvalente: Let's remove raise_if. The original reason for adding it was for disambiguation of plain raise

exla/test/exla/defn/api_test.exs
```diff
@@ -396,6 +391,87 @@ defmodule EXLA.Defn.APITest do
...
+    defn custom_runtime_raise(value, predicate) do
+      if predicate do
+        runtime_raise RuntimeRaiseError,
+          message: "custom runtime check failed",
+          value: :preserved
+      else
+        value
+      end
+    end
+
```
polvalente: we might want a test where we have a custom call raise and catch inside itself to ensure that we only bubble up uncaught raises

exla/test/exla/defn/api_test.exs
```diff
@@ -396,6 +391,87 @@ defmodule EXLA.Defn.APITest do
...
+    defn halt_on_nth(x, n) do
+      {x, i, _n} =
+        while {x, i = 0, n}, i < 10 do
+          i =
+            if i == n do
+              runtime_raise "Halting on selected iteration"
```
polvalente: Maybe have one of these include the iteration as an argument? Not sure if the tensor would survive allocation, though. Seems like an interesting experiment.

nx/lib/nx/defn/compiler.ex
```diff
@@ -186,9 +186,17 @@ defmodule Nx.Defn.Compiler do
...
+           end) do
+        {:value, result} ->
+          result
+
+        {:raise, spec} ->
+          Nx.Defn.Expr.apply_runtime_raise(spec)
+      end
     after
       if previous_backend do
         Process.put(Nx.Shared.backend_pdict_key(), previous_backend)
```
polvalente: Do we still need this throw-catch mechanism given that runtime_raise is a new Expr node?
I believe we can have the compiler introduce the raise as a runtime thing when it is reached during execution.

nx/lib/nx/defn/kernel.ex
Chapaman: Done

exla/test/exla/defn/api_test.exs
Chapaman: the tensor does survive when passing through `io_call`

nx/lib/nx/defn/compiler.ex
Chapaman: after our talk, i'll have a ghost tensor to flag so that it is included on the result but removed before the final return - inspired by what is done on callback pid on exla

polvalente: I think this is due to how tracing is implemented. runtime_raise should be functionally equivalent to an io_call that raises, but it has a special node to not require the reassignment and passing a tensor as argument

exla/lib/exla/defn.ex
```diff
@@ -248,26 +248,44 @@ defmodule EXLA.Defn do
     infeeds = Map.new(infeeds)
 
     {:ok, outfeed_pid} =
-      Outfeed.start_child(executable, outfeed, Process.group_leader(), infeeds)
+      Outfeed.start_child(
+        executable,
+        outfeed,
+        Process.group_leader(),
+        infeeds
+      )
```
josevalim: We can revert this change!

exla/lib/exla/defn.ex
```diff
@@ -248,26 +248,44 @@ defmodule EXLA.Defn do
...
 
     run_options = Keyword.put(run_options, :callback_server_pid, outfeed_pid)
 
     {:ok, runner} =
       EXLA.Defn.Runner.start_link(lock, fn ->
-        EXLA.Executable.run(executable, [Enum.reverse(buffers)], run_options)
+        try do
+          EXLA.Executable.run(executable, [Enum.reverse(buffers)], run_options)
+        after
+          send(outfeed_pid, :stop)
```
josevalim: Why is this necessary? Shouldn't we keep it in the outfeed as before?

exla/lib/exla/defn.ex
```diff
@@ -248,26 +248,44 @@ defmodule EXLA.Defn do
...
 
-        Enum.map(results, fn result ->
-          EXLA.Defn.Buffers.to_nx!(result, outputs, executable.mesh)
-        end)
+        case {down_reason, runner_result} do
+          {{:shutdown, %EXLA.Defn.CallbackError{} = error}, _runner_result} ->
+            :erlang.raise(error.kind, error.reason, error.stacktrace)
+
+          {_down_reason, {:error, kind, reason, stacktrace}} ->
+            :erlang.raise(kind, reason, stacktrace)
```
josevalim: I don't understand why we have both the CallbackError and the {:error, kind, reason, stacktrace}. Could they be unified into one?

exla/lib/exla/executable.ex
```diff
@@ -47,8 +47,14 @@ defmodule EXLA.Executable do
     callback_server_pid = Keyword.get(options, :callback_server_pid)
     inputs = prepare_runtime_callback_inputs(inputs, callback_server_pid)
 
-    for data_and_device_id <- run(client, ref, device_id, inputs, options) do
-      decompose_output(data_and_device_id, output_typespecs, client, mesh)
+    case run(client, ref, device_id, inputs, options) do
+      {:error, message} ->
+        raise RuntimeError, message: IO.iodata_to_binary(message)
```
josevalim: Let's return {:error, message} to the caller and let the caller choose to raise.

exla/lib/exla/executable.ex
```diff
@@ -47,8 +47,14 @@ defmodule EXLA.Executable do
...
+
+      results ->
+        for data_and_device_id <- results do
+          decompose_output(data_and_device_id, output_typespecs, client, mesh)
+        end
```
josevalim: Let's make this proper tagged tuples:

```suggestion
      {:ok, results} ->
        {:ok, for data_and_device_id <- results do
          decompose_output(data_and_device_id, output_typespecs, client, mesh)
        end}
```

josevalim:
I have dropped some comments but it seems the implementation of Nx.Defn.Expr is still a bit too complex. Could runtime_raise be equivalent to:

```
defmacro runtime_raise(arg) do
  quote do
    io_call(fn -> Kernel.raise(arg) end)
  end
end
```

or similar?

On the other hand, we may have issues with tensors and inspect and what not, so an alternative is:

```
defmacro runtime_raise(arg) when is_binary(arg) do
  quote do
    io_call(fn -> Kernel.raise(arg) end)
  end
end
```

where we require raise to be a compile-time string.

exla/lib/exla/defn.ex
Chapaman: removed

Chapaman:
> I have dropped some comments but it seems the implementation of Nx.Defn.Expr is still a bit too complex. Could runtime_raise be equivalent to:
> 
> ```
> defmacro runtime_raise(arg) do
>   quote do
>     io_call(fn -> Kernel.raise(arg) end)
>   end
> end
> ```
> 
> or similar?
> 
> On the other hand, we may have issues with tensors and inspect and what not, so an alternative is:
> 
> ```
> defmacro runtime_raise(arg) when is_binary(arg) do
>   quote do
>     io_call(fn -> Kernel.raise(arg) end)
>   end
> end
> ```
> 
> where we require raise to be a compile-time string.

Dropped the Expr tracing stuff. runtime_raise is now io_call(Nx.tensor(0), fn _ -> Kernel.raise(msg) end). The message is a compile-time string to avoid dealing with tensors/inspect, is this ok?

polvalente (approved):

exla/lib/exla/defn/callback_error.ex
```diff
@@ -0,0 +1,9 @@
+defmodule EXLA.Defn.CallbackError do
+  @moduledoc false
+  defexception [:kind, :reason, :stacktrace]
+
+  @impl true
+  def message(%{kind: kind, reason: reason}) do
+    Exception.format_banner(kind, reason, [])
```
josevalim:
```suggestion
  def message(%{kind: kind, reason: reason, stacktrace: stacktrace}) do
    Exception.format_banner(kind, reason, stacktrace)
```

nx/lib/nx/defn/kernel.ex
```diff
@@ -1521,15 +1523,71 @@ defmodule Nx.Defn.Kernel do
...
+  See `runtime_raise/1` for when this runs relative to `raise/2`.
+  """
+  defmacro runtime_raise(exception, arguments) do
+    __defn__!(:runtime_raise, 2)
+
+    quote do
+      io_call(Nx.tensor(0), fn _ ->
+        Elixir.Kernel.raise(unquote(exception), unquote(arguments))
+      end)
+    end
```
josevalim: Let's remove this one for know because the lines between what the arguments mean can get really blurry. For example, what about this:

    runtime_raise(FooBar, a + b)

`a + b` is a tensor expression in this case, which is most likely not what people think. What we could do is to have an API like this:

    runtime_raise(Foo, a + b, fn a_plus_b -> "cannot have value: #{inspect(a + b)}" end)

but then people can just raise inside `io_call`.

josevalim (approved):
Just two minor comments, one about removing more code, and we can ship it!

polvalente (approved):

josevalim:

PR #1828, opened 2026-08-31
fix(grad): seed the backward pass from the expression's own type

### LINKED ISSUE #1827: grad returns wrong values for f64 functions containing while (backward pass is seeded as f32)
### Problem

`grad` seeds the backward pass with a hard-coded `{:f, 32}` regardless of the type
being differentiated. Straight-line code promotes that seed away on the first
multiply; a `while` carry cannot, because its types must match exactly.

```elixir
defmodule Seed do
  import Nx.Defn

  defn looped(x) do
    {_i, r} =
      while {i = 0, r = x}, i < 3 do
        {i + 1, Nx.multiply(r, 2.0)}
      end

    Nx.sum(r)
  end
end

x = Nx.tensor([1.0, 2.0, 3.0], type: :f64)

Seed.looped(x)
#=> #Nx.Tensor<f64  48.0>                 forward pass is fine

Nx.Defn.grad(&Seed.looped/1).(x)
#=> #Nx.Tensor<f32[3]  [0.0, 2.5, 0.0]>   expected f64[3] [8.0, 8.0, 8.0]
```

Three doublings give `Σ8x`, so the gradient is `8.0` everywhere; central
differences agree to 2.1e-8.

| body | grad type, f64 input |
|---|---|
| straight-line | `f64` |
| `cond` / `if` | `f64` |
| `while` | `f32` |

The result is wrong, not just single-precision. `Nx.to_binary/1` of the returned
gradient is byte-identical to the leading bytes of the correct f64 buffer:

| n | returned value | `to_binary` equals first `4n` bytes of correct f64 buffer |
|---|---|---|
| 1 | `[0.0]` | yes |
| 2 | `[0.0, 2.5]` | yes |
| 3 | `[0.0, 2.5, 0.0]` | yes |
| 4 | `[0.0, 2.5, 0.0, 2.5]` | yes |
| 5 | `[0.0, 2.5, 0.0, 2.5, 0.0]` | yes |

Under EXLA it fails loudly instead of silently:

```
error: expect operands to be compatible with body block return types but got
'tensor<i32>', 'tensor<3xf64>', 'tensor<f32>', 'tensor<3xf32>', ...
vs 'tensor<i32>', 'tensor<3xf64>', 'tensor<f32>', 'tensor<3xf64>', ...
```

The cause is `lib/nx/defn/grad.ex:114`:

```elixir
defp constant(float, %T{shape: shape} = t) do
  names = List.duplicate(nil, tuple_size(shape))
  Expr.constant(%{t | names: names, type: {:f, 32}}, float, [])
end
```

### Solution

Seed from the expression's own type:

```elixir
Expr.constant(%{t | names: names, type: Nx.Type.to_real(t.type)}, float, [])
```

`constant/2` has one caller, `grad.ex:98`. `Nx.Type.to_real/1` is
behaviour-preserving everywhere except the defect:

| input | `to_real` | change |
|---|---|---|
| `{:f, 64}` | `{:f, 64}` | fixes this |
| `{:f, 32}`, `{:bf, 16}` | unchanged | none |
| `{:s, _}`, `{:u, _}` | `{:f, 32}` | none |
| `{:c, 64}` / `{:c, 128}` | `{:f, 32}` / `{:f, 64}` | none — grads of complex functions stay real |

Measured on `main` at `37901d74` with that one line changed: the `while` case
returns `f64 [8.0, 8.0, 8.0]`, the EXLA error above is gone, `nx` is 2766 passed
(unchanged) and `exla` is 1418/1419 (unchanged — the one failure is
`EXLA.DeviceMemorySharingTest`, an IPC-handle failure already present before the
change).

`Nx.Type.to_floating/1` looks like the more natural choice and is wrong here:
`to_floating({:c, 64})` is `{:c, 64}`, so grads of complex functions come back
complex. That changes five existing tests on dtype alone — `pad works on
complex`, `broadcast computes gradient`, `slice computes gradient`, `squeeze
computes gradient for complex`, and `as_type/bitcast as_type passes through for
non-downcasting calls`.

#### Siblings checked

| site | affected | evidence |
|---|---|---|
| `grad.ex:114` `constant/2` | yes — this issue | the only hard-coded `{:f, 32}` under `lib/nx/defn/` |
| `svd.ex:254-255` `min_precision_type/1` | no | it is a floor, not a cap; `{:f, 64}` maps to `{:f, 64}` |
| `svd.ex:255` catch-all, complex input | unreachable | `Nx.LinAlg.svd/2` raises `not yet implemented for complex inputs` |

One question, since the mapping above turns on it: is a real-valued grad for a
complex-valued function the intended convention, or a consequence of this same
hard-coded seed? `to_real` preserves today's behaviour either way; if it is not
intended, that is a separate change.

Happy to open a PR with the fix and a regression test.

cash-mckeeman:
Closes #1827.

### Problem

`grad` seeds the backward pass with a hard-coded `{:f, 32}` regardless of the
type being differentiated. Straight-line code promotes that seed away on the
first multiply; a `while` carry cannot, because its types must match exactly.

```elixir
defn looped(x) do
  {_i, r} =
    while {i = 0, r = x}, i < 3 do
      {i + 1, Nx.multiply(r, 2.0)}
    end

  Nx.sum(r)
end

x = Nx.tensor([1.0, 2.0, 3.0], type: :f64)

Nx.Defn.grad(&looped/1).(x)
#=> #Nx.Tensor<f32[3]  [0.0, 2.5, 0.0]>   expected f64[3] [8.0, 8.0, 8.0]
```

| body | grad type, f64 input |
|---|---|
| straight-line | `f64` |
| `cond` / `if` | `f64` |
| `while` | `f32` |

Under EXLA this is an MLIR error — `tensor<3xf32>` against `tensor<3xf64>` in the
loop carry. Under `Nx.Defn.Evaluator` there is no error and the values are wrong.

### Solution

Seed from the expression's own type:

```elixir
Expr.constant(%{t | names: names, type: Nx.Type.to_floating(t.type)}, float, [])
```

`constant/2` has one caller, `grad.ex:98`, so this changes the seed and nothing
else. `Nx.Type.to_floating/1` preserves whatever the input already is:

| input | `to_floating` | effect |
|---|---|---|
| `{:f, 64}` | `{:f, 64}` | fixes this |
| `{:f, 32}`, `{:bf, 16}` | unchanged | none |
| `{:s, _}`, `{:u, _}` | `{:f, 32}` | none |
| `{:c, 64}` / `{:c, 128}` | `{:c, 64}` / `{:c, 128}` | grads of complex functions are now complex, where they used to be real |

That last row is a behaviour change, settled in review. Five existing assertions
compared complex gradients against real expectations and now compare against
complex ones — `pad works on complex`, `broadcast computes gradient`, `slice
computes gradient`, `squeeze computes gradient for complex`, and `as_type/bitcast
as_type passes through for non-downcasting calls`. Their values are unchanged;
only the dtype they are compared against moved.

### Siblings checked

| site | affected | evidence |
|---|---|---|
| `grad.ex:114` `constant/2` | yes — this fix | the only hard-coded `{:f, 32}` under `lib/nx/defn/` |
| `svd.ex:254-255` `min_precision_type/1` | no | a floor, not a cap; `{:f, 64}` maps to `{:f, 64}` |
| `svd.ex:255` catch-all, complex input | unreachable | `Nx.LinAlg.svd/2` raises `not yet implemented for complex inputs` |

### Verification

Two tests are added. Against the unpatched seed both fail; against `to_real`
only the complex one fails, which is what pins `to_floating` specifically — the
f64 case is fixed by either.

- `nx`: 2766 → 2768 passed, zero warnings, `mix format --check-formatted` clean
- `exla`: 1419 passed, 49 excluded
- the EXLA error above is gone; the same jitted grad returns `f64 [8.0, 8.0, 8.0]`

Branched from `main` at `37901d74`; independent of any other open PR.

### DIFF
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -111,7 +111,7 @@ defmodule Nx.Defn.Grad do
 
   defp constant(float, %T{shape: shape} = t) do
     names = List.duplicate(nil, tuple_size(shape))
-    Expr.constant(%{t | names: names, type: Nx.Type.to_real(t.type)}, float, [])
+    Expr.constant(%{t | names: names, type: Nx.Type.to_floating(t.type)}, float, [])
   end
 
   # Align heterogenous vectorized inputs to the union of vec axes — matching
--- a/nx/lib/nx/defn/grad.ex
+++ b/nx/lib/nx/defn/grad.ex
@@ -111,7 +111,7 @@ defmodule Nx.Defn.Grad do
 
   defp constant(float, %T{shape: shape} = t) do
     names = List.duplicate(nil, tuple_size(shape))
-    Expr.constant(%{t | names: names, type: {:f, 32}}, float, [])
+    Expr.constant(%{t | names: names, type: Nx.Type.to_real(t.type)}, float, [])
   end
 
   # Align heterogenous vectorized inputs to the union of vec axes — matching
--- a/nx/test/nx/defn/grad_test.exs
+++ b/nx/test/nx/defn/grad_test.exs
@@ -3678,6 +3678,26 @@ defmodule Nx.Defn.GradTest do
       assert grad_while_constant(tensor) == Nx.tensor([1.0, 1.0, 1.0, 1.0, 1.0])
     end
 
+    defn grad_while_f64(t) do
+      grad(t, fn t ->
+        {_, t} =
+          while {i = 0, t}, i < 3 do
+            {i + 1, t * 2.0}
+          end
+
+        Nx.sum(t)
+      end)
+    end
+
+    test "preserves the input type across the loop carry" do
+      # Three doublings, so Σ(8t) and ∂/∂t = 8 everywhere.
+      tensor = Nx.tensor([1.0, 2.0, 3.0], type: :f64)
+      result = grad_while_f64(tensor)
+
+      assert Nx.type(result) == {:f, 64}
+      assert result == Nx.tensor([8.0, 8.0, 8.0], type: :f64)
+    end
+
     defn grad_while_param(t, x) do
       value_and_grad(x, fn x ->
         {_, t} =
--- a/nx/test/nx/defn/grad_test.exs
+++ b/nx/test/nx/defn/grad_test.exs
@@ -3678,7 +3678,7 @@ defmodule Nx.Defn.GradTest do
       assert grad_while_constant(tensor) == Nx.tensor([1.0, 1.0, 1.0, 1.0, 1.0])
     end
 
-    defn grad_while_f64(t) do
+    defn grad_while_doubling(t) do
       grad(t, fn t ->
         {_, t} =
           while {i = 0, t}, i < 3 do
@@ -3691,11 +3691,18 @@ defmodule Nx.Defn.GradTest do
 
     test "preserves the input type across the loop carry" do
       # Three doublings, so Σ(8t) and ∂/∂t = 8 everywhere.
-      tensor = Nx.tensor([1.0, 2.0, 3.0], type: :f64)
-      result = grad_while_f64(tensor)
+      tensor = Nx.tensor([1.0, 2.0, 3.0], type: {:f, 64})
+      result = grad_while_doubling(tensor)
 
       assert Nx.type(result) == {:f, 64}
-      assert result == Nx.tensor([8.0, 8.0, 8.0], type: :f64)
+      assert result == Nx.tensor([8.0, 8.0, 8.0], type: {:f, 64})
+    end
+
+    test "preserves complex types across the loop carry" do
+      result = grad_while_doubling(~VEC[1+1i 2-2i 3])
+
+      assert Nx.type(result) == {:c, 64}
+      assert result == ~VEC[8 8 8]c64
     end
 
     defn grad_while_param(t, x) do
--- a/nx/test/nx/defn/grad_test.exs
+++ b/nx/test/nx/defn/grad_test.exs
@@ -2246,39 +2246,39 @@ defmodule Nx.Defn.GradTest do
     defn grad_sum_broadcast(t), do: grad(t, &Nx.sum(Nx.broadcast(&1, {3, 2, 2})))
 
     test "computes gradient" do
-      for multiplier <- [1, Complex.new(0, 1)] do
+      for {multiplier, type} <- [{1, {:f, 32}}, {Complex.new(0, 1), {:c, 64}}] do
         assert grad_sum_broadcast({3, 2, 2} |> Nx.iota() |> Nx.multiply(multiplier)) ==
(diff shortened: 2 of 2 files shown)


nx/lib/nx/defn/grad.ex
```diff
@@ -111,7 +111,7 @@ defmodule Nx.Defn.Grad do
 
   defp constant(float, %T{shape: shape} = t) do
     names = List.duplicate(nil, tuple_size(shape))
-    Expr.constant(%{t | names: names, type: {:f, 32}}, float, [])
+    Expr.constant(%{t | names: names, type: Nx.Type.to_real(t.type)}, float, [])
```
```suggestion
    Expr.constant(%{t | names: names, type: Nx.Type.to_floating(t.type)}, float, [])
```

I think this is the better definition. to_real would downcast a complex number

nx/test/nx/defn/grad_test.exs
```diff
@@ -3678,6 +3678,26 @@ defmodule Nx.Defn.GradTest do
...
+    end
+
+    test "preserves the input type across the loop carry" do
+      # Three doublings, so Σ(8t) and ∂/∂t = 8 everywhere.
+      tensor = Nx.tensor([1.0, 2.0, 3.0], type: :f64)
+      result = grad_while_f64(tensor)
+
+      assert Nx.type(result) == {:f, 64}
+      assert result == Nx.tensor([8.0, 8.0, 8.0], type: :f64)
+    end
```
polvalente: Can we also add a test for complex types?

nx/lib/nx/defn/grad.ex
cash-mckeeman: Applied. The five tests that asserted real gradients for complex functions now assert complex — values unchanged, only the dtype they compare against.

nx/test/nx/defn/grad_test.exs
cash-mckeeman: Added `preserves complex types across the loop carry`. It also happens to be what separates to_floating from to_real: the f64 case passes under both, so a complex input is the only thing that pins the choice.

cash-mckeeman:
CI is red only on the `1.20.2/28.5` legs — `1.17.3/27.3` is green on the same commit. The failures are `Nx.LinAlgTest "eigh properties for matrices with different eigenvalues"` and `EXLA.MemoryTrackingTest "tracks memory allocation and deallocation"`. Neither calls `grad`.

I measured `eigh` over 600 seeded draws with and without this change — worst eigenvalue error is identical, `4.57763671875e-5` (f32) and `7.62939453125e-5` (c64).

Both tests reseed every run: `Nx.Random.key(System.unique_integer())`, plus a bare `:rand.uniform()` for the eigenvalue signs. `Windows qr property` was the red on the previous run here and passes on this one.

The grad tests pass on the failing leg — `2767/2768`, `eigh` the only failure.

polvalente (approved):

PR #1830, opened 2026-09-04
fix(grad): take while gradient carry types from the carry

### LINKED ISSUE #1829: grad of a while with more than one carry fails to compile under EXLA
`grad` through a `while` stopped compiling under EXLA at f64 once the backward pass started seeding from the differentiated expression's type (#1828, mine). Anything whose carry holds more than one element hits it:

```elixir
defn acc_grad(a) do
  grad(a, fn a ->
    {_i, _a, acc} =
      while {i = 0, a = a, acc = a}, i < 3 do
        {i + 1, a, acc + a}
      end

    acc
  end)
end

a = Nx.tensor(1.0, type: :f64)

acc_grad(a)                #=> 4.0 -- the evaluator is fine
EXLA.jit(&acc_grad/1).(a)  #=> ** (RuntimeError) expect operands to be compatible
                           #   with body block return types ... 'tensor<f32>' ...
```

The gradient carries are built with `Nx.broadcast/2`, which takes the shape from the carry's initial but the type from the incoming gradient. With a single-element carry those happen to agree; with more than one they don't, so the loop's operands stop matching its body's return types.

Before #1828 this compiled, but returned an f32 gradient for an f64 function so it was latent before. It did run, though, and now it doesn't. Two related things fall out of the same place: a carry holding a tensor of a different type to the input (an f64 parameter alongside an f32 series) never compiled at all, and f16/bf16 gradients come back as f32 instead of keeping their type.

#1830 has a fix.

cash-mckeeman:
Closes #1829.

`grad` through a `while` stopped compiling under EXLA at f64 once #1828 (mine) made the seed follow the differentiated expression's type:

```elixir
defn acc_grad(a) do
  grad(a, fn a ->
    {_i, _a, acc} =
      while {i = 0, a = a, acc = a}, i < 3 do
        {i + 1, a, acc + a}
      end

    acc
  end)
end

EXLA.jit(&acc_grad/1).(Nx.tensor(1.0, type: :f64))
#=> ** (RuntimeError) expect operands to be compatible with body block return types
```

The gradient carries were built with `Nx.broadcast/2`, which takes the shape from the carry's initial and the type from the incoming gradient. With one element in the carry those agree; with more than one they don't. This takes the type from the initial instead.

That alone isn't enough, because the body can compute a gradient in a wider type than the carry it flows into — an f64 parameter carried alongside an f32 series still failed. This differentiates the body, widens the carries to whatever it produced, and differentiates again. 

The tests are in exla rather than nx because `Nx.Defn.Evaluator` computes the right answer in every one of these cases; only the MLIR lowering rejects a mismatched carry, which is why #1828's own test didn't catch this.

`nx` 2781 and `exla` 1435, both green.

### DIFF
--- a/exla/test/exla/defn/expr_test.exs
+++ b/exla/test/exla/defn/expr_test.exs
@@ -1515,6 +1515,49 @@ defmodule EXLA.Defn.ExprTest do
                      cond_inside_while_vectorized(Nx.vectorize(Nx.tensor([1, 2, 3]), :a), 3)
                    end
     end
+
+    defn grad_while_accumulator(a) do
+      grad(a, fn a ->
+        {_i, _a, acc} =
+          while {i = 0, a = a, acc = a}, i < 3 do
+            {i + 1, a, acc + a}
+          end
+
+        acc
+      end)
+    end
+
+    test "computes the gradient of a multi-element carry" do
+      for type <- [bf: 16, f: 16, f: 32, f: 64] do
+        # acc = a + 3a, so the gradient is 4 and keeps the input's type.
+        result = grad_while_accumulator(Nx.tensor(1.0, type: type))
+
+        assert Nx.type(result) == type
+        assert_equal(result, Nx.tensor(4.0, type: type))
+      end
+    end
+
+    defn grad_while_carried_series(a, s) do
+      grad(a, fn a ->
+        {_i, acc, _s} =
+          while {i = 0, acc = a, s = s}, i < 3 do
+            {i + 1, acc + Nx.sum(s), s}
+          end
+
+        acc
+      end)
+    end
+
+    test "computes the gradient when a carry has a different type to the input" do
+      for type <- [bf: 16, f: 16, f: 32, f: 64, s: 64] do
+        # acc = a + 3 * sum(s), so the gradient is 1 whatever s holds.
+        s = Nx.broadcast(Nx.tensor(2, type: type), {4})
+        result = grad_while_carried_series(Nx.tensor(1.0, type: {:f, 64}), s)
+
+        assert Nx.type(result) == {:f, 64}
+        assert_equal(result, Nx.tensor(1.0, type: {:f, 64}))
+      end
+    end
   end
 
   defn while_inside_if(pred, x) do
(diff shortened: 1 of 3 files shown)


exla/test/exla/defn/expr_test.exs
```diff
@@ -1515,6 +1515,27 @@ defmodule EXLA.Defn.ExprTest do
...
+
+        acc
+      end)
+    end
+
+    test "computes the gradient of a multi-element carry at f64" do
+      # acc = a + 3a, so the gradient is 4. With more than one element in the
+      # carry, the gradient carries cannot take their type from the incoming
+      # gradient -- the loop's operands stop matching its body's return types.
+      result = grad_while_accumulator(Nx.tensor(1.0, type: {:f, 64}))
```
polvalente: I suspect that f16 and other lower precision stuff might also be suffering from type promotion issues. Can we have this test be parameterized for other float types?

cash-mckeeman: Parameterized over [bf: 16, f: 16, f: 32, f: 64], and you were right: f16 and bf16 gradients were coming back as f32. Added a second test for a carry whose type differs from the input.

exla/test/exla/defn/expr_test.exs
```diff
@@ -1515,6 +1515,49 @@ defmodule EXLA.Defn.ExprTest do
...
+
+        acc
+      end)
+    end
+
+    test "computes the gradient when a carry has a different type to the input" do
+      for type <- [bf: 16, f: 16, f: 32, f: 64, s: 64] do
+        # acc = a + 3 * sum(s), so the gradient is 1 whatever s holds.
+        s = Nx.broadcast(Nx.tensor(2, type: type), {4})
+        result = grad_while_carried_series(Nx.tensor(1.0, type: {:f, 64}), s)
```
polvalente:
```suggestion
        result = grad_while_carried_series(Nx.tensor(1.0, type: {:f, 8}), s)
```

This shouldn't cause type failures as well

polvalente (approved):

PR #1835, opened 2026-09-07
fix(nx): read float literals at the precision of the expression around them

cash-mckeeman:
A bare float literal is annotated `{:f, 32}`, and that annotation is the precision it gets read back at. Binary operations promote it; nothing else does.

```elixir
defn pick(pred, t), do: Nx.select(pred, 0.1, t)

pick(Nx.tensor(1, type: :u8), Nx.tensor([1.0], type: :f64))
#=> #Nx.Tensor<f64[1] [0.10000000149011612]>
```

Same for `clip`, `pad`, `stack`, `reduce`, `dot`, and comparisons, which is how tolerances get written:

```elixir
defn converged?(t), do: t <= 1.0e-8

converged?(Nx.tensor(9.999999969612645e-9, type: :f64))
#=> #Nx.Tensor<u8 0>   # the probe is below 1.0e-8 and above its f32 round
```

The promotion lived in `binary_expr/5`. Every node is built through `expr/4`, so it moves there and the two calls in `binary_expr/5` become redundant. Comparisons answer in `{:u, 8}`, which can't say what precision to read the constant at, so the operands do.

Second commit: `reshape/2` didn't fold a constant the way `broadcast/4` does, so vectorized operands lost the promotion on the way through. Folding it also drops the wrapper nodes, so two tests asserting inspected expressions now assert smaller graphs.

One question — eager calls are untouched, so `Nx.multiply(Nx.f64(2), 0.7)` still gives `1.399999976158142`. Fixing that reaches into `Nx.Shared.binary_type/2`. Want it?

### DIFF
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -1652,23 +1652,23 @@ defmodule Nx.Defn.Expr do
   end
 
   # A float literal is annotated {:f, 32} but carries a full precision Elixir
-  # float, and the annotation is the precision it gets read back at. Comparisons
-  # answer in {:u, 8}, so there the operands say what that precision is.
+  # float, and the annotation is the precision it gets read back at. The output
+  # type and the other operands decide what that should be. The literal's own
+  # annotation is the thing being corrected, so it does not get a say, and an
+  # op answering in a tuple has no type to contribute.
   defp constant_read_type(%T{type: type}, args) do
-    if Nx.Type.float?(type) do
-      type
-    else
-      args
-      |> collect_float_types()
-      |> case do
-        [] -> type
-        [first | rest] -> Enum.reduce(rest, first, &Nx.Type.merge/2)
-      end
+    types = operand_float_types(args)
+    types = if match?({:tuple, _}, type), do: types, else: [type | types]
+
+    case types do
+      [] -> type
+      [first | rest] -> Enum.reduce(rest, first, &Nx.Type.merge/2)
     end
   end
 
-  defp collect_float_types(args) do
+  defp operand_float_types(args) do
     Enum.flat_map(args, fn
+      %T{data: %Expr{op: :constant}} -> []
       %T{type: type} -> if Nx.Type.float?(type), do: [type], else: []
       _ -> []
     end)
--- a/nx/lib/nx/defn/expr.ex
+++ b/nx/lib/nx/defn/expr.ex
@@ -1353,6 +1353,8 @@ defmodule Nx.Defn.Expr do
   end
 
   defp expr(tensor, context, op, args) do
+    args = upcast_float_constants(args, constant_read_type(tensor, args))
+
     %{tensor | data: %Expr{id: id(), op: op, args: args, context: context}, donatable?: false}
   end
 
@@ -1644,15 +1646,40 @@ defmodule Nx.Defn.Expr do
         |> Nx.to_number()
         |> then(&constant(out, &1))
 
-      c1 ->
-        expr(out, context, op, [maybe_upcast_float_constant(arg1, out.type), arg2])
-
-      c2 ->
-        expr(out, context, op, [arg1, maybe_upcast_float_constant(arg2, out.type)])
-
       true ->
         expr(out, context, op, [arg1, arg2])
     end
+  end
+
+  # A float literal is annotated {:f, 32} but carries a full precision Elixir
+  # float, and the annotation is the precision it gets read back at. Comparisons
+  # answer in {:u, 8}, so there the operands say what that precision is.
+  defp constant_read_type(%T{type: type}, args) do
+    if Nx.Type.float?(type) do
+      type
+    else
+      args
+      |> collect_float_types()
+      |> case do
+        [] -> type
+        [first | rest] -> Enum.reduce(rest, first, &Nx.Type.merge/2)
+      end
+    end
+  end
+
+  defp collect_float_types(args) do
+    Enum.flat_map(args, fn
+      %T{type: type} -> if Nx.Type.float?(type), do: [type], else: []
+      _ -> []
+    end)
+  end
+
+  defp upcast_float_constants(args, type) do
+    Enum.map(args, fn
+      %T{data: %Expr{op: :constant}} = t -> maybe_upcast_float_constant(t, type)
+      list when is_list(list) -> upcast_float_constants(list, type)
+      other -> other
+    end)
   end
 
   defp maybe_upcast_float_constant(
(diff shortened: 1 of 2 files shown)


nx/lib/nx/defn/expr.ex
```diff
@@ -1639,17 +1646,43 @@ defmodule Nx.Defn.Expr do
...
+        [] -> type
+        [first | rest] -> Enum.reduce(rest, first, &Nx.Type.merge/2)
+      end
+    end
+  end
+
+  defp collect_float_types(args) do
+    Enum.flat_map(args, fn
+      %T{type: type} -> if Nx.Type.float?(type), do: [type], else: []
+      list when is_list(list) -> collect_float_types(list)
```
polvalente: I don't think we should be traversing nested lists here. These are generally option arguments and not tensor arguments. Do we have a test that fails if we remove this?

nx/lib/nx/defn/expr.ex
```diff
@@ -1639,17 +1646,43 @@ defmodule Nx.Defn.Expr do
...
+    if Nx.Type.float?(type) do
+      type
+    else
+      args
+      |> collect_float_types()
+      |> case do
+        [] -> type
+        [first | rest] -> Enum.reduce(rest, first, &Nx.Type.merge/2)
+      end
+    end
```
polvalente, line 1683: I think this function should just use `tensor` and `args` directly through the `else` block. Otherwise this might end up staying at a lower precision.

nx/lib/nx/defn/expr.ex
```diff
@@ -1090,7 +1090,12 @@ defmodule Nx.Defn.Expr do
   @impl true
   def reshape(out, tensor) do
     tensor = to_expr(tensor)
-    expr(out, tensor.data.context, :reshape, [tensor])
+
+    if c = maybe_constant(tensor) do
+      constant(out, c)
+    else
+      expr(out, tensor.data.context, :reshape, [tensor])
+    end
```
polvalente, line 1098: Ideally this would be a separate PR

nx/lib/nx/defn/expr.ex
polvalente: General question for the PR: what do we want to happen if we get :c64 and :c128 vs a float constant? What happens if an :f64 constant encounters a :c64 tensor?

nx/lib/nx/defn/expr.ex
cash-mckeeman: I'll split this out

nx/lib/nx/defn/expr.ex
cash-mckeeman: Removed from `collect_float_types` — comparisons are the only caller and they're binary, so it never fired. 

Kept in `upcast_float_constants`, where `concatenate` and `stack` pass their tensors in a list; without it `Nx.stack([t_f64, 0.1])` regresses. Nothing covered that, so I added a test.

nx/lib/nx/defn/expr.ex
```diff
@@ -1639,17 +1641,42 @@ defmodule Nx.Defn.Expr do
...
       true ->
         expr(out, context, op, [arg1, arg2])
     end
   end
 
+  # A float literal is annotated {:f, 32} but carries a full precision Elixir
+  # float, and the annotation is the precision it gets read back at. The output
+  # type and the other operands decide what that should be. The literal's own
+  # annotation is the thing being corrected, so it does not get a say, and an
+  # op answering in a tuple has no type to contribute.
```
polvalente:
```suggestion
  # widen precision of float literals as needed
```

nx/lib/nx/defn/expr.ex
```diff
@@ -1639,17 +1641,42 @@ defmodule Nx.Defn.Expr do
...
+  defp constant_read_type(%T{type: type}, args) do
+    types = operand_float_types(args)
+    types = if match?({:tuple, _}, type), do: types, else: [type | types]
```
polvalente: If a tuple can appear there, I think this would justify recursion into the tuple arguments

nx/lib/nx/defn/expr.ex
```diff
@@ -1639,17 +1641,42 @@ defmodule Nx.Defn.Expr do
...
+
+    case types do
+      [] -> type
+      [first | rest] -> Enum.reduce(rest, first, &Nx.Type.merge/2)
+    end
```
polvalente: I think if we include the tuple argument types, we can always use the reduce here

nx/lib/nx/defn/expr.ex
```diff
@@ -1639,17 +1641,42 @@ defmodule Nx.Defn.Expr do
...
+  end
+
+  defp operand_float_types(args) do
+    Enum.flat_map(args, fn
+      %T{data: %Expr{op: :constant}} -> []
+      %T{type: type} -> if Nx.Type.float?(type), do: [type], else: []
+      _ -> []
+    end)
```
polvalente, line 1691: I think we should return all types for simplicity

nx/lib/nx/defn/expr.ex
```diff
@@ -1639,17 +1641,42 @@ defmodule Nx.Defn.Expr do
...
+  end
+
+  defp upcast_float_constants(args, type) do
+    Enum.map(args, fn
+      %T{data: %Expr{op: :constant}} = t -> maybe_upcast_float_constant(t, type)
+      list when is_list(list) -> upcast_float_constants(list, type)
```
polvalente: I still think traversing lists here is not the way to go

nx/lib/nx/defn/expr.ex
cash-mckeeman: I think in other parts of the repo, the constant adapts to the tensor (`Nx.Type.merge_number/2`) with the exception being where the tensor cannot hold the constant type - the fallback in that function catches cases where a tensor with integer type widens to `{:f, 32}`.

That feels right to me in general and in this PR. I tried to faithfully implement according to that principle.

polvalente: Agreed!

cash-mckeeman: Digging through this a bit, though, I found that when something calls `to_tensor` that the `Nx.Type.merge/2` function is called. And that function runs against the `{:f, 32}` placeholder given by `Nx.Type.infer/1`. 

All code paths get that placeholder, but it's `to_tensor` that is destructive in the sense that it rounds the input right after, I believe. Going through `defn`, the type is just stamped and the input can be recovered. 

Maybe that's the real fix needed; making those two code paths agree.

nx/lib/nx/defn/expr.ex
cash-mckeeman: I'll take another pass

cash-mckeeman:
Simplified to avoid `{:tuple, size}` - it's what `elem/2` takes as an argument, and what `while` and `cond` return. `Nx.Type.float?/1` is already false for it.

Also folded the check and the promotion into one function, so `expr/4` only asks for a literal's precision when there's a literal.

polvalente (approved):

